PR #100, opened 2014-04-11
Add hidden property support

ritch:
This fixes https://github.com/strongloop/loopback/issues/233

/to @bajtos @raymondfeng


crandmck:
Added to docs per #832.

Issue #234, opened 2014-08-19
Assigning default values: null vs. undefined


fabien:
Currently any default value (default: ... in LDL) will be assigned as long as the value is undefined. If the value has been set to null, the default won't be assigned.

This introduces the following side-effect: due to the `removeUndefined()` calls during `save()`, you cannot actually persist 'undefined' to the data store. AFAIK most databases will represent missing data as null anyway. As such, you cannot ever get back to a state where the default value will be applied - it will remain null.

superkhau:
@fabien Has the issue been resolved? Can I close it?

PR #238, opened 2014-08-20
hasOne and belongsTo update/destroy methods on scope

fabien:
Note the following semantics:
- destroy on a hasOne relation will nullify the foreignKey (see also: #234)
- destroy on a belongsTo relation will effectively destroy the related item completely

Is this what's expected? Or should we introduce a flag to configure this?


mariohmol:
Isnt a upsert for this? Like.. user.Address.upsert(..) instead of checking if is new do create or if not use update..

PR #309, opened 2014-09-24
Allow to filter by null value

manuelnaranjo:
There are cases where it's interesting to find all the elements that have a
property set to null, for example when a there's a relation and you want to
find all those objects without the relation.

I found out that mssql, mysql and I think also postresql provide this feature, just not the memory connector.

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -403,6 +403,10 @@ function applyFilter(filter) {
   }
 
   function test(example, value) {
+    if (example === null) {
+      return value === undefined || value === null;
+    }
+
     if (typeof value === 'string' && (example instanceof RegExp)) {
       return value.match(example);
     }


raymondfeng:
See https://github.com/strongloop/loopback/issues/1009 too.

raymondfeng:
@manuelnaranjo Would you please sign https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler?

manuelnaranjo:
@raymondfeng sure just did

bajtos:
Please add a unit-test to [test/basic-querying.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/84448bb0dcfee4f26c9a051aa43bea22dcf0bab3/test/basic-querying.test.js) covering this feature. The tests in this file are run by all connectors, thus we will know whether they support this feature correctly.

Having said that, this patch may be going in a wrong direction per the discussion in strongloop/loopback#1009.

Amir-61:
Looking into this old abandoned PR. @manuelnaranjo would you like to continue working on this PR? If so please add test coverage. With that being said it seems this may conflict with pending feature: https://github.com/strongloop/loopback/issues/1009

manuelnaranjo:
No sorry I haven't used loopback in a while, you can close it I don't have
the time to work on it

superkhau:
Closing as abandoned.

PR #321, opened 2014-10-09
Provide feedback (result count) for updateAll and destroyAll

### LINKED ISSUE #294: No actual feedback for delete(All)
AFAIK none of the delete/destroy methods provide useful feedback whether the delete was completed, and if so, what it affected (in case of deleteAll/where). If nothing matched, there's no way to tell. Taking this further, should this return an error in the case of removeById?

fabien:
This currently targets the Memory connector and fixes #294 - it appears that the MongoDB connector already does the right thing. We'd have to check the other connectors (running test/manipulation.test.js will fail if unsupported).

This is the first step towards more meaningful feedback from the (HTTP) API as well (needs loopback update, for remoting return values).

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -527,17 +527,22 @@ Memory.prototype.destroyAll = function destroyAll(model, where, callback) {
   }
   var cache = this.collection(model);
   var filter = null;
+  var count = 0;
   if (where) {
     filter = applyFilter({where: where});
     Object.keys(cache).forEach(function (id) {
       if (!filter || filter(this.fromDb(model, cache[id]))) {
+        count++;
         delete cache[id];
       }
     }.bind(this));
   } else {
+    count = Object.keys(cache).length;
     this.collection(model, {});
   }
-  this.saveToFile(null, callback);
+  this.saveToFile(null, function(err) {
+    callback && callback(err, count);
+  });
 };
 
 Memory.prototype.count = function count(model, callback, where) {
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -564,16 +564,20 @@ Memory.prototype.update =
     filter = applyFilter({where: where});
 
     var ids = Object.keys(cache);
+    var count = 0;
     async.each(ids, function (id, done) {
       var inst = self.fromDb(model, cache[id]);
       if (!filter || filter(inst)) {
+        count++;
         self.updateAttributes(model, id, data, done);
       } else {
         process.nextTick(done);
       }
     }, function (err) {
       if (!err) {
-        self.saveToFile(null, cb);
+        self.saveToFile(null, function(err) {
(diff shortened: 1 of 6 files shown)


bajtos:
- #502 added feedback on the number of records modified by `updateAll`.
- #511 added feedback on the number of records deleted by `deleteAll`.

The only remaining piece in this patch is reporting whether "deleteById" deleted the record or whether the record was already deleted. 

Considering that this PR has been stale for more than 4 months, I am closing it.

@fabien please send a new pull request when you get time to implement feedback in "deleteById".

fabien:
@bajtos @raymondfeng how should we handle the return value for `prototype.destroy`? see https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1763

We should do one of the following:

`cb(err, count > 0)`

OR:

`cb(err ? err : (count === 0 ? new Error('Not deleted ...') : undefined))`

raymondfeng:
Reporting as an error seems to be more appropriate to me.

fabien:
@raymondfeng want me to create a PR?

bajtos:
Reporting as an error is IMO a breaking change. When I want to delete a record, then the fact that it was already deleted is usually not important.

bajtos:
Don't forget to update `destroyById` while you are changing `prototype.delete`, so that the API is consistent.

fabien:
@bajtos so you're not interested in knowing if it succeeded the first time? ...

fabien:
If it's a breaking change (because the error will propagate to the REST API), I'm find with `cb(err, count > 0)` to get a boolean result (count as a number doesn't make any sense IMO).

raymondfeng:
Now propotype.delete takes an `options` object. Can we do the following?

1. By default, it calls back with `cb(null, count > 0);`
2. We allow an option property such as `reportFailureAsError` to report an error.

fabien:
@raymondfeng what will be the error message?

raymondfeng:
Something like 'No instance found for id <id>'

bajtos:
Can you enlighten me and describe a situation when you want the delete operation to fail when the record is already not there?

Hiding this behaviour behind a flag is a reasonable compromise allowing users to choose the flavour they prefer. Although it won't be available via REST API (at least not now), so perhaps a per-model option (setting) or a global loopback option would make more sense?

fabien:
Well, for the same reasons you'd like to know how many items you deleted doing `deleteAll` - it confirms that something was there and now isn't anymore. Perhaps someone else deleted it in the meantime, while you were under the impression it was still there when you decided to delete it. It sounds like a contrived example, but in general you'd like to know if the action you performed was effective (even if the end result is the same, it is gone either way).

We can always add such a setting or global option later. If it weren't for @raymondfeng to bring the Error option up, I would have omitted error reporting for this first pass anyway.

PR #324, opened 2014-10-10
Include scope

fabien:
This allows the usage of all the filter params (where, order, fields, include, ...) on included relations.

The syntax is as follows  `{ relation: '<relation>', scope: { ... } }`:

```
Category.find({
  include: {relation: 'posts', scope:{
    fields: ['title'], include: ['author'],
    order: 'title DESC'
  }}
}, function(err, categories) {
  console.log(categories)
});

User.find({
  include: {posts: {relation: 'categories', scope:{
    order: 'seq DESC'
  }}}
}, function(err, users) {
  console.log(users)
});
```

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -15,7 +15,7 @@ function normalizeInclude(include) {
     // Build an array of key/value pairs
     var newInclude = [];
     var rel = include.rel || include.relation;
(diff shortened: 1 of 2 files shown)


JpCapdevila:
Will this work in the REST API?

raymondfeng:
I tried `belongsTo` User and it works for me.

Issue #325, opened 2014-10-10
GeoPoint constructor does not support "object" data type as an input parameter


mikstur:
Hi,

In the geo.js file https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/geo.js#L117, when passing in an object such as:
{"lat": 12.3, "lng": 45.6}, the server crashes. 

I received this error while trying to query a model's related model as follows:
- Model A has two GeoPoint fields
- Model B has no GeoPoint fields but other strings and numbers
- Model A "hasMany" Model B.

When calling ModelA.find({include:"ModelB"},function(err,results){}); the server crashes and console prints out 'lat must be a number when creating a GeoPoint'.

Discussion here: https://groups.google.com/forum/#!topic/loopbackjs/6cLhZcOiuSE

Cheers.

PR #331, opened 2014-10-16
GeoPoint constructor does not support JSON string object data type as an input parameter

mikstur:
https://github.com/strongloop/loopback-datasource-juggler/issues/325


raymondfeng:
Fixed by https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/geo.js#L135

PR #333, opened 2014-10-16
Update validateUniqueness to support custom primaryKeys.

tma-isbx:
We have some models that have some non-'id' primary keys and they were failing validations when we were trying to save them.

Not sure how to sign the CLA, but I don't have a problem with it.


raymondfeng:
@tma-isbx Any update?

bajtos:
Closing as abandoned.

PR #359, opened 2014-11-26
add hooks module and support pre && post

violet-day:
Here is usage.

``` js
Post.pre('_save', function (next) {
  //do stuff
 next();
});
```


bajtos:
#367 implemented "operation hooks". See #397 for per-method hooks that are closer to what was proposed here.

violet-day:
@bajtos Thank you for ur work,I have used this new API to refactor my project.

PR #364, opened 2014-12-02
docs.json: add lib/model.js

bajtos:
I noticed that `ModelBaseClass` was missing from our API docs. The class provides some useful methods like `ModelBaseClass.defineProperty(prop, params)`.

screen shot 2014-12-02 at 8 29 36

/to @raymondfeng @crandmck please review

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -1,10 +1,11 @@
 {
     "content": [
         "lib/datasource.js",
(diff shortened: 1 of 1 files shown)


pulkitsinghal:
thanks @bajtos

PR #368, opened 2014-12-09
ValidationError messages include field value

bitmage:
I modified the formatErrors() function to include the current value of the fields which have failed validation.

The format looks like this:

```
The `Person` instance is not valid. Details: `id: 123456` can\'t be set.
```

Please let me know if that format is good, and I will update the test expectations to match.

### DIFF
--- a/lib/validations.js
+++ b/lib/validations.js
@@ -735,7 +735,7 @@ function ValidationError(obj) {
   this.message = util.format(
     'The %s instance is not valid. Details: %s.',
       context ? '`' + context + '`' : 'model',
-      formatErrors(obj.errors) || '(unknown)'
+      formatErrors(obj) || '(unknown)'
   );
 
   this.statusCode = 422;
@@ -760,10 +760,21 @@ function ValidationError(obj) {
 util.inherits(ValidationError, Error);
 
 var errorHasStackProperty = !!(new Error).stack;
+var inspect = require('util').inspect;
 
-function formatErrors(errors) {
+function formatProperty(obj, propertyName) {
+  var value = obj[propertyName];
+  if ((typeof value) === 'function')
+    return propertyName
+  else if ((typeof value) === 'string')
+    return propertyName + ': ' + inspect(value.substring(0, 32));
+  else
+    return propertyName + ': ' + inspect(value);
+}
+
+function formatErrors(obj) {
   var DELIM = '; ';
-  errors = errors || {};
+  var errors = obj.errors || {};
   return Object.getOwnPropertyNames(errors)
     .filter(function(propertyName) {
       return Array.isArray(errors[propertyName]);
@@ -771,7 +782,7 @@ function formatErrors(errors) {
     .map(function(propertyName) {
       var messages = errors[propertyName];
       return messages.map(function(msg) {
-        return '`' + propertyName + '` ' + msg;
+        return '`' + formatProperty(obj, propertyName) + '` ' + msg;
       }).join(DELIM);
     })
     .join(DELIM);
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -2112,7 +2112,7 @@ describe('relations', function () {
         should.exist(err);
         err.name.should.equal('ValidationError');
         var msg = 'The `Passport` instance is not valid.';
-        msg += ' Details: `name` can\'t be blank.';
+        msg += ' Details: `name: undefined` can\'t be blank.';
         err.message.should.equal(msg);
         done();
       });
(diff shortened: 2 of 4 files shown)


bajtos:
What's the status of this work? Are you @bitmage still willing to finish it?

bitmage:
Oh yeah, sorry!  I will take a look in the next day or two and try to get things wrapped up.

lib/validations.js
```diff
@@ -760,10 +760,21 @@ function ValidationError(obj) {
...
+
+function formatProperty(obj, propertyName) {
+  var value = obj[propertyName];
+  if ((typeof value) === 'function')
+    return propertyName
+  else if ((typeof value) === 'string')
+    return propertyName + ': ' + inspect(value.substring(0, 32));
+  else
+    return propertyName + ': ' + inspect(value);
+}
```
bitmage: What do you think of this?  I noticed we were producing some strange output in the case of nested documents (which have a type of `function`).  I don't know how to output a sensible output in that case... but I think it's an edge case that can be left for another day.

For strings, objects, and undefined/null we now have some pretty clear output.

test/relations.test.js
```diff
@@ -2569,7 +2569,7 @@ describe('relations', function () {
       Person.create({ name: 'Wilma', addresses: addresses }, function(err, p) {
         err.name.should.equal('ValidationError');
         var expected = 'The `Person` instance is not valid. ';
-        expected += 'Details: `addresses` contains invalid item: `work` (`street` can\'t be blank).';
+        expected += 'Details: `addresses: [ { id: \u001b[32m\'home\'\u001b[39m, street: \u001b[32m\'Home Street\'\u001b[39m },\n  { id: \u001b[32m\'work\'\u001b[39m, street: \u001b[32m\'\'\u001b[39m } ]` contains invalid item: `work` (`street` can\'t be blank).';
```
bajtos: This is wrong. Error messages must not include ANSI color escape sequences.

lib/validations.js
bajtos: Handling non-scalar values (objects, arrays, functions) will be always tricky. You can take a look at [chai.js](https://github.com/chaijs/chai/blob/5fc486b5733d67a02ead53de6557741d4c57c276/lib/chai/utils/objDisplay.js#L31-L42) which basically prints only top-level keys of the object.

As for function values, `JSON.stringify` removes them from the output, which solves the problem. On the other hand, it does not handle circular references, which may be a problem.

test/relations.test.js
bajtos: Notice how an invalid item in an array is handled: the message contains `id` of the item, and then information about the error in this item only.

This patch should build on top of that and print only the invalid item in the message. Example:

```
// Option A
Details: `addresses` contains invalid item: `work` (`street: undefined`) can't be blank
```

bajtos: > Notice how an invalid item in an array is handled: the message contains id of the item, and then information about the error in this item only.

> This patch should build on top of that and print only the invalid item in the message. Example:

On the second thought, that would be a much more involved change, let's leave it for another day.

However, you still need to remove colour codes.

bitmage: Ok.  Why are the color codes getting inserted in the first place?  I'm using 'inspect' from node core API.  I had never noticed it to print colors before.  Is one of the Loopback dependencies monkey patching it to add colors?

I decided to use 'inspect' because it limits how deep it will parse the tree.  This seemed like desirable behavior which JSON.stringify does not offer.

bitmage:
Your Model prototype is [implementing the inspect method](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model.js#L442) and configuring it to output colors.

Any time a model is console.log'd or inspected it will run this method.  This overrides whatever top-level configuration is used, so  `inspect(value, {colors: false})` in the validation code doesn't work.

I can fall back to JSON.stringify as you originally suggested, but it's a blocking operation, and I'm still concerned about the lack of a depth limit.

bajtos:
Closing in favour of #389.

bajtos:
> Any time a model is console.log'd or inspected it will run this method. This overrides whatever top-level configuration is used, so inspect(value, {colors: false}) in the validation code doesn't work.

Good catch! There are several ways how to fix the problem, e.g. by passing `customInspect:false` to `util.inspect` or calling `toObject()` on Model instances before calling `util.inspect`. I'll check possible solutions and fix the problem in #389.

bajtos:
On Node v0.11.14+, one can return an object from the custom inspect function. That way all options are honoured. For Node v0.10 and older, I have disabled color output.

PR #379, opened 2014-12-22
fix sorting of undefined values in memory connector

cvette:
undefined values should be moved to the end. The current sorting function returns 0, if a or b is undefined.


raymondfeng:
@cvette Thank you for the patch. Can you add a unit test?

PR #383, opened 2014-12-28
replace deprecated function __defineGetter__

bitmage:
`__defineGetter__` is deprecated:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__defineGetter__

Replacing it with Object.defineProperty retains the same functionality but is ES5 standard:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Specifications.

This was causing a failure in IE.


bitmage:
Updates on this?  The tests are failing for an unrelated reason - floating point errors with GPS coordinates.

PR #390, opened 2015-01-12
Fix virtual id get function.

BerkeleyTrue:
Fix id property returning undefined in certain situations.


slnode:

PR #389, opened 2015-01-12
Include property value in the error message

bajtos:
When building a list of errors for `ValidationError.message`, include
the values of invalid properties too.

In order to keep the message reasonably short, the values are truncated
at approx 32 characters.

See the unit-tests for examples of error messages.

Supersedes #368.
A part of strongloop/strong-arc#562

/to @raymondfeng @bitmage please review
/cc @fabien @ritch FYI

PR #395, opened 2015-01-15
Fix detection of `util.inspect` version

bajtos:
Improve the detection to work in browser too. This fixes a regression
introduced in e4fc3878 (#389)

/cc @rmg @raymondfeng I'll merge without review to push the fix to npmjs ASAP.


bajtos:
Released as `loopback-datasource-juggler@2.14.1`

rmg:
Awesome :+1:

Issue #367, opened 2014-12-09
Provide consistent hook APIs for methods mixed into model classes from connectors


raymondfeng:
When a LoopBack model is attached a data source, the underlying connector will mix in a list of methods on the model class. If the connector is backed by one of the databases, the methods are the CRUD operations. At the moment, the juggler only has hard-coded before/after hooks that have serious issues:

1. The before/after hook only allows one handler
2. The triggering of hooks are hard-coded in the lib/dao.js
3. Not all paths are covered

See open issues or pull requests:
- https://github.com/strongloop/loopback-datasource-juggler/pull/363
- https://github.com/strongloop/loopback-datasource-juggler/pull/359
- https://github.com/strongloop/loopback-datasource-juggler/issues/262
- https://github.com/strongloop/loopback-datasource-juggler/issues/202
- https://github.com/strongloop/loopback-datasource-juggler/issues/159
- https://github.com/strongloop/loopback/issues/614
- https://github.com/strongloop/loopback-datasource-juggler/issues/142

Tasks:

 - [x] Finish and land #403
 - [x] Land `before delete` hook #412
 - [x] Update docs draft for `before delete` hook
 - [x] Write documentation for the new hook API.
 - [x] Rename `query` to access (#418) and update docs
 - ~~Clean up change-detection events, let them use the new API - see https://github.com/strongloop/loopback/issues/1048~~
 - ~~Add `test/persistence-hooks.suite.js` to connector tests: Not strictly needed, since all hooks are implemented inside juggler.~~

Out of scope (follow-up):
  - Per-method hooks #397
  - Advance hook use-cases #417

bajtos:
@alFReD-NSH I am starting to work on this today.

>  I assume you guys prefer to use loopback-phase for this.

Not necessarily.

STRML:
What would you like to use, then? It should be a pretty simple impl with
hooks-js.

bajtos:
> @STRML What would you like to use, then? It should be a pretty simple impl with
hooks-js.

First I need to get more familiar with the way how the current hooks work, including various quirky edge-cases. Then I'd like to come up with a proposal how the hooks should work instead and discuss it  here. 

Whether we will end up using hooks-js or a different module is an implementation detail. I'll worry about that after we agree on what is the desired behaviour.

bajtos:
I summarized the current status of hooks in a [gist](https://gist.github.com/bajtos/b8157112452eeaf984c9).

[...]

PR #398, opened 2015-01-19
use connector.findOrCreate() if possible

clarkorz:
When using mongodb as database, I often use the native `findAndModify()` to perform a `find or create` (using `$setOnInsert` and `upsert` option). Since `findAndModify()` is an atomic operation, it won't cause conflict in some cases.

This PR tests that if the `connector` support `findOrCreate`, and if so then use it, otherwise `find()` or `create()`.

**Note**:
Since this is an atomic operation now, should the `create` hooks be triggered?

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -582,6 +582,9 @@ Memory.prototype.update =
     async.each(ids, function (id, done) {
       var inst = self.fromDb(model, cache[id]);
       if (!filter || filter(inst)) {
+        // The id value from the cache is string
+        // Get the real id from the inst
+        id = self.getIdValue(model, inst);
         self.updateAttributes(model, id, data, done);
       } else {
         process.nextTick(done);
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -228,6 +228,15 @@ Memory.prototype.updateOrCreate = function (model, data, callback) {
   });
 };
 
+Memory.prototype.findOrCreate = function (model, query, data, callback) {
+  this.all(model, query, function (err, list) {
+    if (err || (list && list[0])) return callback(err, list && list[0], false);
+    this.create(model, data, function (err) {
+      callback(err, data, true);
+    });
+  }.bind(this));
+};
+
 Memory.prototype.save = function save(model, data, callback) {
   var id = this.getIdValue(model, data);
   var cachedModels = this.collection(model);
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -603,6 +603,9 @@ Memory.prototype.updateAttributes = function updateAttributes(model, id, data, c
     }
   }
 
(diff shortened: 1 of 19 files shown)


lib/connectors/memory.js
```diff
@@ -228,6 +228,15 @@ Memory.prototype.updateOrCreate = function (model, data, callback) {
...
 
+Memory.prototype.findOrCreate = function (model, query, data, callback) {
+  this.all(model, query, function (err, list) {
+    if (err || (list && list[0])) return callback(err, list && list[0], false);
+    this.create(model, data, function (err) {
+      callback(err, data, true);
+    });
+  }.bind(this));
+};
+
```
bajtos: Isn't is better to leave this out and let juggler to use the default unoptimized implementation for the memory connector?

lib/dao.js
```diff
@@ -332,14 +332,64 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, callback) {
...
+    data = obj.toObject(true);
+
+    // validation required
+    obj.isValid(function (valid) {
+      if (valid) {
+        findOrCreate();
+      } else {
+        callback(new ValidationError(obj), obj);
+      }
+    }, data);
```
bajtos: This looks like a lot of copy-and-pasted code, is there a way how to refactor it into a shared method?

bajtos:
@raymondfeng could you please review ASAP? This patch affects my work on the new hooks infrastructure in https://github.com/strongloop/loopback-datasource-juggler/issues/367.

> Note: Since this is an atomic operation now, should the create hooks be triggered?

I'd say we can leave that out of this patch in favour of #367

lib/connectors/memory.js
clarkorz: @bajtos you are right. I'm lazy. This only for testing the case that connector *has* `findOrCreate()`. But now, the tests can't cover the case that connector does NOT have `findOrCreate()`.

I should extend the memory connector and implement `findOrCreate()` in a standalone test.

bajtos: > I should extend the memory connector and implement findOrCreate() in a standalone test.

That would be much better indeed :+1:

bajtos:
@clarkorz please rebase the patch on top of the current master. #403 modified the way how `findOrCreate` works, it added new hooks ([preliminary docs](https://gist.github.com/bajtos/b8157112452eeaf984c9#file-new-hooks-docs-md)).

raymondfeng:
@clarkorz Any progress?

clarkorz:
Ops...sorry for mess it up. I have dropped this PR and create a new one here https://github.com/strongloop/loopback-datasource-juggler/pull/419

bajtos:
Please try to cleanup (rewrite) the git history using `git rebase -i` and/or `git cherry-pick` and keep the same PR next time.

Issue #384, opened 2015-01-04
Is it possible to embed denormalized properties for consistent structure of response that with or without `include`?


clarkorz:
Thinking there is an *Order* model, which belongs to a `buyer` (*User* model), a `seller`(*User* model), and a `product` (*Product* model). For `GET /orders`, each order instance need to include buyer's name, seller's name, and product's name.

For now, I could define `fields` scope option on each relations. e.g.:
```
"buyer": {
  "type": "belongsTo", 
  "model": "User", 
  "scope": { "fields": [ "name" ], "limit": 10 }
}
```

The problem is,  `GET /orders` involves  `1 + 10 * 3` times of db queries, and these `name` properties rarely changes.  So, I would like to denormalize these properties to *Order* model. 

The juggler support this by `properties` option on relation definition. e.g.:
```
"buyer": {
  "type": "belongsTo", 
  "model": "User", 
  "scope": { "fields": [ "name" ], "limit": 10 },
  "properties": { "name": "buyerName" },
  "options": { "invertProperties": true }
}
```
The response of `Order.find()` will be: 
```js
// Order.find()
[{
...
  buyerId: 1, buyerName: "john",
  sellerId: 1, sellerName: "joe",
  productId: 1, productName: "pats",
...
}]
```
If I do `include` some relations, the response will be different:
```js
// Order.find({ include: ['buyer', 'seller', 'product'] })
[{
...
  buyerId: 1, buyerName: "john", buyer: { id: 1, name: "john", ... },
  sellerId: 1, sellerName: "joe", seller: { id: 1, name: "joe", ... },
  productId: 1, productName: "pats", product: { id: 1, name: "pats", ... },
...
}]
```
Would it be better that denormalized properties are embedded into to the root object, so the response of `Order.find()` is similar as `Order.find({ include: ['buyer', 'seller', 'product'] })` ?
```
// Order.find() equivalent to following query but without extra db query
// Order.find( {
//    include: {
//      buyer: { "fields": [ "name" ] },
//      seller: { "fields": [ "name" ] },
//      product: { "fields": [ "name" ] }
//    }
//  })
{
...
  buyerId: 1, buyer: { id: 1, name: "john", ... },
  sellerId: 1, seller: { id: 1, name: "joe", ...},
  productId: 1, product: { id: 1, name: "pats", ...},
...
}
```

PR #399, opened 2015-01-19
support embeds data for belongsTo relation

clarkorz:
PR for https://github.com/strongloop/loopback-datasource-juggler/issues/384

If ModelA belongsTo ModelB, this PR will include some properties of ModelB on instances of ModelA as embedded object. 

The effect is similar with using an `include` filter with `fields` option, but without extra db queries. Which is useful when denormalizing some properties.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -13,6 +13,8 @@ var List = require('./list');
 var Hookable = require('./hooks');
 var validations = require('./validations');
 var _extend = util._extend;
+var utils = require('./utils');
+var fieldsToArray = utils.fieldsToArray;
 
 // Set up an object for quick lookup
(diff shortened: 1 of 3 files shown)


raymondfeng:
@clarkorz This feature is completely broken for relational DBs which won't persist any properties that are not defined in the model definition.

raymondfeng:
@clarkorz Did you see my comment? I had to disable the test at https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/relations.test.js#L1791-L1794 so that connectors for relational DBs can pass.

clarkorz:
@raymondfeng sorry I miss the message, I will investigate it right away

PR #400, opened 2015-01-20
Add support for protected properties.

chrene:
Protected properties allows you to define properties for an object which should not be included when nested. For instance if you have an Author object and a Book object. A book has a relation to with Author, and book is public API. Author will have personal information such as social security number etc, and they can now be "protected" such that anyone looking up the author of the book will not get those information back.

tldr: It's like the "hidden" feature, but it's only hidden when nested

### DIFF
--- a/lib/list.js
+++ b/lib/list.js
@@ -72,11 +72,11 @@ List.prototype.push = function (obj) {
   return item;
 };
 
-List.prototype.toObject = function (onlySchema, removeHidden) {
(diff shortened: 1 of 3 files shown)


slnode:

raymondfeng:
Would the `protected` flag applies to included models too? `include` can be used to pull in related models over relations. See http://docs.strongloop.com/display/LB/Include+filter.

chrene:
I'm not sure yet, I'll test it out

Issue #404, opened 2015-01-21
destroyAll with where break compatibility with loopback-connector-rest


tario:
I am using REST datasource and I am trying to make a simple delete from strongloop API explorer, it fails crashing the server with exception "TypeError: object is not a function"
It was reported on lookback-connector-rest, but maybe should be fixed on loopback-datasource-juggler
https://github.com/strongloop/loopback-connector-rest/issues/21

Stacktrace:
https://github.com/strongloop/loopback-connector-rest/blob/master/lib/rest-connector.js#L271
https://github.com/strongloop/loopback-connector-rest/blob/master/lib/rest-model.js#L53
...

The callback object used on rest-connector.js:L271 is defined here (Stacktrace): 
https://github.com/strongloop/loopback-connector-rest/blob/master/lib/rest-connector.js#L268
https://github.com/strongloop/loopback-connector-rest/blob/master/lib/rest-connector.js#L353
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L909

Note this line of code on dao.js
```javascript
    this.getDataSource().connector.destroyAll(this.modelName, where, function (err, data) {
```

It passes the "query object" as second argument to destroyAll function, destroyAll only forward the arguments in order and the second argument arriving on RestConnector.prototype.responseHandler is the same queryobject, later this object is used as "function" and then it fails with the error "Object is not a function"

Maybe I can fix this on my own fork, but I don't know what would be the best solution according your project to share it with you.

PD: beside this, the design ideas and the implementation of strongloop is pretty AWESOME

Edit: additional data, I found the change on b136a8fc, after version 2.9.0, maybe downgrading to that version of the npm would work fine for me

raymondfeng:
@tario Thank you for reporting the inconsistency. IIRC, the `where` argument was added after the REST connector was built. Can you try to submit a patch to fix it in https://github.com/strongloop/loopback-connector-rest/issues/21?

tario:
I will send a Pull Request, in what branch you need the fix ?

tario:
I have yet another thing that seems like a wrong behavior related to delete, I have a strongloop "client" consumming another strongloop "server" using REST datasource, each time I delete an item requesting delete to client, server returns status code 204 No Content (According to this, it is ok http://www.restapitutorial.com/lessons/httpmethods.html), but client strongloop interprets it like an error and returns 500 Internal Server Error on the first request.

[...]

PR #406, opened 2015-01-22
#404: fix incompatible use of remove by id breaking compatibility with loopback-connector-rest

tario:
This fix issue #404

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -375,7 +375,17 @@ DataAccessObject.exists = function exists(id, cb) {
  */
 DataAccessObject.findById = function find(id, cb) {
   if (stillConnecting(this.getDataSource(), this, arguments)) return;
-  this.findOne(byIdQuery(this, id), cb);
+
+  this.getDataSource().connector.find(this.modelName, id, function (err, data) {
+    var obj = null;
+    if (data) {
+      if (!getIdValue(this, data)) {
+        setIdValue(this, data, id);
+      }
+      obj = new this(data, {applySetters: false, persisted: true});
+    }
+    cb(err, obj);
+  }.bind(this));
 };
 
 DataAccessObject.findByIds = function(ids, cond, cb) {
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -926,13 +926,13 @@ DataAccessObject.remove = DataAccessObject.deleteAll = DataAccessObject.destroyA
 DataAccessObject.removeById = DataAccessObject.destroyById = DataAccessObject.deleteById = function deleteById(id, cb) {
   if (stillConnecting(this.getDataSource(), this, arguments)) return;
   var Model = this;
-  
-  this.remove(byIdQuery(this, id).where, function(err) {
+
+  this.getDataSource().connector.destroy(this.modelName, id, function (err) {
     if ('function' === typeof cb) {
       cb(err);
     }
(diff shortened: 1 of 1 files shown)


slnode:

lib/dao.js
```diff
@@ -926,13 +926,13 @@ DataAccessObject.remove = DataAccessObject.deleteAll = DataAccessObject.destroyA
...
-  
-  this.remove(byIdQuery(this, id).where, function(err) {
+
+  this.getDataSource().connector.destroy(this.modelName, id, function (err) {
     if ('function' === typeof cb) {
       cb(err);
     }
     if(!err) Model.emit('deleted', id);
-  });
+  }.bind(this));
```
raymondfeng: This change will break other things, such as the default scope. We translate deleteById to deleteAll with an id filter to make the 'delete' consistent. I understand that some connectors may choose to provide a specific implementation of destroy (by id). I need to find out what is the best way to honor that.

tario: Maybe, we should fix loopback-connector-rest to accept the where on deleteAll instead of merging this change?

But how could rest connector do a "where" ?

raymondfeng: A workaround in the REST connector is to detect the where object with a single `id` property and translate it as `delete by id`.

bajtos:
FWIW, this patch no longer applies cleanly since #403 has been landed.

> A workaround in the REST connector is to detect the where object with a single id property and translate it as delete by id.

The algorithm detecting whether the `where` object is a single id can be useful for change replication too.

@raymondfeng What's the plan for this pull request? Is there something @tario can improve to get it landed? If a  different implementation is needed, then I am proposing to close this PR as rejected.

raymondfeng:
@tario Can you check if https://github.com/strongloop/loopback-datasource-juggler/pull/403 helps?

tario:
#403 doesnt work for me, I just test my issue using     
```javascript
...
"loopback-datasource-juggler": "git://github.com/strongloop/loopback-datasource-juggler.git#f9b0ac482c4833c7fbe10f970a03f3b77486e014",
...
```
on my package.json, and it fails (when I try to delete something, the server crashes with "TypeError: object is not a function")

For my particular case, the issue is already solved, I am using my modified version from here:
https://github.com/tario/loopback-datasource-juggler/commits/master

Thanks for all the hints!

bajtos:
@slnode test please
@tario we need you to sign the CLA before your patch can be merged (if we decide to accept it): https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler

jouke:
I have submitted a pull request for loopback-connector-rest that fixes this issue. It should really be fixed in there and not in dao.js imho

slnode:

jannyHou:
@tario Could you assign the CLA and add some test cases to verify your change? Thanks.

slnode:

slnode:

slnode:

jannyHou:
Looks like it's fixed in https://github.com/strongloop/loopback-connector-rest/issues/21 @tario thank you for the contribution :) are you still keen to work on this pr?

jannyHou:
I am closing this PR since it's fixed in strongloop/loopback-connector-rest#21 , feel free to reopen it if there is an update or someone still run into the problem. Thanks for understanding.

PR #409, opened 2015-01-27
Fixed nullCheck in validations for undefined attributes

billinghamj:
The current code `!(attr in this)` was not working (and I have written a unit test showing this). I have replaced it with a typeof undefined check. This then passes the unit test.

I have also written a couple of other unit tests to check other possible cases to ensure my change does not impact those.

### DIFF
--- a/lib/validations.js
+++ b/lib/validations.js
@@ -589,8 +589,7 @@ var defaultMessages = {
 };
 
 function nullCheck(attr, conf, err) {
-  var isNull = this[attr] === null || !(attr in this);
-  if (isNull) {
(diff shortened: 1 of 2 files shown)


slnode:

billinghamj:
Good point - I've made that change.

billinghamj:
I can roll the commits together into one if you'd prefer

raymondfeng:
@billinghamj +1 to squash the commits into one. Thanks!

billinghamj:
👍

billinghamj:
Thanks

Issue #159, opened 2014-07-01
DataAccessObject.upsert/updateOrCreate - no hooks triggered


fabien:
Contrary to create(), upsert() does not call any before/after hooks at all.

In addition, it appears to be impossible for any of the beforeHooks to cancel the pending action (save for example). `next(new Error('Halted'))` apparently doesn't work.

bajtos:
I am closing this issue, we are moving away from `beforeUpdate/afterUpdate` hooks in favour of [Operation hooks](http://docs.strongloop.com/display/LB/Operation+hooks) and [Per-method hooks](https://github.com/strongloop/loopback-datasource-juggler/issues/397).

Please upgrade your project to use the new `before save` and/or `after save` hooks and open a new GH issue if you run into any problem.

jmls:
does this mean that we have to write our own upsert hooks to find the model id we have a string ID ?

I am trying to PUT (upsert) data over existing records , but am getting an error back 

code":11000,"errmsg":"insertDocument :: caused by :: 11000 E11000 duplicate key error index

which suggests that the record is being created rather than updated

bajtos:
> does this mean that we have to write our own upsert hooks to find the model id we have a string ID ?
@jmls Definitely not. AFAIK, LoopBack automatically converts (coerces) the model id to the correct type. What database are you using? I remember there were some issues with ID coercion in the MongoDB connector.

jmls:
I am using mongo ...

On 17 February 2015 at 15:29, Miroslav Bajtoš <notifications@github.com>
wrote:

> does this mean that we have to write our own upsert hooks to find the
> model id we have a string ID ?
> @jmls <https://github.com/jmls> Definitely not. AFAIK, LoopBack
> automatically converts (coerces) the model id to the correct type. What
> database are you using? I remember there were some issues with ID coercion
> in the MongoDB connector.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/strongloop/loopback-datasource-juggler/issues/159#issuecomment-74685822>
> .
>



-- 
Julian Lyndon-Smith
IT Director,
dot.r
http://www.dotr.com

bajtos:
> I am using mongo ...

@jmls could you please open a new issue in [loopback-connector-mongodb](https://github.com/strongloop/loopback-connector-mongodb/issues/new) and provide steps how to reproduce the problem?

PR #411, opened 2015-01-28
Ensure hooks get called on upsert. Fix for #159

mrfelton:
This is @jcsmesquita's fix for the issue in the form of a pull request so that it can be more easily tested by people. In my limited testing it does appear to work, although my understanding of loopback's internals is not good enough to know if this is the right way to resolve this.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -265,48 +265,21 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
   if (stillConnecting(this.getDataSource(), this, arguments)) {
     return;
   }
-  var self = this;
   var Model = this;
   if (!getIdValue(this, data)) {
     return this.create(data, callback);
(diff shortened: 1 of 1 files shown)


slnode:

bajtos:
Hi @mrfelton, thank you for the pull request. We are moving away from the beforeCreate/afterCreate hooks - see #403, #412 and [preliminary docs](https://gist.github.com/bajtos/b8157112452eeaf984c9#file-new-hooks-docs-md). Can you please try them out and let us know in #367 whether these new hooks fix the problem for you? 

The solution proposed in this patch is incorrect as it never calls `connector.updateOrCreate`. It is preferable to use connector-provided implementation of `updateOrCreate` whenever possible, as it is more performant and usually provides an atomic implementation, which is something the fall-back method implemented in DAO cannot guarantee.

PR #412, opened 2015-01-29
Implement intent hook `before delete`

bajtos:
Methods `DAO.deleteAll` and `DAO.prototype.delete` now invoke `before delete` hook too. The hook receives `ctx.where` describing models to be deleted.

Connect #367

/cc @ritch FYI

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -995,7 +995,11 @@ DataAccessObject.remove = DataAccessObject.deleteAll = DataAccessObject.destroyA
       { Model: Model, query: query },
       function(err, ctx) {
         if (err) return cb(err);
-        doDelete(ctx.query.where);
+        var context = { Model: Model, where: ctx.query.where };
+        Model.notifyObserversOf('before delete', context, function(err, ctx) {
+          if (err) return cb(err);
+          doDelete(ctx.where);
+        });
       });
   }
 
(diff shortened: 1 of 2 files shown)


raymondfeng:
What's behavior of next(err) and next(null, result) if the hook decides to respond with err or result and bypass the 'delete'? I don't see test coverage for that.

bajtos:
Good catch. I have added test for `next(err)`.

Regarding `next(null, result)`: I checked the discussion in #165 which introduced this options, I have read the linked mailing list thread, and did not find any description of the use-case when this is needed. Could you please refresh my memory? 

At the moment, none of the new hooks supports `next(err, result)`.  I strongly prefer to avoid this third overload, as the overload is very difficult to implement consistently for all hooks. For example, the `query` hook is called by `updateAttributes`, `updateAll` and `deleteAll`. Each of these functions pass different data to their callback, thus the hook cannot provide a single `result` that would be correct in all cases.

IMO this third overload belongs to per-method hooks (#397).

raymondfeng:
`next(null, result)` can be used for the following use cases:

1. Caching: returns the result from cache instead of calling the target method
2. Try to find before create/update
3. Implement soft delete

raymondfeng:
The code LGTM.

bajtos:
> next(null, result) can be used for the following use cases:
>
> - Caching: returns the result from cache instead of calling the target method
> - Try to find before create/update
> - Implement soft delete
>
> We can add this behavior in a separate PR.

I have created a new issue #417 to track this requirement, because I have already run out of the time allocated for hooks in this print.

PR #413, opened 2015-01-29
Make sure batch create calls back with correct data

raymondfeng:
/to @bajtos or @ritch 
/cc @glesage

See https://github.com/strongloop/loopback/issues/1031


bajtos:
LGTM.

bajtos:
Please defer a new juggler release until the implementation of the new hooks is complete, at which point a minor release should be carved out.

PR #414, opened 2015-01-30
Fix hasOne remoting

raymondfeng:
/to @bajtos or @ritch 

The PR is required to enable remoting for hasOne relations


bajtos:
LGTM. Please defer a new juggler release until the implementation of the new hooks is complete, at which point a minor release should be carved out.

PR #415, opened 2015-01-30
Fix regression in `.save()` from 1fd6eff

bajtos:
The commit 1fd6eff (intent-based hooks) introduced a subtle regression in `.save()` method where dynamic property setters were invoked twice.

This patch fixes the problem by moving pre-save data normalization into `before save` callback.

The second commit removes redundant `toObject` call which I discovered by reviewing the diff of 1fd6eff. Note that I was not able to write a unit-test that would fail due to this duplicate call.

Connect #367

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1132,21 +1132,19 @@ DataAccessObject.prototype.save = function (options, callback) {
     options.throws = false;
   }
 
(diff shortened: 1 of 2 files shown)


raymondfeng:
LGTM

PR #416, opened 2015-01-31
Fix id type issue for update

raymondfeng:
/to @ritch  or @bajtos 

The PR fixes https://github.com/strongloop/loopback/issues/1058

### DIFF
--- a/test/memory.test.js
+++ b/test/memory.test.js
@@ -7,11 +7,11 @@ var async = require('async');
 var should = require('./init.js');
 var Memory = require('../lib/connectors/memory').Memory;
 
-describe('Memory connector', function () {
+describe('Memory connector', function() {
   var file = path.join(__dirname, 'memory.json');
 
   function readModels(done) {
-    fs.readFile(file, function (err, data) {
+    fs.readFile(file, function(err, data) {
       var json = JSON.parse(data.toString());
       assert(json.models);
       assert(json.ids.User);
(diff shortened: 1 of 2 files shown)


test/memory.test.js
```diff
@@ -7,11 +7,11 @@ var async = require('async');
 var should = require('./init.js');
 var Memory = require('../lib/connectors/memory').Memory;
 
-describe('Memory connector', function () {
+describe('Memory connector1', function() {
```
bajtos: why `connector`?

test/memory.test.js
```diff
@@ -19,82 +19,113 @@ describe('Memory connector', function () {
...
+              }
+              readModels(function(err, json) {
+                if (err) {
+                  return done(err);
+                }
+                assert.equal(Object.keys(json.models.User).length, 2);
+                var user = JSON.parse(json.models.User[ids[1]]);
+                assert.equal(user.name, 'John');
+                assert(user.id === ids[1]);
+                User.update({id: user.id}, {name: 'John1'},
```
bajtos: This test is already too long and testing way too many things.

Don't add your new test code into this existing test as it makes things even worse. Please create a new test focused only on the use case you are fixing.

raymondfeng:
@bajtos Tests are refactored.

bajtos:
Thanks. It would be even better if the tests don't depend on each other, but we can improve that later.

LGTM.

pulkitsinghal:
used the master v2.15.0 and it works

Issue #417, opened 2015-02-02
Advanced hook use-cases


bajtos:
- Caching: returns the result from cache instead of calling the target method
- Try to find before create/update
- Implement soft delete

See the discussion in #412. This feature may depend on #397.

clarkbw:
> Implement soft delete

see strongloop/loopback#1241 about soft delete

PR #418, opened 2015-02-02
Rename hook "query" to "access"

bajtos:
The name "query" creates incorrect assumption that hook handlers may return the result of a query to bypass database access (see #417). That is far from true, since this hook is called also by methods like `deleteAll` or `updateAll` that don't perform any SELECT query.

Connect strongloop/loopback-datasource-juggler#367

/cc @crandmck FYI. I'll update the documentation myself when this is landed.


raymondfeng:
LGTM

PR #419, opened 2015-02-03
use optimized findOrCreate if possible

clarkorz:
When using mongodb as database, I often use the native `findAndModify()` to perform a `find or create` (using `$setOnInsert` and `upsert` option). Since `findAndModify()` is an atomic operation, it won't cause conflict in some cases.

This PR tests that if the `connector` support `findOrCreate`, and if so then use it, otherwise `find()` or `create()`.

This PR use the new hook system introduced in https://github.com/strongloop/loopback-datasource-juggler/pull/403

### DIFF
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -274,26 +274,25 @@ module.exports = function(dataSource, should) {
           });
       });
 
-      // TODO(bajtos) Enable this test for all connectors that
-      // provide optimized implementation of findOrCreate.
-      // The unoptimized implementation does not trigger the hook
-      // when an existing model was found.
-      it.skip('triggers `before save` hook when found', function(done) {
-        TestModel.observe('before save', pushContextAndNext());
+      if (dataSource.connector.findOrCreate) {
+        it('triggers `before save` hook when found', function(done) {
+          TestModel.observe('before save', pushContextAndNext());
 
-        TestModel.findOrCreate(
-          { where: { name: existingInstance.name } },
-          { name: existingInstance.name },
-          function(err, record, created) {
-            if (err) return done(err);
-            observedContexts.should.eql(aTestModelCtx({ instance: {
-              id: record.id,
-              name: existingInstance.name,
-              extra: undefined
-            }}));
-            done();
-          });
-      });
+          TestModel.findOrCreate(
+            { where: { name: existingInstance.name } },
+            { name: existingInstance.name },
+            function(err, record, created) {
+              if (err) return done(err);
+              record.id.should.eql(existingInstance.id);
+              observedContexts.should.eql(aTestModelCtx({ instance: {
+                id: getLastGeneratedUid(),
+                name: existingInstance.name,
+                extra: undefined
+              }}));
+              done();
+            });
+        });
+      }
 
       it('triggers `before save` hook when not found', function(done) {
         TestModel.observe('before save', pushContextAndNext());
@@ -1248,6 +1247,10 @@ module.exports = function(dataSource, should) {
 
     function uid() {
       lastId += 1;
+      return '' + lastId;
+    }
+
+    function getLastGeneratedUid() {
       return '' + lastId;
     }
   });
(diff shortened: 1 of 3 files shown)


bajtos:
I didn't find anything suspicious.

Please enable the following test for the connectors providing optimised version of `findOrCreate`: [`it(triggers "before save" hook when found')`](https://github.com/strongloop/loopback-datasource-juggler/blob/1f0f4b0f02ee294f13abfd530f50d793f7828197/test/persistence-hooks.suite.js#L277-L296). You can use the condition `if (dataSource.connector.findOrCreate)` to check whether the test should be run or skipped.

I'll let @raymondfeng to make the final approval.

clarkorz:
@bajtos I change the test a little. Please confirm.

bajtos:
@clarkorz your change is rather suspicious. Why is it necessary to include `.toObject()` call? Take a look at the next test `it triggers "before save" hook when not found`, which does not need `toObject` call at all.

bajtos:
Let me elaborate little bit more in a line comment.

test/persistence-hooks.suite.js
```diff
@@ -274,26 +274,29 @@ module.exports = function(dataSource, should) {
...
+              observedContexts.should.not.eql(aTestModelCtx({ instance: {
+                id: record.id,
+                name: existingInstance.name,
+                extra: undefined
+              }}));
+              record.toObject(true).should.eql({
+                id: existingInstance.id,
+                name: existingInstance.name,
+                extra: undefined
+              });
```
bajtos: The purpose of this test is to test the values passed to the hook (`observedContexts`), not the value of the argument passed to `findOrCreate` callback (`record`).

The check of `record.toObject()` belongs to a different test suite, most likely `test/manipulation.test.js`.

It may be possible to modify the instance properties checked by `observedContexts.should.eql(aTestModelCtx({ instance: {` to make this test pass. I am ok with that as long is the change makes sense. However, it is more likely that the test is not passing because of a problem in the implementation of `findOrCreate`. Especially since an almost identical test `it triggers "before save" hook when not found` below is passing.

clarkorz:
@bajtos Could you review again?  And I don't figure out why the merge build failed, it reports some git errors.

test/persistence-hooks.suite.js
```diff
@@ -274,26 +274,26 @@ module.exports = function(dataSource, should) {
...
-              extra: undefined
-            }}));
-            done();
-          });
-      });
+          var id = uid();
+
+          TestModel.findOrCreate(
+            { where: { name: existingInstance.name } },
+            { id: id, name: existingInstance.name },
```
bajtos: The tests is supposed to update an existing instance, overriding `id` is not a correct thing to do in such case.

Looking at the original test again, the `instance.id` in the before-save hook should not match `record.id`, because the `record.id` is not know at the time when the hook is triggered. Instead, the id is a new unique value that is discarded later.

IMO the test should assert the last generated id value, which is pretty easy to get via `'' + lastId`. We should also check that the `record` returned by `findOrCreate` has the correct (existing) id.

```js
TestModel.observe('before save', pushContextAndNext());

TestModel.findOrCreate(
  { where: { name: existingInstance.name } },
  { name: existingInstance.name },
  function(err, record, created) {
    if (err) return done(err);
    created.id.should.eql(existingInstance.id);
    observedContexts.should.eql(aTestModelCtx({ instance: {
      id: getLastGeneratedUid(),
      name: existingInstance.name,
      extra: undefined
    }}));
    done();
  });

// at the bottom of the file, close to `function uid()`:
function getLastGeneratedUid() {
  return '' + lastId;
}
```

Thoughts?

clarkorz: good idea

clarkorz:
@bajtos finish~

bajtos:
Thank you @clarkorz, the patch LGTM now.

@raymondfeng Could you please take a look at the boilerplate-like code in the optimized branch? I am not very familiar with this part of the juggler, thus I might have missed something.

raymondfeng:
@clarkorz I resolved the merge conflict and pushed to https://github.com/strongloop/loopback-datasource-juggler/tree/clarkorz-featrue/support-connector-findOrCreate. Can you check before we merge it into master?

clarkorz:
@raymondfeng LGTM

We might handle the `options` param further in another PR later.

PR #420, opened 2015-02-03
Create model foreign key matching type of opposite part of relation (even if it has a custom field type)

aol-nnov:
Initially this fix was proposed in strongloop/loopback-connector-postgresql#57 but @raymondfeng advised to move it into loopback-datasource-juggler.
So, here it is. :)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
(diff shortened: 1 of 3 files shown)


slnode:

aol-nnov:
@raymondfeng sorry for the delay - a bit busy here... I will add the testcase asap today or tomorrow.

aol-nnov:
@raymondfeng here it is, finally ;)

raymondfeng:
ok to test.

PR #424, opened 2015-02-04
Update dao.js

siddhant-mohan:
In function done callback was called with er object instead of err.

https://github.com/strongloop/loopback-datasource-juggler/issues/425


slnode:

raymondfeng:
Thank you for the patch. I fixed the issue before I saw this PR. See https://github.com/strongloop/loopback-datasource-juggler/commit/79eabf3cf97b5db3aaf257986f3258a984fcb4cc

PR #430, opened 2015-02-05
Return 400 when client provides an incorrect value

PradnyaBaviskar:
to / @bajtos,
Made following changes...
1) Set status code to 400, when client provides incorrect value.
2) Added test cases to verify the change.

Related: #709

Can you please review?


bajtos:
Merged, thank you for the contribution.

PR #431, opened 2015-02-05
Fix the perf around should.not.equal for complex objects

raymondfeng:
/to @bajtos 

The PR fixes the performance issue by providing a custom message to `should.not.equal` calls. The `should` module always asserts the `equal` first, creates an `AssertionError`, and then tests the `not` condition. See See https://github.com/shouldjs/should.js/blob/master/lib/assertion.js#L47-L68. For complex objects, the [getMessage](https://github.com/shouldjs/should.js/blob/master/lib/assertion.js#L202-L208) is very expensive. It takes more than 4 seconds for a `DataSource` object that has references to many models.

I'll create an issue to should.js too.

### DIFF
--- a/test/schema.test.js
+++ b/test/schema.test.js
@@ -13,16 +13,16 @@ describe('dataSource', function () {
   });
 
   it('should clone existing model', function () {
-    // Workaround for a possible bug in `should`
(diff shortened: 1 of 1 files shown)


bajtos:
:+1: :shipit:

Issue #293, opened 2014-09-06
Shouldn't Model inherit properties from Base?


fabien:
Currently, a model definition does not contain the properties of a base model:

```javascript
var User = ds.define('User', {name: String});
var Customer = ds.define('Customer', {vip: Boolean}, {base: 'User'});
console.log(Customer.definition.properties['name']); // undefined
```

PR #436, opened 2015-02-07
Make sure base properties/settings are merged into the submodel

raymondfeng:
/to @ritch or @bajtos 
/cc @fabien 

See https://github.com/strongloop/loopback-datasource-juggler/issues/293


bajtos:
LGTM.

PR #439, opened 2015-02-10
Remove workaround for perfomance degradation in should.js

btd:
it is not very important PR (so if you like how it is working now, feel free to close it), just wanted to point that perfomance degradation was fixed (i am ashamed that i have not noticed this before).


slnode:

PR #443, opened 2015-02-12
Always use findOrCreate when creating a through model

clarkorz:
fix a problem that mentioned in strongloop/loopback-connector-mongodb#92

Notes:
1. The earlier commits add `.editorconfig`, I have separated the changes to a standalone commit beside the real patch 
2. I don't see the necessary to add new tests since they have already covered, and pass tests of strongloop/loopback-connector-mongodb#92
3. to @raymondfeng , although this PR fix the id problem, but for a `HasManyThrough#create`, it does **create** a **new** model every time, hence the through model will always been **created**, it would never been **found**. So the query is unnecessary if the connector doesn't support optimized `findOrCreate`. Is that ok?

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -954,13 +954,14 @@ HasManyThrough.prototype.create = function create(data, done) {
     var fk2 = keys[1];
 
     function createRelation(to, next) {
-      var d = {};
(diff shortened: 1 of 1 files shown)


clarkorz:
@raymondfeng please review. I think current tests have already covered so I don't add new test for this PR.

PR #435, opened 2015-02-06
Add an optional `options` argument to all CRUD methods

raymondfeng:
/to @bajtos @ritch 

This PR adds an optional `options` argument to all CRUD methods to make the APIs consistent. Some of them already start to take the `options` object after @bajtos implemented the intent-based hooks.

With the `options` argument, it's now possible to pass in more flags, such as `cascadeDelete`. 

Please note the connectors are yet to support the extra `options` argument.

PR #447, opened 2015-02-16
Promises Support

partap:
Added transparent support for Promises in the loopback model methods as suggested by @bajtos in strongloop/loopback#418.

I tried to leave the existing functionality intact as much as possible.  All of the original unit tests pass with my mods.

The code for the core methods is done and I've added tests to manipulation.test.js for the promise variants.  The code for the related models is in there but is mostly untested right now.

I didn't do anything with EmbedsMany for now because most of the methods have an implicit synchronous mode when you leave off the callback.  Will probably have to make explicit async versions. 

Thanks to @raymondfeng and @shlomiassaf for helping locate code for core methods related models!

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -9,8 +9,10 @@ exports.defineCachedRelations = defineCachedRelations;
 exports.sortObjectsByIds = sortObjectsByIds;
 exports.setScopeValuesFromWhere = setScopeValuesFromWhere;
 exports.mergeQuery = mergeQuery;
+exports.createPromiseCallback = createPromiseCallback;
 
 var traverse = require('traverse');
+var Promise = require('bluebird');
 
 function safeRequire(module) {
   try {
@@ -340,3 +342,16 @@ function sortObjectsByIds(idName, ids, objects, strict) {
   
   return heading.concat(tailing);
 };
+
+function createPromiseCallback() {
+  var cb;
+  var promise = new Promise(function (resolve, reject) {
+    cb = function (err, data) {
+      if (err) return reject(err);
+      return resolve(data);
+    };
+  });
+  cb.promise = promise;
+  return cb;
+}
+
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -9,20 +9,9 @@ exports.defineCachedRelations = defineCachedRelations;
 exports.sortObjectsByIds = sortObjectsByIds;
 exports.setScopeValuesFromWhere = setScopeValuesFromWhere;
 exports.mergeQuery = mergeQuery;
-exports.createPromiseCallback = createPromiseCallback;
+exports.createPromiseCallback = createPromiseCallback
 
 var traverse = require('traverse');
-var debug = require('debug')('loopback:utils');
-
-if (typeof Promise === 'undefined') {
-  try {
-    var Promise = require('bluebird');
-  } catch (e) {
-    console.log('Run npm install bluebird to enable Promises in loopback.')
-  }
-}
-
-
 
 function safeRequire(module) {
   try {
@@ -78,7 +67,7 @@ function mergeQuery(base, update, spec) {
   }
   spec = spec || {};
   base = base || {};
-
+  
   if (update.where && Object.keys(update.where).length > 0) {
     if (base.where && Object.keys(base.where).length > 0) {
       base.where = {and: [base.where, update.where]};
@@ -97,38 +86,38 @@ function mergeQuery(base, update, spec) {
       base.include[update.include] = saved;
     }
   }
-
+  
   if (spec.collect !== false && update.collect) {
     base.collect = update.collect;
   }
-
+  
   // Overwrite fields
   if (spec.fields !== false && update.fields !== undefined) {
     base.fields = update.fields;
   } else if (update.fields !== undefined) {
     base.fields = [].concat(base.fields).concat(update.fields);
   }
-
+  
   // set order
   if ((!base.order || spec.order === false) && update.order) {
     base.order = update.order;
   }
-
+  
   // overwrite pagination
   if (spec.limit !== false && update.limit !== undefined) {
     base.limit = update.limit;
   }
-
+  
   var skip = spec.skip !== false && spec.offset !== false;
-
+  
   if (skip && update.skip !== undefined) {
     base.skip = update.skip;
   }
-
+  
   if (skip && update.offset !== undefined) {
     base.offset = update.offset;
   }
-
+  
   return base;
 }
 
@@ -323,16 +312,16 @@ function sortObjectsByIds(idName, ids, objects, strict) {
   ids = ids.map(function(id) {
       return (typeof id === 'object') ? String(id) : id;
   });
-
+  
   var indexOf = function(x) {
     var isObj = (typeof x[idName] === 'object'); // ObjectID
     var id = isObj ? String(x[idName]) : x[idName];
     return ids.indexOf(id);
   };
-
+  
   var heading = [];
   var tailing = [];
-
+  
   objects.forEach(function(x) {
     if (typeof x === 'object') {
       var idx = indexOf(x);
@@ -340,7 +329,7 @@ function sortObjectsByIds(idName, ids, objects, strict) {
       idx === -1 ? tailing.push(x) : heading.push(x);
     }
   });
-
+  
   heading.sort(function(x, y) {
     var a = indexOf(x);
     var b = indexOf(y);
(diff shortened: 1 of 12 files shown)


slnode:

partap:
I used petkaantonov/bluebird because that's what I'm most familiar with, but I tried to keep the use as simple as possible so it will work with other promises/a+ libraries as well.  

Might want to use a smaller library or make the bluebird dependency optional for non-promise users...

slnode:

partap:
oops.

partap:
Oh wow, I didn't even realize that Promises were native in io.js until just now...

partap:
A lot of problems with the scope access... I will upload a new patch shortly.

partap:
More relation tests included now.  HasManyThrough relation was using "done" instead of "cb" so I missed promisifying those methods the first time.

Looking good now, for the most part...I'm done testing hasMany and hasManyThrough.

Also, a bunch of my tests were broken before because they didn't explicitly return the nested promises.  Coffeescript making me lazy I guess :-/

lib/utils.js
```diff
@@ -9,8 +9,10 @@ exports.defineCachedRelations = defineCachedRelations;
 exports.sortObjectsByIds = sortObjectsByIds;
 exports.setScopeValuesFromWhere = setScopeValuesFromWhere;
 exports.mergeQuery = mergeQuery;
+exports.createPromiseCallback = createPromiseCallback;
 
 var traverse = require('traverse');
+var Promise = require('bluebird');
```
raymondfeng: Can we check if the global `Promise` exists (`typeof Promise !== undefined`) and use `bluebird` as the fallback? For example, iojs already supports `Promise` natively.

raymondfeng:
@partap Great effort! Thank you for contributing. It looks good in general. I'll leave the in-depth review to @bajtos.

shlomiassaf:
@partap +1

pandaiolo:
Nice, I'll definitely use this +1 @partap

shlomiassaf:
@partap, @raymondfeng 

I will take this as an example, but it might exist elsewhere.

The original SL code assumes a CB parameter is mandatory.
It must be the last  parameter, last from total of 2 to 4, but it must be.

Adding `Promise` support changes the assumption hence might change the logic.

Look [here](https://github.com/partap/loopback-datasource-juggler/blob/5cadb2057644de7e403fdf1db50b7c15838f4d92/lib/dao.js#L1500-L1513)

```
DataAccessObject.update =
DataAccessObject.updateAll = function (where, data, options, cb) {
  if (stillConnecting(this.getDataSource(), this, arguments)) return;

  assert(arguments.length >= 1, 'At least one argument is required');
  if (options === undefined && cb === undefined) {
    if (typeof data === 'function') {
      // updateAll(data, cb);
      cb = data;
      data = where;
      where = {};
    }
  } else if (cb === undefined) {
    if (typeof options === 'function') {
      // updateAll(query, data, cb)
      cb = options;
      options = {};
    }
  }

  data = data || {};
  options = options || {};

  assert(typeof where === 'object', 'The where argument must be an object');
  assert(typeof data === 'object', 'The data argument must be an object');
  assert(typeof options === 'object', 'The options argument must be an object');

```

Now there is a case where  `(options === undefined && cb === undefined)`
But also __data ===  undefined__

This situation is not handled in the logic since it was not possible.
Now, after the `assert` change it is possible.

Now, if this situation happen it will cause `data` to be undefined, luckily fail the double check in the next assert test.

You need to handle it.

lib/utils.js
bajtos: +1, use global `Promise` when it's available. See https://github.com/strongloop/strong-remoting/pull/179

bajtos: Actually, I don't like the idea of shipping "bluebird" with loopback-datasource-juggler. We should let the user to decide what promise implementation he wants to use and always use `global.Promise`.

There are several options how can `createPromiseCallback` handle the case when `global.Promise` is not defined:
 1. Throw an error explaining the problem. Problem: this is a breaking change, as `User.create(data)` will now throw even though the user does not care about the result of the operation.
 2. Return a no-op callback (same as now) without a promise. Problem: user will get a runtime error while accessing `.then` of undefined, it is totally not clear how to fix the problem.
 3. Return a no-op callback with a dummy promise-like object that throws when `.then` or `.catch` is called.

I am proposing to go with the third option, as it is backward compatible and allows us to print a helpful error message.

A mock implementation:

```js
function createPromiseCallback() {
  var cb;

  if (!global.Promise) {
    cb = function(){};
    cb.promise = {};
    Object.defineProperty(cb.promise, 'then', { get: throwPromiseNotDefined });
    Object.defineProperty(cb.promise, 'catch', { get: throwPromiseNotDefined });
    return cb;
  }

  var promise = new Promise(function (resolve, reject) {
    cb = function (err, data) {
      if (err) return reject(err);
      return resolve(data);
    };
  });
  cb.promise = promise;
  return cb;
}

function throwPromiseNotDefined() {
  throw new Error(
    'Your Node runtime does support ES6 Promises. ' +
    'Set "global.Promise" to your preferred implementation of promises.');
}
```

bajtos:
@pandaiolo Thank you for the pull request, you did a great job!

The change set is huge and thus challenging to review. Can you please split this patch into multiple smaller pull requests? For example, the first PR can add promise support to DAO methods and include infrastructure changes like `util. createPromiseCallback`. The second can add promises to datasource methods, the third to relations, etc. It will also make it easier for us to split the task of reviewing the changes in detail, e.g. I am comfortable with reviewing DAO methods, but I need another pair of eyes to review relation methods, because I am not familiar with them.

lib/utils.js
bajtos: #449 was landed, please rebase on top of the current master and remove the infrastructure changes that are no longer needed.

partap: Should I take bluebird out of the npm dependencies?  Or add it to optionalDependencies, maybe?

bajtos: Take it out of npm dependencies, see #449.

partap:
@raymondfeng Thanks! 

I've just wanted promise support in this library for so long, especially for related models... It was actually relatively easy once I knew where to look in the codebase.  Most methods only needed 2-3 lines added/changed.  
Testing was time-consuming, and revealed some special cases, but actually stepping through them and reading the code has given me a deeper understanding of how the library works behind the scenes, which is a plus for me.

partap:
Hmm...github is saying there are merge conflicts now....

I just pulled master ... 

@bajtos It looks like we both modified utils.js for native Promise support in different ways.  I resolved it with your version, so it should be ok now... hope it didn't make too much of a mess in the commits.

partap:
@bajtos 
I just noticed your comment above about splitting this into multiple pull requests...  

I'll see what I can do...might be a bit complicated.
I guess the way to go would be to reset my local repo to the current master, and commit my working directory changes into multiple new branches?  The related models would need to branch off the DataAccessObject branch...

partap:
@shlomiassaf Hmm, looks like you're right...thanks!  There wasn't a test for that case.

And yes, there were several methods with this type of argument resolution issue... I got all the ones that were flagged by the unit tests, but it's possible there are more that I missed...

partap:
@shlomiassaf @raymondfeng 
I rewrote the section based on `arguments.length` rather than `*arg* === undefined`.  I think it's easier to follow that way.  Also I think it might better handle the case where an undefined variable is passed in by mistake , e.g.:
```
MyModel.updateAll(where, data)
```
if  `data` is undefined, and I was just checking for `data === undefined && options === undefined && cb === undefined`, it  could overwrite the entire dataset with the `where` values...

raymondfeng:
@partap We check `undefined` instead of `arguments.length` by intention. See @bajtos's comment in https://github.com/strongloop/loopback-datasource-juggler/pull/435

partap:
@raymondfeng Hmm... I see what you mean.  I actually ran into that exact problem in one of the scope methods that was using `arguments.length`.

This particular method is tricky though... 3 possible object args with the only required one in the middle.
I could switch it back to checking `undefined`, but with a special `arguments.length` check for that potentially disastrous situation.

partap:
@bajtos @raymondfeng 
I rebased and separated the code for DAO and related models... closing this PR now.

bajtos:
> Hmm...github is saying there are merge conflicts now....
> I just pulled master ...

When pulling changes from master, you need to rebase your branch on top of the master (`git pull --rebase && git push -f`). By default, `git pull` runs merge and creates a merge commit, which messes up the commit history.

> I'll see what I can do...might be a bit complicated.
> I guess the way to go would be to reset my local repo to the current master, and commit my working directory changes into multiple new branches? The related models would need to branch off the DataAccessObject branch...

Here is what I would do: create a new feature branch for DAO, base it on top of the current master from strongloop/loopback-datasource-juggler. Then you can run `git cherry-pick {sha}` to copy the commits relevant to DAO from this all-in-one branch into your new per/area branch.

partap:
@bajtos 
Yeah, I figured it out...just make sure nobody bases anything on *this* branch, I guess  :)

New PRs:
https://github.com/strongloop/loopback-datasource-juggler/issues/451
https://github.com/strongloop/loopback-datasource-juggler/issues/452

partap:
@raymondfeng 
1. Yeah, my bad...I saw the typo right after I pushed it.  It's fixed now.
2. The extra check is to guard against the case of `updateAll(where, data)`, where `data` is an uninitialized variable.  I'm thinking it would be preferable for a bug in somebody's code to do nothing rather than accidentally overwrite the entire data set with the `where` conditions...

partap:
Actually, I could put the same check into the second case:
`if (typeof data === 'function' && arguments.length === 2)`

I'm done with this branch though...

partap:
I've got a new rebased branch that contains the full Promises patch (DAO and related). 
https://github.com/partap/loopback-datasource-juggler/tree/promises-rebase
Any updates will be propagated through the separate DAO and Related branches for their respective PRs....

PR #449, opened 2015-02-17
Promises for operation hooks

bajtos:
- Allow the observer functions passed to `ModelBaseClass.observe` to return a promise instead of calling the callback.

- Support both promise and callback styles in `ModelBaseClass.notifyObserversOf`.

Related: https://github.com/strongloop/strong-remoting/pull/179 and https://github.com/strongloop/loopback-datasource-juggler/pull/447
Connect strongloop/loopback#418

/to @raymondfeng Please review.
/cc @partap This patch provides the infrastructure needed for promise support, it's based on my comments in your PR. My plan is to land this PR first, so that you can use this infrastructure in your PR.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -567,7 +567,15 @@ ModelBaseClass.notifyObserversOf = function(operation, context, callback) {
 
     async.eachSeries(
       observers,
-      function(fn, next) { fn(context, next); },
+      function notifySingleObserver(fn, next) {
+        var retval = fn(context, next);
+        if (retval && typeof retval.then === 'function') {
+          retval.then(
+            function() { next(); },
+            next // error handler
+          );
+        }
+      },
       function(err) { callback(err, context) }
     );
   });
--- a/lib/model.js
+++ b/lib/model.js
@@ -561,6 +561,8 @@ ModelBaseClass.observe = function(operation, listener) {
 ModelBaseClass.notifyObserversOf = function(operation, context, callback) {
   var observers = this._observers && this._observers[operation];
 
+  if (!callback) callback = utils.createPromiseCallback();
+
   this._notifyBaseObservers(operation, context, function doNotify(err) {
     if (err) return callback(err, context);
     if (!observers || !observers.length) return callback(null, context);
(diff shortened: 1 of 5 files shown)


raymondfeng:
LGTM

ritch:
@bajtos are you updating the docs / examples (atleast creating examples) for this?

bajtos:
@ritch Good catch, I added a task to strongloop/loopback#418 to update docs & examples once we have all pieces in place.

ctcampb3:
@bajtos, I see that [Using promises](https://loopback.io/doc/en/lb2/Using-promises.html#loopback-support-for-promises) says you can use promises with [Operation hooks](https://loopback.io/doc/en/lb2/Operation-hooks.html). But I'm missing where the example is. It took me a few days to figure out that you don't need to return next() in the promise chain. Is there any way we could have an example out there so others don't go through the same struggle?

bajtos:
@ctcampb3 I am sorry to hear that you lost so much time trying to figure this out :( Perhaps the following example would be a good one to update?
https://github.com/strongloop/loopback-example-app-logic#define-a-model-operation-hook

BTW I think we have a similar problem in remoting hooks too, see https://github.com/strongloop/strong-remoting/issues/406#issuecomment-314127151 and https://github.com/strongloop/strong-remoting/issues/406#issuecomment-313516324. It may be possible to improve our code to detect the situation where `next` was called and a promise was returned too and warn the user.

ctcampb3:
@bajtos , I think those issues look pretty similar. For the example I thought it might be good to have it on the [Using promises](https://loopback.io/doc/en/lb2/Using-promises.html#loopback-support-for-promises) page under the **LoopBack support for promises** section. Having an example with a warning here would make it more clear. 
Something generic like:

When returning a Promise instead of a callback the callback function should not be called at all.  For example:
```
MyModel.observe('before save', function filterProperties(ctx, next) {
  // Resolving with a callback
  next();
});
MyModel.observe('before save', function filterProperties(ctx, next) {
  // Resolving with a promise
  return Promise.resolve();
});
```

bajtos:
@ctcampb3 makes sense. Would you mind contributing this doc improvement yourself? See this page to get you started: http://loopback.io/doc/en/contrib/doc-contrib.html Please mention my handle in the pull request to make sure I notice it.

PR #451, opened 2015-02-18
Promisify DAO

partap:
Rebased and isolated DAO changes from https://github.com/strongloop/loopback-datasource-juggler/pull/447

Connects strongloop/loopback#418

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -148,7 +148,10 @@ function noCallback(err, result) {
  *   - instance (null or Model)
  */
 DataAccessObject.create = function (data, options, cb) {
-  if (stillConnecting(this.getDataSource(), this, arguments)) return;
+  var connectionPromise = stillConnecting(this.getDataSource(), this, arguments);
+  if (connectionPromise) {
+    return connectionPromise;
+  }
 
   var Model = this;
   var self = this;
@@ -167,9 +170,9 @@ DataAccessObject.create = function (data, options, cb) {
     }
   }
 
-  cb = cb || noCallback;
   data = data || {};
   options = options || {};
+  cb = cb || (Array.isArray(data) ? noCallback : utils.createPromiseCallback());
 
   assert(typeof data === 'object', 'The data argument must be an object or array');
   assert(typeof options === 'object', 'The options argument must be an object');
@@ -276,12 +279,27 @@ DataAccessObject.create = function (data, options, cb) {
     }, obj, cb);
   }
 
+  // Does this make any sense? How would chaining be used here? -partap
+
   // for chaining
-  return obj;
+  return cb.promise || obj;
 };
 
 function stillConnecting(dataSource, obj, args) {
-  return dataSource.ready(obj, args);
+  if (typeof args[args.length-1] === 'function') {
+    return dataSource.ready(obj, args);
+  }
+
+  // promise variant
+  var promiseArgs = Array.prototype.slice.call(args);
+  promiseArgs.callee = args.callee
+  var cb =  utils.createPromiseCallback();
+  promiseArgs.push(cb);
+  if (dataSource.ready(obj, promiseArgs)) {
+    return cb.promise;
+  } else {
+    return false;
+  }
 }
 
 /**
@@ -298,8 +316,9 @@ function stillConnecting(dataSource, obj, args) {
 // 'upsert' will be used as the name for strong-remoting to keep it backward
 // compatible for angular SDK
 DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data, options, cb) {
-  if (stillConnecting(this.getDataSource(), this, arguments)) {
-    return;
+  var connectionPromise = stillConnecting(this.getDataSource(), this, arguments);
+  if (connectionPromise) {
+    return connectionPromise;
   }
 
   if (options === undefined && cb === undefined) {
@@ -316,7 +335,7 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
     }
   }
 
-  cb = cb || noCallback;
+  cb = cb || utils.createPromiseCallback();
   data = data || {};
   options = options || {};
 
@@ -404,6 +423,7 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
       });
     }
   }
+  return cb.promise;
 };
 
 /**
@@ -418,10 +438,19 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
  * @param {Function} cb Callback called with (err, instance, created)
  */
 DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) {
-  if (stillConnecting(this.getDataSource(), this, arguments)) return;
+  var connectionPromise = stillConnecting(this.getDataSource(), this, arguments);
+  if (connectionPromise) {
+    return connectionPromise;
+  }
 
-  assert(arguments.length >= 2, 'At least two arguments are required');
-  if (options === undefined && cb === undefined) {
+  assert(arguments.length >= 1, 'At least one argument is required');
+  if (data === undefined && options === undefined && cb === undefined) {
+    assert(typeof query === 'object', 'Single argument must be data object');
+    // findOrCreate(data);
+    // query will be built from data, and method will return Promise
+    data = query;
+    query = {where: data};
+  } else  if (options === undefined && cb === undefined) {
     if (typeof data === 'function') {
       // findOrCreate(data, cb);
       // query will be built from data
@@ -437,7 +466,7 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
     }
   }
 
-  cb = cb || noCallback;
+  cb = cb || utils.createPromiseCallback();
   query = query || {where: {}};
   data = data || {};
   options = options || {};
@@ -466,11 +495,19 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
         if (created) {
           Model.notifyObserversOf('after save', { Model: Model, instance: obj },
             function(err) {
-              cb(err, obj, created);
+              if (cb.promise) {
+                cb(err, [obj, created]);
+              } else {
+                cb(err, obj, created);
+              }
               if (!err) Model.emit('changed', obj);
             });
         } else {
-          cb(err, obj, created);
+          if (cb.promise) {
+            cb(err, [obj, created]);
+          } else {
+            cb(err, obj, created);
+          }
         }
       });
   }
@@ -481,9 +518,10 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
     try {
       this._normalize(query);
     } catch (err) {
-      return process.nextTick(function () {
+      process.nextTick(function () {
         cb(err);
       });
+      return cb.promise;
     }
 
     this.applyScope(query);
@@ -521,12 +559,23 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
   } else {
     Model.findOne(query, options, function (err, record) {
       if (err) return cb(err);
-      if (record) return cb(null, record, false);
+      if (record) {
+        if (cb.promise) {
+          return cb(null, [record, false]);
+        } else {
+          return cb(null, record, false);
+        }
+      }
       Model.create(data, options, function (err, record) {
-        cb(err, record, record != null);
+        if (cb.promise) {
+          cb(err, [record, record != null]);
+        } else {
+          cb(err, record, record != null);
+        }
       });
     });
   }
+  return cb.promise;
 };
 
 /**
@@ -537,7 +586,10 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
  * @param {Function} cb Callback function called with (err, exists: Bool)
  */
 DataAccessObject.exists = function exists(id, options, cb) {
-  if (stillConnecting(this.getDataSource(), this, arguments)) return;
+  var connectionPromise = stillConnecting(this.getDataSource(), this, arguments);
+  if (connectionPromise) {
+    return connectionPromise;
+  }
 
   assert(arguments.length >= 1, 'The id argument is required');
   if (cb === undefined) {
@@ -548,7 +600,7 @@ DataAccessObject.exists = function exists(id, options, cb) {
     }
   }
 
-  cb = cb || noCallback;
+  cb = cb || utils.createPromiseCallback();
   options = options || {};
 
   assert(typeof options === 'object', 'The options argument must be an object');
@@ -559,10 +611,11 @@ DataAccessObject.exists = function exists(id, options, cb) {
       cb(err, err ? false : count === 1);
     });
   } else {
-    return process.nextTick(function() {
+    process.nextTick(function() {
       cb(new Error('Model::exists requires the id argument'));
     });
   }
+  return cb.promise;
 };
 
 /**
@@ -580,7 +633,10 @@ DataAccessObject.exists = function exists(id, options, cb) {
  * @param {Function} cb Callback called with (err, instance)
  */
 DataAccessObject.findById = function find(id, options, cb) {
-  if (stillConnecting(this.getDataSource(), this, arguments)) return;
+  var connectionPromise = stillConnecting(this.getDataSource(), this, arguments);
+  if (connectionPromise) {
+    return connectionPromise;
+  }
 
   assert(arguments.length >= 1, 'The id argument is required');
   if (cb === undefined) {
@@ -590,18 +646,20 @@ DataAccessObject.findById = function find(id, options, cb) {
       options = {};
     }
   }
-  cb = cb || noCallback;
+  cb = cb || utils.createPromiseCallback();
   options = options || {};
 
   assert(typeof options === 'object', 'The options argument must be an object');
   assert(typeof cb === 'function', 'The cb argument must be a function');
 
   if (id == null || id === '') {
-    return process.nextTick(function() {
+    process.nextTick(function() {
       cb(new Error('Model::findById requires the id argument'));
     });
+  } else {
+    this.findOne(byIdQuery(this, id), options, cb);
   }
-  this.findOne(byIdQuery(this, id), options, cb);
+  return cb.promise;
 };
 
 /**
@@ -626,7 +684,7 @@ DataAccessObject.findByIds = function(ids, query, options, cb) {
     }
   }
 
-  cb = cb || noCallback;
+  cb = cb || utils.createPromiseCallback();
   options = options || {};
   query = query || {};
 
@@ -638,7 +696,7 @@ DataAccessObject.findByIds = function(ids, query, options, cb) {
   var pk = idName(this);
   if (ids.length === 0) {
     process.nextTick(function() { cb(null, []); });
-    return;
+    return cb.promise;;
   }
 
   var filter = { where: {} };
@@ -647,6 +705,7 @@ DataAccessObject.findByIds = function(ids, query, options, cb) {
   this.find(filter, options, function(err, results) {
     cb(err, err ? results : utils.sortObjectsByIds(pk, ids, results));
   });
+  return cb.promise;
 };
 
 function convertNullToNotFoundError(ctx, cb) {
@@ -662,7 +721,7 @@ function convertNullToNotFoundError(ctx, cb) {
 
 // alias function for backwards compat.
 DataAccessObject.all = function () {
-  DataAccessObject.find.apply(this, arguments);
+  return DataAccessObject.find.apply(this, arguments);
 };
 
 var operators = {
@@ -994,7 +1053,10 @@ DataAccessObject._coerce = function (where) {
  */
 
 DataAccessObject.find = function find(query, options, cb) {
-  if (stillConnecting(this.getDataSource(), this, arguments)) return;
+  var connectionPromise = stillConnecting(this.getDataSource(), this, arguments);
+  if (connectionPromise) {
+    return connectionPromise;
+  }
 
   if (options === undefined && cb === undefined) {
     if (typeof query === 'function') {
@@ -1010,7 +1072,7 @@ DataAccessObject.find = function find(query, options, cb) {
     }
   }
 
-  cb = cb || noCallback;
+  cb = cb || utils.createPromiseCallback();
   query = query || {};
   options = options || {};
 
@@ -1023,9 +1085,10 @@ DataAccessObject.find = function find(query, options, cb) {
   try {
     this._normalize(query);
   } catch (err) {
-    return process.nextTick(function () {
+    process.nextTick(function () {
       cb(err);
     });
+    return cb.promise;
   }
 
   this.applyScope(query);
@@ -1073,7 +1136,7 @@ DataAccessObject.find = function find(query, options, cb) {
       });
 
       // already handled
-      return;
+      return cb.promise;
     }
   }
 
@@ -1135,6 +1198,7 @@ DataAccessObject.find = function find(query, options, cb) {
       self.getDataSource().connector.all(self.modelName, query, allCb);
     });
   }
+  return cb.promise;
 };
 
 /**
@@ -1146,7 +1210,10 @@ DataAccessObject.find = function find(query, options, cb) {
  * @param {Function} cb Callback function called with (err, instance)
  */
 DataAccessObject.findOne = function findOne(query, options, cb) {
-  if (stillConnecting(this.getDataSource(), this, arguments)) return;
+  var connectionPromise = stillConnecting(this.getDataSource(), this, arguments);
+  if (connectionPromise) {
+    return connectionPromise;
+  }
 
   if (options === undefined && cb === undefined) {
     if (typeof query === 'function') {
@@ -1160,7 +1227,7 @@ DataAccessObject.findOne = function findOne(query, options, cb) {
     }
   }
 
-  cb = cb || noCallback;
+  cb = cb || utils.createPromiseCallback();
   query = query || {};
   options = options || {};
 
@@ -1173,6 +1240,7 @@ DataAccessObject.findOne = function findOne(query, options, cb) {
     if (err || !collection || !collection.length > 0) return cb(err, null);
     cb(err, collection[0]);
   });
+  return cb.promise;
 };
 
 /**
@@ -1190,7 +1258,10 @@ DataAccessObject.findOne = function findOne(query, options, cb) {
  * @param {Function} [cb] Callback called with (err)
  */
 DataAccessObject.remove = DataAccessObject.deleteAll = DataAccessObject.destroyAll = function destroyAll(where, options, cb) {
-  if (stillConnecting(this.getDataSource(), this, arguments)) return;
+  var connectionPromise = stillConnecting(this.getDataSource(), this, arguments);
+  if (connectionPromise) {
+    return connectionPromise;
+  }
 
   var Model = this;
 
@@ -1206,7 +1277,7 @@ DataAccessObject.remove = DataAccessObject.deleteAll = DataAccessObject.destroyA
     }
   }
 
-  cb = cb || noCallback;
+  cb = cb || utils.createPromiseCallback();
   where = where || {};
   options = options || {};
 
@@ -1267,6 +1338,7 @@ DataAccessObject.remove = DataAccessObject.deleteAll = DataAccessObject.destroyA
       });
     }
   }
+  return cb.promise;
 };
 
 function whereIsEmpty(where) {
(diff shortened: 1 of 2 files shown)


slnode:

lib/dao.js
```diff
@@ -205,9 +205,11 @@ DataAccessObject.create = function (data, options, cb) {
         }
         data[i] = results[i].result;
       }
-      cb(errors, data);
+      if (cb.promise) {
+        cb(null, data);
+      } else cb(errors, data);
```
bajtos: This does not look right, you are basically swallowing errors in the promise case.

In the original PR, I was suggesting to drop support for the array version of `create` in the promise mode. Later, I have realised that such change can break the existing code calling the array version of create with no callback.

Here is a better solution: when in the array and promise mode, build a single array where each item is either the model instance created or the error object when create failed. If there was at least one error, then reject the promise. If all models were created successfully, then resolve the promise with the array.

Please make sure this case is covered by unit-tests (I did not check the test code yet).

Thoughts?

lib/dao.js
```diff
@@ -590,18 +598,19 @@ DataAccessObject.findById = function find(id, options, cb) {
...
   assert(typeof cb === 'function', 'The cb argument must be a function');
 
   if (id == null || id === '') {
-    return process.nextTick(function() {
+    process.nextTick(function() {
       cb(new Error('Model::findById requires the id argument'));
     });
   }
-  this.findOne(byIdQuery(this, id), options, cb);
+  else this.findOne(byIdQuery(this, id), options, cb);
```
bajtos: Coding style: open a new block and put `this.findOne` on a new line. `else` is on the same line as the closing brace of the previous block.

```js
} else {
  this.findOne(...);
}
```

lib/dao.js
```diff
@@ -1477,25 +1496,34 @@ DataAccessObject.update =
 DataAccessObject.updateAll = function (where, data, options, cb) {
   if (stillConnecting(this.getDataSource(), this, arguments)) return;
 
-  assert(arguments.length >= 2, 'At least two arguments are required');
-  if (options === undefined && cb === undefined) {
+  assert(arguments.length >= 1, 'At least one argument is required');
+
+  if (data === undefined && options === undefined && cb === undefined && arguments.length === 1) {
```
bajtos: Can we drop `arguments.length === 1`? Why is that check necessary?

test/manipulation.test.js
```diff
@@ -57,6 +57,20 @@ describe('manipulation', function () {
       });
     });
 
+    it('should create instance (promise variant)', function (done) {
+      Person.create({name: 'Anatoliy'})
+      .then (function (p) {
```
bajtos: Coding style: indent `.then` by two spaces, no space between opening brace.

```js
Person.create(...)
  .then(function(p) {
     // ...
  })
```

test/manipulation.test.js
```diff
@@ -57,6 +57,20 @@ describe('manipulation', function () {
...
+        p.name.should.equal('Anatoliy');
+        should.exist(p);
+        return Person.findById(p.id)
+        .then (function (person) {
+          person.id.should.eql(p.id);
+          person.name.should.equal('Anatoliy');
+          done();
+        });
+      }).catch(done);
```
bajtos: Coding style: place `.catch` on a new line.

```js
Person.create(...)
  .then(function(p) {
    // ...
  })
  .catch(done);
```

test/manipulation.test.js
```diff
@@ -69,6 +83,19 @@ describe('manipulation', function () {
...
+      var p = new Person({name: 'Anatoliy'});
+      p.name.should.equal('Anatoliy');
+      p.isNewRecord().should.be.true;
+      p.save()
+      .then (function(inst) {
+        inst.isNewRecord().should.be.false;
+        inst.should.equal(p);
+        done();
+      }).catch(done);
+
```
bajtos: Extra empty line, please remove.

test/manipulation.test.js
bajtos: Please fix this in all other tests too.

test/manipulation.test.js
bajtos: Please fix this in all other tests too.

test/manipulation.test.js
```diff
@@ -91,6 +118,20 @@ describe('manipulation', function () {
...
+      .then (function (p) {
+        should.not.exist(p);
+        done();
+      })
+      .catch(function (err) {
+        err.should.be.instanceof(ValidationError);
+        err.statusCode.should.equal(422);
+        err.details.messages.id.should.eql(['can\'t be set']);
+        done();
+      });
```
bajtos: This is weird. Are you expecting that `Person.create` succeeds or fails? Looking at the test above, the operation should fail. In which case you need to assert in the `then` block that the operation should have failed.

```js
Person.create({ id: 123456 })
  .then(function(p) { 
    return done(new Error('Operation should have failed.'));
  })
  .catch(function(err) {
    // ...
  });
```

Also have you seen this test fail? AFAIK, `catch` converts any errors thrown from the handler into a promise rejection, thus you need to add one more `catch` handler at the end. And since `then` accepts both success and failure handlers, 
the code can look like this:


```
Person.create({ id: 123456 })
  .then(
    function(p) { done(new Error('Person.create should have failed.')); },
    function(err) {
      // asserts
    })
  .catch(done);
```

test/manipulation.test.js
```diff
@@ -104,6 +145,22 @@ describe('manipulation', function () {
...
+      .then (function(inst) {
+        should.not.exist(inst);
+        done();
+      })
+      .catch(function (err) {
+        err.should.be.instanceof(ValidationError);
+        err.statusCode.should.equal(422);
+        err.details.messages.id.should.eql(['can\'t be set']);
+        done();
+      });
```
bajtos: The same problem as described in my [comment above](#r25005283)

test/manipulation.test.js
```diff
@@ -193,6 +297,29 @@ describe('manipulation', function () {
...
+      ];
+      Person.create(batch)
+      .then(function(ps) {
+        should.exist(ps);
+        ps.should.be.instanceOf(Array);
+        ps.should.have.lengthOf(batch.length);
+        ps[0].should.be.eql({id: 'a', name: 'A'});
+        done();
+      }).catch(done);
+    });
```
bajtos: The `beforeCreate` hook is not really specific to the callback vs. promise implementation, right? In which case I prefer to omit this promise-variant test.

test/manipulation.test.js
```diff
@@ -215,6 +342,29 @@ describe('manipulation', function () {
...
+            id: created.id,
+            name: 'a-name',
+            gender: undefined
+          });
+          done();
+        });
+      }).catch(done);
+
+    });
+
```
bajtos: Same here, "preserve properties with undefined value" is not related to callback vs. promise, can we omit this test please?

test/manipulation.test.js
```diff
@@ -261,6 +435,30 @@ describe('manipulation', function () {
...
+        .catch(function (err) {
+          should.exist(err);
+          p.save({validate: false})
+          .then(function (d) {
+            should.exist(d);
+            done();
+          }).catch(done);
+        });
+      });
+    });
```
bajtos: The same problem as described in my [comment above](#r25005283), if the operation is supposed to fail, then the `.then()` block should assert on that.

test/manipulation.test.js
```diff
@@ -293,6 +491,24 @@ describe('manipulation', function () {
...
+        return created.save()
+        .then(function(saved) {
+          saved.password.should.equal('bar-2');
+          return StubUser.findById(created.id)
+          .then(function(found) {
+            found.password.should.equal('bar-2');
+            done();
+          });
+        });
+      }).catch(done);
```
bajtos: "preserve properties with dynamic setter" is not related to callback vs. promise, can we omit this test?

test/manipulation.test.js
```diff
@@ -329,12 +557,24 @@ describe('manipulation', function () {
...
+    it('should ignore undefined values on updateAttributes (promise variant)', function(done) {
+      person.updateAttributes({'name': 'John', age: undefined})
+      .then(function(p) {
+        return Person.findById(p.id)
+        .then(function(p) {
+          p.name.should.equal('John');
+          p.age.should.equal(15);
+          done();
+        });
+      }).catch(done);
```
bajtos: "ignore undefined values" is not related to callback vs. promise, can we omit this test?

test/manipulation.test.js
```diff
@@ -357,6 +609,18 @@ describe('manipulation', function () {
...
+      StubUser.updateOrCreate({ id: 'newid', password: 'foo' })
+      .then(function(created) {
+        created.password.should.equal('foo-1');
+        return StubUser.findById(created.id)
+        .then(function(found) {
+          found.password.should.equal('foo-1');
+          done();
+        });
+      }).catch(done);
+    });
```
bajtos: "preserve properties with dynamic setter" is not related to callback vs. promise, can we omit this test?

test/manipulation.test.js
```diff
@@ -373,6 +637,22 @@ describe('manipulation', function () {
...
+        .then(function(updated) {
+          updated.password.should.equal('bar-2');
+          return StubUser.findById(created.id)
+          .then(function(found) {
+            found.password.should.equal('bar-2');
+            done();
+          });
+        });
+      }).catch(done);
+    });
```
bajtos: "preserve properties with dynamic setter" is not related to callback vs. promise, can we omit this test?

test/manipulation.test.js
```diff
@@ -397,6 +677,27 @@ describe('manipulation', function () {
...
+        .then(function(updated) {
+          updated.toObject().should.have.properties({
+            id: instance.id,
+            name: 'updated name 2',
+            gender: undefined
+          });
+          done();
+        });
+      }).catch(done);
+    });
```
bajtos: "preserve properties with undefined value" is not related to callback vs. promise, can we omit this test?

bajtos:
Hmm, this change has one major flaw. Most of the methods starts with checking whether the we are connected to the database, and if we are not, then the operation is deferred until the connection is established. When that happens, all methods return undefined at the moment.

```
if (stillConnecting(this.getDataSource(), this, arguments)) return;
```

We need to come up with a way how to return a promise in this case too.

Something like the following:

```js
if (var p = stillConnecting(this.getDataSource(), this, arguments)) return p;
```

And inside `stillConnecting`:

```js
function stillConnecting() {
  if (typeof arguments[argument.length-1] === 'function') {
    return dataSource.ready(obj, arguments);
  }

  // promise variant
  var args = Array.prototype.slice.call(arguments);
  var cb =  utils.createPromiseCallback();
  args.push(cb);
  dataSource.ready(obj, args);
  return cb.promise;
}
```

lib/dao.js
partap: This was the "partial success" case I mentioned...  At the time I just
decided to always resolve the promise, and then you would need to check the
error property of each result (which is what happens in the unit test).  I
don't really like it either...

I hadn't gotten around to changing the code since your last comment, but I
think you were correct in saying to drop promise support for the array
version.  (It shouldn't break existing code)

Actually, I think array creation with promises would be better handled at a
higher level, where the application developer can decide on the behavior.
e.g.:  with bluebird, you could do something like either
Promise.all(myArray).map(MyModel.create).all() or
Promise.all(myArray).map(MyModel.create).settled(), depending on your
requirements...

lib/dao.js
partap: contrived case:

var where, data;
where = {customerType: 'Educational'};
// data = {discount: educationalDiscount};

// .... other stuff

Customer.updateAll(where, data)
.then(...)

The data assignment was accidentally left commented out by a developer
while they experimented with other values or something. When it is run,
instead of throwing an error or doing nothing at all, it treats where as
data and overwrites the entire dataset.

I admit, it's not Promise-related, and you can't protect every developer
from every stupid thing they do, but it's not immediately obvious what's
happened because 2 values were passed to the method...

test/manipulation.test.js
partap: I can change it, but let me defend this style first... :)

Firstly, .then is a chained function from the previous statement that
returned a promise in the same context, same as .catch.  There can be
multiple .thens chained together, and I like the indentation to actually
mean a scope change...

https://www.dropbox.com/s/6ofmy3jx6i6pyku/Screenshot%202015-02-19%2014.24.53.png?dl=0
https://www.dropbox.com/s/mgf4k5oiob9hx1g/Screenshot%202015-02-19%2014.26.01.png?dl=0

Does the extra indentation in the second version make it any clearer?  I
prefer the first version where each indentation is clearly a change in
context.

Taken to extremes, indenting chained .thens can actually reduce
readability... Here's one of the tests in relations.test.js, a common
promise pattern:
https://www.dropbox.com/s/xdijmj9340szrdv/Screenshot%202015-02-19%2014.40.54.png?dl=0

and here it is with indented .thens:
https://www.dropbox.com/s/b3muldxq44k7xh8/Screenshot%202015-02-19%2014.42.16.png?dl=0

...that's clearly ridiculous, so maybe just indent the first .then
https://www.dropbox.com/s/7iqqsda2gd1w8hg/Screenshot%202015-02-19%2014.51.53.png?dl=0

Better, but now the indentation rule seems more arbitrary and subjective to
me... I can live with it, though, if you insist :)

Secondly, the space after .then is my mistake,  The space after function is
not... if there is no space, it looks more like it is calling a function
rather than defining it (my editor actually highlights them differently).

https://www.dropbox.com/s/zkw2jwywyxeky6p/Screenshot%202015-02-19%2014.10.36.png?dl=0
https://www.dropbox.com/s/zm1mzh00zk28e2h/Screenshot%202015-02-19%2014.11.58.png?dl=0

...also I notice that *most* other places in the file, the callbacks are
defined with a space after function (e.g.: function (done) 2 lines above)

test/manipulation.test.js
partap: catch(function (err) {...}) is basically syntactic sugar for then(null,
(err) {...}).
In this case, your second catch statement wouldn't do anything unless
another error is thrown from the first error handler, which it isn't.

In general, I like to use catch rather than a second argument to then
because it's more visible than just a comma and thus easier to see its
intent as an error handler.

test/manipulation.test.js
partap: Originally I just wanted to make sure that it worked without inspecting all
the test cases fully.
I think I just duplicated every test that I could rewrite using promises
rather than callbacks... we can probably take out a bunch of them.

I was a bit more selective in the relations tests, once I realized that
there were hundreds of them....

partap:
I noticed this as well...  I'm running up against this problem in my main
project now.

partap:
Hmm, tried your stillConnecting patch....
It's a little more complicated...dataSource.ready needs args.callee to be
set, and we need to return false rather than the promise if it *is*
ready....
The only problem I see now is that we are creating an extra promise for
every call now whether we need it or not and (usually) discarding it.

test/manipulation.test.js
ritch: > I think I just duplicated every test that I could rewrite using promises
rather than callbacks... we can probably take out a bunch of them.

So I've noticed a lot of talk about whitespace which I'd like to point out as basically annoying but could really care less if fixed before this PR lands... but this (duplicating a large number of tests) is concerning. @bajtos @raymondfeng any ideas on how to test promise support without duplicating so many tests?

ritch:
```js
if (stillConnecting(this.getDataSource(), this, arguments)) return;

> We need to come up with a way how to return a promise in this case too.

This is a bit out of scope for this PR... but I'd like to see this instead of some hack to get promises to work (which is possible by returning the promise which behaves like the `stillConnecting()` function).

```js
dao.create = function(data, cb) {
  var done = createPromiseOrCallback(arguments);

  this.ready(function(err) {
    if(err) return done(err);
    connector.create(data, done);
  });
}


dao.ready = function(options) {
  var connector = this.connector;

  if(connector.connected) {
    return cb();
  } else {
    var t = setTimeout(function() {
      cb(new Error('timeout'));
    });

    connector.once('connected', function() {
      clearTimeout(t);
      cb();
    });    
  }
}
```

test/manipulation.test.js
partap: It's only 20 extra tests for DAO, and, as pointed out, many of them are
redundant and can be removed.

Relations are a different matter... there are a lot of different code paths
to test.  I might be able to trim that down too, but as is, I've only added
about 70 tests to the existing 240.

test/manipulation.test.js
partap: Looks like I misread your comment before. 
You're right.  `should.not.exist(p)` is not a sufficient check that the method failed.  I'll update it.

partap:
Oops. Guess I should have waited to rebase...  
@bajtos I think I've addressed all your concerns for the time being.

I think we should handle the stillConnecting problem in a cleaner way, but this works for now.

partap:
@ritch 
I like that flow better.  Yeah, probably out of scope for this PR though.

partap:
@bajtos
How can we test this stillConnecting hack?  I'm still not familiar enough with the loopback architecture to understand exactly when or how the connection happens. 

It doesn't appear to be working properly in my main project in the case that it is still connecting...

partap:
Ok, found it ... `arguments` was being mixed up with `args`.
Would still be nice to get this case into the tests somehow.

bajtos:
> Would still be nice to get this case into the tests somehow.

Here is a sample code demonstrating how to write such tests:

```js
describe('Connector with slow connect', function() {
  var SlowMemory = {
    initialize: function(ds, cb) {
      ds.connector = new Memory(ds.settings);
      ds.connecting = true;
      ds.connected = false;
      ds._connectNow = cb;
    }
  };

  it('works', function(done) {
    var order = ['start'];

    var ds = new DataSource({ connector: SlowMemory });
    var TestModel = ds.createModel('TestModelSlow');

    async.parallel([
      function create(cb) {
        TestModel.create({}, function(err, m) {
          order.push('method');
          cb(err);
        });
      },
      function connect(cb) {
        setTimeout(function() {
          order.push('connect');
          ds._connectNow();
          cb();
        }, 20);
      }],
      function completed(err) {
        if (err) return done(err);
        order.should.eql(['start', 'connect', 'method']);
        done();
      });
  });
});
```

lib/dao.js
```diff
@@ -148,7 +148,8 @@ function noCallback(err, result) {
  *   - instance (null or Model)
  */
 DataAccessObject.create = function (data, options, cb) {
-  if (stillConnecting(this.getDataSource(), this, arguments)) return;
+  var p;
+  if (p = stillConnecting(this.getDataSource(), this, arguments)) return p;
```
bajtos: Is there any particular reason why you are initializing `p` inside the `if` condition, considering that the variable is declared outside `if`? Linters will complain that your code may contain a typo, because many cases where `if` contains a single `=` were meant to use `==` or `===`.

I strongly prefer the following style:

```js
var p = stillConnecting(this.getDataSource(), this, arguments);
if (p) return p;
```

Also the name `p` is rather non-descriptive, how about `connectedPromise` or `promiseIfConnecting`?

bajtos:
Great job.

@ritch I like you proposal too and agree that it's out of scope of this PR.

> @partap The only problem I see now is that we are creating an extra promise for
every call now whether we need it or not and (usually) discarding it.

Yes, this is not great. IMO, we should get rid of `stillConnecting` per Ritchie's proposal instead of trying to hack the current solution based on `stillConnecting`.

Since the promises come with 50-100% performance penalty anyways, micro-optimisations like avoiding extra promise allocations are most likely not worth the effort now.

lib/dao.js
```diff
@@ -167,9 +168,9 @@ DataAccessObject.create = function (data, options, cb) {
     }
   }
 
-  cb = cb || noCallback;
   data = data || {};
   options = options || {};
+  cb = cb || (Array.isArray(data) ? noCallback : utils.createPromiseCallback());
```
bajtos: In the `Array.isArray(data)` case, can we return a dummy promise object that will report a meaningful error when used via `.then` or `.catch`? The code is already in `utils.createPromiseCallback`, it just needs to be extracted into a standalone method.

```js
function createThrowingPromise(message) {
  var promise = {};
  Object.defineProperty(promise, 'then', { get: function() { throw new Error(message); });
  Object.defineProperty(promise, 'catch', { get: function() { throw new Error(message); });
  return promise;
};

// and then in dao.create()
if (!cb) {
  if (Array.isArray(data)) {
    cb = noCallback;
    cb.promise = utils.createThrowingPromise(
      'The array variant of Model.create() does not support promise-based API.');
  } else {
    cb = utils.createPromiseCallback();
  }
}
```

lib/dao.js
bajtos: :+1:

lib/dao.js
bajtos: My counter-example is that when a method forwards it's arguments, perhaps amending some of them along the way, then it may pass `undefined` value.

However, in this specific case, I am ok with the solution here.

test/manipulation.test.js
bajtos: This will be always subjective. The rule I prefer is that whenever a statement is spread on more than one line, all subsequent lines are indented by two spaces, i.e. "just indent the first .then". This is also the rule we apply in supertest-based tests:

```js
supertest(app).get('/')
  .set(...)
  .send(..)
  .expect(...)
  .end(done);
```

> The space after function is not... if there is no space, it looks more like it is calling a function
rather than defining it (my editor actually highlights them differently).

Again, there isn't an objective way how to decide this. We are following the [Google coding style](https://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml?showone=Function_Declarations_Within_Blocks#Function_Declarations_Within_Blocks) (which is what Node follows too AFAIK) where the anonymous functions are always declared as `function(...args...)`, no space after the `function` keyword.

> I notice that *most* other places in the file, the callbacks are defined with a space after function 

Yes, we were not consistent in keeping the coding style consistent. We had the same problem in strongloop/loopback, where we eventually decided to add `jscs` as a pretest script and fix all style violations manually.

test/manipulation.test.js
bajtos: > In this case, your second catch statement wouldn't do anything unless
another error is thrown from the first error handler, which it isn't.

The second catch handler is invoked when the assert in the first *success handler* throws an `AssertionError`.

test/manipulation.test.js
bajtos: > any ideas on how to test promise support without duplicating so many tests?

I'd say write only one test per each method to test the happy path. Since the implementation is using callback under the hood, I think it's ok to assume that it would work in other cases that are already covered by callback-based tests.

bajtos:
> I'd like to see this (see my example below) instead of some hack to get promises to work (which is possible by returning the promise which behaves like the stillConnecting() function).

I was thinking about this some more. IMO the `connecting` check should be an internal implementation detail of the connectors, the datasource should not need to deal with that. Most well-implemented connectors don't have a disconnected state, either because they are using connection pooling (e.g. MySQL) or because they don't maintain a persistent connection (REST, SOAP).

The base module loopback-connector can provide some utility functions to make it easier for connector implementors to implement "wait until connected" functionality where needed.

bajtos:
@partap another thing we need to address: at the moment, the callback of `findOrCreate` accepts three args - `err`, `obj`, `created`. However, the promise function `resolve` takes only one argument and discards the rest.

I see two solutions to this problem:

 1. Inside `makePromiseCallback`, if there is more than two arguments, resolve the promise with an array `[arg2, arg3,...]`.

 2. Make `findOrCreate` smarter and pass an object to the callback when in a promise mode: `cb(null, { instance: obj, created: created })`.

@raymondfeng @ritch thoughts?

lib/dao.js
partap: Hehe...I just copied the code from your previous comment :)
I had to pull the `var p` out of the `if` statement....but yeah, I definitely prefer this style too.

lib/dao.js
partap: I did it this way to preserve the existing behavior, returning data when called with an array and no callback. 
Probably best not to break any existing API behavior, at least in this PR...

partap:
@bajtos 
I missed the findOrCreate situation, thanks.  Looks like another missing case in the test suite...

I'm fine with either an array or an object being passed back... either one will require something extra from the developer on the receiving end.

Actually, on second thought, I'm leaning towards the array... with then with bluebird you could do something like:
```
User.findOrCreate(...)
.spread(function (user, created) {
// ...
});

File: test/manipulation.test.js (hunk 4 above)
Comment (partap) [inline test/manipulation.test.js]: @bajtos 
(I know, I know...enough about the whitespace...)

I do agree it's subjective, which is why I was pointing out my own reasoning...

FWIW, neither Google or the [Node style guide](https://github.com/felixge/node-style-guide) explicitly says *not* to put a space after `function` ...although I admit,  they don't use a space when examples contain an anonymous function...

[Douglas Crockford](http://javascript.crockford.com/code.html#function) explicitly recommends the space after function: "If the space is omited, then it can appear that the function's name is function, which is an incorrect reading."

....and not to beat a dead horse, but it looks like you are not blindly following Google's style guide.  Google and Node disagree on chaining style, and your style recommendation matches neither one (it's closer to Node's though):

[Node](https://github.com/felixge/node-style-guide#method-chaining):
```
Person
  .create(...)
  .then(...)
  .catch(...);
```
 [Google](https://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml?showone=Code_formatting#Code_formatting):
```
Person.create(...).
    then(...).
    catch(...);
```
*Eww...* ;)

Anyways, it's your codebase...style it how you like :)

Personally, I think your chaining style is fine for js (mine works better for coffeescript...less parentheses required), but I think the space after `function` should be allowed, or even recommended...  A quick search in loopback-datasource-juggler shows that you are already doing it that way quite a bit more than without the space throughout the code, jscs or not...

(OK, I'm done now...)

File: lib/dao.js (hunk 3 above)
Comment (partap) [inline lib/dao.js]: >
> My counter-example is that when a method forwards it's arguments, perhaps
> amending some of them along the way, then it may pass undefined value.
>
I would think that would be another example that strengthens my case,
actually.

> However, in this specific case, I am ok with the solution here.
>
Cool...  I think you'd agree that this particular method *is* a special
case.... most methods don't have the possible side effect of overwriting
your entire dataset if the arguments are misinterpreted...

File: lib/dao.js (hunk 18 above)
Comment (bajtos) [inline lib/dao.js]: > returning data when called with an array and no callback.

Oh, I missed that, you are of course correct. Don't introduce any breaking change :+1:

**Comment by bajtos** [comment]
> Missed the findOrCreate situation, thanks. Looks like another missing case in the test suite...
> I'm leaning towards the array...

OK, let's use the array.

Just note that `Promise.prototype.spread` is not a part of Promises/A+ or ES6.

File: test/manipulation.test.js (hunk 4 above)
Comment (bajtos) [inline test/manipulation.test.js]: On the second thought, I think @ritch is right that the whitespace is not a deal breaker, especially since we don't have a linter set up yet. Let's focus on fixing the issues in the functionality to get this patch landed.

File: lib/dao.js (hunk 3 above)
Comment (bajtos) [inline lib/dao.js]: > Cool...  I think you'd agree that this particular method *is* a special
case.... most methods don't have the possible side effect of overwriting
your entire dataset if the arguments are misinterpreted...

Yes, definitely.

**Comment by bajtos** [comment]
I have grepped the file for more cases where the callback accepts 3 arguments and did not find any except `findOrCreate`.

As far as I am concerned, there are two outstanding issue to get addressed before this patch can be landed:

 - Clean `if (p = stillConnecting(this.getDataSource(), this, arguments))`
 - Fix `findOrCreate` to resolve the promise with both `obj` and `created`

Did I miss anything?

**Comment by partap** [comment]
@bajtos There were no tests at all for findOrCreate so I added some... 
Fixed the case of a single data argument and no callback with promise.

Is this all we need to merge this PR?

**Comment by bajtos** [comment]
The code LGTM, great work! I ran loopback's tests using this branch and all passed :+1: 

@raymondfeng @ritch would you like to take one final look before we merge?

@partap could you please squash all commits into a single one a provide a descriptive commit message following the [50/72 rule](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)? The message body should describe the new API and mention all edge cases (batch create, return value of `create()`, etc.)

**Comment by partap** [comment]
@bajtos 
Does this look ok to you?
```
Add Promises to DAO

When a callback is omitted from a DAO method, return a Promise that
resolves to the value normally passed to the callback of that method.
If a callback is provided, behave normally.

This API will use native ES6 promises if available.  If not available,
or to force the use of another Promise library, you must assign the
global.Promise object.

e.g.:
global.Promise = require('bluebird')

Class methods affected:

- create
- updateOrCreate / upsert
- findOrCreate
- exists
- find
- findOne
- findById
- findByIds
- remove / deleteAll / destroyAll
- removeById / deleteById / destroyById
- count
- update / updateAll

Prototype methods affected:

- save
- delete / remove / destroy
- updateAttribute
- updateAttributes
- reload

Exceptions / edge cases:

- create() used to return the data object that was passed in, even if
  no callback was provided.  Now, if a callback is provided, it will
  return the data object, otherwise it will return a Promise.

- If create() is provided an array of data objects for creation, it
  will continue to always return the array. This batch creation mode
  does not support promises.

- findOrCreate() has a callback of the form: cb(err, instance, created),
  with the extra parameter indicating whether the instance was created
  or not. When called with its promise variant, the resolver will
  receive a single array parameter: [instance, created]
```

bajtos:
@partap Perfect :+1:

partap:
Squashed.

bajtos:
@slnode test please

bajtos:
Landed, thank you very much for this valuable contribution! :tada: :clap:

partap:
Woo!

satyadeepk:
Extremely useful stuff @partap and the loopback team! 
Would be great if you could update the main docs and the apidocs to reflect these promisified methods.

PR #452, opened 2015-02-18
Related Models Promisification

partap:
Rebased and isolated related models promise code from https://github.com/strongloop/loopback-datasource-juggler/pull/447

The tests currently rely on promise-capable DAO from https://github.com/strongloop/loopback-datasource-juggler/pull/451, but the code itself should work separately.

Connects strongloop/loopback#418

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -758,9 +758,12 @@ HasMany.prototype.findById = function (fkId, cb) {
   filter.where[idName] = fkId;
   filter.where[fk] = modelInstance[pk];
 
+  cb = cb || utils.createPromiseCallback();
+
   if (filter.where[fk] === undefined) {
     // Foreign key is undefined
-    return process.nextTick(cb);
+    process.nextTick(cb);
+    return cb.promise;
   }
   this.definition.applyScope(modelInstance, filter);
 
@@ -784,6 +787,7 @@ HasMany.prototype.findById = function (fkId, cb) {
       cb(err);
     }
   });
+  return cb.promise;
 };
 
 /**
@@ -795,6 +799,8 @@ HasMany.prototype.exists = function (fkId, cb) {
   var fk = this.definition.keyTo;
   var pk = this.definition.keyFrom;
   var modelInstance = this.modelInstance;
+  cb = cb || utils.createPromiseCallback();
+
 
   this.findById(fkId, function (err, inst) {
     if (err) {
@@ -810,6 +816,7 @@ HasMany.prototype.exists = function (fkId, cb) {
       cb(null, false);
     }
   });
+  return cb.promise;
 };
 
 /**
@@ -818,12 +825,14 @@ HasMany.prototype.exists = function (fkId, cb) {
  * @param {Function} cb The callback function
  */
 HasMany.prototype.updateById = function (fkId, data, cb) {
+  cb = cb || utils.createPromiseCallback();
   this.findById(fkId, function (err, inst) {
     if (err) {
       return cb && cb(err);
     }
     inst.updateAttributes(data, cb);
   });
+  return cb.promise;
 };
 
 /**
@@ -832,6 +841,7 @@ HasMany.prototype.updateById = function (fkId, data, cb) {
  * @param {Function} cb The callback function
  */
 HasMany.prototype.destroyById = function (fkId, cb) {
+  cb = cb || utils.createPromiseCallback();
   var self = this;
   this.findById(fkId, function(err, inst) {
     if (err) {
@@ -840,6 +850,7 @@ HasMany.prototype.destroyById = function (fkId, cb) {
     self.removeFromCache(inst[fkId]);
     inst.destroy(cb);
   });
+  return cb.promise;
 };
 
 var throughKeys = function(definition) {
@@ -875,6 +886,8 @@ HasManyThrough.prototype.findById = function (fkId, cb) {
   var modelInstance = this.modelInstance;
   var modelThrough = this.definition.modelThrough;
 
+  cb = cb || utils.createPromiseCallback();
+
   self.exists(fkId, function (err, exists) {
     if (err || !exists) {
       if (!err) {
@@ -897,6 +910,7 @@ HasManyThrough.prototype.findById = function (fkId, cb) {
       cb(err, inst);
     });
   });
+  return cb.promise;
 };
 
 /**
@@ -910,6 +924,8 @@ HasManyThrough.prototype.destroyById = function (fkId, cb) {
   var pk = this.definition.keyFrom;
   var modelInstance = this.modelInstance;
   var modelThrough = this.definition.modelThrough;
+
+  cb = cb || utils.createPromiseCallback();
 
   self.exists(fkId, function (err, exists) {
     if (err || !exists) {
@@ -928,6 +944,7 @@ HasManyThrough.prototype.destroyById = function (fkId, cb) {
       modelTo.deleteById(fkId, cb);
     });
   });
+  return cb.promise;
 };
 
 // Create an instance of the target model and connect it to the instance of
@@ -942,7 +959,7 @@ HasManyThrough.prototype.create = function create(data, done) {
     done = data;
     data = {};
   }
-  done = done || function(){};
+  done = done || utils.createPromiseCallback();
 
   var modelInstance = this.modelInstance;
 
@@ -984,6 +1001,7 @@ HasManyThrough.prototype.create = function create(data, done) {
     else
       async.map(to, createRelation, done);
   });
+  return done.promise;
 };
 
 
@@ -1004,6 +1022,9 @@ HasManyThrough.prototype.add = function (acInst, data, done) {
     data = {};
   }
   var query = {};
+
+  data = data || {};
+  done = done || utils.createPromiseCallback();
 
   // The primary key for the target model
   var pk2 = definition.modelTo.definition.idName();
@@ -1033,6 +1054,7 @@ HasManyThrough.prototype.add = function (acInst, data, done) {
     }
     done(err, ac);
   });
+  return done.promise;
 };
 
 /**
@@ -1060,9 +1082,12 @@ HasManyThrough.prototype.exists = function (acInst, done) {
 
   definition.applyScope(this.modelInstance, filter);
 
+  done = done || utils.createPromiseCallback();
+
   modelThrough.count(filter.where, function(err, ac) {
     done(err, ac > 0);
   });
+  return done.promise;
 };
 
 /**
@@ -1091,12 +1116,15 @@ HasManyThrough.prototype.remove = function (acInst, done) {
 
   definition.applyScope(this.modelInstance, filter);
 
+  done = done || utils.createPromiseCallback();
+
   modelThrough.deleteAll(filter.where, function (err) {
     if (!err) {
       self.removeFromCache(query[fk2]);
     }
     done(err);
   });
+  return done.promise;
 };
 
 
@@ -1186,6 +1214,7 @@ RelationDefinition.belongsTo = function (modelFrom, modelTo, params) {
     get: function() {
       var relation = new BelongsTo(definition, this);
       var relationMethod = relation.related.bind(relation);
+      relationMethod.getAsync = relation.getAsync.bind(relation);
       relationMethod.update = relation.update.bind(relation);
       relationMethod.destroy = relation.destroy.bind(relation);
       if (!polymorphic) {
@@ -1221,6 +1250,7 @@ BelongsTo.prototype.create = function(targetModelData, cb) {
     cb = targetModelData;
     targetModelData = {};
   }
+  cb = cb || utils.createPromiseCallback();
 
   this.definition.applyProperties(modelInstance, targetModelData || {});
 
@@ -1241,6 +1271,7 @@ BelongsTo.prototype.create = function(targetModelData, cb) {
       cb && cb(err);
     }
   });
+  return cb.promise;
 };
 
 BelongsTo.prototype.build = function(targetModelData) {
@@ -1250,6 +1281,7 @@ BelongsTo.prototype.build = function(targetModelData) {
 };
 
 BelongsTo.prototype.update = function (targetModelData, cb) {
+  cb = cb || utils.createPromiseCallback();
   var definition = this.definition;
   this.fetch(function(err, inst) {
     if (inst instanceof ModelBaseClass) {
@@ -1259,12 +1291,16 @@ BelongsTo.prototype.update = function (targetModelData, cb) {
         + ' is empty'));
     }
   });
+  return cb.promise;
 };
 
 BelongsTo.prototype.destroy = function (cb) {
   var modelTo = this.definition.modelTo;
   var modelInstance = this.modelInstance;
   var fk = this.definition.keyFrom;
+
+  cb = cb || utils.createPromiseCallback();
+
   this.fetch(function(err, targetModel) {
     if (targetModel instanceof ModelBaseClass) {
       modelInstance[fk] = null;
@@ -1277,6 +1313,7 @@ BelongsTo.prototype.destroy = function (cb) {
         + ' is empty'));
     }
   });
+  return cb.promise;
 };
 
 /**
@@ -1394,6 +1431,21 @@ BelongsTo.prototype.related = function (refresh, params) {
   }
 };
 
+
+/**
+ * Define a Promise-based method for the belongsTo relation itself
+ * - order.customer.get(cb): Load the target model instance asynchronously
+ *
+ * @param {Function} cb Callback of the form function (err, inst)
+ * @returns {Promise | Undefined} returns promise if callback is omitted
+ */
+BelongsTo.prototype.getAsync = function (cb) {
+  cb = cb || utils.createPromiseCallback();
+  this.related(true, cb);
+  return cb.promise;
+}
+
+
 /**
  * A hasAndBelongsToMany relation creates a direct many-to-many connection with
  * another model, with no intervening model. For example, if your application
@@ -1504,6 +1556,7 @@ RelationDefinition.hasOne = function (modelFrom, modelTo, params) {
     get: function() {
       var relation = new HasOne(definition, this);
       var relationMethod = relation.related.bind(relation)
+      relationMethod.getAsync = relation.getAsync.bind(relation);
       relationMethod.create = relation.create.bind(relation);
       relationMethod.build = relation.build.bind(relation);
       relationMethod.update = relation.update.bind(relation);
@@ -1559,6 +1612,8 @@ HasOne.prototype.create = function (targetModelData, cb) {
     targetModelData = {};
   }
   targetModelData = targetModelData || {};
+  cb = cb || utils.createPromiseCallback();
+
   targetModelData[fk] = modelInstance[pk];
   var query = {where: {}};
   query.where[fk] = targetModelData[fk];
@@ -1579,9 +1634,11 @@ HasOne.prototype.create = function (targetModelData, cb) {
         + modelTo.modelName));
     }
   });
+  return cb.promise;
 };
 
 HasOne.prototype.update = function(targetModelData, cb) {
+  cb = cb || utils.createPromiseCallback();
   var definition = this.definition;
   var fk = this.definition.keyTo;
   this.fetch(function(err, targetModel) {
@@ -1593,9 +1650,11 @@ HasOne.prototype.update = function(targetModelData, cb) {
         + ' is empty'));
     }
   });
+  return cb.promise;
 };
 
 HasOne.prototype.destroy = function (cb) {
+  cb = cb || utils.createPromiseCallback();
   this.fetch(function(err, targetModel) {
     if (targetModel instanceof ModelBaseClass) {
       targetModel.destroy(cb);
@@ -1604,6 +1663,7 @@ HasOne.prototype.destroy = function (cb) {
         + ' is empty'));
     }
   });
+  return cb.promise;
 };
 
 /**
@@ -1625,6 +1685,7 @@ HasMany.prototype.create = function (targetModelData, cb) {
     targetModelData = {};
   }
   targetModelData = targetModelData || {};
+  cb = cb || utils.createPromiseCallback();
 
   var fkAndProps = function(item) {
     item[fk] = modelInstance[pk];
@@ -1650,6 +1711,7 @@ HasMany.prototype.create = function (targetModelData, cb) {
       cb && cb(err);
     }
   });
+  return cb.promise;
 };
 /**
  * Build a target model instance
@@ -1739,6 +1801,20 @@ HasOne.prototype.related = function (refresh, params) {
     params[fk] = modelInstance[pk];
     self.resetCache();
   }
+};
+
+/**
+ * Define a Promise-based method for the hasOne relation itself
+ * - order.customer.getAsync(cb): Load the target model instance asynchronously
+ *
+ * @param {Function} cb Callback of the form function (err, inst)
+ * @returns {Promise | Undefined} Returns promise if cb is omitted
+ */
+
+HasOne.prototype.getAsync = function (cb) {
+  cb = cb || utils.createPromiseCallback();
+  this.related(true, cb);
+  return cb.promise;
 };
 
 RelationDefinition.embedsOne = function (modelFrom, modelTo, params) {
@@ -1896,6 +1972,7 @@ EmbedsOne.prototype.create = function (targetModelData, cb) {
   }
 
   targetModelData = targetModelData || {};
+  cb = cb || utils.createPromiseCallback();
 
   var inst = this.callScopeMethod('build', targetModelData);
 
@@ -1921,6 +1998,7 @@ EmbedsOne.prototype.create = function (targetModelData, cb) {
       updateEmbedded();
     }
   }
+  return cb.promise;
 };
 
 EmbedsOne.prototype.build = function (targetModelData) {
@@ -1963,25 +2041,29 @@ EmbedsOne.prototype.update = function (targetModelData, cb) {
   var embeddedInstance = modelInstance[propertyName];
   if (embeddedInstance instanceof modelTo) {
     embeddedInstance.setAttributes(data);
+    cb = cb || utils.createPromiseCallback();
     if (typeof cb === 'function') {
       modelInstance.save(function(err, inst) {
         cb(err, inst ? inst[propertyName] : embeddedInstance);
       });
     }
   } else if (!embeddedInstance && cb) {
-    this.callScopeMethod('create', data, cb);
+    return this.callScopeMethod('create', data, cb);
   } else if (!embeddedInstance) {
-    this.callScopeMethod('build', data);
+    return this.callScopeMethod('build', data);
   }
+  return cb.promise;
 };
 
 EmbedsOne.prototype.destroy = function (cb) {
   var modelInstance = this.modelInstance;
   var propertyName = this.definition.keyFrom;
   modelInstance.unsetAttribute(propertyName, true);
+  cb = cb || utils.createPromiseCallback();
   modelInstance.save(function (err, result) {
     cb && cb(err, result);
   });
+  return cb.promise;
 };
 
 RelationDefinition.embedsMany = function embedsMany(modelFrom, modelTo, params) {
@@ -2379,6 +2461,7 @@ EmbedsMany.prototype.create = function (targetModelData, cb) {
     targetModelData = {};
   }
   targetModelData = targetModelData || {};
+  cb = cb || utils.createPromiseCallback();
 
   var embeddedList = this.embeddedList();
 
@@ -2406,6 +2489,7 @@ EmbedsMany.prototype.create = function (targetModelData, cb) {
       updateEmbedded();
     }
   }
+  return cb.promise;
 };
 
 EmbedsMany.prototype.build = function(targetModelData) {
@@ -2464,6 +2548,7 @@ EmbedsMany.prototype.add = function (acInst, data, cb) {
     cb = data;
     data = {};
   }
+  cb = cb || utils.createPromiseCallback();
 
   var self = this;
   var definition = this.definition;
@@ -2499,6 +2584,7 @@ EmbedsMany.prototype.add = function (acInst, data, cb) {
       cb(null, null);
     }
   });
+  return cb.promise;
 };
 
 /**
@@ -2529,6 +2615,8 @@ EmbedsMany.prototype.remove = function (acInst, cb) {
 
   belongsTo.applyScope(modelInstance, filter);
 
+  cb = cb || utils.createPromiseCallback();
+
   modelInstance[definition.name](filter, function(err, items) {
     if (err) return cb(err);
 
@@ -2540,6 +2628,7 @@ EmbedsMany.prototype.remove = function (acInst, cb) {
       cb(err);
     });
   });
+  return cb.promise;
 };
 
 RelationDefinition.referencesMany = function referencesMany(modelFrom, modelTo, params) {
@@ -2640,6 +2729,7 @@ ReferencesMany.prototype.related = function(receiver, scopeParams, condOrRefresh
 
   var actualCond = {};
   var actualRefresh = false;
+
   if (arguments.length === 3) {
     cb = condOrRefresh;
   } else if (arguments.length === 4) {
@@ -2682,6 +2772,8 @@ ReferencesMany.prototype.findById = function (fkId, cb) {
 
   this.definition.applyScope(modelInstance, filter);
 
+  cb = cb || utils.createPromiseCallback();
+
   modelTo.findByIds(ids, filter, function (err, instances) {
     if (err) {
       return cb(err);
@@ -2708,6 +2800,7 @@ ReferencesMany.prototype.findById = function (fkId, cb) {
       cb(err);
     }
   });
+  return cb.promise;
 };
 
 ReferencesMany.prototype.exists = function (fkId, cb) {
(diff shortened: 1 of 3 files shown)


slnode:

partap:
@bajtos 
I've updated the relations to use, e.g.,  `customer.orders.get()` for promise-based retrieval to avoid the awkwardness of `customer.orders(true)`

partap:
@bajtos 
Rather than rewrite ~70 tests in relations.test.js to only use promises for relations and not dao, I suggest copying the dao.js file from the promises-dao into the working tree here to get the tests to run...

partap:
@bajtos
There is a ton of trailing whitespace in relation-definition.js.  
I've been dancing around it with my commits... Could you take care of it in the master branch?

bajtos:
> There is a ton of trailing whitespace in relation-definition.js. I've been dancing around it with my commits... Could you take care of it in the master branch?

Done.

lib/relation-definition.js
```diff
@@ -1276,8 +1313,12 @@ BelongsTo.prototype.related = function (refresh, params) {
   var scopeQuery = null;
   
   if (arguments.length === 1) {
+    if (typeof refresh === 'boolean') {
+      params = utils.createPromiseCallback();
+    } else {
```
bajtos: This makes `Product.categories(true)` return a promise, correct? I thought we agreed that `Product.categories()` will not be promise-enabled at all. I am proposing to revert all changes made in the `related()` method.

lib/relation-definition.js
```diff
@@ -1369,6 +1411,20 @@ BelongsTo.prototype.related = function (refresh, params) {
...
+ * Define a Promise-based method for the belongsTo relation itself
+ * - order.customer.get(refresh): Load the target model instance asynchronously
+ *
+ * @param refresh
+ * @returns {Promise}
+ */
+BelongsTo.prototype.get = function (refresh) {
+  var cb = utils.createPromiseCallback();
+  return this.related(refresh, cb);
+}
```
bajtos: We should not introduce promise-only APIs, all methods should support both promises and callbacks.

Also the parameter `refresh` is not very intuitive to use, I'd prefer to have two methods instead, e.g. `.get()` and `.getCached()`. For the purpose of this pull request, I am ok with adding the former variant only and wait for user feedback before we add more.

```js
BelongsTo.prototype.get = function (cb) {
  var cb = utils.createPromiseCallback();
  return this.related(true, cb);
};
```

@raymondfeng @ritch thoughts?

lib/relation-definition.js
```diff
@@ -1662,8 +1728,13 @@ HasOne.prototype.related = function (refresh, params) {
   var modelInstance = this.modelInstance;
 
   if (arguments.length === 1) {
-    params = refresh;
-    refresh = false;
+    if (typeof refresh === 'boolean') {
+      params = utils.createPromiseCallback();
+    }
```
bajtos: Same comment as above, `related()` should remain callback-only.

lib/relation-definition.js
```diff
@@ -1714,6 +1785,19 @@ HasOne.prototype.related = function (refresh, params) {
...
+ * Define a Promise-based method for the hasOne relation itself
+ * - order.customer.get(refresh): Load the target model instance asynchronously
+ *
+ * @param {Boolean} refresh Reload from the data source
+ * @returns {Promise}
+ */
+
+HasOne.prototype.get = function (refresh) {
+  var cb = utils.createPromiseCallback();
+  return this.related(refresh, cb);
```
bajtos: Same comment as above, this should support both callbacks and promises, and always pass `refresh=true`.

lib/scope.js
```diff
@@ -39,16 +39,21 @@ ScopeDefinition.prototype.targetModel = function(receiver) {
...
   var actualRefresh = false;
   var saveOnCache = true;
-  if (arguments.length === 3) {
+  if (arguments.length === 3 || arguments.length === 4) {
+    if (cb === undefined) {
+      if (typeof condOrRefresh === 'function') {
     cb = condOrRefresh;
-  } else if (arguments.length === 4) {
+      } else {
+        cb = utils.createPromiseCallback();
```
bajtos: Same comment as above, this should remain callback-only.

lib/scope.js
```diff
@@ -162,16 +168,21 @@ function defineScope(cls, targetClass, name, params, methods, options) {
...
+
       f._targetClass = targetModel.modelName;
       if (f._scope.collect) {
         f._targetClass = i8n.camelize(f._scope.collect);
       }
-      
+
+      f.get = function (refresh) {
+        var cb = utils.createPromiseCallback();
+        return definition.related(self, f._scope, refresh, cb);
```
bajtos: Same comment as above, this should support both callbacks and promises, and always pass `refresh=true`.

test/relations.test.js
```diff
@@ -90,6 +91,18 @@ describe('relations', function () {
         });
       });
 
+      it('should create record on scope with promises', function (done) {
+        Book.create()
+        .then (function (book) {
+          return book.chapters.create()
+          .then (function (c) {
```
bajtos: Coding style: indent `.then` by two spaces, no space before opening bracket. Please apply the fix everywhere in the test file.

```js
return book.chapters.create()
  .then(function(c) {
    //...
  })
  .catch(done);
```

test/relations.test.js
```diff
@@ -90,6 +91,18 @@ describe('relations', function () {
...
+            should.exist(c);
+            c.bookId.should.equal(book.id);
+            done();
+          });
+        }).catch(done);
```
bajtos: Coding style: place `.catch` on a new line. Please apply the fix everywhere in the test file.

lib/relation-definition.js
raymondfeng: +1 to allow callback. Isn't `order.user(cb)` the same as `order.user.get(cb)`? I wonder what's the motivation to add a `get`.

bajtos: It is almost the same, except it does not support zero-arg sync version returning a cached copy.

At the moment, `order.user()` returns a cached copy of the related user, thus it's not possible to return a promise instead. See https://github.com/strongloop/loopback/issues/418#issuecomment-74442529 and https://github.com/strongloop/loopback/issues/418#issuecomment-74691763

I find `order.user(false).then...` and `order.user(true).then...` difficult to read, what the heck is true/false supposed to mean without reading the docs? Therefore I proposed to add a new clean API that does not suffer from this problem.

raymondfeng: I see.

lib/relation-definition.js
partap: Oops.
Yeah, I added the `get()` methods but didn't remove promises from the `related()` methods.
...although it doesn't really hurt anything, I guess...

lib/relation-definition.js
partap: Sure, makes sense... I'll add callback support to `get()`.

lib/scope.js
partap: OK, but note that all the `related` functions default to `refresh=false` if refresh is omitted.  
No idea why...

partap:
Oh snap... removing all promise support from `related()` breaks `get()`

partap:
...and removing the refresh arg from `get` also breaks it... it could be a filter condition.

lib/relation-definition.js
partap: Seems to uncover a bug with hasOne... it's failing with promises, and if I change the normal test to use `refresh=true`, it also fails:
```
  1) relations hasOne with scope can be used to query data:
     Uncaught TypeError: Cannot read property 'id' of null
```

bajtos: I guess we can mark the failing test with `it.skip` for now and investigate later. Unless @raymondfeng has time to help you with this now?

bajtos:
@partap what's the status of this PR, is it ready for final review?

partap:
@bajtos I believe so.

partap:
@bajtos 
Rebased on latest master with DAO Promise support now.
I redid the `get` methods to work the same way as `related`, rather than forcing `refresh=true`...it's a little messy with the argument resolution because of the `arguments` checks, but I think it would probably be better to change the default for `refresh` in `related` than in `get`, and that would be beyond the scope of this PR.

bajtos:
> I redid the get methods to work the same way as related, rather than forcing refresh=true...it's a little messy with the argument resolution because of the arguments checks, but I think it would probably be better to change the default for refresh in related than in get, and that would be beyond the scope of this PR.

The "related()" API is messy, confusing and broken, I have intentionally asked to use a better approach in the new "get()" method. See the discussion in https://github.com/strongloop/loopback/issues/256 for more details.

Also boolean-based APIs are really bad, it's very difficult to tell what is the code `Product.categories(true)` doing - is it saying `refresh=true`, `useCache=true`, or something completely else?

bajtos:
If we want to provide a function for getting the data from the cache first, then let's can add `getCached` to make the distinction clear. IMO it can be left out of scope of this PR.

bajtos:
@slnode ok to test

partap:
@bajtos 
Ok, I reverted it back to forcing `refresh=true`.

The test for hasOne with scope fails now.  Needs to be fixed in `related` or somewhere deeper...

partap:
@bajtos 
It looks like the relation is not being created with the scope properties.  The test passes if you add `block: false` to `supplier.account.create`

Not sure exactly how that is supposed to work, but I'm guessing it's supposed to happen around [here](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L1539-L1542)?

Or maybe it's not supposed to be auto-added and it's just a bug in the test?  I can't see how a general query can be used to set properties reliably.  e.g. `{where: {foo: {gt: 5}}}`?

bajtos:
@partap I am not enough familiar with relations to be able to help you here. @raymondfeng and/or @fabien should be more knowledgeable.

lib/relation-definition.js
```diff
@@ -1325,7 +1362,8 @@ BelongsTo.prototype.related = function (refresh, params) {
       if (query.where[pk] === undefined
         || query.where[pk] === null) {
         // Foreign key is undefined
-        return process.nextTick(cb);
+        process.nextTick(cb);
+        return cb.promise;
```
bajtos: AFAICT, `related()` does not support promise API, `cb.promise` is never set and this change can be reverted.

lib/relation-definition.js
```diff
@@ -1356,10 +1394,10 @@ BelongsTo.prototype.related = function (refresh, params) {
           cb(err);
         }
       });
-      return modelInstance[fk];
+      return cb.promise || modelInstance[fk];
```
bajtos: Same comment as above - `related()` should not support promise API, thus `cb.promise` is never set and this change can be reverted.

lib/relation-definition.js
```diff
@@ -1356,10 +1394,10 @@ BelongsTo.prototype.related = function (refresh, params) {
...
     } else {
       cb(null, cachedValue);
-      return cachedValue;
+      return cb.promise || cachedValue;
```
bajtos: Same comment as above - `related()` should not support promise API, thus `cb.promise` is never set and this change can be reverted.

lib/relation-definition.js
```diff
@@ -1701,10 +1763,10 @@ HasOne.prototype.related = function (refresh, params) {
           cb(err);
         }
       });
-      return modelInstance[pk];
+      return cb.promise || modelInstance[pk];
```
bajtos: Again, `related()` should not support promise API, thus `cb.promise` is never set and this change can be reverted.

lib/relation-definition.js
```diff
@@ -1701,10 +1763,10 @@ HasOne.prototype.related = function (refresh, params) {
...
     } else {
       cb(null, cachedValue);
-      return cachedValue;
+      return cb.promise || cachedValue;
```
bajtos: Again, `related()` should not support promise API, thus `cb.promise` is never set and this change can be reverted.

lib/scope.js
```diff
@@ -73,6 +73,7 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
     // Return from cache
     cb(null, self.__cachedRelations[name]);
   }
+  return cb.promise;
```
bajtos: Again, `related()` should not support promise API, thus `cb.promise` is never set and this change can be reverted.

lib/scope.js
```diff
@@ -176,16 +177,28 @@ function defineScope(cls, targetClass, name, params, methods, options) {
...
         f._targetClass = i8n.camelize(f._scope.collect);
       }
-      
+
+      f.get = function (cond, cb) {
+        if (cb === undefined && cond === undefined) {
+          cb = utils.createPromiseCallback();
+          return definition.related(self, f._scope, true, cb);
+        } else if (cb === undefined) {
+          if (typeof cond !== 'function') {
```
bajtos: nitpick: could you please join these two if statements into a single one?

```js
} else if (cb === undefined && typeof cond !== 'function') {
```

test/relations.test.js
```diff
@@ -109,6 +122,25 @@ describe('relations', function () {
         });
       });
 
+      it('should create a batch of records on scope with promises', function (done) {
```
bajtos: Is it a good idea to support promise-enabled batch create, considering the issue we have discussed in the previous PR?

Since the callback-based version calls `async.map` and thus the promise-variant does not require significant changes and simply preserves all quirks of the callback-based version, I am ok with keeping this in, just wanted to check.

test/relations.test.js
```diff
@@ -1726,6 +2322,24 @@ describe('relations', function () {
       });
     });
 
+    it('should allow to create belongsTo model in beforeCreate hook with promises', function (done) {
+      var mind;
+      Fear.beforeCreate = function (next) {
+        this.mind.create()
```
bajtos: `beforeCreate` hook is deprecated, plus I don't see any reason why we should suspect the promisification to work differently when invoked from a hook.

I am proposing to remove this particular unit test.

@raymondfeng thoughts?

test/relations.test.js
```diff
@@ -1849,6 +2500,28 @@ describe('relations', function () {
       });
     });
 
+    it('can be used to query data with promises', function (done) {
```
bajtos: We need to promisify datasource methods too, but that's out of scope of this PR.

test/relations.test.js
```diff
@@ -1947,30 +2671,68 @@ describe('relations', function () {
...
+        });
+      });
+    });
+
+    it('should find record that match scope', function (done) {
+      Account.updateAll({ block: true }, function (err) {
+        Supplier.findById(supplierId, function (err, supplier) {
+          supplier.account(function (err, account) {
+            should.not.exists(account);
+            done();
```
bajtos: Is this comment still true?

test/relations.test.js
```diff
@@ -1947,30 +2671,68 @@ describe('relations', function () {
...
+          should.exist(supplier);
+          return supplier.account.create({accountNo: 'a01', block: false})
+          .then(function (account) {
+            return supplier.account.get()
+          })
+          .then(function (act) {
+            accountId = act.id;
+            should.exist(act);
+            act.should.be.an.instanceOf(Account);
+            act.block.should.be.false;
```
bajtos: Please keep the original callback-base test around too.

test/relations.test.js
```diff
@@ -3629,6 +4828,7 @@ describe('relations', function () {
           });
           cb(null, summary);
         });
+        return cb.promise;
       };
 
       summarize.shared = true; // remoting
```
bajtos: A general note on writing promise-based code that applies to most of the tests in this pull request: one of the benefits of using promises is that you can keep your code nesting shallow by "lifting" an intermediary promise from a `then` block to the higher level.

```js
Category.findOne()
  .then(function(cat) {
    return cat.jobs.summarize();
  });
  .then(function(summary) {
    // ...
    done();
  })
  .catch(done);
```

Instead of deeply nested

```js
Category.findOne()
  .then(function(cat) {
    return cat.jobs.summarize()
      .then(function(summary) {
        // ...
       done();
      });
  })
  .catch(done);
```

Notice that the first variant reads more like a traditional sequential program, while the second looks more like a callback hell/xmas tree.

While it would be nice to use the first style thorough this pull request, I understand such change may take quite a long time and I can live with the current version too (I won't insist on fixing the coding style).

bajtos:
I reviewed the changes and left few comments above, most of them are trivial details. There are many new unit-tests, which is good for the coverage, but also bad for reviewers. I did only a superficial review of the tests, I'll trust you that you converted the callback-based tests into promise-based tests correctly.

@raymondfeng @fabien you are more familiar with relations than I am, would you like to take a look too?

test/relations.test.js
raymondfeng: I think it's still valuable to keep the test case. One thing we have to think about is if we allow hooks to return a `Promise` (similar as how we handle remote methods).

test/relations.test.js
raymondfeng: +1.

raymondfeng:
@partap The patch mostly LGTM. I wonder if we should separate the promise related tests into its own file, such as `relations.promise.test.js`. But we can do that after the patch is landed.

test/relations.test.js
bajtos: > I think it's still valuable to keep the test case.

Fair enough.

> ne thing we have to think about is if we allow hooks to return a Promise (similar as how we handle remote methods).

That's already done and released - see #449.

bajtos: > I think it's still valuable to keep the test case.

Fair enough.

> One thing we have to think about is if we allow hooks to return a Promise (similar as how we handle remote methods).

That's already done and released - see #449.

lib/relation-definition.js
partap: `get()` works by calling `related()`, so related needs to return the promise.

test/relations.test.js
partap: @bajtos I agree... I did do that style in several places, but I missed a few (rewriting these tests is mind-numbing).  
Also, there are a few that can't be written that way, due to the second `then` block needing access to the scope of the first. (for instance, when that second `then` refers to `cat` )

test/relations.test.js
partap: @bajtos Not sure to which comment you are referring...was it the one about the test failing when `refresh=true`?
I mentioned elsewhere that I believe this test was broken, not the code being tested.
It passed before, when `refresh=false` (default case), because the account was cached, but the test didn't check for `act.block === false`, and indeed, the property was not set.

test/relations.test.js
partap: ...the broken one?

lib/relation-definition.js
bajtos: I find that rather ugly. Why does `related()` need to know implementation details of `get`? I'd much rather keep this knowledge inside `get()`.

```js
var cb = cb || utils.createPromiseCallback();
this.related(true, cb);
return cb.promise;
```

test/relations.test.js
bajtos: Hmm, there is no comment here now, have you removed it recently? I am not familiar with these tests, I am ok to keep this particular block of code as it is now, unless @raymondfeng knows a better solution.

test/relations.test.js
bajtos: Looking at the diff, I can see that there was a test called "should find record that match scope" and that you have removed it. At the same time there is a new test called "should find record that match scope with promises".

I find that confusing. If the original test was broken, but the promise-based version works, then surely the original test can be fixed too?

It seems that "should find record that match scope" is back at lines L2690-L2699, in which case my original comment that started this thread is resolved.

partap:
@bajtos @raymondfeng 
Are we good?  
Mostly the comments I noticed were about taking the `cb.promise` returns out of `related()`, which I did at first, but it broke `get()` so I added them back in.

Technically, I could remove them and create and return new promises in each of the `get()` methods, but I think that would be messier and unnecessary. The only reason to do so would be not not have the word "promise" in any of the `related` methods, even though it seems that API is being deprecated anyhow.   Returning `cb.promise` doesn't affect `related()` at all.  It behaves exactly the same unless it was called with a promise cb.

Personally, I would have preferred leaving the full promise support in `related()`...it doesn't hurt anything, and makes the API more uniform overall...

test/relations.test.js
partap: I think the diff was just confusing that in this section because I grouped the promise tests together rather than interleaving them.

Some of the tests in this section rely on data created in the previous test(s), so I kept the flow between dependent tests rather than rewrite them completely.

bajtos: :+1: makes sense

bajtos:
> Mostly the comments I noticed were about taking the cb.promise returns out of related(), which I did at first, but it broke get() so I added them back in.

I'd like to understand this better. How was `get()` broken after you removed `return cb.promise`? Does it mean that my proposed solution does not work?

```js
var cb = cb || utils.createPromiseCallback();
this.related(true, cb);
return cb.promise;
```

> Technically, I could remove them and create and return new promises in each of the get() methods, but I think that would be messier and unnecessary.

I strongly prefer to keep related code together in a single place (i.e. inside `get()`) rather than sprinkling the implementation all over the place (i.e. split it between `get()` and `related()`).

In other words, in order to implement the new "get" function, you should not need to change "related" methods.

fabien:
@bajtos I can't really get into the discussion above, however, could you tell me if the following will still work after this PR? Otherwise it would be a BC breaking change.

```javascript
Thing.find({ include: 'owner' }, function(err, things) {
  things.each(function(thing) {
    var owner = thing.owner(); // already included, so sync accessor
    console.log(owner.username);
  });
});
```

partap:
@bajtos 
On second thought, I agree with you regarding your `get` solution.  I guess I was making it more complicated in my head...

BTW, Something weird seems to be going on with github's email notifications... For instance, I just now got your first note about `get`, and also your reply to my message from last night re-iterating the `get` solution. :-/

partap:
@fabien 
That code should definitely still work.  
Let me check to make sure there is a test case that covers that.

partap:
@bajtos 
I fixed a couple possible problems in the tests and flattened some promise chains where possible.

I started to update with your indention style but decided against it since it would only affect a few tests, and sort of masked the actual changes.  At least the indention style is consistent for now...  Then if you decide to enforce a style checker to the repo you can do it all at once...

bajtos:
LGTM. @partap please squash the commits and provide a descriptive commit message, as you did in #451/

@raymondfeng @fabien would you like to take a look too, before I land this?

fabien:
@partap did you add any particular tests to make sure the scenario I describe still works? If I gather correctly, you'll use `inst.related.get().then(...)` instead of `inst.related()` to discern between the promisified and regular api?

bajtos:
> If I gather correctly, you'll use inst.related.get().then(...) instead of inst.related() to discern between the promisified and regular api?

Yes, that's the intended usage.

Since the `related` getter was not changed at all, I don't feel a strong need to add new tests for that method. They should have been already written by the person who implemented `inst.related` using the data cached from `find()` response, right? Here is an example of such a test: [test/relations.test.js#L397-L422](https://github.com/strongloop/loopback-datasource-juggler/blob/7dc0fcd26fc670a7dde76a39a76ce9885da96e8a/test/relations.test.js#L397-L422).

raymondfeng:
LGTM

fabien:
LGTM

fabien:
Actually, this bit worries me - we already have `get()` on EmbedsMany relations: https://github.com/strongloop/loopback-datasource-juggler/blob/7dc0fcd26fc670a7dde76a39a76ce9885da96e8a/lib/relation-definition.js#L2277

lib/relation-definition.js
```diff
@@ -1369,6 +1406,21 @@ BelongsTo.prototype.related = function (refresh, params) {
...
+
+/**
+ * Define a Promise-based method for the belongsTo relation itself
+ * - order.customer.get(cb): Load the target model instance asynchronously
+ *
+ * @param {Function} cb Callback of the form function (err, inst)
+ * @returns {Promise | Undefined} returns promise if callback is omitted
+ */
+BelongsTo.prototype.get = function (cb) {
+  var cb = cb || utils.createPromiseCallback();
```
bajtos: please remove `var`, the `cb` variable is already defined.

bajtos:
> @fabien Actually, this bit worries me - we already have get() on EmbedsMany relations: https://github.com/strongloop/loopback-datasource-juggler/blob/7dc0fcd26fc670a7dde76a39a76ce9885da96e8a/lib/relation-definition.js#L2277

Good catch. AFAICT, the commit https://github.com/strongloop/loopback-datasource-juggler/commit/6ed7a0a5f231bf6bed56a6b0e8e572fb6cbbe130 adding that method included a unit-test too and the test is still passing, so we should be good?

However, I dislike the fact that `model.related.get()` has different meaning depending on the relation type. What other name can we use instead? If `model.related.find()` is not taken yet, then I feel this is the right name, because most `related` methods accept a filter in the same way as `Model.find()` accepts a filter.

```js
scope.find = function(filter, cb) {
  if (cb === undefined && typeof filter === 'function') {
    // .find(cb)
    cb = filter;
    filter = true;
  } else if (typeof filter !== 'object') {
    // .find() or .find(undefined, cb) but also .find(false, cb)
    filter = true;
  }
  if (!cb) cb = utils.createPromiseCallback();
  this.related(filter, cb);
  return cb.promise;
}
```

Thoughts?

fabien:
@bajtos How about `fetch` - in fact, we already have this implemented: https://github.com/strongloop/loopback-datasource-juggler/blob/7dc0fcd26fc670a7dde76a39a76ce9885da96e8a/lib/relation-definition.js#L290

What if we would re-use this for the promise based API? I think it does the same thing for the callback-based API, unless I'm missing something.

bajtos:
> How about `fetch` - in fact, we already have this implemented.

Good point. Is the `fetch` method available for all relations? If it is, then we can promisify this method and drop the part of this patch that adds `get` method instead.

I am sad that `fetch` sends `refresh: false` by default, by I tired of fighting against this unfortunate API decision. Let's focus on landing this patch and leave any unrelated improvements for another day.

bajtos:
@partap can you please drop the new "get" methods from your code, replace the usage of "get" with "fetch" in your the new tests, and promisify [fetch](https://github.com/partap/loopback-datasource-juggler/blob/cfaff29fc56fb897ba69d83c7f62e1220512b927/lib/relation-definition.js#L290-L292) implementation?

I hope that should be the last change needed before we can land this.

partap:
Sorry, been busy for a while...

@bajtos @fabien 
 I'm not sure what the advantage of this is... why not stick with `get()`?  There's not really a conflict...
`EmbedsMany` is not promisified at all.... all the calls have synchronous returns.  And even if it *was* promisified...it does the same thing as the new `get()`

With the change you are proposing, we would be replacing, eg:
```
customer.orders.get()
```
with 
```
customer.orders.fetch(true)
```
which is the same thing as we had when I promisified `related()`, and has the same problem that drove you to suggest the `get()` method in the first place.
```
customer.orders(true)
```

If you are really uncomfortable with the usage of `get()`, how about renaming it to `find()`? 
It's not currently implemented, so we don't need to worry about keeping backwards compatibility with the arguments.  Also, I would say that, intuitively, it makes more sense in the overall API:
```
customer = Customer.find(...)
customer.orders.find(...)
```

bajtos:
> With the change you are proposing, we would be replacing, eg:

> ```
customer.orders.get()
```

> with

> ```
customer.orders.fetch(true)
```

> which is the same thing as we had when I promisified related(), and has the same problem that drove you to suggest the get() method in the first place.

Yes, that's a valid point. It's just that I'd rather land this patch imperfect than wait even longer to make it perfect. I mean we can always add `get` or `find` later, right?

I am ok with both `get` and `find`, although it may be little bit confusing that `Order.customer.find` is returning a single model instance, when other `find` methods return an array of instances.

Regardless of the decision we make here, I think the method `fetch` should be promisified.

@fabien what are your thoughts, which solution do you prefer?

partap:
@bajtos In that case why not just re-promisify `related` rather than add a new soon-to-be deprecated method to the API? (Right now, `fetch` is marked as an internal)

partap:
@bajtos @fabien 
Looking into this... 
I can promisify `fetch` easily enough, but to make it accessible as a method on the model is presenting a problem...

I can't figure out any way to access and bind [`Relation.prototype.fetch`](https://github.com/strongloop/loopback-datasource-juggler/blob/7dc0fcd26fc670a7dde76a39a76ce9885da96e8a/lib/relation-definition.js#L290)
from the property getter function in `defineScope`:
https://github.com/partap/loopback-datasource-juggler/blob/promises-related/lib/scope.js#L189

Hrm...for `belongsTo` and `hasOne` I could access the base method, like I did for `get`
https://github.com/partap/loopback-datasource-juggler/blob/promises-related/lib/relation-definition.js#L1193
...but not for the common case...

Best I can figure is to create a new `fetch`, similar to how I implemented `get` in `defineScope`, but then it wouldn't be the base `Relation.prototype.fetch`.

I'm thinking this `fetch` business seems more complicated than just re-promisifying `related`, and doesn't add any benefit that I can see...

bajtos:
> I feel like this PR is moving backward...what's wrong with leaving it as-is? As I mentioned before, get isn't actually conflicting with anything, and the API is pleasant to use (I'm already using it in another project)

If I understand @fabien's concerns correctly, embedded relations already provide `get` methods that are returning the results synchronously, while your `get` method returns a promise. This creates API inconsistency that may confuse users.

I agree we we are a bit stuck here :(

Is there any name we can use for your `get`, one that is not used yet (unlike `fetch`, `get` and `find`)? How about `load`?

partap:
Hmm...yeah, I guess it doesn't conflict code-wise, but maybe API-wise...

Although it could be said that it's already a little inconsistent that embedded relations have `get` methods and the others don't.

`load` could be ok... although I'm a little worried about loading up the API with synonyms that do different things.  I remember being confused about the inconsistent use of `delete`/`remove`/`destroy` and `upsert`/`createOrUpdate` (or was it `updateOrCreate`?) ...I think that's all sorted out now...

What about `getAsync`?  Bluebird users should already be familiar with the convention, and it could be used for embedded relations as well, if desired.

bajtos:
> @partap What about getAsync? Bluebird users should already be familiar with the convention, and it could be used for embedded relations as well, if desired.

Let's use `getAsync` then. I am not 100% happy with it, but I don't have any better solution and as the saying goes, perfect is the enemy of good :) We can always add a new alias later, when we find a good one.

BTW thank you for your patience, I appreciate your perseverance and commitment to get this done.

bajtos:
@partap P.S. Please rebase your changes on top of the current master and resolve the merge conflicts. I think it's ok to squash all your commits into a single one while you are at rewriting the history.

partap:
@bajtos 
Rebased and squashed.  
I also added 3 new tests for `getAsync` using callbacks (1 for each implementation), just because there weren't any. To find them, just search for "getAsync with callback"
...Hmm, can't find

partap:
@bajtos It looks ok now.

lib/scope.js
```diff
@@ -176,16 +176,27 @@ function defineScope(cls, targetClass, name, params, methods, options) {
...
-      
+
+      f.getAsync = function (cond, cb) {
+        if (cb === undefined && cond === undefined) {
+          cb = utils.createPromiseCallback();
+          cond = true;
+        } else if (cb === undefined && typeof cond !== 'function') {
+          cb = utils.createPromiseCallback();
+        }
+        definition.related(self, f._scope, cond, cb);
```
bajtos: I think this code does not correctly handle the variant `getAsync(mycb)`. 

It will call `related(self, f._scope, mycb, undefined)`, and because `related` is checking `arguments.length`, it will parse the arguments incorrectly.

I think this is the correct implementation:

```js
  if (cb === undefined) {
    if (cond === undefined) {
      // getAsync()
      cb = utils.createPromiseCallback();
      cond = true;
    } else if (typeof cond !== 'function') {
      // getAsync({where:{}})
      cb = utils.createPromiseCallback();
    } else {
      // getAsync(function(){})
      cb = cond;
      cond = true;
    }
  }
```

Thoughts?

bajtos:
@partap I found an issue the proposed implementation of `scope.getAsync`, see my comment above. I can apply the proposed fix myself to save you time, if you like.

partap:
@bajtos 
Good catch, thanks! 

That case wasn't covered by the tests so I added one... your fix works.

partap:
@bajtos 
Not sure why is says failing check now...  All tests pass on my end.

Hmm.  It appears there is a request from "Jenkins" to allow read-write access to *all* my public and private repos. I can't imagine how that would be necessary.

bajtos:
@slnode test please

bajtos:
> Not sure why is says failing check now... All tests pass on my end.

The build slaves were not working correctly. I have restarted the build.

> Hmm. It appears there is a request from "Jenkins" to allow read-write access to all my public and private repos. I can't imagine how that would be necessary.

Yeah, I agree that should not be required. Can you @rmg comment on this please?

bajtos:
@slnode test please

bajtos:
The CI is broken, I have verified the implementation locally on my machine.

bajtos:
Landed via 2bdcce0d, thank you for this valuable contribution!

esco:
:+1:

rmg:
Jenkins requires that level of access so that it can verify that you are part of the strongloop org and that you have commit access to the repos it is validating you against -- It is not a public CI.

test/relations.test.js
mrfelton: With bluebird you can do that using .spread(). I'm not sure if that is something standard across promise libraries or weather it's just bluebird specific but it's certainly nice to enable you to pass data between multiple promise scopes.

michael-lumley:
Is there any documentation on this?

superkhau:
>Is there any documentation on this, and the correct syntax for implementing it?

@crandmck ^

crandmck:
No, unfortunately I haven't had time to document use of the promise API at all.  There are lots of open issues for the existing docs, so it would have to be prioritized relative to those....  Before I can start on that, though, it would be very helpful to have a few "promise-ized" sample apps.  @superkhau perhaps you could help in that area?  For example (maybe) loopback-getting-started-promise, etc.

superkhau:
@jannyHou Could you create an issue for yourself to create a sample app for this? I believe you did implemented this feature right?

jannyHou:
@superkhau yeah, thanks for reminding me. I create a story in TOB for it: https://github.com/strongloop-internal/scrum-loopback/issues/827

Issue #450, opened 2015-02-17
Refactor model/model-definition/model-buidler to strong-model module


raymondfeng:
We should refactor the model definition code into its own module, such as `strong-model`.

Similar efforts:
- https://github.com/LearnBoost/mongoose/tree/master/lib/schema
- https://github.com/hapijs/joi
- https://github.com/bluefeet/o-js

bosco-li:
Assigning to Apex since the team will be owning the FA in 2017

/cc @0candy @kjdelisle @pthieu

PR #453, opened 2015-02-18
Implement 'Guid' type + fix loading of the GeoPoint type

### LINKED ISSUE #388: Guids
We should add a new type (or option) for automatically generating guids. Something like this:

```js
{myCustomId: {type: 'guid'}}
```

 - [x] implementation
 - [ ] docs

bajtos:
The new type represents a UUID (GUID) string value.

The constructor interprets the value "new" as a request to generate
a new unique id, which allows developers to define auto-generated guid
properties in LDL:

    id: { type: 'guid', id: true, default: 'new' }

Connect #388

/to @raymondfeng @ritch @fabien @kraman  please review

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@ exports.ModelBuilder = exports.LDL = require('./lib/model-builder.js').ModelBuil
 exports.DataSource = exports.Schema = require('./lib/datasource.js').DataSource;
 exports.ModelBaseClass = require('./lib/model.js');
 exports.GeoPoint = require('./lib/geo.js').GeoPoint;
+exports.Guid = require('./lib/guid');
 exports.ValidationError = require('./lib/validations.js').ValidationError;
 
 Object.defineProperty(exports, 'version', {
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -2,6 +2,7 @@ var util = require('util');
 var Connector = require('loopback-connector').Connector;
 var geo = require('../geo');
 var GeoPoint = geo.GeoPoint;
+var Guid = require('../guid');
 var utils = require('../utils');
 var fs = require('fs');
 var async = require('async');
@@ -280,6 +281,9 @@ Memory.prototype.fromDb = function (model, data) {
           break;
         case 'GeoPoint':
           val = GeoPoint(val);
+          break;
+        case 'Guid':
+          val = Guid(val);
           break;
       }
     }
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -1,6 +1,7 @@
 var util = require('util');
 var Connector = require('loopback-connector').Connector;
 var geo = require('../geo');
+var GeoPoint = geo.GeoPoint;
 var utils = require('../utils');
 var fs = require('fs');
 var async = require('async');
@@ -276,6 +277,9 @@ Memory.prototype.fromDb = function (model, data) {
           break;
         case 'Number':
           val = Number(val);
+          break;
+        case 'GeoPoint':
+          val = GeoPoint(val);
           break;
       }
     }
--- a/lib/guid.js
+++ b/lib/guid.js
@@ -10,7 +10,8 @@ function Guid(data) {
   }
 
   if (data === undefined || data === 'new') {
-    data = uuid.v4();
+    // Generate a v1 (time-based) id
+    data = uuid.v1();
   } else if (data instanceof Guid) {
     data = data.value;
   }
--- a/lib/guid.js
+++ b/lib/guid.js
@@ -1,0 +1,33 @@
+var assert = require('assert');
+var uuid = require('node-uuid');
+module.exports = Guid;
+
+var UUID_REGEXP = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+
+function Guid(data) {
+  if (!(this instanceof Guid)) {
+    return new Guid(data);
+  }
+
+  if (data === undefined || data === 'new') {
+    data = uuid.v4();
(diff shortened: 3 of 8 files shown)


kraman:
+1

bajtos:
Cross-posting from #388:

> > How do i set if this should be generated automatically or not

> I'd say the id should be automatically generated iff it is not defined.
> In other words, if the property is of type guid string with id: true and it does not have any value, then the juggler should generate one.

I decided to not implement the logic proposed above. Instead, the user has to explicitly specify `default: 'new'`. The explicit approach should make it more obvious what's happening under the hood, plus it allows the users of this feature to decide where the value should be generated - in loopback or in the (MongoDB) connector.

bajtos:
@raymondfeng I have a question for you: What changes are necessary to support this new type in the connectors? 

How is the MongoDB connector handling ids, does it always use `_id` with an UUID generated by the database? Can we improve the connector to detect `{type: guid, id: true}` and map it transparently to `_id` field? Is it a good idea?

bajtos:
> What changes are necessary to support this new type in the connectors?

I am playing with the idea of using database-specific column type to store the GUID value as raw 128-bit values. Thoughts?

raymondfeng:
Let me start with a general discussion.

I have a plan to refactor the model/property/type definition into its own module to facilitate the definition/construction of JS classes/traits/properties/types. See https://github.com/strongloop/loopback-datasource-juggler/issues/450.

For id property, it's actually pretty involved.

1. What is the type of id (String, Number, ObjectId, or sub-type of them)?
2. Where is the id generated (generated by client, generated by db, or set by app, maybe related to default)? 
3. How is the id generated (increment, uuid, or lookup serial from DB)?
4. How does the id map to a DB column?
5. Is it single or composite?
6. How does id map to CRUD operations?

The LDL needs to have ability to capture the relevant metadata so that Juggler and connectors can collaboratively honor them.

raymondfeng:
Now the comments for the PR:

1. Do we want to introduce a new type or should we add an attribute to extend the id definition? For a new type, connectors need to implement the support of mapping.

2. For `new` as the default, I would prefer to use `$new` to be consistent with `$now`.

3. The connector probably needs to advise what ID types are supported/preferred. If the model id is meant be generated by the DB, we probably should allow the id type to be injected.

ritch:
> I am playing with the idea of using database-specific column type to store the GUID value as raw 128-bit values. Thoughts?

Sounds good.

The offline sync / replication requirements are that loopback can generate the guid in the client... AFAICT this is possible by setting `default: 'new'`.

bajtos:
@raymondfeng thank you for a thoughtful comment. I don't have bandwidth to address the general id refactoring you have mentioned, my goal is to provide GUID ids generated by the client, as mentioned by @ritch. I am ok to align my work here with the long-term vision you have for the ids, just let me know what to change in my pull request.

> Do we want to introduce a new type or should we add an attribute to extend the id definition? For a new type, connectors need to implement the support of mapping.

I have considered both options. Adding a new type has several benefits: validating the values is easy, there is a clear distinction between "regular" strings and GUID values, and we can map GUID properties to 128bit columns in the databases (where feasible).

If we don't want to add a new type, then I am proposing to implement `defaultFn` (#292) and provide a function for generating GUID strings that can be used as `defaultFn`:

```js
id: { type: 'string', id: true, defaultFn: 'GUID' }
```

Which approach do you prefer?

 > For `new` as the default, I would prefer to use `$new` to be consistent with `$now`.

+1, I had the same thought yesterday evening.

> The connector probably needs to advise what ID types are supported/preferred. If the model id is meant be generated by the DB, we probably should allow the id type to be injected.

IMO, this is a new feature that is not strictly needed for GUID properties. AFAIK, if your model doesn't define any `id:true`, then the `id` column is already injected by the juggler and/or the connector, thus we already have a part of this in place.

> The offline sync / replication requirements are that loopback can generate the guid in the client... AFAICT this is possible by setting default: 'new'.

Yes, that's the plan.

> I am playing with the idea of using database-specific column type to store the GUID value as raw 128-bit values.

To keep the initial implementation simple, perhaps we can map to "string" by default, and allow developers to configure a different mapping using the per-database configuration. The latter can be implemented later, when there is a demand for it.

@raymondfeng please let me know what's needed to move this patch forward and get it eventually landed.

raymondfeng:
For the short-term, I prefer to implement `defaultFn` property as it's minimal, reasonable, and powerful.

bajtos:
> For the short-term, I prefer to implement `defaultFn` property as it's minimal, reasonable, and powerful.

As you wish, see #458.

jgwinner:
Hey, sorry to dig this up ... has anyone reconsidered handling GUID's as text strings? 

I've found the SQL that is generated in the app is *nearly* OK, as long as it's formatted this way:


`  sql: "INSERT INTO `cam_detection`(`id`,`user_id`,`date_start`,`rec_frames`,`rec_status`,`rec_filename`,`rec_snapshot`,`date_end`,`last_cloudsync`,`id_cloud`) VALUES('974a7131-d8d6-11e9-8d0c-9714eb8974ea','974a7130-d8d6-11e9-8d0c-9714eb8974ea','2019-09-16 22:40:07.861',0,0,'string','string','2019-09-16 22:40:07.861','2019-09-16 22:40:07.861',0)",`

would be perfect if it was:

`  sql: "INSERT INTO `cam_detection`(`id`,`user_id`,`date_start`,`rec_frames`,`rec_status`,`rec_filename`,`rec_snapshot`,`date_end`,`last_cloudsync`,`id_cloud`) VALUES(0x974a7131d8d611e98d0c9714eb8974ea,0x974a7130d8d611e9-8d0c9714eb8974ea,'2019-09-16 22:40:07.861',0,0,'string','string','2019-09-16 22:40:07.861','2019-09-16 22:40:07.861',0)",`

Thoughts? Or is this too old of a topic?

        == John ==

PR #456, opened 2015-02-19
support: add install-test-deps.js

bajtos:
This script should be used by connectors to install dependencies needed to run the shared test suite.

Usage: Add an npm "install" scripts with the following contents:

    node -e "require('loopback-datasource-juggler/support/install-test-deps')"

### DIFF
--- a/support/install-test-deps.js
+++ b/support/install-test-deps.js
@@ -1,0 +1,15 @@
+#!/usr/bin/env node
+
+if (process.env.npm_config_production) return;
+
+var path = require('path');
+var juggler = path.resolve(path.dirname(__filename), '..');
+var cp = require('child_process');
+
+console.log('Installing dev dependencies in', juggler);
+process.chdir(juggler);
+
+cp.exec('npm install', function(err, stdout, stderr) {
+  if (err) throw err;
+  if (stderr) console.log(stderr);
+});


raymondfeng:
I don't think the script will work on Windows.

bajtos:
> I don't think the script will work on Windows.

Why it should not work? It works for me on Windows 8 in VirtualBox:

```
C:\src\loopback-datasource-juggler (feature/install-test-deps)
λ node support\install-test-deps.js
Installing dev dependencies in C:\src\loopback-datasource-juggler

C:\src\loopback-datasource-juggler (feature/install-test-deps)
λ ls node_modules\
async  bluebird  debug  inflection  lodash  loopback-connector  mocha  qs  should  traverse

raymondfeng:
Never mind. I just realized that child_process.exec() is to invoke a shell script.

raymondfeng:
I still don't like the fact that each connector will have to mess around package.json with the install script (node -e "require('loopback-datasource-juggler/support/install-test-deps')"). These scripts are evil :-).

What about making the logic in the install script of `loopback-datasource-juggler`?

raymondfeng:
I didn't have a chance to test it out with loopback-datasource-juggler but I fixed most connectors' dev deps.

bajtos:
I'm ok with whatever works best for you.

PR #458, opened 2015-02-20
Add a new property option `defaultFn`

bajtos:
The property allows developers to specify that the default value should be retrieved via a named function.

Only two built-in functions are supported at the moment:

 - `guid`, `uuid` - generate a new GUID/UUID
 - `now` - use the current date and time

Support for custom (user-provided) functions is not implemented yet.

Connect #388, this PR supersedes #453.

/cc @ritch @fabien @kraman

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -16,6 +16,7 @@ var validations = require('./validations');
 var _extend = util._extend;
 var utils = require('./utils');
 var fieldsToArray = utils.fieldsToArray;
+var uuid = require('node-uuid');
 
 // Set up an object for quick lookup
 var BASE_TYPES = {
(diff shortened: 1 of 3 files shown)


bajtos:
Note that I am intentionally leaving out support for custom (user-provided) functions, as I don't have enough time in this sprint to implement that. We have the issue #292 to track that feature request.

lib/model.js
```diff
@@ -228,10 +229,32 @@ ModelBaseClass.prototype._initProperties = function (data, options) {
...
       }
     }
 
+    // Set default value using a named function
+    if (propVal === undefined) {
+      var defn = properties[p].defaultFn;
+      switch (defn) {
+        case undefined:
+          break;
+        case 'GUID':
```
raymondfeng: I prefer to use lower cases as it's the convention for function names in JS.

raymondfeng: We should allow `uuid` too.

bajtos:
@raymondfeng fixed, PTAL.

raymondfeng:
LGTM

mike-aungsan:
Any help on this?

This property does not work.

"properties": {
"testUuid": {
"type": "string",
"description": ["Just A Number field to Test defaultFn and Error Message.", 
"..."],
"default": "$uuid",
"required": false
},
}

Anything wrong?

tatemz:
@bajtos - Any update on custom user provided `defaultFn`?

bajtos:
> @tatemz Any update on custom user provided `defaultFn`?

I am afraid this is not a priority for us right now. Can you contribute this feature yourself?

tatemz:
@bajtos - see https://github.com/strongloop/loopback-datasource-juggler/pull/1221

Issue #457, opened 2015-02-20
Test failures in loopback-connector-mssql


raymondfeng:
There are a few issues that cause test failures in `loopback-connector-mssql`:

1. The `should.js@5.0.0`: `should.not.have.property('x')` now fails if the x value if undefined
2. Dev dependency on `bluebird`
3. `test/manipulation.js` has a Number createdAt (https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/manipulation.test.js#L18) which will produce a `big` int that cannot be contained by the SQL server column of type `int`
4. https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/manipulation.test.js#L212 cannot be asserted for RDBs as the column will return null or default values.

bajtos:
Ad 1 - The solution is to change the check to `should.have.property('x', undefined)`. The property is defined on the object.

Ad 2 - We have discussed this in https://github.com/strongloop/loopback-datasource-juggler/commit/9e0a22141aae52e8d8adc0bf7ba30bc97bebeb9a. I'd like to fix this once and forever via #456. You were proposing to "upgrade the connector's dev dependencies to match juggler", which will require us to modify all connectors after each change in juggler test dependencies. Let me know how to move forward.

Ad 3 - Does it mean that whenever a model property uses `Number`, it must come with a database-specific config saying the backing type should be `int`? Or is this only a problem of `createdAt` field, which contains millisecond timestamp that does not fit 32bit `int`?

Ad 4 - If that's the case, then we can change the test and assert that the value is either `undefined` or `null`. IIRC, one of the hook tests were depending on the behaviour described in these tests, so we may need to fix the hook tests in a similar way too.

raymondfeng:
Ad 1 - I agree.

Ad 2 - I'll comment in  https://github.com/strongloop/loopback-datasource-juggler/pull/456

Ad 3 - It's only a problem for timestamp that doesn't fit into 32 bits. I'll change the type to be `Date`.

Ad 4 - We should change the test cases to allow `null` or `undefined`.

raymondfeng:
https://github.com/strongloop/loopback-datasource-juggler/pull/460

PR #459, opened 2015-02-20
Fix createdAt type so that it won't overflow SQL server int

raymondfeng:
/to @bajtos 
See https://github.com/strongloop/loopback-datasource-juggler/issues/457


bajtos:
LGTM :+1:

PR #470, opened 2015-02-24
Enhance updateAll to coerce the data per property definitions

rudzon:
Coerce ```data``` before updating models.
Relates to https://github.com/strongloop/loopback-connector-mysql/pull/74 .


slnode:

PR #472, opened 2015-02-25
failing test for HasManyThrough relations

clarkbw:
I haven’t found the root cause of this yet, but the test illustrates the issue.  As far as I can tell this seems to be a scope / query problem.

`user.following` functions correctly, returning a list containing only other user
however `user2.followers` should be returning another list containing only the first user

I originally found this by trying to use a type of `user.following.findById` which returned an error, not finding the relation.


clarkbw:
Closing this, the issue still exists but I haven't had the time to dig into it further.  If this is still important to the loopback team feel free to reopen and I'll just unsubscribe.

PR #473, opened 2015-02-26
Deprecate model hooks and events

### LINKED ISSUE #617: nin condition with an empty array produces incorrect sql
I'm using loopback with postgresql-connector.
if `idList` is an empty array, I'm getting sql syntax error from postgres, if it is not empty, everything is OK.

``` javascript
var idList = [];
Model.destroyAll({
  and: [
    {
      parentId: 1
    },
    {
      id: {nin: idList}
    }
  ]
});
```

bajtos:
Deprecate the following [Model hooks](http://docs.strongloop.com/display/LB/Model+hooks):

 - beforeValidate
 - afterValidate
 - beforeCreate
 - afterCreate
 - beforeSave
 - afterSave
 - beforeUpdate
 - afterUpdate
 - beforeDestroy
 - afterDestroy

Deprecate the following Model events:

 - changed
 - deletedAll
 - deleted

Closes strongloop/loopback#617.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ OPTS = --growl
 TESTS = test/*.test.js
 
 test:
-	$(TESTER) $(OPTS) $(TESTS)
+	NO_DEPRECATION=loopback-datasource-juggler $(TESTER) $(OPTS) $(TESTS)
 test-verbose:
 	$(TESTER) $(OPTS) --reporter spec $(TESTS)
 testing:
--- a/lib/browser.depd.js
+++ b/lib/browser.depd.js
@@ -1,0 +1,16 @@
+// A lightweight alternative to "depd" that works in the browser
+module.exports = function depd(namespace) {
+  var warned = {};
+  return function deprecate(message) {
+    if (warned[message]) return;
+    warned[message] = true;
+
+    if (process.noDeprecation) {
+      return;
+    } else if (process.traceDeprecation) {
+      console.trace(namespace, 'deprecated', message);
+    } else {
+      console.warn(namespace, 'deprecated', message);
+    }
+  };
+};
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -1,3 +1,5 @@
+var deprecated = require('depd')('loopback-datasource-juggler');
+
 /*!
  * Module exports
  */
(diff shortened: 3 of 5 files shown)


bajtos:
Waffle: Connects to  strongloop/loopback#617

bajtos:
I added one more commit to fix `depd` in browser. We may want to extract "lib/browser.depd.js" into a standalone module in the future, or perhaps contribute the simpler version to `depd`. I don't have enough time for that right now.

@raymondfeng @ritch ping, is this good to be merged?

bajtos:
I made one more change. Initially, event deprecation messages were printed at the time when the deprecated event was emitted. That's a wrong approach, because `depd` will print a stack showing where the event was emitted, which is not helpful in identifying the place which is listening for this deprecated event. I have reworked the implementation to print the deprecation message from `Model.on()` and cleaned up the commit history along the way.

The PR is ready for review and merge.

ritch:
LGTM

bajtos:
For posterity, the CI build passed on 5 out of 6 machines.

screen shot 2015-02-27 at 19 34 24

bajtos:
@crandmck could you please update the docs and reword the warning to a deprecation note? http://docs.strongloop.com/display/LB/Model+hooks

I am not sure if the we have the model events (`Model.on('changed')`, deleted, deletedAll) documented anywhere. If we do, then please update that page as well.

crandmck:
Added deprecation notice.  I removed 
> However, afterInitialize is still useful.
Should that stay in?

Also removed the following events from http://docs.strongloop.com/display/LB/Events and http://docs.strongloop.com/display/LB/Basic+model+object:
-  changed
- deleted
- deletedAll

Please confirm that is correct, and I will also remove from http://apidocs.strongloop.com/loopback/#model.

bajtos:
@crandmck Thank you.

>  > However, afterInitialize is still useful.

> Should that stay in?

Yes please, `afterInitialize` is not deprecated and still useful. It is also different from the rest of the model hooks because it's synchronous (i.e. there is no callback).

> Also removed the following events from http://docs.strongloop.com/display/LB/Events 

Can that page mention that there are Operation hooks that should (and can) be used instead of the deprecated events?

Also perhaps the page http://docs.strongloop.com/display/LB/Basic+model+object should link to "Operation hooks" instead of "Model hooks"?

crandmck:
OK, added note about afterInitialize, and note about using operation hooks instead of the deprecated events.

PR #476, opened 2015-02-28
Make sure inclusion filter is applied to the target model

raymondfeng:
/to @ritch or @bajtos 
/cc @superkhau 

The PR ensures the scope on `include` is applied to the target model instead of the through model.
  
See https://github.com/strongloop/loopback/issues/1076

### DIFF
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -2066,7 +2066,7 @@ describe('relations', function () {
     var Article, TagName, ArticleTag;
     it('can be declared', function (done) {
       Article = db.define('Article', {title: String});
-      TagName = db.define('TagName', {name: String});
+      TagName = db.define('TagName', {name: String, flag: String});
       Article.hasAndBelongsToMany('tagNames');
       ArticleTag = db.models.ArticleTagName;
       db.automigrate(function () {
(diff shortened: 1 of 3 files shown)


test/relations.test.js
```diff
@@ -2139,6 +2139,58 @@ describe('relations', function () {
     it('should set targetClass on scope property', function() {
       should.equal(Article.prototype.tagNames._targetClass, 'TagName');
     });
+
+    it('should apply inclusion fields to the target model', function(done) {
+      Article.create({title: 'a1'}, function (e, article) {
+        should.not.exist(e);
```
bajtos: Consider replacing `should.not.exist(e)` with `if (e) return done(e);`. The latter gives much more helpful error message.

bajtos:
I cannot evaluate changes in the production code, it looks like a magic and a mud of ball at the same time to me. Considering that you have added new tests for the stuff you are fixing and all existing tests are passing, the patch LGTM as far as I am concerned.

superkhau:
@bajtos LOL, I thought it was a ball of mud too before finally giving up and asking @raymondfeng for help. ;) Glad I wasn't the only one.

Issue #474, opened 2015-02-26
Operational hook 'after delete' should receive id(s)


fabien:
One of the main use-cases for this hook - and it's predecessor even `deleted` - is handling 'cascading deletes' (in NoSQL). However, unlike the soon-to-be-deprecated event, the new operational hook has no exact way to know what has been deleted, except for the where condition, which is not really helpful.

JonnyBGod:
In truth this hook should really receive a copy of the object that was deleted (instance).

One other common use case is to remove related data from other services like files for instance.
For most of these use cases a simple object id is not enough as it probably relies on other stored information such as a path and/or a related model.

Without this information I am unable to grasp the utility for this hook.

fabien:
@JonnyBGod I'll try to fix this

JonnyBGod:
Ok. let me know if you need help.

bajtos:
Let me explain why we are not passing the list of deleted ids or instances in the "after delete" hook. AFAIK, database drivers don't report back the list of id (nor full row data) in the response of a DELETE command. We could implement this at LoopBack side, but that would cause performance degradation for users that don't need this functionality.

IMO, this issue is not a problem in the "after delete" hook per se, but instead a problem of the current design of "deleteAll".

BTW note that if you delete a single model only, the id of the model can be obtained from the where query - see https://github.com/strongloop/loopback/blob/84c65de38ac24e012e3eb8c3f7fa281d565101da/lib/persisted-model.js#L1126-L1139. I think it makes sense to expose this method as a public Model API.

Here is my proposal how to solve this problem:
 - Add an option that will change the behaviour of DELETEs to return the list of deleted ids or instances. I think it makes most sense to read this option from Model settings.
 - Modify the implementation of "deleteById" and "deleteAll" to honour this new option and include the list of ids or instances in the "after delete" hook.
 - Unless there is a driver that provides this functionality OOTB, we will have to always implement it at LoopBack side. In that case we can include the list of ids or instances in the "before delete" hook too.

Example model configurations:

```js
// car.json
{
  "name": "Car",
  "options": {
    // query the full instance data
    "queryOnDelete": "full"
  }
}

// bar.json
{
  "name": "Bar",
  "options": {
    // query the id column only
    "queryOnDelete": "id"
  }
}
```

raymondfeng:
[...]

PR #483, opened 2015-03-04
Fix #481 and #474

### LINKED ISSUE #481: updateAttributes operational hooks: use instance, not where
Because it operates on an instance, you know ...

fabien:
http://docs.strongloop.com/display/public/LB/Operation+hooks will need updates

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1867,7 +1867,7 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
 
   var context = {
     Model: Model,
-    where: byIdQuery(Model, getIdValue(Model, inst)).where,
+    instance: inst,
     data: data
   };
 
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1717,12 +1717,12 @@ DataAccessObject.prototype.remove =
 
       Model.notifyObserversOf(
         'access',
-        { Model: Model, query: byIdQuery(Model, id) },
+        { Model: Model, query: byIdQuery(Model, id), instance: self },
         function(err, ctx) {
           if (err) return cb(err);
           Model.notifyObserversOf(
             'before delete',
-            { Model: Model, where: ctx.query.where },
+            { Model: Model, where: ctx.query.where, instance: self },
             function(err, ctx) {
               if (err) return cb(err);
               doDeleteInstance(ctx.where);
@@ -1730,13 +1730,19 @@ DataAccessObject.prototype.remove =
         });
 
       function doDeleteInstance(where) {
+        var context = {
+          Model: Model,
+          where: where,
+          instance: self
+        };
+        
         if (!isWhereByGivenId(Model, where, id)) {
           // A hook modified the query, it is no longer
           // a simple 'delete model with the given id'.
           // We must switch to full query-based delete.
           Model.deleteAll(where, { notify: false }, function(err) {
             if (err) return cb(err);
-            Model.notifyObserversOf('after delete', { Model: Model, where: where }, function(err) {
+            Model.notifyObserversOf('after delete', context, function(err) {
               cb(err);
               if (!err) Model.emit('deleted', id);
             });
@@ -1751,7 +1757,7 @@ DataAccessObject.prototype.remove =
             }
 
             destroyed(function () {
-              Model.notifyObserversOf('after delete', { Model: Model, where: where }, function(err) {
+              Model.notifyObserversOf('after delete', context, function(err) {
                 cb(err);
                 if (!err) Model.emit('deleted', id);
               });
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -539,7 +539,7 @@ module.exports = function(dataSource, should) {
         existingInstance.updateAttributes({ name: 'changed' }, function(err) {
           if (err) return done(err);
           observedContexts.should.eql(aTestModelCtx({
-            where: { id: existingInstance.id },
+            instance: existingInstance,
             data: { name: 'changed' }
           }));
           done();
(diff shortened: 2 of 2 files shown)


raymondfeng:
LGTM. @bajtos PTAL.

bajtos:
I am very strongly opposed to this change, see https://github.com/strongloop/loopback-datasource-juggler/issues/474#issuecomment-77316105 and https://github.com/strongloop/loopback-datasource-juggler/issues/474#issuecomment-77316105.

bajtos:
I am very strongly opposed to this change, see https://github.com/strongloop/loopback-datasource-juggler/issues/474#issuecomment-77316105 and https://github.com/strongloop/loopback-datasource-juggler/issues/481#issuecomment-77314954.

bajtos:
We can reopen this PR if the discussion in #474 and #481 proves me wrong. In that case this patch requires unit-tests demonstrating how the changes made to `ctx.instance` are handled, especially when both `ctx.data` and `ctx.instance` is changed.

fabien:
@bajtos I will adapt the unit tests - this is a critical flaw that needed to be addressed, so I disagree completely with your strong rejection.

bajtos:
@fabien have you read my comments in the other two issues? Your pull request is a hack fixing the wrong thing in my opinion. I really don't want to end up in the same situation as we have with the old model hooks, where it was not possible to fix them and I had to rewrite them from scratch.

Please don't work on this pull request until we come to an agreement on what's the right way how to address your problem, as otherwise you may just waste your time.

In case I did not make myself clear, I truly want to improve the operations hooks to serve your use case. At the same time, I want to find a solution that will not compromise the current design, will not add even more confusion and will not make the future maintenance more difficult.

For example, the question of using "this" vs. "data" is a commonly asked question about the old hooks and I really don't want to introduce the issue back into the new hooks.

fabien:
@bajtos yes, I read your comments, and I can see what you mean. However, I don't think this PR is moving in the wrong way, it just needs to be fleshed out. Let's go ahead constructively.

Let me remind you that the current working state of operation hooks simply cannot do what the old `hooks` and `events` could do - the context currently doesn't provide the instance that was previously available to work with.

First of all, can we agree on: 

Any hook operating on an instance-level should *also* receive the instance in the context.

I agree, at the moment the PR is incomplete, because it doesn't handle the `instance` vs. `data` situation, and besides resolving this, the docs should be clear on this purpose. What about:

If an instance is passed in the context, it should never be used to manipulate data - it is only there to get the current state (for comparison for example). Any changes would go into the `data` property of the context.

bajtos:
> If an instance is passed in the context, it should never be used to manipulate data - it is only there to get the current state (for comparison for example). Any changes would go into the data property of the context.

This is what I disagree with. One of the ideas behind ORM and OOP in general is that the objects should encapsulate the behaviour and hide implementation details. An example is a property with custom setter that creates a normalised version of the input value, or a prototype method that performs some calculation and updates several properties.

Another example is a code that changes some model properties and then would like to get a value computed from these changed properties.

That's why I believe users should use the instance object to manipulate the data in as many places as possible. Because if they had only the `data` object, then they would have to build a new instance in order to use these prototype methods.

Also because we perform a validation of the updated model instance, in many cases we would have to 1) extract data from the instance before calling the hooks, 2) pass the data back to the instance for validation. At the moment, we can skip 1) because hooks are manipulating the instance directly.

> Any hook operating on an instance-level should also receive the instance in the context.

There are two basic ways how the hook may want to use the instance:

A) It wants to modify the data that will be persisted to the database, this is specific to "before save" hook. We must make it very clear whether the save will update the whole record (`ctx.instance` ATM) or perform a partial update (`ctx.data` ATM).

B) It wants to get more information about the model instance affected. IIRC, there are only two cases where we are not exposing this data even though we have it available: `prototype.remove` and `prototype.updateAttributes`.

Let's focus at `updateAttributes` first. For hooks interested in modifying the data (A), the API must make it very clear that there is no instance to manipulate, the hook must update the partial `data` object instead. Therefore we cannot use `ctx.instance` to provide the readonly info for (B), because that's inconsistent with the way all operations work, it means one can no longer check `if(ctx.instance)` to decide whether he has a full instance or partial data for modifications.

When it comes to `prototype.remove`, there are no restrictions similar to what we have in `updateAttributes`. However, I am reluctant to use `ctx.instance` simply because it will be inconsistent with the way how "before save" use `ctx.instance`.

My conclusion is that we need to come up with a new property name that will be set in the where+data case and that will make it very clear that the changes made to this instance will be ignored by the operation. Perhaps `ctx.targetInstance` would be a good name?

Thoughts? I'd like to hear @raymondfeng opinion too.

fabien:
@bajtos this is another way to look at it, but it is fine with me, as long as we make a clear distinction.

From what I've learned during refactoring to operation hooks, it is indeed just the `prototype.updateAttributes` and `prototype.remove` methods that need a fix.

Say we'd go with `targetInstance` which we'd only use current read-only state, wouldn't it be better to have `targetData`, which wouldn't even be an actual instance? Just to be fully clear/safe?

bajtos:
> Say we'd go with targetInstance which we'd only use current read-only state, wouldn't it be better to have targetData, which wouldn't even be an actual instance? Just to be fully clear/safe?

My concern is that converting the instance to data is not free, it takes some time and allocates extra memory. I don't want all users to pay this price.

Can we come up with a better name that would make the intent more clear than `targetInstance`? Perhaps `affectedInstance`?

fabien:
What about `currentInstance` it denotes a sense of unchanged 'state' better than `targetInstance`, I think.

bajtos:
> What about currentInstance it denotes a sense of unchanged 'state' better than targetInstance, I think.

Sounds good too. Let's go with `currentInstance` then and see what @ritch and @raymondfeng think of it.

fabien:
@bajtos now that PR #485 is here - can I work on top of this or do I need to go separate and fix conflicts later (there will be quite some, because I'll work on the same parts of the code now).

fabien:
@bajtos will `currentInstance` only be passed in a `before save` + `updateAttributes` scenario? Can we identify where `currentInstance` comes into play?

fabien:
Closing this in favor of #486

Issue #484, opened 2015-03-05
Shared state for hooks invoked for a single operation


bajtos:
Add a context property that will be preserved across all hooks invoked for a single operation.

For example, both "access", "before save" and "after save" invoked for `Model.create()` should have the same object passed in `ctx.hookState`.

This way the hooks can pass state date between "before" and "after" hook.

/cc @fabien

fabien:
@bajtos I would like to give this - and `targetData` a shot - unless you have time to do this now?

bajtos:
@fabien I don't have time for that now, I'll reassign to you. Can you please send `targetState` in a standalone pull request? I think this part will pretty much unopinionated, thus the pull request can be merged sooner.

What is the idea behind `targetData`? If you are referring to our discussion in #483, then let's keep the comments there and don't spread them across multiple issues.

fabien:
@bajtos yes, let's keep `targetData` at #483

bajtos:
Implemented by #485.

PR #485, opened 2015-03-05
Implement #484 : hookState

fabien:
This implements #484

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -231,9 +231,8 @@ DataAccessObject.create = function (data, options, cb) {
   Model = this.lookupModel(data); // data-specific
   if (Model !== obj.constructor) obj = new Model(data);
 
-  Model.notifyObserversOf('before save', {
-    Model: Model, instance: obj, hookState: hookState
-  }, function(err) {
+  var context = { Model: Model, instance: obj, hookState: hookState };
+  Model.notifyObserversOf('before save', context, function(err) {
     if (err) return cb(err);
 
     data = obj.toObject(true);
@@ -271,9 +270,8 @@ DataAccessObject.create = function (data, options, cb) {
               if (err) {
                 return cb(err, obj);
               }
-              Model.notifyObserversOf('after save', {
-                Model: Model, instance: obj, hookState: hookState
-              }, function(err) {
+              var context = { Model: Model, instance: obj, hookState: hookState };
+              Model.notifyObserversOf('after save', context, function(err) {
                 cb(err, obj);
                 if(!err) Model.emit('changed', obj);
               });
@@ -357,9 +355,8 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
     return this.create(data, options, cb);
   }
 
-  Model.notifyObserversOf('access', { 
-    Model: Model, query: byIdQuery(Model, id), hookState: hookState
-  }, doUpdateOrCreate);
+  var context = { Model: Model, query: byIdQuery(Model, id), hookState: hookState };
+  Model.notifyObserversOf('access', context, doUpdateOrCreate);
 
   function doUpdateOrCreate(err, ctx) {
     if (err) return cb(err);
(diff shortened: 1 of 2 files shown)


fabien:
@bajtos @raymondfeng thoughts?

bajtos:
@raymondfeng @ritch If you have any objections against `hookState`, then speak now. Otherwise I'll review and land this PR later today/early tomorrow CET.

lib/dao.js
```diff
@@ -230,7 +231,9 @@ DataAccessObject.create = function (data, options, cb) {
   Model = this.lookupModel(data); // data-specific
   if (Model !== obj.constructor) obj = new Model(data);
 
-  Model.notifyObserversOf('before save', { Model: Model, instance: obj }, function(err) {
+  Model.notifyObserversOf('before save', {
+    Model: Model, instance: obj, hookState: hookState
+  }, function(err) {
```
bajtos: coding style: please move the context initialisation to a new line and property.

```
var context = ...;
Model.notifyObserversOf('before save', context, function(err) {
```

The same fix should be applied to all other places where the context initialization does not fit a single line.

IIRC, we are already using the style described above in other existing code, so let's keep the style consistent.

test/persistence-hooks.suite.js
```diff
@@ -1095,6 +1095,31 @@ module.exports = function(dataSource, should) {
         });
       });
 
+      it('applies propagates hookState from `before delete` to `after delete` hook', function(done) {
```
bajtos: Applies or propagates? Pick only one please.

Also please add a check that the hookState is propagated from the "access" hook too.

test/persistence-hooks.suite.js
```diff
@@ -1202,9 +1227,14 @@ module.exports = function(dataSource, should) {
...
 
-    function pushContextAndNext() {
+    function pushContextAndNext(fn) {
       return function(context, next) {
+        if (typeof fn === 'function') {
+          fn(context);
+        }
+
         context = deepCloneToObject(context);
+        context.hookState.test = true;
```
bajtos: I find this confusing, why do you need `hookState.test`?

test/persistence-hooks.suite.js
```diff
@@ -1246,6 +1276,9 @@ module.exports = function(dataSource, should) {
 
     function aTestModelCtx(ctx) {
       ctx.Model = TestModel;
+      if (!ctx.hookState) {
+        ctx.hookState = { test: true };
+      }
```
bajtos: Same comment as above, why do you need to set `test:true`?

test/persistence-hooks.suite.js
fabien: This is just a variable that will be set by all invocations of `pushContextAndNext` and thus, it will make sure that all tests will verify that the hookState is working (and available to all known hooks that invoke `pushContextAndNext`, which should be all IIRC).

lib/dao.js
fabien: OK.

test/persistence-hooks.suite.js
fabien: See below, `context.hookState.test` is tested in *any* hook.

fabien:
@bajtos I'm running out of time here, with zero progress today (and lost work from yesterday).

I moved ahead with #486 - how I can I make sure I don't have to redo that?

fabien:
@bajtos reformatted and fixed - except for `hookState.test`, which should remain - perhaps be renamed, but otherwise is required to test all hooks.

ritch:
No qualms here. Just a reminder that we need an example for this.

raymondfeng:
LGTM.

fabien:
OK, can anyone merge then please?

test/persistence-hooks.suite.js
bajtos: This is very convoluted way of testing this feature. The other tests for Operations hook strive to make it very clear what is being tested (what features are supported) and also to make each test as small and isolated as feasible.

bajtos:
I am very unhappy about this merge, since when do we prefer quick merges at the cost of code quality? The code coverage is low, the fact that the state is shared between hooks is tested only for before/after delete hooks. The technique using `hookState.test` is too magic, it will be difficult to understand for anybody not familiar with this pull request (including us in few months) and it make the future maintenance more difficult.

PR #486, opened 2015-03-05
Operation hooks improvements

fabien:
This is a combined PR - because of the expected conflicts/overlap of changes.

It implements the following improvements:

To maintain 'state' between `before` and `after` hooks, a `hookState` (object) property was introduced to `context`.

For consistency, all instance-level operations (`prototype.updateAttributes` and `prototype.delete`) will now have access to the current instance as part of the hook's `context`:

- `prototype.save` + `before save`:  the instance will be available as `instance` (unchanged behavior).

- `prototype.save` + `after save`:  the instance will be available as `instance` (unchanged behavior).

- `prototype.updateAttributes` + `before save`: because the instance should not be manipulated directly, the instance will be available as **currentInstance** to distinguish between the `instance` that *can* be directly modified (it is omitted now). In other words, `currentInstance` provides the current 'state' of the instance, and it should be regarded as immutable. Any modifications should be applied to `context.data`, which incorporates the partial nature of the `updateAttributes` operation.

- `prototype.updateAttributes` + `after save`: in *after* hooks, this difference is not important, and thus, the instance will be simply be available as `instance`.

- `prototype.remove/destroy/delete` + `before save`: again, the difference is irrelevant, so the instance will be available as `instance`.

- `prototype.remove/destroy/delete` + `after save`:  the instance will be available as `instance`.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1732,7 +1732,7 @@ DataAccessObject.prototype.remove =
       assert(typeof options === 'object', 'The options argument should be an object');
       assert(typeof cb === 'function', 'The cb argument should be a function');
 
-      var self = this;
+      var inst = this;
       var Model = this.constructor;
       var id = getIdValue(this.constructor, this);
       var hookState = {};
@@ -1743,7 +1743,9 @@ DataAccessObject.prototype.remove =
 
       Model.notifyObserversOf('access', context, function(err, ctx) {
           if (err) return cb(err);
-          var context = { Model: Model, where: ctx.query.where, hookState: hookState };
+          var context = { 
+            Model: Model, where: ctx.query.where, instance: inst, hookState: hookState
+          };
           Model.notifyObserversOf('before delete', context, function(err, ctx) {
               if (err) return cb(err);
               doDeleteInstance(ctx.where);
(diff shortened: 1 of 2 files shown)


fabien:
@bajtos @raymondfeng please review

fabien:
This should supersede #485

fabien:
The next step would be some 'canned' before hooks which will do the following:

```
Model.observe('before delete', fetchInstances); // OR:

Model.observe('before delete', fetchIds);

function fetchInstances(ctx, next) { // queryOnDelete: 'full'
  ctx.Model.find({ where: ctx.where }, function(err, instances) {
    if (err) return next(err);
    ctx.hookState.instances = instances;
    next();
  });
};

function fetchIds(ctx, next) { // queryOnDelete: 'id'
  ctx.Model.find({ where: ctx.where, fields: ['id'] }, function(err, instances) {
    if (err) return next(err);
    ctx.hookState.ids = _.pluck(instances, 'id');
    next();
  });
};

```

Perhaps we can even allow this syntax: `Model.observe('before delete', 'presetHookFnName');`

And: `Model.registerObserver('presetHookFnName', function(ctx, next) { ... })`.

This would be the basis for https://github.com/strongloop/loopback-datasource-juggler/issues/474#issuecomment-77316105

fabien:
I'll make a new PR

PR #488, opened 2015-03-05
Improve instance-level operation hooks

fabien:
For consistency, all instance-level operations (`prototype.updateAttributes` and `prototype.delete`) will now have access to the current instance as part of the hook's `context`:

- `prototype.save` + `before save`:  the instance will be available as `instance` (unchanged behavior).

- `prototype.save` + `after save`:  the instance will be available as `instance` (unchanged behavior).

- `prototype.updateAttributes` + `before save`: because the instance should not be manipulated directly, the instance will be available as **currentInstance** to distinguish between the `instance` that *can* be directly modified (it is omitted now). In other words, `currentInstance` provides the current 'state' of the instance, and it should be regarded as immutable. Any modifications should be applied to `context.data`, which incorporates the partial nature of the `updateAttributes` operation.

- `prototype.updateAttributes` + `after save`: in *after* hooks, this difference is not important, and thus, the instance will be simply be available as `instance`.

- `prototype.remove/destroy/delete` + `before save`: again, the difference is irrelevant, so the instance will be available as `instance`.

- `prototype.remove/destroy/delete` + `after save`:  the instance will be available as `instance`.

### DIFF
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -535,12 +535,15 @@ module.exports = function(dataSource, should) {
       it('triggers `before save` hook', function(done) {
         TestModel.observe('before save', pushContextAndNext());
 
-        existingInstance.name = 'changed';
+        var currentInstance = deepCloneToObject(existingInstance);
+
         existingInstance.updateAttributes({ name: 'changed' }, function(err) {
           if (err) return done(err);
+          existingInstance.name.should.equal('changed');
           observedContexts.should.eql(aTestModelCtx({
             where: { id: existingInstance.id },
-            data: { name: 'changed' }
+            data: { name: 'changed' },
+            currentInstance: currentInstance
           }));
           done();
         });
@@ -736,10 +739,24 @@ module.exports = function(dataSource, should) {
           { id: existingInstance.id, name: 'updated name' },
           function(err, instance) {
             if (err) return done(err);
-            observedContexts.should.eql(aTestModelCtx({
-              where: { id: existingInstance.id },
-              data: { id: existingInstance.id, name: 'updated name' }
-            }));
+            if (dataSource.connector.updateOrCreate) {
+              // Atomic implementations of `updateOrCreate` cannot
+              // provide full instance as that depends on whether
+              // UPDATE or CREATE will be triggered
+              observedContexts.should.eql(aTestModelCtx({
+                where: { id: existingInstance.id },
+                data: { id: existingInstance.id, name: 'updated name' }
+              }));
+            } else {
+              // currentInstance is set, because a non-atomic `updateOrCreate`
+              // will use `prototype.updateAttributes` internally, which
+              // exposes this to the context
+              observedContexts.should.eql(aTestModelCtx({
+                where: { id: existingInstance.id },
+                data: { id: existingInstance.id, name: 'updated name' },
+                currentInstance: existingInstance
+              }));
+            }
             done();
           });
       });
@@ -1026,7 +1043,8 @@ module.exports = function(dataSource, should) {
         existingInstance.delete(function(err) {
           if (err) return done(err);
           observedContexts.should.eql(aTestModelCtx({
-             where: { id: existingInstance.id }
+           where: { id: existingInstance.id },
+           instance: existingInstance
           }));
           done();
         });
@@ -1068,7 +1086,8 @@ module.exports = function(dataSource, should) {
         existingInstance.delete(function(err) {
           if (err) return done(err);
           observedContexts.should.eql(aTestModelCtx({
-            where: { id: existingInstance.id }
+            where: { id: existingInstance.id },
+            instance: existingInstance
           }));
           done();
         });
@@ -1109,11 +1128,13 @@ module.exports = function(dataSource, should) {
           observedContexts.should.eql([
             aTestModelCtx({ 
               hookState: { foo: 'bar', test: true },
-              where: { id: '1' }
+              where: { id: '1' },
+              instance: existingInstance
             }),
             aTestModelCtx({ 
               hookState: { foo: 'BAR', test: true },
-              where: { id: '1' }
+              where: { id: '1' },
+              instance: existingInstance
             })
           ]);
           done();
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1732,18 +1732,23 @@ DataAccessObject.prototype.remove =
       assert(typeof options === 'object', 'The options argument should be an object');
       assert(typeof cb === 'function', 'The cb argument should be a function');
 
-      var self = this;
+      var inst = this;
       var Model = this.constructor;
       var id = getIdValue(this.constructor, this);
       var hookState = {};
 
       var context = { 
-        Model: Model, query: byIdQuery(Model, id), hookState: hookState 
+        Model: Model, query: byIdQuery(Model, id), hookState: hookState
       };
 
       Model.notifyObserversOf('access', context, function(err, ctx) {
           if (err) return cb(err);
-          var context = { Model: Model, where: ctx.query.where, hookState: hookState };
+          var context = {
+            Model: Model,
+            where: ctx.query.where,
+            instance: inst,
+            hookState: hookState
+          };
           Model.notifyObserversOf('before delete', context, function(err, ctx) {
               if (err) return cb(err);
               doDeleteInstance(ctx.where);
@@ -1757,7 +1762,12 @@ DataAccessObject.prototype.remove =
           // We must switch to full query-based delete.
           Model.deleteAll(where, { notify: false }, function(err) {
             if (err) return cb(err);
-            var context = { Model: Model, where: where, hookState: hookState };
+            var context = {
+              Model: Model,
+              where: where,
+              instance: inst,
+              hookState: hookState
+            };
             Model.notifyObserversOf('after delete', context, function(err) {
               cb(err);
               if (!err) Model.emit('deleted', id);
@@ -1766,14 +1776,19 @@ DataAccessObject.prototype.remove =
           return;
         }
 
-        self.trigger('destroy', function (destroyed) {
-          self._adapter().destroy(self.constructor.modelName, id, function (err) {
+        inst.trigger('destroy', function (destroyed) {
+          inst._adapter().destroy(inst.constructor.modelName, id, function (err) {
             if (err) {
               return cb(err);
             }
 
             destroyed(function () {
-              var context = { Model: Model, where: where, hookState: hookState };
+              var context = {
+                Model: Model,
+                where: where,
+                instance: inst,
+                hookState: hookState
+              };
               Model.notifyObserversOf('after delete', context, function(err) {
                 cb(err);
                 if (!err) Model.emit('deleted', id);
@@ -1893,6 +1908,7 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
     Model: Model,
     where: byIdQuery(Model, getIdValue(Model, inst)).where,
     data: data,
+    currentInstance: inst,
     hookState: hookState
   };
 


fabien:
The next step would be some 'canned' before hooks which will do the following:

```
Model.observe('before delete', fetchInstances); // OR:

Model.observe('before delete', fetchIds);

function fetchInstances(ctx, next) { // queryOnDelete: 'full'
  ctx.Model.find({ where: ctx.where }, function(err, instances) {
    if (err) return next(err);
    ctx.hookState.instances = instances;
    next();
  });
};

function fetchIds(ctx, next) { // queryOnDelete: 'id'
  ctx.Model.find({ where: ctx.where, fields: ['id'] }, function(err, instances) {
    if (err) return next(err);
    ctx.hookState.ids = _.pluck(instances, 'id');
    next();
  });
};

```

Perhaps we can even allow this syntax: `Model.observe('before delete', 'presetHookFnName');`

And: `Model.registerObserver('presetHookFnName', function(ctx, next) { ... })`.

This would be the basis for https://github.com/strongloop/loopback-datasource-juggler/issues/474#issuecomment-77316105

fabien:
@bajtos @raymondfeng @ritch please review

bajtos:
> prototype.remove/destroy/delete + before save: again, the difference is irrelevant, so the instance will be available as instance.

I prefer to keep this consistent with "before save" and use `currentInstance` name.

> prototype.updateAttributes + after save: in after hooks, this difference is not important, and thus, the instance will be simply be available as instance.
> prototype.remove/destroy/delete + after save: the instance will be available as instance.

At one hand, nobody should be really expecting that changes to `ctx.instance` made in an "after" hook will be persisted. On the other hand, it introduces inconsistency. Until now, the recommendation was to check for `ctx.instance` and if it was not there, then the user should expect `ctx.where` and `ctx.data` (or `ctx.where` for "delete" hooks). With the proposal here, this rule no longer apply and one has to use different rules for different hooks.

In the light of the above, I prefer to consistently use `ctx.currentInstance` in all places where this PR adds this new context property.

fabien:
OK, I can live with that. Any other thoughts? If not, I'll apply the changes requested here.

test/persistence-hooks.suite.js
```diff
@@ -736,10 +739,18 @@ module.exports = function(dataSource, should) {
...
           function(err, instance) {
             if (err) return done(err);
-            observedContexts.should.eql(aTestModelCtx({
-              where: { id: existingInstance.id },
-              data: { id: existingInstance.id, name: 'updated name' }
-            }));
+            if (dataSource.connector.findOrCreate) {
+              observedContexts.should.eql(aTestModelCtx({
+                where: { id: existingInstance.id },
+                data: { id: existingInstance.id, name: 'updated name' }
```
bajtos: Typo, you are executing `updateOrCreate` but testing for `findOrCreate`.

Atomic findOrCreate does not set `ctx.currentInstance` - we need to capture this in the documentation.

fabien: Actually, it's not a typo - whenever an optimized (dataSource.connector.findOrCreate) call can be made, the `currentInstance` will not be available. However, in the normal, unoptimized scenario, `updateAttributes` is invoked internally, which *does* set `currentInstance`. This is why I had to make a distinction here.

bajtos:
See the comment above. Other than the comments above, the implementation looks good.

I'd like to hear from @ritch or @raymondfeng whether the name `currentInstance` is good enough (whether it makes clear what's it purpose), as they have fresh eyes.

My internet connection is poor in the evenings so I won't be able to continue this discussion until tomorrow.

fabien:
@bajtos hope we can merge first thing in the morning. I won't be able to do much tonight, and I'm afraid I'll need a full day to port my code over ...

bajtos:
Frankly, expecting non-trivial pull requests to get landed the same day is unreasonable, I even dare to say silly. Please expect at least 2-3 working days.

fabien:
@bajtos meanwhile I'll fiddle with my thumbs, OK? I have this hunch that nothing ever is really trivial in your opinion ... thanks anyway :-1:

ritch:
I'm not worried about the name... my concern is we are adding yet another case you have to handle when using a hook. What would alleviate my concern is if you can use the same assumptions (eg. `where` vs `instance`) and not even know `currentInstance` exists.

Also... Why do we need `currentInstance` to do these:

```js
Model.observe('before delete', fetchInstances); // OR:
Model.observe('before delete', fetchIds);
``` 

or is it **only** in the case of `after save`?

fabien:
@ritch I tend to agree with you. Initially I only had `currentInstance` for the specific `updateAttributes` case. It's the only situation where it really matters.

Ideally, we would just forget about the `currentInstance` denominator and intelligently diff the changes to `instance` when performing `updateAttributes`. In that case `instances` could be the actual instance (===) or a temporary copy, used for accepting the changes, which then would be extracted for the partial update.

If we can agree on such a solution, I'd be happy to work on it in the morning. I really need to know which direction we're heading with this - I'm stuck refactoring a huge code-base to be operation hooks compatible - and this Friday is an important deadline for it. @raymondfeng thoughts?

raymondfeng:
For the `save` context, I think there should be the following properties:

1. Model - the model class (readonly)
2. data - the data object to be sent to DB (this is one that hooks should mutate to influence the `save`)
3. instance - __ideally__ it should be the object instance which the calling prototype method is on (readonly)
4. where - the where clause of the criteria
5. options - the options object (which might be used to propagate transactions later on)

The main concern for 3 is back-compatibility since `instance` is anti-intuitively used in the current implementation.

fabien:
@raymondfeng if I gather correctly, this is similar to https://github.com/strongloop/loopback-datasource-juggler/pull/483#issuecomment-77327773

If an instance is passed in the `save` context, it should **never** be used to manipulate data - it is only there to get the current state (for comparison for example). Any changes would go into the data property of the context. Making it consistent with the batch operations as well.

fabien:
However, working with `data` directly is rather low-level, because at the instance level there may be a setter, computed properties etc. that might change as a result of updated values. Therefor, I would suggest to with the solution above: https://github.com/strongloop/loopback-datasource-juggler/pull/488#issuecomment-77468296

When available, use `instance` directly for any modifications, and in the case of `updateAttributes` extract the diff to perform the partial update as usual.

raymondfeng:
@fabien Do you agree on the following?
For `prototype.save`, `data` is the same as `instance`.
For `prototype.updateAttributes`, `data` is the changes to be applied while `instance` is the model instance itself.  (I think there is a bug in the current implementation which applies changes to the instance before making a DB call).
For `prototype.delete`, `data` is undefined while `instance` is the model instance itself.

bajtos:
Let me cross-post my earlier comments explaining why I think it's bad idea to send data & instance as proposed by @raymondfeng.

https://github.com/strongloop/loopback-datasource-juggler/pull/483#issuecomment-77332635
> If an instance is passed in the context, it should never be used to manipulate data - it is only there to get the current state (for comparison for example). Any changes would go into the data property of the context.

This is what I disagree with. One of the ideas behind ORM and OOP in general is that the objects should encapsulate the behaviour and hide implementation details. An example is a property with custom setter that creates a normalised version of the input value, or a prototype method that performs some calculation and updates several properties.

Another example is a code that changes some model properties and then would like to get a value computed from these changed properties.

That's why I believe users should use the instance object to manipulate the data in as many places as possible. Because if they had only the data object, then they would have to build a new instance in order to use these prototype methods.

Also because we perform a validation of the updated model instance, in many cases we would have to 1) extract data from the instance before calling the hooks, 2) pass the data back to the instance for validation. At the moment, we can skip 1) because hooks are manipulating the instance directly.

https://github.com/strongloop/loopback-datasource-juggler/pull/483#issuecomment-77335731

My concern is that converting the instance to data is not free, it takes some time and allocates extra memory. I don't want all users to pay this price.

---

Also I believe the hook implementations should have the information whether a partial or a full update will be performed, because the partial update may need special handling.

E.g. when applying partial update of "firstName" only on a model with "firstName", "lastName" and computed "fullName",  then the hook has to do one of:

 - reject the update, because we don't have lastName to calculate "fullName" - this is the case for batch updateAll
 - modify the update data and include unmodified "lastName" and recomputed "fullName" - this is the case for instance updateAttributes

---

@fabien I would really appreciate if you can keep the work in the same pull request the next time, so that the conversation stays in single place.

---

> @ritch I'm not worried about the name... my concern is we are adding yet another case you have to handle when using a hook. What would alleviate my concern is if you can use the same assumptions (eg. where vs instance) and not even know currentInstance exists.

See the second part of https://github.com/strongloop/loopback-datasource-juggler/pull/483#issuecomment-77332635

> Any hook operating on an instance-level should also receive the instance in the context.

There are two basic ways how the hook may want to use the instance:

A) It wants to modify the data that will be persisted to the database, this is specific to "before save" hook. We must make it very clear whether the save will update the whole record (`ctx.instance` ATM) or perform a partial update (`ctx.data` ATM).

B) It wants to get more information about the model instance affected. IIRC, there are only two cases where we are not exposing this data even though we have it available: `prototype.remove` and `prototype.updateAttributes`.

Let's focus at `updateAttributes` first. For hooks interested in modifying the data (A), the API must make it very clear that there is no instance to manipulate, the hook must update the partial `data` object instead. Therefore we cannot use `ctx.instance` to provide the readonly info for (B), because that's inconsistent with the way all operations work, it means one can no longer check `if(ctx.instance)` to decide whether he has a full instance or partial data for modifications.

When it comes to `prototype.remove`, there are no restrictions similar to what we have in `updateAttributes`. However, I am reluctant to use `ctx.instance` simply because it will be inconsistent with the way how "before save" use `ctx.instance`.

My conclusion is that we need to come up with a new property name that will be set in the where+data case and that will make it very clear that the changes made to this instance will be ignored by the operation. Perhaps `ctx.targetInstance` would be a good name?

bajtos:
One more thought: while the current design of operation hooks may be perceived as too complex, it reflect the complexity of the DAO methods provided by LoopBack and forces users to think more about the different scenarios in which the hook may be invoked.

I am concerned that if we simplify the API too much, it may make it too easy to shoot oneself in the foot.

raymondfeng:
I think there are some confusions about the terms we use. Let me first clarify what I use:

1. instance - the receiver of the method if it is a prototype method, for example, the `instance` will be `order` for `order.updateAttributes()`.
2. data - the data to be sent to the underlying connector for DB operations. Please note the `data` can be a plain JS object or an instance of the target model class (I believe that @bajtos uses `instance` for this case).

bajtos:
> @raymond For `prototype.save`, `data` is the same as `instance`.

This will break the contract that `ctx.instance` is immutable and does not change, because changes to `ctx.data` will be propagated to `ctx.instance`.

Other than that, I think this may be a reasonable compromise. I'd rather use a different name for `ctx.instance`, e.g. `ctx.targetInstance`, in order to preserve backwards compatibility with the current implementation.

Should the context include both `where` and `ctx.targetInstance`, or are these two properties mutually exclusive?

fabien:
As I see it there are two ways to provide a consistent API - we'd have to choose one:

1. whenever `instance` is available, you'd use it directly (`save` and `updateAttributes`, the latter will do a diff and perform the partial update)

2. whenever `data` is available, you'd use it directly. If there's a `instance` (heck, even `currentInstance`) available, you can choose to inspect it for your modifications to `data`

bajtos:
@raymondfeng IIUC, if we go with your proposal, then `ctx.data` is an instance of the target model class in case of a full update, and a plain JS object in case of a partial update?

In other words, the following is correct?

```js
var isFullUpdate = ctx.data instanceof ctx.Model;
```

raymondfeng:
> var isFullUpdate = ctx.data instanceof ctx.Model;
Yes.

fabien:
After giving this more thought, I think option 1 (and thus https://github.com/strongloop/loopback-datasource-juggler/pull/488#issuecomment-77513842) works most consistently. Principle Of Least Surprise, you know ...

raymondfeng:
What about the following?

1. Add `self` to refer to the JS object instance that owns the prototype method
2. `self` and `where` are exclusive
3. Keep `instance` and `data` as-is (even though I really feel the term of `instance` is confusing)

bajtos:
> For prototype.updateAttributes, data is the changes to be applied while instance is the model instance itself. (I think there is a bug in the current implementation which applies changes to the instance before making a DB call).

As far as I understand the current implementation, the changes are applied to the instance so that the instance can be validated. Which I find a bit weird, because we can't guarantee that the data we see in LoopBack will be the same that will end up in the database, since updateAttributes is making only a partial update.

Perhaps if we have `prototype.update`, performing a full property update, then we can remove the validation from `updateProperties` and keep it very lightweight partial update, similar to `updateAll`.

Another benefit of `prototype.update` is that the hooks will naturally receive `ctx.instance`, because we are sending the full instance to the database.

fabien:
Note that `instance` is not actually immutable (not enforced) and it is currently the recommended way to modify during `prototype.save`. If we can extend this to `updateAttributes` and handle the partial update transparently.

fabien:
@raymondfeng I personally like `instance` better than `self`.

raymondfeng:
@fabien same here.

bajtos:
> If we can extend this to updateAttributes and handle the partial update transparently.

As I commented elsewhere, this has performance costs. Are we willing to place them on all users, even on those that don't need the functionality provided by hooks?

Also note that you will need to detect the cases where the hook added more properties or wants to remove properties from the partial data object.

We were discussing this approach in the past, it ended up with things like `ctx.instance` and `ctx.listOfPropertiesToUpdate`, which was rather ugly.

fabien:
OK, what about option 2. above then?

bajtos:
@fabien I am wondering why is `prototype.updateAttributes` so important to you. Are you relying on the update being partial? If there was `prototype.update` method I have described above, could you use it in your code instead of `prototype.updateAttributes`? Because you know, it would provide `ctx.instance` out of the box, as that's consistent with the current design of hooks.

raymondfeng:
> Perhaps if we have prototype.update, performing a full property update, then we can remove the validation from updateProperties and keep it very lightweight partial update, similar to updateAll

Is it necessary as we already have `prototype.save`? If yes, I'm fine to add `prototype.update` and maps it to `PUT /:id`. We should then map `prototype.updateAttributes` to `PATCH /:id`.

fabien:
@bajtos I think partial updates are mandatory, and what you'd expect if you update data. You don't always have a full object at your disposal to send back over the wire. And moving partial updates to PATCH is a huge BC break @raymondfeng

raymondfeng:
@bajtos prototype.updateAttributes is very important. In most cases, we are doing partial updates. For example, changing the e-mail, changing the balance.

raymondfeng:
> And moving partial updates to PATCH is a huge BC break

I'm thinking ahead on LoopBack 3.0 :-)

fabien:
@bajtos @raymondfeng please leave `update` vs. `updateAttributes` to a separate discussion, however related it might seem to solve this. Let's focus on current API and maintaining functionality that we need to restore with the new operation hooks.

fabien:
I've updated https://github.com/strongloop/loopback-datasource-juggler/pull/488#issuecomment-77516113 with a common pattern I faced. If we'd go for option 2, and only ever allow modifications of `cx.data` we'll have consistency out of the box.

fabien:
With option 2, you can essential do a diff (as @bajtos pointed out as being expensive to enable across the board) yourself, and use it to set `ctx.data`.

bajtos:
> Is it necessary as we already have prototype.save? 

`prototype.save` does not accept a (partial) object with data to apply, therefore you can't use it via REST API.

```js
// mock up
DAO.prototype.update = function(data, options, cb) {
  this.setAttributes(data);
  this.save(options, cb);
}
```

> If yes, I'm fine to add prototype.update and maps it to PUT /:id. We should then map prototype.updateAttributes to PATCH /:id.

I am all for that, but since it's a breaking change, it has to wait until 3.0.

> `prototype.updateAttributes` is very important. In most cases, we are doing partial updates. For example, changing the e-mail, changing the balance.

I know, partial updates are very powerful. However, with great power comes great responsibility, and one has to be more careful about the application (and hook) design.

On the other hand, the typical usage of the updateAttributes in the Angular SDK is to call `instance.$save`, which will replace all data anyways.

So I'd like to have full update that can be used by clients not caring about partial updates, with the benefit that they hooks can stay simple in this case, as they have `ctx.instance` available.

Then there is partial update via `updateAttributes`, in which case the hooks must be designed more carefully.

> @bajtos @raymondfeng please leave update vs. updateAttributes to a separate discussion, however related it might seem to solve this. Let's focus on current API and maintaining functionality that we need to restore with the new operation hooks.

The reason I am asking is that if you can switch to `update` and we can keep the "before save" hook of updateAttributes unchanged, then there is less things to discuss.

> self and where are exclusive

That is a breaking change, isn't it?

Is exclusive `self` that much better than the initial proposal that started this pull request, which was `ctx.targetInstance` or `ctx.currentInstance` that is optionally added to the existing context?

fabien:
@raymondfeng LB 3.0 should not even break BC so boldly as this, I think. It would change the semantics of the current implementation - it would be better to introduce a new remote call instead.

fabien:
Can we stick to the discussion please? I want to hear about option 2. as proposed.

raymondfeng:
@fabien Can you summarize option 2 here? I'm lost :-)

fabien:
@raymondfeng sure, it's easy to get lost here indeed

What I meant:

Whenever `ctx.data` is available, you'd use it directly. If there's a `ctx.instance` available, you can choose to inspect it **read only** for your modifications to data. You **never** modify `ctx.instance` directly (`updateAttributes`), or at least we don't recommend it (`save`), ever.

This is highly consistent, and saves you from this pattern:

```
if (ctx.instance) {
   setSomeAttributes(ctx.instance);
} else {
   setSomeAttributes(ctx.data);
}
```

And should perform more elaborate manipulation of `cx.data` you can perform this yourself (you can essentialy do a diff - as @bajtos pointed out as being expensive to enable across the board - yourself, and use it to set cx.data. Should you need to ...)

raymondfeng:
@fabien In your proposal, does `instance` point to `this` or it's whatever we set today?

fabien:
@raymondfeng yes, `ctx.instance` is the current instance (before op hooks, it was `this`), but of course it's only there for `prototype` calls. It's can also act as a flag to distinguish between this, if needed.

fabien:
@bajtos @raymondfeng I updated my comment above

raymondfeng:
@fabien I think we have the same proposal. But @bajtos is concerned that changing `instance` to `this` is a breaking change. That's why I started to propose `self`.

fabien:
@raymondfeng great that we agree - however, I would not change `instance` to anything else. It's the name we use throughout LB (think, remote method hooks/remoting).

raymondfeng:
Yes, but the fact is that we use `instance` in the recently added hooks for a different thing. Reverting it back to `this` is going to a breaking change.

fabien:
We're not going back to `this` are we? I don't think there will be breaking changes, with the proposal above. The only case `ctx.instance` could be manipulated was during `save` - this will still work as before, so no BC break there. However, we need to express in the docs that this is not recommended any longer and that `cx.instance` should be regarded as **immutable** from now on (once we settle this).

raymondfeng:
@fabien For static methods, such as create, what is ctx.instance in your proposal?

fabien:
@raymondfeng it's undefined of course - that's why `cx.instance` can help distinguish between static and prototype methods

raymondfeng:
@fabien ok. that's good but it is a breaking change. IIRC, ctx.instance is set to the data instance to be created during `create` today. (To be honest, I'm fine to take it even it's breaking to stop damages)

fabien:
@raymondfeng I think in that case, as you explain with `create`, we should go ahead and break it, because it's inconsistent and a throw-back to the old confusing situation (data vs this manipulation).

fabien:
On second thought, we *should* have an instance with `create`, but we just need to make sure that the docs explicitly say not to manipulate it directly.

raymondfeng:
What ctx.instance will be for `create`?

fabien:
Trouble with ctx.instance in this case is that there actually is none, because create will go straight to the datasource/connector, right?

raymondfeng:
yes

fabien:
OK, seems clear then: unless there's an actual instance (as with prototype methods), it should not be defined at all, omitted from the context.

fabien:
While technically, this will be BC, in practice I think it will not matter much as long as one had implemented the (anti-)pattern I mentioned above, I faced this multiple times while porting over to op hooks, so I could actually handle single and batch operations correctly.

fabien:
Finally, in `create` + `after save`, of course you will have `ctx.instance` available after all.

fabien:
We probably should add a `isNew` flag though, to gain back `before/afterCreate` functionality from before.

fabien:
@raymondfeng can I move ahead with this implementation?

raymondfeng:
@bajtos Do you agree?

fabien:
We can discuss about the last bit (`create`), as long as we can agree on that `ctx.data` is for manipulating, and `ctx.instance` is available for non-static methods (save/updateAttributes) then we already made things a lot more consistent.

bajtos:
Your discussion touched multiple points, I need more time to compose my reply.

bajtos:
Let me start with what I believe is the problem we are trying to solve in this pull request: @fabien would like to get more information about the instance being manipulated via "delete" and "updateAttributes". This information is already available inside LoopBack, but it is not exposed to the hooks at the moment.

I don't recall any specific example showing that the current way of modifying operation's data via ctx.instance/ctx.data does not cover all major use cases. Please correct me if I am wrong.

In the light of that, I still think it's better to add a new property to the hooks as was originally proposed in this pull request, rather than make major refactoring of the hooks, possibly introducing breaking changes, possibly breaking scenarios that worked well with the current implementation but which won't work anymore after the change.

Here are my objections against the last proposal made here:

> if we only ever allow modifications of `ctx.data` we'll have consistency out of the box.

If `ctx.data` is a plain JS object, then we have to pay extra cost of converting a model instance to this data object and then back into the model instance for validation. We are also making it more difficult to invoke any model methods on the modified data, as the user has to apply the changes from data to a model instance too.

If `ctx.data` is a plain JS object for partial database updates and a model instance for full database updates, then I am concerned that most users won't realise this difference until it bites them through a bug in their application. I.e. we make an API that seems to provide a consistent way of manipulating inconsistent data, which will lead users to falsely believe that their code will correctly handle all different use cases and there will be absolutely no warning that they may be wrong.

> your implementation of operation hooks contains obvious oversights, as has been pointed out by others, yet you went for deprecation. Now you bring even more things to the discussion, without resolving what used to be possible, and now is impossible ...

Fair enough. If we believe that the deprecation of old hooks was made too early, then let's remove the deprecation until we fix all issues in the new hooks that we are aware of. That will also remove the pressure to get the issues fixed by yesterday, because frankly, a hasty fix is likely to cause more harm than good.

> Whenever `ctx.data` is available, you'd use it directly. If there's a `ctx.instance` available, you can choose to inspect it read only for your modifications to `ctx.data`. You never modify `ctx.instance` directly (updateAttributes), or at least we don't recommend it (save), ever.

This is a breaking change in the sense that the following code will stop working correctly.

```js
if (ctx.instance) {
  ctx.instance.modified = new Date();
} else {
  ctx.data.modified = new Date();
}
```

When running `updateAttributes`, `ctx.instance` is set and thus the code will change the wrong object.

Note that this code is used as a canonical example in the current documentation for [before save hook](http://docs.strongloop.com/display/LB/Operation+hooks#Operationhooks-beforesave) and I used it in multiple answers in the mailing list, thus I am pretty certain this pattern is already used by several users.

It also seems to me that your proposal in https://github.com/strongloop/loopback-datasource-juggler/pull/488#issuecomment-77517754 optimises the API to be easily used in batch and/or partial updates, while sacrificing ease of use when working with instances only. This is not necessarily a wrong thing, we have to make a tradeoff here, just that IMO we should optimise for ease of use when working with instances.

Setting aside my objections against a single `ctx.data` property, breaking backwards compatibility can be avoided by using different names for the new properties:

```
ctx.changeset - the entity to manipulate
ctx.targetInstance - the model instance affected by the operation (where available)

ctx.data - link to ctx.changeset for backwards compatibility (batch/partial updates)
ctx.instance - link to ctx.changeset for backwards compatibility (full updates of a single instance)
```

> @fabien Trouble with ctx.instance in this case is that there actually is none, because create will go straight to the datasource/connector, right?
> @raymondfeng yes

Please check the code before jumping to conclusions. The current [implementation of create](https://github.com/strongloop/loopback-datasource-juggler/blob/6b0456870729e2e07e8df02d6ad02f1eed001412/lib/dao.js#L222-L256) builds a new Model instance using the data provided in the argument, and then calls `toObject(true)` to build the data passed to the connector.

So yes, there is a model instance involved even in the case of a static methods like `create`.

One of the reasons for such behaviour is that we need a model instance in order to validate the data. Also note that the "before save" hook is intentionally invoked before the validation, so that hooks can add any (computed) required fields and also so that the validation is run on the final data being sent to the database, including any changes made by the hooks.

> We probably should add a isNew flag though, to gain back before/afterCreate functionality from before.

That part is already being discussed here: https://github.com/strongloop/loopback/issues/838

fabien:
@bajtos to summarize:

ctx.changeset - the entity to manipulate **so entity can actually be `data` or `instance`, depending on the operation?**
ctx.targetInstance - the model instance affected by the operation (where available)

ctx.data - link to ctx.changeset for backwards compatibility (batch/partial updates)
ctx.instance - link to ctx.changeset for backwards compatibility (full updates of a single instance)

fabien:
In other words:

`prototype.save` or `create`:
- ctx.changeset === ctx.instance
- ctx.targetInstance is available (and === ctx.instance)

`prototype.updateAttributes`:
- ctx.changeset === ctx.data
- ctx.targetInstance is available (and === ctx.instance)

PS it's not just about reverting deprecation - the fact is that by deprecating people have started actively porting their code over to use op hooks - hence the 'surge' of new issues here recently. 

And once you set out to port your code-base (which in my case contains *a lot* of code) that is affected by this, you cannot follow your development planning when you hit a roadblock like this. Hence the urge to resolve it - postponing is not an option, when you plan to spend time on it now.

bajtos:
> ctx.changeset - the entity to manipulate - so entity can actually be data or instance, depending on the operation?

Yes in the sense that allowing either data or instance is better than always converting instance to a plain object.

I am not happy that this API is hiding the distinction between a partial and a full update, but if the rest of you agree that this is an acceptable limitation, and you are willing to educate users about this catch, then I can live with that.

> prototype.save or create:
> - ctx.changeset === ctx.instance
> - ctx.targetInstance is available (and === ctx.instance)

and ctx.data is undefined

> prototype.updateAttributes:
> - ctx.changeset === ctx.data
> - ctx.targetInstance is available (and === ctx.instance)

Not exactly.

- ctx.changeset === ctx.data
- ctx.targetInstance === the instance being updated, which is the information you are missing in the current implementation
- ctx.where === the where filter for backwards compatibility, basically equal to ({ id: ctx.targetInstance.id })
- ctx.instance === undefined

> PS it's not just about reverting deprecation - the fact is that by deprecating people have started actively porting their code over to use op hooks - hence the 'surge' of new issues here recently.
> And once you set out to port your code-base (which in my case contains a lot of code) that is affected by this, you cannot follow your development planning when you hit a roadblock like this. Hence the urge to resolve it - postponing is not an option, when you plan to spend time on it now.

First of all, I fully understand the pain you are going through.

However, I still believe that you could have taken a different approach here:
 - First of all, you could have made some due diligence or a small spike first, to get a better understanding of the amount of work needed for upgrade and any blocking issues you need to get fixed.
 - Then you could let us know in advance so that we can together find a time when we have enough time to address the issues you find along the way.

fabien:
@bajtos why can't we just go back the original PR intentions?

What's missing from the PR or what needs to be fixed? I don't think there's any other way that treat `updateAttributes` differently from the other `save` operations, and thus, use `currentInstance`. 

To me, that doesn't mean that other operations should use `currentInstance` - they should just be `instance` for `delete` related hooks. 

So a proper `before save` hook example would be something like:

```
Model.observe('before save', function(ctx, next) {
  if (ctx.instance) { // 'prototype.save' or 'create'
    ctx.instance.updatedAt = new Date();
    ctx.instance.createdAt = ctx.instance.createdAt || ctx.instance.updatedAt;
  } else {
    ctx.data.updatedAt = new Date();
    var createdAt = ctx.data.createdAt;
    if (ctx.currentInstance) { // 'updateAttributes'
      createdAt = createdAt || ctx.currentInstance.createdAt;
    }
    ctx.data.createdAt = createdAt || ctx.data.updatedAt;
  }
});
```
```

**Comment by fabien** [comment]
@bajtos I did perform a small spike first to test the waters, but due to the diversity of the kinds of mixins that rely on the old hooks and events, the real issues only surfaced halfway through.

Even though it may seem like I'm criticizing you for your work, I am happy with the way you paved to the future with this. And honestly, I didn't realize that a few touchups like this PR introduces would have such an impact to discuss for days. I still stand behind this PR, as it's the least likely to BC and it only introduces a minimal *additional* API option to the context, that was sorely missing before.

**Comment by fabien** [comment]
Refactored example `before save` timestamping hook:

```
Model.observe('before save', function(ctx, next) {
  updateTimestamps(this.instance || this.data, this.currentInstance);
});

function updateTimestamps(changeset, instance) {
  changeset.updatedAt = new Date();
  var createdAt = changeset.createdAt;
  if (instance) { // 'prototype.updateAttributes'
    createdAt = createdAt || instance.createdAt;
  }
  changeset.createdAt = createdAt || changeset.updatedAt;
};
```

**Comment by bajtos** [comment]
> why can't we just go back the original PR intentions?

That's a call for @raymondfeng and/or @ritch, they had objections against your original intentions.

> So a proper before save hook example would be something like...

Makes me wonder, why do you need to include the property `createdAt` in a partial update? If you omit it from the data, then the connector will not update the field, right?

```js
Model.observe('before save', function(ctx, next) {
  if (ctx.instance) { // 'prototype.save' or 'create'
    ctx.instance.updatedAt = new Date();
    ctx.instance.createdAt = ctx.instance.createdAt || ctx.instance.updatedAt;
  } else {
    ctx.data.updatedAt = new Date();
    // make sure we don't accidentally change the creation time
    delete ctx.data.createdAt;
  }
  next();
});
```

What is missing in this example is the case when we are creating a new instance via optimised "updateOrCreate", where the hook gets data+where. But since we cannot tell whether we are going to UPDATE or CREATE, there is no way how to handle this correctly anyway.

bajtos:
>  I still stand behind this PR, as it's the least likely to BC and it only introduces a minimal additional API option to the context, that was sorely missing before.

Yes, that's why I like the original intent of this PR too, providing we can make small name tweaks as discussed in the beginning.

However, it's important for me to get buy-in from @ritch and @raymondfeng too. My original concern was that the name `currentInstance` may not be clear/descriptive enough, but as you can see, the issue seems to be more complex.

test/persistence-hooks.suite.js
```diff
@@ -736,10 +739,18 @@ module.exports = function(dataSource, should) {
           { id: existingInstance.id, name: 'updated name' },
           function(err, instance) {
             if (err) return done(err);
-            observedContexts.should.eql(aTestModelCtx({
-              where: { id: existingInstance.id },
-              data: { id: existingInstance.id, name: 'updated name' }
-            }));
+            if (dataSource.connector.updateOrCreate) {
```
bajtos: Please add a comment explaining why we get different behaviour depending on the connector implementation.

fabien:
I think @raymondfeng and @ritch mainly had objections to introducing another kind of property `currentInstance` - which I partially agree with - I dislike it for `delete` hooks, so essentially, I would only want it for `updateAttributes`

RE example: this is true, however, the code above allows you to manually 'override' the createdAt value if you wanted it - not discarding it just like that. Your example is fine though.

test/persistence-hooks.suite.js
fabien: OK, noted (but not pushed yet).

bajtos:
BTW I have reviewed the code, see one comment above. The rest looks good to me.

> I think @raymondfeng and @ritch mainly had objections to introducing another kind of property currentInstance - which I partially agree with - I dislike it for delete hooks, so essentially, I would only want it for updateAttributes

I understand that. My POV is that I prefer the consistency across hook context, so that we can say "if you have ctx.where then you may have ctx.currentInstance too".

However, if you agree with @ritch and @raymondfeng that for the two "delete" hooks it makes more sense to use `ctx.instance` and `ctx.where`, then I am ok with that too. As I am thinking about it for more, it makes sense too, because then you can do "if (ctx.instance) else use ctx.where", which is another pattern we already use elsewhere.

Here is my final conclusion:
 - updateAttributes and "before save" hook - we must use `ctx.currentInstance` to prevent breaking change
 - "before delete" and "after delete" hooks - use either "ctx.instance" or "ctx.currentInstance", whatever makes most sense to the majority of you.

fabien:
@bajtos conclusion: :+1: (only use `ctx.currentInstance` with `updateAttributes`)

bajtos:
> - only use ctx.currentInstance with updateAttributes
> - not just to prevent breaking change, but mainly because it would be wrong (and a noop) to do the partial update if you went ahead and changed the ctx.instance itself during updateAttributes: change ctx.data instead, just like a batch operation would
> - I think the pattern "if (ctx.instance) else use ctx.where" makes a lot of sense
> - if you've missed the current instance from the initial op hooks implementation, you'll appreciate ctx.currentInstance - if not, well, no harm done by introducing it now
> - the prototype.delete/remove/destroy hooks will just receive instance in addition to where
> - important: update the docs to reflect this additional context property

Agreed :+1: 

I was pondering over what's the best variant for delete hooks, and I came to the same conclusion as you did - "ctx.instance" is better for delete, because it's consistent with the approach "if (ctx.instance) else use ctx.where".

Let's wait for @raymondfeng and @ritch to take a look too. Also please squash the commits into a single one and provide a descriptive commit message (perhaps based on your last comment above) before landing the patch.

fabien:
@bajtos OK - I will wait a bit and then squash and force-push later. FYI I've been testing this patch with the parts of my code that I ported to op hooks, and most were still consistent and required no change - hence, BC is indeed kept AFAIK.

fabien:
@raymondfeng @ritch  if you happen to have a moment to review this - I would like to know if I can rework my code based on this PR. Doesn't matter if it's not merged yet, I just need to know if this is the way forward.

raymondfeng:
@fabien Does `currentInstance` point to `this` for `prototype.updateAttributes`? Is it the only case that we'll set up `currentInstance`?

fabien:
@raymondfeng exactly, it refers to `this` (the instance used) and only for `updateAttributes`. This is because we need to make this distinction there (partial update).

raymondfeng:
Is the PR up to date? 

1. I don't see `ctx.instance` for `prototype.updateAttributes`. Is it by design. 
2. Does the PR remove `ctx.instance` for static methods?

fabien:
@raymondfeng 

1. as explained, in the case of `prototype.updateAttributes` its' called `currentInstance`; there will not be an `instance` - otherwise it would still not be clear that one should work with `ctx.data` then.

See: https://github.com/fabien/loopback-datasource-juggler/blob/feature/op-hooks/lib/dao.js#L1911

2. no, there are no further changes - except for `instance` with `delete` hooks  - BC is completely preserved.

raymondfeng:
I guess I can live with that even though I'm not happy that `ctx.instance` is NOT pointing to `this` for prototype methods. 

We should build some examples to illustrate the common usages of hooks.

fabien:
@raymondfeng it is only for `prototype.updateAttributes` - `prototype.delete` WILL have `ctx.instance`.

Yes, I might be able to extract some examples for this. As I said, I have a lot of code that I need to refactor now, and I think my tests and mixins cover a lot of ground with this.

fabien:
One final alternative solution `partialData` for `prototype.updateAttributes`:

```javascript
Model.observe('before save', function(ctx, next) {
  if (ctx.instance) { // prototype.save and create
    ctx.instance.updatedAt = new Date();
  } else if (ctx.partialData) { // prototype.updateAttributes
    console.log('last update: ', ctx.instance.updatedAt);
    ctx.partialData.updatedAt = new Date();
  } else {
    ctx.data.updatedAt = new Date();
  }
});
```

fabien:
@raymondfeng actually, forget about the alternative solution above - it doesn't make sense after all.

fabien:
Or we could introduce a boolean flag `isPartial` or `isAtomic` instead:

```javascript
Model.observe('before save', function(ctx, next) {
  if (ctx.isPartial) { // prototype.updateAttributes
    console.log('last update: ', ctx.instance.updatedAt);
    ctx.data.updatedAt = new Date();
  } else if (ctx.instance) { // prototype.save and create
    ctx.instance.updatedAt = new Date();
  } else {
    ctx.data.updatedAt = new Date();
  }
});
```

Which could be simplified as:

```javascript
Model.observe('before save', function(ctx, next) {
  if (ctx.isPartial || !ctx.instance) { // prototype.updateAttributes or any other update
    if (ctx.instance) console.log('last update: ', ctx.instance.updatedAt);
    ctx.data.updatedAt = new Date();
  } else if (ctx.instance) { // prototype.save and create
    ctx.instance.updatedAt = new Date();
  }
});
```

But it would be really important to stress that `ctx.isPartial` means that `ctx.instance` is immutable (or noop).

fabien:
@raymondfeng I'm off now, but I need to port my code over the weekend. It'd be great to hear what the final verdict would be - if you can live with this PR as-is, that would be fine, too. If you think `isPartial` is the way to go, let's decide. Thanks.

raymondfeng:
I actually prefer to always make changes in hooks to `ctx.data`, which can be alias to `ctx.instance` when `ctx.instance` is the source of truth. Do we have any cases that both `ctx.data` and `ctx.instance` are needed?

fabien:
@raymondfeng I'm afraid @bajtos will be against that. However, I could live with your proposal, as discussed earlier as well. The most consistent approach then, would be:

- any modification should **only** be performed on `ctx.data`
- whenever appropriate, pass along the `cx.instance` (create, prototype.updateAttributes, prototype.delete)
- `ctx.data` will always alias to `ctx.instance`, except for `prototype.updateAttributes`
- `ctx.data` will only be available to `save` hooks, not `delete hooks` (wouldn't make much sense)

raymondfeng:
I might have a better solution to this issue now. What about the following?

1. For hooks triggered by an prototype method, such as `updateAttributes`, `save`, and `remove`, we bind `this` to the hook function so that it can alway access the current object as `this`. Hooks triggered by a static method, such as `create` or `updateAll`, will not be bound.

2. `ctx.data` always point to the data to be sent to a connector for persistence for `before` hooks. It's the source of truth for input to DB operations. I think we should set `ctx.data` to the `filter` argument for `find` or `findOne`. (`ctx.data` is mutable)

3. `ctx.where` is set for hooks triggered by a static method with a `where` clause. I'm fine to have `ctx.id` if the method is `by id`. There will be no `ctx.where` for hooks triggered by a prototype method. (`ctx.where` is mutable) (I'm not sure if we should call it `ctx.query` to unify `where` and `filter`). 

4. `ctx.options` is set for methods called with `options` argument.

5. `ctx.result` always point to the data received from a connector for `after` hooks. It's the source of truth for output via callback or promise. `ctx.result` varies depending on the CRUD method. (`ctx.result` is mutable)

fabien:
@raymondfeng @bajtos to get a better feel of the pain points, I went ahead with this PR and ported my code over. I've dumped part of this here: https://gist.github.com/fabien/9f64a306f312c8caa444

Everything works, and I think this PR reflects the minimum change to make this happen, while keeping BC and not adding even more complexity.

My conclusion: porting is definitely non-trivial, because of the fact that many different operations are now squashed into one handler. This is bad or good, depending on how you look at it - you can see this also by the patterns that emerge.

However, even though it goes against the philosophy of these new hooks, it would certainly be valuable to have `ctx.operation` and the current type of operation at hand ('create', 'update', ...).

@raymondfeng I don't like `this` to be bound - it's messy. Also it prevents the common pattern that you see (for better or worse, I guess): `var source = ctx.currentInstance || ctx.instance`.

1. please don't introduce `this` - we have enough things to watch for in `ctx` as it is
2. again, even though it might seem consistent to squash similar concepts into the same variable or operation, in practice it's usually better to be explicit (regarding `data === filter`).
3. another ctx property to watch for: `id`
4. I'm fine with this
5. I don't see why this is needed - more complexity, I think

In short, let's keep BC and keep the required change (this PR) minimal. I think I'm entitled speak up on this matter, considering the number of hooks I have implemented and tested in my application(s) codebase. Perhaps we can further optimize the API later, based on the common patterns in my Gist? 

Please consider merging this PR unless there are strong objections not to.

raymondfeng:
> 1. please don't introduce this - we have enough things to watch for in ctx as it is

I'm fine as long as we have definitive and consistent way to access the current model instance for hooks triggered by a prototype method.

> 2. again, even though it might seem consistent to squash similar concepts into the same variable or operation, in practice it's usually better to be explicit (regarding data === filter).

I realized that the filter is already visible as `ctx.query` for `access` hooks. I'm good.

> another ctx property to watch for: id ? let's keep ctx.where

I'm good to only have `ctx.where`.

> I'm fine with this

Good.

> I don't see why this is needed - more complexity, I think

By looking at the code, I don't see a way for after hooks to know the output from the operation, such as `number of records affected`. I cannot set up a `after find` hook either.

raymondfeng:
@fabien Thank you for putting together the examples. They are helpful to put the discussion in context.

I'm really worried about the magic you had to put into your hook logic as seen from the gist with the current context structure. Let me pick a few examples:

```
var source = ctx.currentInstance || ctx.instance || ctx.data || {};
var target = ctx.instance || ctx.data || {};
```

__Why do I have to check so many ctx properties to understand what's source and what's target?__

```
var inst = ctx.currentInstance || ctx.instance;
```
__Why do I have to always get the instance conditionally?__

```
Model.observe('before save', function(ctx, next) {
    var source = ctx.currentInstance || ctx.instance;
    var target = ctx.instance || ctx.data || {};
    if (source) {
        Model.applyTimestamps(target, source.isNewRecord());
        if (source[createdProp] && createdProp) target[createdProp] = source[createdProp];
        if (source[updatedProp] && updatedProp) target[updatedProp] = source[updatedProp];
    } else if (ctx.data) {
        Model.applyTimestamps(ctx.data);
    }
    next();
});
```

__IIUC, the logic is automatically refresh the timestamps for create and update. I hate to write this magic. It will be great if I can just write `Model.applyTimestamps(ctx.data, ctx.operation === 'create')`.__

To be honest, I will be really puzzled to craft so many conditional statements in a hook as you have illustrated.

fabien:
@raymondfeng I'm afraid this the (sad) reality that came with https://github.com/strongloop/loopback-datasource-juggler/pull/403

The most complex (I think) is `before save` - it covers many different scenarios - I hope I don't forget any... 

For example, I want to enforce `foo = 'bar'` to be set:

```
Model.observe('before save', function(ctx, next) {
  if (ctx.currentInstance) { // `prototype.updateAttributes`
    ctx.data.foo = 'bar';
  } else if (ctx.instance) { // `prototype.save`, `create`, `updateOrCreate`
    ctx.instance.foo = 'bar';
  } else { // anything else: `updateAll` ...
    ctx.data.foo = 'bar';
  }
  next();
});
```
Or, if I don't want to override the value, and just provide it as a fallback:
```
Model.observe('before save', function(ctx, next) {
  if (ctx.currentInstance) { // `prototype.updateAttributes`
    ctx.data.foo = ctx.currentInstance.foo || 'bar';
  } else if (ctx.instance) { // `prototype.save`, `create`, `updateOrCreate`
    ctx.instance.foo = ctx.instance.foo || 'bar';
  } else { // anything else: `updateAll` ...
    ctx.data.foo = ctx.data.foo || 'bar';
  }
  next();
});
```

In other words, while the idea of unifying all 'save' related operations with the new hooks implementation sounds great in theory, it turns out that it didn't really resolve the old concerns. That said, being able to act upon `updateAll` is great - this was sorely lacking before.

raymondfeng:
@fabien Maybe I have mixed up two concerns here:

1. Your PR as a short term solution to fill the gap for your cases based on the current hook design
2. Ideas to fix the current hook design introduced by #403 

In the two examples you have, I really think we should just do `ctx.data.foo = 'bar'`

fabien:
@raymondfeng yes, as I noted before, I agree with this notion: `ctx.data` should be only mutable, whereas `ctx.instance` should be treated as immutable. I'm afraid this will be a breaking change though, but probably the wisest decision to take now, and benefit from on the long term.

fabien:
@raymondfeng @bajtos I propose to land this PR and then refactor at a later stage - it would be a bit too much to expect those who ported their code over to the new hooks to do it differently in such short timespan. Perhaps towards v.3 is more appropriate?

raymondfeng:
I'm fine to land your PR as a short-term workaround. But the complexity and counter-intuition on how to use the `ctx` object really bothers me. We should fix it ASAP and probably release the changes in `loopback-datasource-juggler@3.0.0`.

raymondfeng:
BTW, I think `loopback-datasource-juggler` deserves a new major release with the `Promise` and `Hook` features.

bajtos:
> However, even though it goes against the philosophy of these new hooks, it would certainly be valuable to have ctx.operation and the current type of operation at hand ('create', 'update', ...).

I'd like to implement https://github.com/strongloop/loopback-datasource-juggler/issues/397 for that.

raymondfeng:
@bajtos +1 to implement the per method (https://github.com/strongloop/loopback-datasource-juggler/issues/397) based hooks. That will give developers maximum control and flexibility.

But even with that, the intent-based hooks are needing some fixes to make it consistent and intuitive.

bajtos:
> BTW, I think loopback-datasource-juggler deserves a new major release with the Promise and Hook features.

IMO it's ok to release a minor version, because we are not breaking backwards compatiblity. That's the semver way. It also allows us to release a major version when we learn enough about the new hooks and find issues that require a major change.

raymondfeng:
> IMO it's ok to release a minor version, because we are not breaking backwards compatibility. That's the semver way. It also allows us to release a major version when we learn enough about the new hooks and find issues that require a major change.

Clarification: I was not questioning what have been released. I meant to say if we fix the hooks with breaking changes, it's reasonable to release it together with the promise as a major release.

bajtos:
I don't want to repeat all my arguments why I still think the current design of operation hooks is good, I feel we are running in circles here.

Let's land this pull request to fix the @fabien's immediate problems (after the commit history is cleaned up) and start a new discussion on what is a better solution for this kind of hooks, hopefully getting more community input this time.

bajtos:
> > IMO it's ok to release a minor version, because we are not breaking backwards compatibility. That's the semver way. It also allows us to release a major version when we learn enough about the new hooks and find issues that require a major change.

> Clarification: I was not questioning what have been released. I meant to say if we fix the hooks with breaking changes, it's reasonable to release it together with the promise as a major release.

:+1:, makes sense.

bajtos:
> I'm really worried about the magic you had to put into your hook logic as seen from the gist with the current context structure. Let me pick a few examples:

> ```js
var source = ctx.currentInstance || ctx.instance || ctx.data || {};
var target = ctx.instance || ctx.data || {};
```

> Why do I have to check so many ctx properties to understand what's source and what's target?

Because there are subtle differences between "full update of an instance", "updateAttributes" and "updateAll".

However, I am ok with adding property getters `ctx.source` and `ctx.target` to provide a syntax-sugar API for users that don't care about these details.

fabien:
@bajtos I'll clean it up so it can be merged. Thanks! :+1:

raymondfeng:
For `final fix` to the intent-based hooks, I propose that we use a use-case driven approach. Please add or link your use cases to the wiki page: https://github.com/strongloop/loopback-datasource-juggler/wiki/Use-cases-for-intent-based-CRUD-hooks

bajtos:
@raymondfeng is there anything else to discuss in this PR? Can I merge it once @fabien cleans up the commit history?

raymondfeng:
Go ahead pls.

fabien:
@bajtos I'll push the code in the morning ... sorry about the delay!

fabien:
@bajtos cleaned up - can you merge please?

bajtos:
@fabien can you please beef up the commit message and explain the changes? If not then I'll try to do that myself tomorrow, so that this is finally landed.

bajtos:
Landed via  bc54104

fabien:
@bajtos I'm a bit late to the party - thanks for landing this patch and beefing up the commit!

bajtos:
You are welcome, thank you for your patience :) I'd like to wait with the release until #452 is landed too, so that we don't release minor versions too often.

PR #489, opened 2015-03-06
Cannot create model without id property (https://github.com/strongloop/loopback/issues/1181)

komarik0:


slnode:

komarik0:
I'm not sure what exactly you mean.
But at least for 2.19.2 version you will have id in any model which extends Model. I mean every model will have id. So TransientModel will have id and all sub-models will have id too without my fix.

komarik0:
In case you mean that sub-models of TransientModel will have id with my fix - it's not true. Because TransientModel does not have id it will not be added during extend. And settings are merged before  calling define function of sub-model. So sub-model will have idInjection=false and will not have id.

bajtos:
@komarik0 please add a unit-test that fails without your fix and passes with your fix in place. Also add a unit-test that will verify that the scenario described in @raymondfeng's comment keeps working. These two tests will help us to better understand the context of this change.

clarkorz:
@komarik0 There isn't a pre-defined `TransientModel` in loopback currently, but you can define it by your self in `server/server.js`, like this:

```js
// Define loopback extensions, must before `boot()`
loopback.TransientModel = loopback.modelBuilder.define('TransientModel', {}, { idInjection: false });
```

I use `TransientModel` as base model for most of my embed models, like `Address`, `Image`, etc. They don't have an `id` property at all.

komarik0:
@clarkorz, @bajtos, I didn't know that I can define models like that. In general I want to be able to create a model without id using json definition. Is it planned to add some TransientModel as a default model to loopback? Which can be used as a base model without id in json definition.
In case it's planned I think my changes are needless.

jesucarr:
@clarkorz, it looks like your trick doesn't only works for embedsOne relations. If I try in an embedsMany I get this error:

    ValidationError: The `MyModel` instance is not valid. Details: `myRelation` contains duplicate `id`

slnode:

banane1234:
> Can one of the admins verify this patch?
Please

slnode:

bajtos:
@komarik0 
> In general I want to be able to create a model without id using json definition.

You can specify `idInjection: false` in your JSON file in the same way as other model settings:

```json
{
  "name": "MyModel",
  "base": "Model",
  "idInjection": false
}
```

> Is it planned to add some TransientModel as a default model to loopback? Which can be used as a base model without id in json definition.

Not really. However, you can define your own TransientModel and use it as a base class for you models.

`common/models/transient-model.json`
```json
{
  "name": "TransientModel",
  "base": "Model",
  "idInjection": false
}
```

`common/models/my-model.json`
```json
{
  "name": "MyModel",
  "base": "TransientModel"
}
```


@jesucarr 
> it looks like your trick doesn't only works for embedsOne relations. If I try in an embedsMany I get this error:
> ```
ValidationError: The `MyModel` instance is not valid. Details: `myRelation` contains duplicate `id` 
```

This may be an error in the implementation of embedded relations. Please open a new issue and include a small app we can use to reproduce the issue - see http://loopback.io/doc/en/contrib/Reporting-issues.html#bug-report.

---

I am closing this pull request as rejected.

PR #494, opened 2015-03-06
Allow submodel to hide base properties

raymondfeng:
/to @ritch @bajtos 

The PR allows sub-models to hide properties from its base. For example,
```json
"Customer": {
  "base": "User",
  "name": "String",
  "username": null // or false
}
```

### DIFF
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -241,6 +241,14 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
   ModelClass.getter = {};
   ModelClass.setter = {};
 
+  // Remove properties that reverted by the subclass
+  for (var p in properties) {
+    if (properties[p] === null || properties[p] === false) {
+      // Hide the base property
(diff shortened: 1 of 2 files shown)


raymondfeng:
@bajtos @ritch ping.

ritch:
LGTM

bajtos:
This proposal has one serious flaw: it's breaking the assumptions behind inheritance, subclassed models no longer fulfil the "is-a" relationship. I.e. if a car has an engine, you can create a sub class that does not have any engine at all. The methods in base classes cannot assume they will have all properties when they are subclassed.

We should at least mention this caveat in the documentation. Which brings up another question - have you @raymondfeng updated docs and examples? Is there a follow-up task for that?

raymondfeng:
Well, the base properties are still there in the prototype. They are just hidden from the model.definition.properties list so that all connectors only see the `public` ones from the sub-model. 

I'll update the docs.

PR #498, opened 2015-03-11
fix foreign key dataType bug

didikeke:
Use PK name of foreign class  to get columnMetadata.
Details: https://github.com/strongloop/loopback-connector-mysql/issues/84


slnode:

PR #500, opened 2015-03-11
Add abilities to remove and clear observers - Operation Hooks.

0angelic0:
Users need to keep function references to be able to remove them from operation hooks.

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -579,6 +579,30 @@ ModelBaseClass.observe = function(operation, listener) {
 };
 
 /**
+ * Unregister an asynchronous observer for the given operation (event).
+ * @param {String} operation The operation name.
+ * @callback {function} listener The listener function.
+ * @end
+ */
+ModelBaseClass.removeObserver = function(operation, listener) {
+  if (!this._observers[operation]) return;
+  
+  var index = this._observers[operation].indexOf(listener);
+  if (index != -1) this._observers[operation].splice(index, 1);
+};
+
+/**
+ * Unregister all asynchronous observers for the given operation (event).
+ * @param {String} operation The operation name.
+ * @end
+ */
+ModelBaseClass.clearObservers = function(operation) {
+  if (!this._observers[operation]) return;
+
+  this._observers[operation].length = 0;
+};
+
+/**
  * Invoke all async observers for the given operation.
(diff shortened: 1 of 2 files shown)


slnode:

bajtos:
@slnode ok to test

bajtos:
@0angelic0 thank you for the pull request, the implementation LGTM.  I need you to sign our CLA before it can be merged. See the "details" link in the row of the failed "clahub" check.

However, I am wondering why do you want to remove an observer? IMO, the observers are more like static compile-time metadata than something that should be changed at runtime.

Another point to consider: models are inheriting observers, i.e. `notify` calls all observers registered in parent classes too. Your patch allows developers to disable an observer only at the place where it was registered, it is not possible to disable an observer in certain subclasses only. I don't know if that's something you need or not, I just want to point it out for your consideration.

bajtos:
@0angelic0 are you still keen to get this landed? Can you please sign our [CLA](https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler) and answer my comment above?

> However, I am wondering why do you want to remove an observer? IMO, the observers are more like static compile-time metadata than something that should be changed at runtime.

> Another point to consider: models are inheriting observers, i.e. notify calls all observers registered in parent classes too. Your patch allows developers to disable an observer only at the place where it was registered, it is not possible to disable an observer in certain subclasses only. I don't know if that's something you need or not, I just want to point it out for your consideration.

0angelic0:
@bajtos I want my models to change their behavior at runtime when time based events triggered.

The concept of inheriting observers is confusing me at first. Because I thought in dynamic behavior perspective and don't know what the purpose of this. but after you said you thought observers are more like static compile-time metadata, so, this makes sense now.

However, The dynamic behavior is still the thing that I want from operation hooks. 

The feature about to disable an observer in certain subclasses only that you mentioned seems nice to have, and I also thought of something like overriding observers. But these are for the future and we'll have to make sure that Loopback built-in models don't rely on any operation hooks.

Now I don't have inherited custom models yet (except Loopback built-in models, e.g. User, PersistedModel). So, this patch just works for me right now.

bajtos:
Fair enough. I have landed the patch, thank you for the contribution!

PR #502, opened 2015-03-12
Memory connector returns updated records count

superkhau:
The memory connector now includes the number of updated items in its callback
when updating records.

Connected to strongloop/loopback#1167

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -585,9 +585,11 @@ Memory.prototype.update =
     filter = applyFilter({where: where});
 
     var ids = Object.keys(cache);
+    var count = 0;
     async.each(ids, function (id, done) {
       var inst = self.fromDb(model, cache[id]);
       if (!filter || filter(inst)) {
+        count++;
         // The id value from the cache is string
         // Get the real id from the inst
         id = self.getIdValue(model, inst);
@@ -596,9 +598,8 @@ Memory.prototype.update =
         process.nextTick(done);
       }
     }, function (err) {
-      if (!err) {
-        self.saveToFile(null, cb);
-      }
+      if (err) return cb(err);
+      self.saveToFile({count: count}, cb);
     });
   };
 
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -1,4 +1,5 @@
 // This test written in mocha+should.js
+var async = require('async');
 var should = require('./init.js');
 
 var db, Person;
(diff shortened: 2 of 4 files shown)


superkhau:
Some code taken from #321

superkhau:
@slnode test please

superkhau:
@bajtos PTAL

bajtos:
Cross-posting from #503.

I need you to verify two scenarios:

```js
User.update({ id: 1, name: 'foo' }, { name: 'baz' }, cb);
// check that user.name was changed to 'baz'
// check that 1 updated record was reported

User.update({ id: 1, name: 'bar' }, { name: 'baz' }, cb);
// check that user.name was NOT changed
// check that 0 updated records were reported
```

lib/connectors/memory.js
```diff
@@ -582,7 +584,9 @@ Memory.prototype.update =
       }
     }, function (err) {
       if (!err) {
-        self.saveToFile(null, cb);
+        self.saveToFile(null, function(err) {
+          cb && cb(err, count);
+        });
```
bajtos: Cross-posting from #503: IIRC `saveToFile` passes the first arg as to the final callback, so this block can be probably simplified to `self.saveToFile(count, cb)`.

test/manipulation.test.js
```diff
@@ -875,6 +877,66 @@ describe('manipulation', function () {
...
+      async.forEach(connectors, function(connector, callback) {
+        var Model = getModel(connector);
+        async.waterfall([
+          function setup(cb) {
+            Model.create({id: 1, name: 'foo'}, cb);
+          },
+          function verify(model, cb) {
+            model.name.should.equal('foo');
+            Model.update({name: 'foo'}, {name: 'baz'}, function(err, count) {
+              model.name.should.equal('baz');
```
superkhau: @bajtos This test is failing. I'm not sure why yet, but it doesn't rename `model`.

superkhau: Nevermind, I was just not querying the results properly. Fixed.

superkhau:
I realize after writing all these tests, I now realize `basic-querying-test` is run by all connectors too. Therefore I am proposing we remove those `manipulation` tests from #503 and simply put all tests into `basic-querying-test` instead. Thoughts?

test/manipulation.test.js
```diff
@@ -875,6 +877,76 @@ describe('manipulation', function () {
...
+        Model.updateAll.should.be.a.Function;
+      });
+    });
+
+    it('should not update the model when no match is found', function(done) {
+      // skip the transient connector
+      var index = connectors.indexOf('transient');
+      connectors.splice(index, 1);
+
+      async.each(connectors, function(connector, callback) {
```
bajtos: No no, you misunderstood me. Tests in this file don't deal with different connectors, they just provide a test using `global.getDataSource`. Each connector then includes this file in its test-suite and provides a correct implementation of `global.getDataSource` that returns a data-source backed by that connector.

So don't iterate over all connectors here, just run the test code.

For  the future reference, if we were note removing the loop, then it's better to move the loop *outside* the test and define one test per each item looped over, using a unique test name including the name of the item (connector). The current implementation gives useless assertion errors - it will tell you that `count is not exactly 0 but 1 instead`, but you have no idea which connector caused the failure.

bajtos:
> I realize after writing all these tests, I now realize basic-querying-test is run by all connectors too. Therefore I am proposing we remove those manipulation tests from #503 and simply put all tests into basic-querying-test instead. Thoughts?

First of all, it's best to keep all tests for a single method/feature together in one place. So yes, we should keep all `update/updateAll` tests in one place too.

However, I'd rather keep basic-querying-test for queries (find and friends) only. Therefore I am proposing to move existing update/updateAll tests from basic-querying-test to manipulation tests, where they should have been from the beginning.

test/manipulation.test.js
superkhau: Makes sense. Noted! ;)

superkhau:
I don't understand how the tests are organized, but moving these tests to `manipulation.test.js` makes sense after your explanation.

superkhau:
@slnode test please

superkhau:
@bajtos PTAL again.

test/manipulation.test.js
```diff
@@ -875,6 +875,44 @@ describe('manipulation', function () {
         e.should.be.eql(new Error('Invalid date: X'));
       }
     });
+  });
+
+  describe('update/updateAll', function() {
+    var person;
+
+    before(function (done) {
```
bajtos: Please use `beforeEach`, that way the tests don't interdepend on each other.

bajtos:
Would you mind moving the existing updateAll tests from `basic-querying` over to `manipulation`, so that all updateAll tests are in the same place?

https://github.com/strongloop/loopback-datasource-juggler/blob/7dc0fcd26fc670a7dde76a39a76ce9885da96e8a/test/basic-querying.test.js#L625-L677

It's ok to say on, I can live without that too.

superkhau:
@slnode test please

superkhau:
@slnode test please

superkhau:
@bajtos PTAL again. I moved `basic-querying` update tests into `manipulation` and cleaned it up a little to make it match my updates (removed duplicate tests).

test/manipulation.test.js
```diff
@@ -901,6 +902,103 @@ describe('manipulation', function () {
...
+            people.should.have.length(5);
+            done();
+          });
+        });
+      });
+    });
 
+    it('should ignore where conditions with undefined values',
+        function(done) {
+      Person.update({name: 'foo'}, {name: undefined}, function(err, count) {
```
raymondfeng: This test will fail with mongodb. The data needs to have at least one property value. `{name: undefined}` will becomes `{}`.

superkhau: @raymondfeng Thanks for the input. @bajtos How should we move forward? This was just a simple copy/paste from https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/basic-querying.test.js#L657-L668 modified for `manipulation.test.js`.

raymondfeng: Please note https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/basic-querying.test.js#L657-L668 has been fixed

test/manipulation.test.js
```diff
@@ -901,6 +902,103 @@ describe('manipulation', function () {
...
+      }, {
+        name: 'baz',
+        age: 21
+      }, {
+        name: 'qux',
+        age: 22
+      }, {
+        name: 'quux',
+        age: 23
+      }];
```
bajtos: Can you use real english names instead of "foo" and "bar" please?

bajtos:
LGTM once you address @raymondfeng's comment above per his instructions.

test/manipulation.test.js
superkhau: Sure. Out of curiousity though, are real names better than usual programming placeholder names?

test/manipulation.test.js
superkhau: Added in the extra property, Mongo connector is all green now. ;)

bajtos:
@slnode test please

bajtos:
We should probably rework this implementation and pass the count as the third callback argument, see https://github.com/strongloop/loopback/issues/1167#issuecomment-83193587 and https://github.com/strongloop/loopback/issues/1167#issuecomment-83372314.

test/manipulation.test.js
bajtos: I guess using real names is more important when developing an application using [User-Centered Design](http://en.wikipedia.org/wiki/User-centered_design) and [Personas](http://en.wikipedia.org/wiki/Persona_(user_experience)) and it's less important when developing a framework. I think it's a good habit.

bajtos: More importantly, you should use values that make it easy to distinguish between valid and invalid values, or any other important constraint used in the test. So for example if your test data uses english first names, and then you have `name: 'does-not-exist'` in your test, then it's immediately clear what's going on. On the other hand, if your test data has "foo" and "bar", then it's not immediately clear whether `name: 'baz'` is referring to an existing name or not.

raymondfeng:
LGTM

bajtos:
> We should probably rework this implementation and pass the count as the third callback argument, see strongloop/loopback#1167 (comment) and strongloop/loopback#1167 (comment).

Ignore that comment, you are already passing an object as the callback argument, thus we are free to add whatever new properties to it as needed.

superkhau:
@slnode test please

PR #503, opened 2015-03-12
Add `updateAll` implementation tests

superkhau:
- Ensure `updateAll` is implemented for each connector
- Ensure each connector returns the number of updated records

Connected to strongloop/loopback#1167

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -568,11 +568,13 @@ Memory.prototype.update =
     var filter = null;
     where = where || {};
     filter = applyFilter({where: where});
+    var count = 0;
 
     var ids = Object.keys(cache);
     async.each(ids, function (id, done) {
       var inst = self.fromDb(model, cache[id]);
       if (!filter || filter(inst)) {
+        count++;
         // The id value from the cache is string
         // Get the real id from the inst
         id = self.getIdValue(model, inst);
@@ -582,7 +584,9 @@ Memory.prototype.update =
       }
     }, function (err) {
       if (!err) {
-        self.saveToFile(null, cb);
+        self.saveToFile(null, function(err) {
+          cb && cb(err, count);
+        });
       }
     });
   };
--- a/test/connector.test.js
+++ b/test/connector.test.js
@@ -1,0 +1,41 @@
+var async = require('async');
+var should = require('./init.js');
+var DataSource = require('../').DataSource;
+
+describe('connectors', function() {
+  var connectors = ['memory', 'transient'];
+
+  it('should implement the `updateAll` function', function(done) {
+    connectors.forEach(function(connector) {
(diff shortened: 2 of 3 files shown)


superkhau:
@bajtos I'm not too if this implementation is doing what you expect, but I wanted to start with something so we can discuss and make modifications to. PTAL and let me know if I'm on the right track.

lib/connectors/memory.js
```diff
@@ -582,7 +584,9 @@ Memory.prototype.update =
       }
     }, function (err) {
       if (!err) {
-        self.saveToFile(null, cb);
+        self.saveToFile(null, function(err) {
+          cb && cb(err, count);
+        });
```
bajtos: IIRC `saveToFile` passes the first arg as to the final callback, so this block can be probably simplified to `self.saveToFile(count, cb)`.

test/connector.test.js
```diff
@@ -0,0 +1,41 @@
...
+  it('should return the number of updated records', function(done) {
+    async.forEach(connectors, function(connector, callback) {
+      var ds = new DataSource({connector: connector});
+      var User = ds.define('User');
+      async.waterfall([
+        function(cb) {
+          User.create({id: 1, name: 'foo'}, cb);
+        },
+        function(model, cb) {
+          User.update({name: 'baz'}, cb);
```
bajtos: I need you to verify two scenarios:

```js
User.update({ id: 1, name: 'foo' }, { name: 'baz' }, cb);
// check that user.name was changed to 'baz'
// check that 1 updated record was reported

User.update({ id: 1, name: 'bar' }, { name: 'baz' }, cb);
// check that user.name was NOT changed
// check that 0 updated records were reported
```

bajtos:
The implementation looks correct, thank you. Please move the tests to `manipulation.test.js` per (https://github.com/strongloop/loopback/issues/1167#issuecomment-78433983 and address my comment above.

test/connector.test.js
```diff
@@ -0,0 +1,41 @@
...
+        },
+        function(count, cb) {
+          connector === 'transient' ?
```
bajtos: Since the transient connector never changes the data, I would expect that we are not running any DAO tests for it, in which case you don't need this condition.

bajtos:
Now I am confused, what is the difference between this PR and #502? I am proposing to close this PR and move forward with #502.

superkhau:
|Now I am confused, what is the difference between this PR and #502? I am proposing to close this PR and move forward with #502.

#502 was just created to let you merge the updated count feature without having to wait for the other connector tests (ie. to unblock you in case you were blocked specifically by that feature).

lib/connectors/memory.js
superkhau: Fixed. Simplified the confusing logic too.

PR #508, opened 2015-03-16
Make sure id properties cannot be changed

raymondfeng:
/to @ritch or @bajtos 
/cc @fabien 

This PR prevents id values to be changed during prototype.attributes.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1904,6 +1904,19 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
   }
   data = removeUndefined(data);
 
+  // Make sure id(s) cannot be changed
+  var idNames = Model.definition.idNames();
+  for (var i = 0, n = idNames.length; i < n; i++) {
+    var idName = idNames[i];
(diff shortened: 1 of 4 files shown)


bajtos:
LGTM.

bajtos:
I am wondering if this is a breaking change?

raymondfeng:
This is to cover a hole that could violates the data integrity. If an existing app `use` this `feature` by accident, we should break it to get the security issue exposed sooner than later.

nicholasareed:
@raymondfeng this breaks for mongodb users (or for any databases who have an ObjectId for the id). Perhaps a warning should popup (instead of an Error) and 

    delete data[idName];

or modify the if statement to be 

    if (data[idName] !== undefined && data[idName] !== inst[idName] && data[idName].toString() !== inst[idName].toString()) {

raymondfeng:
This should have been fixed by https://github.com/strongloop/loopback-datasource-juggler/pull/522.

nicholasareed:
Appears that it was, sorry I missed the connection! Where can I find the process for how it gets moved to npm? The install there is the broken version.

PR #509, opened 2015-03-16
Enable "between" filter for memory db connector

dbarbeau:
Hello,

I happen to be using the memory datasource a lot but couldn't get "between" requests to work as documented here:
http://docs.strongloop.com/display/public/LB/Where+filter#Wherefilter-between

After trying every combination I could think of I finally figured that it could be a bug on loopback's side.
So here is my modest contribution.

Daniel

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -455,7 +455,12 @@ function applyFilter(filter) {
       if ('neq' in example) {
         return compare(example.neq, value) !== 0;
       }
-
+      
+      if ('between' in example ) {
+        return ( testInEquality({gte:example.between[0]}, value) && 
+                 testInEquality({lte:example.between[1]}, value) );
+      }
+      
       if (example.like || example.nlike) {
 
         var like = example.like || example.nlike;
--- a/test/memory.test.js
+++ b/test/memory.test.js
@@ -224,6 +224,38 @@ describe('Memory connector', function() {
       });
     });
 
+    it('should successfully extract 5 users from the db', function(done) {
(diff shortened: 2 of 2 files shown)


slnode:

raymondfeng:
@dbarbeau Thank you for the patch. Would you please add a test case?

dbarbeau:
@raymondfeng : There's my attempt at adding test cases. Hope that's the way to do it ;)

raymondfeng:
ok to test

bajtos:
@dbarbeau we should have asked before landing this patch, but could you please sign our CLA? https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler

dbarbeau:
Did it work? I get this message "You'll need to sign into GitHub. Maybe next time?" although I am signed in, and the cla application is registered.

bajtos:
> > @bajtos we should have asked before landing this patch, but could you please sign our CLA? https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler

> @dbarbeau Did it work? I get this message "You'll need to sign into GitHub. Maybe next time?" although I am signed in, and the cla application is registered.

I don't think so, I don't see you in the list. @rmg could you PTAL?

rmg:
I believe I've sorted out the CLA auth issue now. @dbarbeau can you try again?

dbarbeau:
Hum, nope, it still returns the same "Maybe next time?" message.

dbarbeau:
Ok now it worked :)

PR #510, opened 2015-03-16
Reject CREATE with a duplicate id

bajtos:
Modify the memory connector to reject requests to create a record with a duplicate id.

Add a unit-test to verify this behaviour across all connectors.

Connect to https://github.com/strongloop/loopback/issues/986

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -210,6 +210,13 @@ Memory.prototype.create = function create(model, data, callback) {
   if(!this.collection(model)) {
     this.collection(model, {});
   }
+
+  if (this.collection(model)[id]) {
+    return process.nextTick(function() {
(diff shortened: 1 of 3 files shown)


lib/connectors/memory.js
```diff
@@ -210,6 +210,11 @@ Memory.prototype.create = function create(model, data, callback) {
   if(!this.collection(model)) {
     this.collection(model, {});
   }
+
+  if (this.collection(model)[id]) {
+    return callback(new Error('Duplicate entry for ' + model + '.' + idName));
```
raymondfeng: We should make the callback inside `process.nextTick` so that the async expectation is guaranteed.

bajtos: Good point, fixed.

bajtos:
LGTY now?

raymondfeng:
LGTM

PR #511, opened 2015-03-18
deleteAll returns number of deleted records

bajtos:

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -545,17 +545,20 @@ Memory.prototype.destroyAll = function destroyAll(model, where, callback) {
   }
   var cache = this.collection(model);
   var filter = null;
+  var count = 0;
   if (where) {
     filter = applyFilter({where: where});
     Object.keys(cache).forEach(function (id) {
       if (!filter || filter(this.fromDb(model, cache[id]))) {
+        count++;
         delete cache[id];
       }
     }.bind(this));
   } else {
+    count = Object.keys(cache).length;
     this.collection(model, {});
   }
-  this.saveToFile(null, callback);
+  this.saveToFile({ count: count }, callback);
 };
 
 Memory.prototype.count = function count(model, callback, where) {
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -735,6 +735,59 @@ describe('manipulation', function () {
     it('should destroy filtered set of records');
   });
 
+  describe('deleteAll/destroyAll', function () {
+    beforeEach(function clearOldData(done) {
+      Person.deleteAll(done);
+    });
+
+    beforeEach(function createTestData(done) {
+      Person.create([
+        { name: 'John' },
+        { name: 'Jane' }
+      ], done);
+    });
+
+    it('should only delete instances that satisfy the where condition', function (done) {
+      Person.deleteAll({name: 'John'}, function (err, data) {
+        if (err) return done(err);
+        data.count.should.equal(1);
+        Person.find({where: {name: 'John'}}, function (err, data) {
+          if (err) return done(err);
+          data.should.have.length(0);
+          Person.find({where: {name: 'Jane'}}, function (err, data) {
+            if (err) return done(err);
+            data.should.have.length(1);
+            done();
+          });
+        });
+      });
+    });
+
+    it('should report zero deleted instances', function (done) {
+      Person.deleteAll({name: 'does-not-match'}, function (err, data) {
+        if (err) return done(err);
+        data.count.should.equal(0);
+        Person.count(function(err, count) {
+          if (err) return done(err);
+          count.should.equal(2);
+          done();
+        });
+      });
+    });
+
+    it('should delete all instances when "where" is not provided', function(done) {
+      Person.deleteAll(function (err, data) {
+        if (err) return done(err);
+        data.count.should.equal(2);
+        Person.count(function(err, count) {
+          if (err) return done(err);
+          count.should.equal(0);
+          done();
+        });
+      });
+    });
+  });
+
   describe('initialize', function () {
     it('should initialize object properly', function () {
       var hw = 'Hello word',
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -58,7 +58,7 @@ describe('basic-querying', function () {
     });
 
   });
-  
+
   describe('findByIds', function () {
     var createdUsers;
     before(function(done) {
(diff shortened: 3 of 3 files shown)


bajtos:
This patch partially supersedes #321.

superkhau:
Connect to strongloop/loopback#1167

test/manipulation.test.js
```diff
@@ -707,6 +707,59 @@ describe('manipulation', function () {
...
+    it('should only delete instances that satisfy the where condition', function (done) {
+      Person.deleteAll({name: 'John'}, function (err, count) {
+        if (err) return done(err);
+        count.should.equal(1);
+        Person.find({where: {name: 'John'}}, function (err, data) {
+          if (err) return done(err);
+          data.length.should.equal(0);
+          Person.find({where: {name: 'Jane'}}, function (err, data) {
+            if (err) return done(err);
+            data.length.should.equal(1);
```
superkhau: `data.should.have.length(1);`

test/manipulation.test.js
```diff
@@ -707,6 +707,59 @@ describe('manipulation', function () {
...
+    beforeEach(function createTestData(done) {
+      Person.create([
+        { name: 'John' },
+        { name: 'Jane' }
+      ], done);
+    });
+
+    it('should only delete instances that satisfy the where condition', function (done) {
+      Person.deleteAll({name: 'John'}, function (err, count) {
+        if (err) return done(err);
```
superkhau: should.not.exist(err);

test/manipulation.test.js
```diff
@@ -707,6 +707,59 @@ describe('manipulation', function () {
...
+        count.should.equal(1);
+        Person.find({where: {name: 'John'}}, function (err, data) {
+          if (err) return done(err);
```
superkhau: `should.not.exist(err);`

test/manipulation.test.js
```diff
@@ -707,6 +707,59 @@ describe('manipulation', function () {
...
+          data.length.should.equal(0);
+          Person.find({where: {name: 'Jane'}}, function (err, data) {
+            if (err) return done(err);
```
superkhau: `should.not.exist(err);`

test/manipulation.test.js
```diff
@@ -707,6 +707,59 @@ describe('manipulation', function () {
...
+            data.length.should.equal(1);
+            done();
+          });
+        });
+      });
+    });
+
+    it('should report zero deleted instances', function (done) {
+      Person.deleteAll({name: 'does-not-match'}, function (err, count) {
+        if (err) return done(err);
```
superkhau: `should.not.exist(err);`

test/manipulation.test.js
```diff
@@ -707,6 +707,59 @@ describe('manipulation', function () {
...
+        count.should.equal(0);
+        Person.count(function(err, count) {
+          if (err) return done(err);
```
superkhau: `should.not.exist(err);`

test/manipulation.test.js
```diff
@@ -707,6 +707,59 @@ describe('manipulation', function () {
...
+          count.should.equal(2);
+          done();
+        });
+      });
+    });
+
+    it('should delete all instances when "where" is not provided', function(done) {
+      Person.deleteAll(function (err, count) {
+        if (err) return done(err);
```
superkhau: `should.not.exist(err);`

test/manipulation.test.js
```diff
@@ -707,6 +707,59 @@ describe('manipulation', function () {
...
+        count.should.equal(2);
+        Person.count(function(err, count) {
+          if (err) return done(err);
```
superkhau: `should.not.exist(err);`

test/manipulation.test.js
bajtos: `should.not.exist(err)` gives crappy error messages, I am intentionally using `if (err) return done(err)`.

test/manipulation.test.js
bajtos: Fixed.

bajtos:
We should probably rework this implementation and pass the count as the third callback argument, see https://github.com/strongloop/loopback/issues/1167#issuecomment-83193587 and https://github.com/strongloop/loopback/issues/1167#issuecomment-83372314.

lib/connectors/memory.js
```diff
@@ -538,17 +538,20 @@ Memory.prototype.destroyAll = function destroyAll(model, where, callback) {
...
+        count++;
         delete cache[id];
       }
     }.bind(this));
   } else {
+    count = Object.keys(cache).length;
     this.collection(model, {});
   }
-  this.saveToFile(null, callback);
+  this.saveToFile(count, callback);
```
raymondfeng: I propose that we wrap `count` into an object to keep it flexible for future extensions, for example,

> this.saveToFile({count: count}, callback);

bajtos: +1, I had the same idea today. It will also allow us to preserve backwards compatibility with connectors that already pass an object to the callback.

bajtos:
@raymondfeng reworked, PTAL.

test/manipulation.test.js
superkhau: lol, i see. I will change my tests in #502 to match for better error messages then. ;)

raymondfeng:
LGTM

fabien:
@bajtos I don't see why you only implemented this change on the connector but did not update dao.js here: https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1763 by returning: `cb(err, count > 1)` to indicate success

raymondfeng:
cb(err, count) is even better.

bajtos:
An oversight on my side, would you mind submitting a pull request to fix the problem?

> cb(err, count) is even better.

Let's keep the API consistent please and send `cb(err, { count: count })` or `cb(err, { wasDeleted: count === undefined ? undefined : count > 0 })`.

The complex condition for `wasDeleted` is needed to preserve the current behaviour, where the value `undefined` means that the connector does not provide that flag yet.

bajtos:
>  I don't see why you only implemented this change on the connector but did not update dao.js here: https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1763 by returning: cb(err, count > 0) to indicate success

Actually, this requires change on multiple places to ensure that the callback of `prototype.delete` receives always a new argument. And the change should come with an implementation in the memory connector too, to ensure the happy path when no hook modifies the delete query works too.

raymondfeng:
> Let's keep the API consistent please and send cb(err, { count: count }) or cb(err, { wasDeleted: count === undefined ? undefined : count > 0 }).

If the current API doesn't call back with any feedback, +1 to `cb(err, {count: count}}`.

doublemarked:
What could we do to see this sorted out? Seems like it was close?

bajtos:
> @doublemarked What could we do to see this sorted out? Seems like it was close?

Well, you can contribute the implementation and submit a pull request.

PR #515, opened 2015-03-19
Add ctx.isNewInstance for "save" hooks

bajtos:
"before save" hooks provide "ctx.isNewInstance" whenever "ctx.instance" is set. Possible values:
 - true for all CREATE operations
 - false for all UPDATE operations
 - undefined for "prototype.save"

"after save" hooks provide "ctx.isNewInstance" whenever "ctx.instance" is set. Possible values:
 - true after all CREATE operations
 - false after all UPDATE operations
 - undefined after "updateOrCreate" and "prototype.save"

Note: both "updateOrCreate" and "prototype.updateAttributes" don't provide `ctx.instance` to "before save" hooks, therefore `ctx.isNewInstance` it not provided either.

Connect to strongloop/loopback#838

/cc @fabien @clarkorz 

Note that we will need to implement this feature in other connectors too, but unfortunately I don't have bandwidth for that now.

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -225,11 +225,13 @@ Memory.prototype.updateOrCreate = function (model, data, callback) {
   var self = this;
   this.exists(model, self.getIdValue(model, data), function (err, exists) {
     if (exists) {
-      self.save(model, data, callback);
+      self.save(model, data, function(err, data) {
+        callback(err, data, { isNewInstance: false });
+      });
     } else {
       self.create(model, data, function (err, id) {
         self.setIdValue(model, data, id);
-        callback(err, data);
+        callback(err, data, { isNewInstance: true });
       });
     }
   });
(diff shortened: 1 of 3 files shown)


bajtos:
I'll need to update this PR if #516 is landed first.

fabien:
LGTM

raymondfeng:
LGTM. 

As a general note, I don't like the fact that we overload the same property names for `before save` and `after save`. For example, for `before save`, `ctx.isNewInstance` denotes that an instance needs to be created, but for `after save`, `ctx.isNewInstance` indicates that an instance has been created. 

I'm fine to land this PR as is and leave the general concerns to be addressed in future design refinements.

bajtos:
@crandmck could you please update the "Operations hook" docs page?

crandmck:
Added #526

shlomiassaf:
@bajtos,

PUT request with the ID property omitted is actually a POST and the `ctx.instance` property is present.
However, PUT with an existing ID property that does not exist in the database gets treated as a new instance  since `updateOrCreate` is used but `before save` event is handled as an UPDATE, hence no `ctx.instance`.

This is inconsistent.

If a model enforces a policy/security/logic on new instances, a user might use a PUT disguised as a POST and bypass it.

PR #520, opened 2015-03-20
Implement scope.findById

fabien:
@raymondfeng we might add this to the LB REST API later. What do you think?


raymondfeng:
LGTM

fabien:
@raymondfeng thanks!

fabien:
@raymondfeng we could probably add `findOne` as well - thoughts?

PR #521, opened 2015-03-20
Return true/false result for destroyById/prototype.destroy

fabien:
See also: #321 - note that the memory connector has been fixed to have `destroy` also return `{ count: 1 }` feedback - this should also be done with the other connectors.

/cc @raymondfeng @bajtos @superkhau

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1413,11 +1413,17 @@ DataAccessObject.removeById = DataAccessObject.destroyById = DataAccessObject.de
 
   var Model = this;
 
-  this.remove(byIdQuery(this, id).where, options, function(err) {
+  this.remove(byIdQuery(this, id).where, options, function(err, data) {
+    var deleted = data && data.count > 0;
     if ('function' === typeof cb) {
-      cb(err);
+      if (options.reportFailureAsError && !err && !deleted) {
+        err = new Error('No instance with id ' + id + ' found for ' + Model.modelName);
+        cb(err, false);
+      } else {
+        cb(err, deleted);
+      }
     }
-    if(!err) Model.emit('deleted', id);
+    if(!err && deleted) Model.emit('deleted', id);
   });
   return cb.promise;
 };
@@ -1760,8 +1766,13 @@ DataAccessObject.prototype.remove =
           // A hook modified the query, it is no longer
           // a simple 'delete model with the given id'.
           // We must switch to full query-based delete.
-          Model.deleteAll(where, { notify: false }, function(err) {
-            if (err) return cb(err);
+          Model.deleteAll(where, { notify: false }, function(err, data) {
+            if (err) return cb(err, false);
+            var deleted = data && data.count > 0;
+            if (options.reportFailureAsError && !deleted) {
+              err = new Error('No instance with id ' + id + ' found for ' + Model.modelName);
+              return cb(err, false);
+            }
             var context = {
               Model: Model,
               where: where,
@@ -1769,17 +1780,20 @@ DataAccessObject.prototype.remove =
               hookState: hookState
             };
             Model.notifyObserversOf('after delete', context, function(err) {
-              cb(err);
-              if (!err) Model.emit('deleted', id);
+              cb(err, deleted);
+              if (!err && !deleted) Model.emit('deleted', id);
             });
           });
           return;
         }
 
         inst.trigger('destroy', function (destroyed) {
-          inst._adapter().destroy(inst.constructor.modelName, id, function (err) {
-            if (err) {
-              return cb(err);
+          inst._adapter().destroy(inst.constructor.modelName, id, function (err, data) {
+            if (err) return cb(err, false);
+            var deleted = data && data.count > 0;
+            if (options.reportFailureAsError && !deleted) {
+              err = new Error('No instance with id ' + id + ' found for ' + Model.modelName);
+              return cb(err, false);
             }
 
             destroyed(function () {
@@ -1790,8 +1804,8 @@ DataAccessObject.prototype.remove =
                 hookState: hookState
               };
               Model.notifyObserversOf('after delete', context, function(err) {
-                cb(err);
-                if (!err) Model.emit('deleted', id);
+                cb(err, deleted);
+                if (!err && !deleted) Model.emit('deleted', id);
               });
             });
           });
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1416,14 +1416,15 @@ DataAccessObject.removeById = DataAccessObject.destroyById = DataAccessObject.de
   this.remove(byIdQuery(this, id).where, options, function(err, data) {
     var deleted = data && data.count > 0;
     if ('function' === typeof cb) {
-      if (options.reportFailureAsError && !err && !deleted) {
+      if (options.strict && !err && !deleted) {
         err = new Error('No instance with id ' + id + ' found for ' + Model.modelName);
         cb(err, false);
       } else {
         cb(err, deleted);
       }
     }
-    if(!err && deleted) Model.emit('deleted', id);
+    if (options.strict && !deleted) return;
+    if(!err) Model.emit('deleted', id);
   });
   return cb.promise;
 };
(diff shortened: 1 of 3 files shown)


lib/dao.js
```diff
@@ -1413,11 +1413,17 @@ DataAccessObject.removeById = DataAccessObject.destroyById = DataAccessObject.de
 
   var Model = this;
 
-  this.remove(byIdQuery(this, id).where, options, function(err) {
+  this.remove(byIdQuery(this, id).where, options, function(err, data) {
+    var deleted = data && data.count > 0;
     if ('function' === typeof cb) {
-      cb(err);
+      if (options.reportFailureAsError && !err && !deleted) {
```
bajtos: Can we come up with a better name please? `count === 0` is not a failure, it just mean that the record has been already deleted by somebody else before us.

lib/dao.js
```diff
@@ -1413,11 +1413,17 @@ DataAccessObject.removeById = DataAccessObject.destroyById = DataAccessObject.de
...
+        err = new Error('No instance with id ' + id + ' found for ' + Model.modelName);
+        cb(err, false);
+      } else {
+        cb(err, deleted);
+      }
     }
-    if(!err) Model.emit('deleted', id);
+    if(!err && deleted) Model.emit('deleted', id);
```
bajtos: This is IMO a breaking change.

I am proposing to use the same flag that is controlling whether `count==0` triggers an error to control this behaviour. In fact, when that flag is set and no records were deleted, then this branch is never executed and thus this line does not need any change.

lib/dao.js
```diff
@@ -1413,11 +1413,17 @@ DataAccessObject.removeById = DataAccessObject.destroyById = DataAccessObject.de
...
-  this.remove(byIdQuery(this, id).where, options, function(err) {
+  this.remove(byIdQuery(this, id).where, options, function(err, data) {
+    var deleted = data && data.count > 0;
     if ('function' === typeof cb) {
-      cb(err);
+      if (options.reportFailureAsError && !err && !deleted) {
+        err = new Error('No instance with id ' + id + ' found for ' + Model.modelName);
+        cb(err, false);
+      } else {
+        cb(err, deleted);
```
bajtos: Returning a `true/false` value is an anti-pattern and it is also difficult to extend. I am proposing to send an object instead.

test/crud-with-options.test.js
```diff
@@ -354,9 +354,81 @@ describe('crud-with-options', function () {
...
   describe('deleteById', function() {
-    it('should allow deleteById(id)', function () {
-      User.deleteById(1);
+
+    beforeEach(seed);
+
+    it('should allow deleteById(id) - success', function (done) {
+      User.findOne(function (e, u) {
+        User.deleteById(u.id, function(err, deleted) {
+          should.not.exist(err);
```
bajtos: Please use `if (err) return done(err)`, it gives more friendly test failure messages.

bajtos:
The implementation looks good in general, I left a bunch of comments regarding details that I'd like to improve. @raymondfeng PTAL too.

lib/dao.js
bajtos: Perhaps `options.failWhenNotFound`, `options.mustExist` or `options.strict`.

Another approach is to use a property with `true` as a default value, e.g. `options.ignoreMissing`, `options.lenient`.

lib/dao.js
bajtos: How about passing directly the `data` object returned by the connector, in the same way as we already do elsewhere?

fabien: I'm against having an object in this case - how can success/fail better be expressed than using a boolean? I don't want to end up doing: `data && data.success` or something silly like that just to check if something was succesful (that cannot really have any other feedback than that, so why should this be extendable?).

fabien:
@bajtos the flag I propose: `strict`

lib/dao.js
bajtos: Also note that `!deleted` will be triggered also in the case when the connector does not report back the number of deleted records. You should use `deleted === false` instead.

lib/dao.js
```diff
@@ -1413,11 +1413,17 @@ DataAccessObject.removeById = DataAccessObject.destroyById = DataAccessObject.de
 
   var Model = this;
 
-  this.remove(byIdQuery(this, id).where, options, function(err) {
+  this.remove(byIdQuery(this, id).where, options, function(err, data) {
+    var deleted = data && data.count > 0;
     if ('function' === typeof cb) {
-      cb(err);
+      if (options.reportFailureAsError && !err && !deleted) {
+        err = new Error('No instance with id ' + id + ' found for ' + Model.modelName);
```
bajtos: Please set an appropriate http status code via `err.statusCode`, e.g. "404 Not Found". This needs to be fixed in other places too.

lib/dao.js
bajtos: That's the thing, you cannot project the future and know what we will need later. A made-up example: let's say we add support for cascading deletes that is not implemented by all connectors and you want to signal back whether the cascading delete was performed or not. You can get `{ count: 1, cascaded: true }`, `{count: 1, cascaded: false }`, but also `{ count: 0 }` or `{ count: undefined }`. 

> I don't want to end up doing: data && data.success

Then put this check into dao.js, so that the cb arg is always set.

```
cb(err, { count: data ? data.count : undefined })
```

raymondfeng: @fabien I wonder if you will have to return an object as soon as you have cascade set to `true`. For example, you might want to know how many child objects are deleted.

raymondfeng:
Naming wise, I'm fine with with `strict` (default to `false`) or `force` (default to `true`).

bajtos:
> @fabien  the flag I propose: strict
> @raymondfeng Naming wise, I'm fine with with strict (default to false) or force (default to true).

Let's go with `strict` (default to `false`) then.

fabien:
@bajtos let me know if you want those commits squashed

raymondfeng:
Yes, please squash the commits. 

BTW, do the tests have dependencies on connector implementations?. Some of the tests from this module are also run with connectors.

bajtos:
Things to address:
 - the error objects should have a reasonable HTTP `statusCode` property and ideally also a machine-readable `code`. E.g. `statusCode: 404, code: 'NOT_FOUND'`
 - the callback argument should be an object (`{ deleted: false }`) for future extensibility as discussed in the comments above

I am not happy about the `should.not.exist(err);` pattern, but I am not going to repeat myself.

bajtos:
Note that some of the PRs that @superkhau is working on are changing the callback arguments from `err, data` to `err, data, info`. This patch may need to reflect that.

raymondfeng:
Now we settle the callback for `updateAll` and `deleteAll` to be cb(err, info).

lib/dao.js
```diff
@@ -1413,10 +1413,17 @@ DataAccessObject.removeById = DataAccessObject.destroyById = DataAccessObject.de
 
   var Model = this;
 
-  this.remove(byIdQuery(this, id).where, options, function(err) {
+  this.remove(byIdQuery(this, id).where, options, function(err, data) {
+    var deleted = data && data.count > 0;
     if ('function' === typeof cb) {
-      cb(err);
+      if (options.strict && !err && !deleted) {
+        err = new Error('No instance with id ' + id + ' found for ' + Model.modelName);
```
raymondfeng: Let's do `err.statusCode = 404`.

lib/dao.js
```diff
@@ -1760,8 +1767,13 @@ DataAccessObject.prototype.remove =
           // A hook modified the query, it is no longer
           // a simple 'delete model with the given id'.
           // We must switch to full query-based delete.
-          Model.deleteAll(where, { notify: false }, function(err) {
-            if (err) return cb(err);
+          Model.deleteAll(where, { notify: false }, function(err, data) {
+            if (err) return cb(err, false);
+            var deleted = data && data.count > 0;
+            if (options.strict && !deleted) {
+              err = new Error('No instance with id ' + id + ' found for ' + Model.modelName);
```
raymondfeng: Same as above.

raymondfeng:
@fabien ping.

fabien:
@raymondfeng I'll see if I can work on this later.

bajtos:
> @fabien I'll see if I can work on this later.

Ping, what's the status of this work? If you won't have time to work on this in the next few weeks, then I am proposing to close this pull request for now. You can always reopen it when you get more time.

bajtos:
> Ping, what's the status of this work? If you won't have time to work on this in the next few weeks, then I am proposing to close this pull request for now. You can always reopen it when you get more time.

@fabien ping again. Would you mind if I close this pull request, since it seems to be abandoned?

/cc @raymondfeng

bajtos:
Closing in favour of #679.

PR #525, opened 2015-03-21
Enable more CRUD remoting methods for embedsOne

fabien:
Aparently only `__get__` was supported, while Loopback's `hasOneRemoting` already supported the other standard CRUD methods, leading to an exception in remoting's `SharedMethod.invoke` call.

There's a PR over at `loopback` with additional integration tests covering these methods.


fabien:
See also: https://github.com/strongloop/loopback/pull/1235

PR #528, opened 2015-03-23
Fix regression in DAO.prototype.save

### LINKED ISSUE #523: Premature merge of PR #515 - connectors not ready for changed API
@bajtos The merge of #515 introduces an unexpected API change - the connector should return the updated data, as part of 3 instead of 2 arguments. While the memory connector has been (silently) updated, the MongoDB connector (for example) has not been updated, nor is there a branch/PR with these changes pending.

https://github.com/strongloop/loopback-datasource-juggler/commit/84d538c1cac9eab7f2faa1029f74aead4fa3ddc9#diff-d717a01e40d8164976e4468af1bce663R1590

/cc @raymondfeng

bajtos:
ffcaa4e added a "data" argument to the callback function, which shadowed the original data with the data returned by a connector. Not all connectors are returning a data object though, in which casethe model instance ("this" object) is updated with wrong values.

This patch fixes the problem by renaming the second callback argument to "unusedData".

While I was working on this fix, I also improved the error messages reported by recently added unit-tests and replaced 

    foo.count.should.equal(X)
    Uncaught TypeError: Cannot read property 'should' of undefined

with more friendly

    foo.should.have.property('count', X)`
    Uncaught AssertionError: expected 0 to have property count

I am landing this without waiting for review to fix broken master ASAP. I'll address any comments in a follow-up pull request.

@fabien @raymondfeng @superkhau

Close #523

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1587,7 +1587,7 @@ DataAccessObject.prototype.save = function (options, cb) {
       inst.trigger('save', function (saveDone) {
         inst.trigger('update', function (updateDone) {
           data = removeUndefined(data);
-          inst._adapter().save(modelName, inst.constructor._forDB(data), function (err, data, result) {
+          inst._adapter().save(modelName, inst.constructor._forDB(data), function (err, unusedData, result) {
             if (err) {
               return cb(err, inst);
             }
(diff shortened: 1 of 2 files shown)


bajtos:
@slnode test please

bajtos:
Both MongoDB and MySQL tests now pass, there are only expected failures about features not implemented yet.

PR #530, opened 2015-03-23
Clean up update/delete manipulation tests

superkhau:
- Finalize delete/update tests to prepare for testing in all connectors

Connect to strongloop/loopback#1167

@bajtos @raymondfeng Merging this without code review as these are mainly cosmetic updates to ensure I have a solid baseline to run `manipulation.test.js` in each connector individually.

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -756,54 +756,65 @@ describe('manipulation', function () {
     it('should destroy filtered set of records');
   });
 
-  describe('deleteAll/destroyAll', function () {
-    beforeEach(function clearOldData(done) {
-      Person.deleteAll(done);
+  describe('when deleteAll/destroyAll executes', function () {
+    it('should be defined as function', function() {
+      Person.deleteAll.should.be.a.Function;
+      Person.destroyAll.should.be.a.Function;
     });
 
-    beforeEach(function createTestData(done) {
-      Person.create([
-        { name: 'John' },
-        { name: 'Jane' }
-      ], done);
-    });
+    describe('with multiple instances in the database', function() {
+      beforeEach(function deleteFixtures(done) {
+        Person.deleteAll(done);
+      });
 
-    it('should only delete instances that satisfy the where condition', function (done) {
-      Person.deleteAll({name: 'John'}, function (err, result) {
-        if (err) return done(err);
-        result.should.have.property('count', 1);
-        Person.find({where: {name: 'John'}}, function (err, data) {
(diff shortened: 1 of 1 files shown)


test/manipulation.test.js
```diff
@@ -756,54 +756,65 @@ describe('manipulation', function () {
...
-      Person.deleteAll({name: 'John'}, function (err, result) {
-        if (err) return done(err);
-        result.should.have.property('count', 1);
-        Person.find({where: {name: 'John'}}, function (err, data) {
+      beforeEach(function createFixtures(done) {
+        Person.create([{
+          name: 'John'
+        }, {
+          name: 'Jane'
+        }], done);
```
bajtos: Please preserve the formatting next time, it makes patches easier to review.

Also AFAIK we don't use the style you introduced above, which is another reason for not reformatting code to this new style.

test/manipulation.test.js
```diff
@@ -756,54 +756,65 @@ describe('manipulation', function () {
...
+      });
+
+      it('should only delete instances that satisfy the where condition',
+          function(done) {
+        Person.deleteAll({name: 'John'}, function(err, info) {
           if (err) return done(err);
-          data.should.have.length(0);
-          Person.find({where: {name: 'Jane'}}, function (err, data) {
+          info.count.should.equal(1);
```
bajtos: You have reverted 1fbaf4e38 and introduced back the confusing failure messages `Uncaught TypeError: Cannot read property 'should' of undefined` :-1: :heavy_multiplication_x: :100:

test/manipulation.test.js
superkhau: Sounds good, is there a style guide I can follow somewhere?

test/manipulation.test.js
superkhau: Did not know we code based on the error output, I will revert it back and convert all tests in update/delete to match the one style.

superkhau: Actually when I look at this again, we are modifying the second arg in this case to return `{count: 0}` instead, meaning `data.should.have.length(0)` won't work since its not an array being returned anymore. Any suggestions for this?

bajtos: `data` is a callback argument for `Person.find`, not `Person.deleteAll`.

superkhau: Right! I thought it was deleteAll. ;p Changing this one back.

superkhau: Actually nevermind it's already fixed on master.

PR #532, opened 2015-03-24
Ensure boolean value for isNewInstance

rmg:
The tests are testing for true/false, not truthy/falsy, so the value
should be coerced to match.

@bajtos I found this while tinkering with strongloop/loopback-connector-mongodb#116

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -417,7 +417,7 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
             var context = {
               Model: Model,
               instance: obj,
-              isNewInstance: result ? result.isNewInstance : undefined,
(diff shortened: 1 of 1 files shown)


bajtos:
There are three valid values: true, false and undefined meaning either it's not possible to tell or that the connector does not provide the value yet. Which is the case of the mongodb connector I believe?

/cc @superkhau

bajtos:
@rmg here is the algorithm I need to preserve:

 - `result` or `result.isNewInstance` is not defined => `ctx.isNewInstance = undefined`
 - `result.isNewInstance` is truthy => `ctx.isNewInstance = true`
 - `result.isNewInstance` is false => `ctx.isNewInstance = false`

rmg:
I made this change to match the test suite. I guess it's the test suite that needs updating instead.

PR #533, opened 2015-03-24
Return info in the third arg for update and delete

superkhau:
This change is required to make the API consistent across all connectors. Since
the postgres connector already uses the second arg for `data.rows`, we use the
third arg in all connectors to return metadata.

Connect to strongloop/loopback-connector-mysql#90

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1375,16 +1375,16 @@ DataAccessObject.remove = DataAccessObject.deleteAll = DataAccessObject.destroyA
 
     }
 
-    function done(err, data) {
+    function done(err, data, info) {
       if (err) return cb(err);
 
       if (options.notify === false) {
-        return cb(err, data);
+        return cb(err, data, info);
       }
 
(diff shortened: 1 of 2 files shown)


superkhau:
@bajtos @raymondfeng PTAL

bajtos:
LGTM.

IIRC, we need to fix `updateAll` too, right?

bajtos:
@superkhau please check why the CI build is failing before you merge.

bajtos:
Actually, this is not LGTM. You forgot to fix the memory connector. The value passed as the second callback argument should remain the same as in the last published version (which is `null` IIRC).

raymondfeng:
I don't think we should introduce the 3rd argument for callbacks. `info` should be part of the 2nd argument. The current implementation only expects the `err`. So we have the freedom to refine the 2nd argument.

superkhau:
Closing as it has been decided we can modify the second arg for postgres connector instead. See strongloop/loopback#1167

PR #538, opened 2015-03-26
Remove all usages of lodash

bajtos:
This commit is dropping lodash in favour of a hand-written implementation based on ES5 Array methods. As a result, the size of the (unminified) loopback browser bundle is decreased by approx 360KB (20%).

See https://github.com/strongloop/loopback/issues/989

/to @raymondfeng @fabien @ritch please review


raymondfeng:
LGTM

PR #542, opened 2015-03-27
Add model setting "persistUndefinedAsNull"

bajtos:
When the setting "persistUndefinedAsNull" is true, the model will use `null` instead of `undefined` in all property values:

 - Known optional model properties are set to `null` when no value  was provided.

 - When setting model properties, `undefined` is always converted  to `null`. This applies to both known (model-defined) properties  and additional (custom, dynamic) properties.

 - The instance method `toObject()` converts `undefined` to `null` too. Because `toJSON()` calls `toObject()` under the hood, the change applies to `toJSON()` too.

Connect to strongloop/loopback#1215.

Once landed, the recommended model configuration will become

```json
{
  "base": "PersistedModel",
  "strict": true,
  "persistUndefinedAsNull": true,
  "trackChanges": true
}
```

/to @ritch @raymondfeng please review
/cc @fabien @BerkeleyTrue

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1878,7 +1878,7 @@ DataAccessObject.prototype.setAttributes = function setAttributes(data) {
 };
 
 DataAccessObject.prototype.unsetAttribute = function unsetAttribute(name, nullify) {
-  if (nullify) {
+  if (nullify || this.constructor.definition.settings.persistUndefinedAsNull) {
     this[name] = this.__data[name] = null;
   } else {
     delete this[name];
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -429,6 +429,12 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
         } else if (typeof DataType === 'string') {
           DataType = modelBuilder.resolveType(DataType);
         }
+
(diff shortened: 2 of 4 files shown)


bajtos:
I am not 100% happy with the option name. If you can propose a better name then I'll change it. Few ideas I have considered: `nullForUndefined: true`, `undefinedIsNull: true`, `storeUndefinedAsNull: true`.

There are two differences between model-level `defaultToNull:false` and property-level `default:null`.
 - Obviously, property-level `default: null` must be configured for each optional property, which is tedious and easy to forget.
 - `default: null` is applied only by `ModelBaseClass.prototype._initProperties`. It does not prevent
  the user from explicitly specifying `undefined` value.

In the light of the above, it may better to rename the option to `storeUndefinedAsNull`, even though it's a rather long name.

Thoughts?

raymondfeng:
What about `persistUndefinedAsNull`?

bajtos:
> What about `persistUndefinedAsNull`?

Sounds good to me, I'll rework.

bajtos:
@raymondfeng Fixed, PTAL.

bajtos:
@slnode test please

raymondfeng:
LGTM

ritch:
:+1:

PR #543, opened 2015-03-27
Fix the test cases so that they be run with the mssql connector

raymondfeng:
/to @bajtos 

For SQL server, if the PK is an identity, the id cannot be set for insert.

We also need to fix https://github.com/strongloop/loopback-datasource-juggler/blob/feature/fix-test-case/test/manipulation.test.js#L308 because not all connectors produce a message that contains `duplicate`.


bajtos:
LGTM.

PR #544, opened 2015-03-27
Allow nesting properties to be queried for memory connector

raymondfeng:
/to @bajtos or @ritch 

See https://github.com/strongloop/loopback/issues/517

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -295,6 +295,21 @@ Memory.prototype.fromDb = function (model, data) {
   return data;
 };
 
+function getValue(obj, path) {
+  if (obj == null) {
+    return undefined;
+  }
+  var keys = path.split('.');
+  var val = obj;
+  for (var i = 0, n = keys.length; i < n; i++) {
+    val = val[keys[i]];
+    if (val == null) {
+      return val;
+    }
+  }
+  return val;
+}
+
 Memory.prototype.all = function all(model, filter, callback) {
   var self = this;
   var nodes = Object.keys(this.collection(model)).map(function (key) {
@@ -361,12 +376,14 @@ Memory.prototype.all = function all(model, filter, callback) {
     var undefinedA, undefinedB;
 
     for (var i = 0, l = this.length; i < l; i++) {
-      undefinedB = b[this[i].key] === undefined && a[this[i].key] !== undefined;
-      undefinedA = a[this[i].key] === undefined && b[this[i].key] !== undefined;
+      var aVal = getValue(a, this[i].key);
+      var bVal = getValue(b, this[i].key);
+      undefinedB = bVal === undefined && aVal !== undefined;
+      undefinedA = aVal === undefined && bVal !== undefined;
 
-      if (undefinedB || a[this[i].key] > b[this[i].key]) {
+      if (undefinedB || aVal > bVal) {
         return 1 * this[i].reverse;
-      } else if (undefinedA || a[this[i].key] < b[this[i].key]) {
+      } else if (undefinedA || aVal < bVal) {
         return -1 * this[i].reverse;
       }
     }
@@ -400,7 +417,7 @@ function applyFilter(filter) {
           }
         }
       }
-      if (!test(where[key], obj && obj[key])) {
+      if (!test(where[key], getValue(obj, key))) {
         pass = false;
       }
     });
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -241,11 +241,17 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
   ModelClass.getter = {};
   ModelClass.setter = {};
 
-  // Remove properties that reverted by the subclass
   for (var p in properties) {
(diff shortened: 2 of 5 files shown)


bajtos:
My main concern is about backwards compatibility. Is model-builder rejecting property names containing dots? If not, then we must distinguish between `nested.property` (referring to a property of a nested object) and `dot.name` (referring to a top-level property name containing a dot character).

raymondfeng:
`slc loopback` checks property names.

bajtos:
> `slc loopback` checks property names.

That's not sufficient, `slc loopback` is not the only way how to define models.

Here is an example that works for me now and most likely won't work with your patch:

```js
var Person = loopback.createModel('Person',
  {
    id: { type: 'string', id: true },
    'country.name': 'string'
  });

var app = loopback();
app.dataSource('db', {  connector: 'memory' });
app.model(Person, { dataSource: 'db' });

Person.create(
  { id: 'an-id', 'country.name': 'USA', 'custom.name': 'test' },
  function(err, created) {
    console.log('created', created);

    Person.find({ 'country.name': 'USA' }, function(err, found) {
      console.log('found', found);
    });
  });
```

And the output:

```
created { id: 'an-id', 'country.name': 'USA', 'custom.name': 'test' }
found [ { id: 'an-id', 'country.name': 'USA', 'custom.name': 'test' } ]
```

I am proposing to modify `getValue` to support both `nested.property` and `dot.name`:

```js
function getValue(obj, path) {
  if (obj == null) {
    return undefined;
  }

 var val = obj[path];
  if (val !== undefined) {
    return val;
  }

  val = obj;
  var keys = path.split('.');
  // etc.
}
```

Please add unit-tests to verify this new case.

Consider checking the connectors whether they support dot-names columns too. If they do, then we should add unit-tests to `test/datatype.test.js`, `test/basic-querying.test.js` and `test/manipulation.test.js` to verify that dot-named properties work everywhere as expected.

If you think dot-names properties should not be supported, then I am ok with that too, as long as we  modify model builder to throw an error when a property name contains any unsupported characters. (Report a deprecation warning in 2.x in order to preserve backwards compatibility.)

raymondfeng:
To avoid confusions, the dot notation should be reserved for nesting properties. It should NOT be used as property names.

bajtos:
> To avoid confusions, the dot notation should be reserved for nesting properties. It should NOT be used as property names.

I did a quick check of other connectors.

- MySQL tries to create a column with un-escaped dot in the name, which fails with `ER_WRONG_TABLE_NAME: Incorrect table name 'country'`

- MongoDB happily creates the record, however the `find` operation returns an empty list.

I suppose we can claim that property names containing a dot are invalid/unsupported usage and thus don't consider your patch as a breaking change, even though it can break existing application. @raymondfeng Is that what you are proposing?

In that case I am proposing to add a deprecation warning whenever user creates a property containing a dot in the name, so that people are at least notified about this issue. I can do it myself in a new pull request. Any objections?

raymondfeng:
Instead of deprecation, I think we should report it as an error.

bajtos:
@raymondfeng I added deprecation messages as a short-term solution per our discussion yesterday. PTAL and merge the patch if you are ok with the implementation.

zane-insta:
@bajtos You mentioned that MongoDB returns an empty list with the find method. I am noticing this as well with v2.25.1. 
Is this expected to happen or is this patch supposed to make the find method work as well?
If this is expected do you know of a workaround that I can use in the short term to get the dot notation find working with MongoDB?

Specifically I am trying to filter location items based on whether or not they are within a rectangular boundary.

bajtos:
@zane-insta My comment was referring to property names containing a dot, for example a "Place" model with two top-level properties called "location.lat" and "location.lng":

```json
{
  "location.lat": 10,
  "location.lng": 20
}
```

AFAIK, this patch is not changing the behaviour of the MongoDB connector.

Please open a new GH issue in https://github.com/strongloop/loopback-connector-mongodb and describe steps how we can reproduce the issue locally on our machines, ideally per our instructions for [Reporting issues](https://github.com/strongloop/loopback/wiki/Reporting-issues#bug-report).

Possibly useful link: [How do I query a property with a dot in it](http://stackoverflow.com/questions/12878212/how-do-i-query-a-property-with-a-dot-in-it).

zane-insta:
Hey @bajtos I actually found a much better way of getting the query I wanted. I just used the $geoWithin MongoDB query like so ...

```
var filter = {

...

  where: {
      location: {
         geoWithin: {
             $box: [
                 [<bottom left coordinates>],
                 [<top right coordinates>]
             ]
         }
      }
   }
}
```

I am not sure if you guys officially support the geoWithin query or not. 
If so would you like me to open an issue for documentation on the mongoDB connector?

eugenehp:
hey @zane-insta have you moved further with the $geoWithin?

zane-insta:
Hey @eugenehp I have just been using the filter I described above without any issues. I am still not sure if it is officially supported or not. I am planning on writing testcases for my code and dealing with it if/when they fail in the future.

PR #548, opened 2015-03-30
Fix persistUndefinedAsNull tests w/ SQL connectors

bajtos:
Check whether the connector enables strict mode and don't assert on dynamic properties in such case. This should fix test failures in SQL connectors like [loopback-connector-mssql](http://ci.strongloop.com/job/loopback-connector-mssql/81).

Connect to strongloop/loopback#1215

/to @raymondfeng please review.


raymondfeng:
LGTM

PR #549, opened 2015-03-30
Allow passing in hookState from options

fabien:
It can be helpful to pass in `hookState`. Example scenario:

```
Model.observe('before save', function Foo(ctx, next) {
   if (!ctx.hookState.value) return next(); // conditionally run hook
   ctx.data.foo = ctx.hookState.value;
   next();
}); 

Model.updateAll({ ... }, { ... }, { hookState: { value: 'bar' } }, function(err, result) {
    ...
});

```
/cc @raymondfeng @bajtos

### DIFF
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -473,8 +473,8 @@ module.exports = function(dataSource, should) {
           observedContexts.should.eql(aTestModelCtx({ instance: {
             id: existingInstance.id,
             name: 'changed',
-            extra: undefined
-          }}));
+            extra: undefined,
+          }, options: { throws: false, validate: true } }));
           done();
         });
       });
@@ -524,8 +524,8 @@ module.exports = function(dataSource, should) {
               name: 'changed',
               extra: undefined
             },
-            isNewInstance: false
-          }));
+            isNewInstance: false,
+            options: { throws: false, validate: true } }));
           done();
         });
       });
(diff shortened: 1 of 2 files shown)


fabien:
My exact use-case: I'm using `before save` to filter out any unwanted attributes, but there are times where you would want to skip filtering on a case-by-case basis (set: `hookState.skipFilter: true`). Because `options` are not passed in when using the remoting api, it's safe to do this whenever needed.

bajtos:
It makes me wonder whether it's better to pass `hookState` from the `options` argument, or whether it's better to pass the `options` argument as `ctx.options`?
/cc @raymondfeng 

@fabien Anyways, the implementation proposed here is ok, please add a unit-test for every method that is being changed/affected.

test/persistence-hooks.suite.js
```diff
@@ -1297,6 +1297,26 @@ module.exports = function(dataSource, should) {
...
+          { name: 'updated name' },
+          { hookState: { foo: 'bar' } },
+          function(err) {
+            if (err) return done(err);
+            var expected = { foo: 'bar', test: true };
+            observedContexts.should.eql(aTestModelCtx({
+              where: { id: existingInstance.id },
+              data: { name: 'updated name' },
+              hookState: { foo: 'bar', test: true }
+            }));
```
bajtos: Don't assert on things that are already verified by other tests, keep the test as focused as possible and assert only on the `hookState`.

```js
observedContexts.hookState.should.eql({ foo: 'bar', test: true });
```

fabien:
@bajtos I'll await @raymondfeng opinion - once we decide I will squish and force-push.

Regarding `options` vs. `options.hookState`, I wanted to prevent any clashes with actual `options`.

raymondfeng:
I proposed before to make `options` available on the `ctx` object. The `options` can be used to fine control the CRUD, such as transaction. If that happens, there is no need to deal with `options.hookState`.

fabien:
@raymondfeng I've changed it to make `options` available to the `ctx` - all tests will check for this now, so it's not needed to add individual unit-tests, similar to when we introduced `hookState`.

raymondfeng:
LGTM

bajtos:
@crandmck could you please update the doc page "Operations hook" and mention the new context property `ctx.options`?

Issue #547, opened 2015-03-29
updateOrCreate fails to update a model with id 0


pulkitsinghal:
This is a corner-case: `updateOrCreate` fails to update an existing model with a numeric ID that has a value of zero because the following lines of code evaluate zero as false, even though a valid `id` exists.

```
  if (!id) {
    return this.create(data, options, cb);
  }
```

https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L363-L366

raymondfeng:
Good catch. We should fix it. Do you want to submit a patch?

PR #550, opened 2015-03-30
check if id does not exist a bit more explicitly

pulkitsinghal:
if left as-is, ids with value of zero were treated as non-existent. https://github.com/strongloop/loopback-datasource-juggler/issues/547


slnode:

PR #553, opened 2015-04-01
Save parent model of embedded relations

fabien:
This fixes the situation where you may only have built an instance, and you're trying to persist an embedded model on it. Before, only `updateAttributes` would be performed, which then complained about a not existing model. 

(more specifically, in my case I would validate the existence of an embedsOne item, which would create a catch-22 type situation, when trying to save the parent model first).


raymondfeng:
LGTM

PR #554, opened 2015-04-01
Fix isNewInstance tests + code cleanup

bajtos:
Fix test for "after save" called on save/CREATE - make it correctly trigger the code path where the connector decides whether a new record is created or an existing one is updated.

Rename callback argument "result" to "info" to make it consistent with other places where we use "info" too.

Remove traling whitespace.


raymondfeng:
LGTM

PR #561, opened 2015-04-08
Allow custom name mapping for discovered models

raymondfeng:


bajtos:
Please add unit-tests to cover the two new options:
 - `options.nameMapper === null`
 - `typeof options.nameMapper === 'function'`

raymondfeng:
@bajtos Tests added

test/discover.test.js
```diff
@@ -0,0 +1,134 @@
...
+        done();
+      });
+    });
+
+});
+
+
+
+
+
```
bajtos: Please remove these extra empty lines.

bajtos:
Perhaps `test/discovery.test.js` would be a better filename than `test/discover.test.js`?

bajtos:
Two minor comments, the rest LGTM. You can land the patch after you address my first comment. Feel free to ignore the second one.

PR #562, opened 2015-04-09
Add new strict mode "validate"

bajtos:
When a model is configured with `strict: 'validate'`, any dynamic properties not included in the schema trigger a validation error.

IMO, this is a much better solution that `strict: 'throw'`, as the latter may crash the whole application via an unhandled error.

/cc @ritch @fabien

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -120,14 +120,26 @@ ModelBaseClass.prototype._initProperties = function (data, options) {
         enumerable: false,
         configurable: true,
         value: false
-      }
+      },
     });
+
+    if (strict === 'validate') {
+      Object.defineProperty(this, '__unknownProperties', {
+        writable: true,
+        enumerable: false,
+        configrable: true,
+        value: []
+      });
+    }
   } else {
     this.__cachedRelations = {};
     this.__data = {};
     this.__dataSource = options.dataSource;
     this.__strict = strict;
     this.__persisted = false;
+    if (strict === 'validate') {
+      this.__unknownProperties = [];
+    }
   }
 
   if (options.persisted !== undefined) {
@@ -207,6 +219,8 @@ ModelBaseClass.prototype._initProperties = function (data, options) {
         }
       } else if (strict === 'throw') {
         throw new Error('Unknown property: ' + p);
+      } else if (strict === 'validate') {
+        this.__unknownProperties.push(p);
       }
     }
   }
@@ -616,7 +630,7 @@ ModelBaseClass.observe = function(operation, listener) {
  */
 ModelBaseClass.removeObserver = function(operation, listener) {
   if (!this._observers[operation]) return;
-  
+
   var index = this._observers[operation].indexOf(listener);
   if (index != -1) this._observers[operation].splice(index, 1);
 };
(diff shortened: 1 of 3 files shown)


lib/model.js
```diff
@@ -120,14 +120,26 @@ ModelBaseClass.prototype._initProperties = function (data, options) {
         enumerable: false,
         configurable: true,
         value: false
-      }
+      },
     });
+
+    if (strict === 'validate') {
+      Object.defineProperty(this, '__discardedProperties', {
```
raymondfeng: Maybe `__unknownProperties` is a better name.

lib/validations.js
```diff
@@ -453,6 +456,14 @@ Validatable.prototype.isValid = function (callback, data) {
       });
     });
 
+    if (reportDiscardedProperties) {
+      for (var ix in inst.__discardedProperties) {
+        var key = inst.__discardedProperties[ix];
+        inst.errors.add(key, defaultMessages['absence'], 'absence');
```
raymondfeng: I find `absence` is misleading. Really, the validation should report `unknown`.

test/loopback-dl.test.js
```diff
@@ -456,6 +456,15 @@ describe('DataSource define model', function () {
     done(null, User);
   });
 
+  it('should report validation error for unknown properties if strict=validate', function() {
```
ritch: Shouldn't this part really be a describe?

> if strict=validate

...something like

```js
describe('strict validation mode', ...
```

...or similar.

superkhau: :+1:

ritch:
Keep in mind this is outside the scope of this PR (so I am fine moving this to an issue for discussion):

But I don't see the point in the `strict: 'throw'` mode. We should just completely get rid of

```js
model.isValid() // sync
```

and only support

```js
model.isValid(cb) // async/callback
```

Removing the need to throw.

bajtos:
> Keep in mind this is outside the scope of this PR (so I am fine moving this to an issue for discussion):

> But I don't see the point in the strict: 'throw' mode. We should just completely get rid of

> ```js
model.isValid() // sync
```

> and only support

> ```js
model.isValid(cb) // async/callback
```

> Removing the need to throw.

:+1: 

Note that `strict: 'throw'` is handled in `Model._initProperties`, not in `Model.isValid`.

I am pondering the idea of making a following change in 3.0:

 - `strict` is not defined or is `false` - dynamic properties are supported (no change from 2.0)
 - `strict` is `true` - dynamic properties are stripped and a validation error is reported (same as 'validate' in 2.0)
 - The other two modes from 2.0 (`strict: true` silently discarding values and  `strict: 'throw'`) should be removed.

Let's discuss this in a standalone GH issue: https://github.com/strongloop/loopback/issues/1299

lib/validations.js
bajtos: To me `unknown` means "unknown error", not "unknown property". I have reworked the code to `unknown-property`.

bajtos:
I have reworked the patch and addressed all comments.

@ritch @raymondfeng LGTY now?

ritch:
Looks good. @raymondfeng ?

raymondfeng:
LGTM

PR #564, opened 2015-04-15
Extend findById to accept an optional filter object

raymondfeng:
/to @bajtos or @ritch

### DIFF
--- a/test/crud-with-options.test.js
+++ b/test/crud-with-options.test.js
@@ -1,7 +1,7 @@
 // This test written in mocha+should.js
 var should = require('./init.js');
 var async = require('async');
-var db, User, options;
+var db, User, options, filter;
 
 describe('crud-with-options', function () {
 
@@ -17,6 +17,7 @@ describe('crud-with-options', function () {
       vip: {type: Boolean}
     });
     options = {};
+    filter = {fields: ['name', 'id']};
 
     db.automigrate(done);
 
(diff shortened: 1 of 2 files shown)


test/crud-with-options.test.js
```diff
@@ -62,15 +88,49 @@ describe('crud-with-options', function () {
...
+          User.findById(u.id, filter, function(err, u) {
+            should.exist(u);
+            should.not.exist(err);
+            u.should.be.an.instanceOf(User);
+            u.should.have.property('name', 'x');
+            u.should.have.property('email', undefined);
+            done();
+          });
+        });
+      });
```
bajtos: Can you please add a copy of this test using Promise-based API, to ensure that `findById(id, filter).then()` works correctly?

bajtos:
I wish strong-remoting allowed mapping of query arguments to object properties, e.g. `?include=[] -> filter.include=[]`. Without that, we are getting a URL like `GET /api/cars/123?filter[include]=...`. At least it's consistent with the query parameters for `GET /api/cars`.

The implementation LGTM.

raymondfeng:
@bajtos Added a promise-style test

bajtos:
:+1:

ahmed-abdulmoniem:
@bajtos @raymondfeng What about `where` in the filter object?

j-mateo:
@ahmed-abdulmoniem This is a findByID method, short for "where "Id" equals $id". I suppose if you wanted to use the where filter, you could just use find?

PR #565, opened 2015-04-16
Validate model on updateOrCreate (upsert).

bajtos:
Fix the implementation of updateOrCreate (a.k.a. upsert) to validate the model before calling the connector.

In order to preserve backwards compatibility, validation errors are only logged via console.warn by default.

The correct behaviour, where validation errors fail the updateOrCreate operation, can be enabled via new model setting "validateUpsert".

Connect to #262

My plan is to also modify `slc loopback:model` to enable this setting by default for all new models.

/cc @offlinehacker @fabien @clarkorz

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -403,12 +403,22 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
         Model.applyProperties(update, inst);
         Model = Model.lookupModel(update);
 
-        // FIXME(bajtos) validate the model!
-        // https://github.com/strongloop/loopback-datasource-juggler/issues/262
+        inst.isValid(function(valid) {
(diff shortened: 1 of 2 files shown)


raymondfeng:
LGTM

mrfelton:
This needs to be documented on http://docs.strongloop.com/display/public/LB/Model+definition+JSON+file#ModeldefinitionJSONfile-Options

bajtos:
> This needs to be documented

Done.

@crandmck PTAL, feel free to make any edits as needed: http://docs.strongloop.com/pages/diffpagesbyversion.action?pageId=3836230&selectedPageVersions=67&selectedPageVersions=66

crandmck:
Reviewed.  LGTM.

PR #568, opened 2015-04-22
validations: treat `NaN` as a blank value

bajtos:
When a required number property is set to NaN, for example as a result of coersion (`Number([1,2,3])`), the "presence" validation now correctly reports an error.

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/243

### DIFF
--- a/lib/validations.js
+++ b/lib/validations.js
@@ -628,6 +628,7 @@ function blank(v) {
   if (typeof v === 'undefined') return true;
   if (v instanceof Array && v.length === 0) return true;
   if (v === null) return true;
+  if (typeof v === 'number' && isNaN(v)) return true;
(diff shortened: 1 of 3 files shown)


lib/validations.js
```diff
@@ -628,6 +628,7 @@ function blank(v) {
   if (typeof v === 'undefined') return true;
   if (v instanceof Array && v.length === 0) return true;
   if (v === null) return true;
+  if (typeof v === 'number' && isNaN(v)) return true;
```
bajtos: Note that `isNaN("NaN")` returns true, that's why we need to check the type. `Number.isNaN` was introduced in ES6 AFAIK, thus it is not available in ES5 browsers (like phantomjs 1.x).

raymondfeng:
LGTM

PR #569, opened 2015-04-22
Merge inclusion in default scope with inclusion in query.

ningsuhen:
It should merge include filter in default scope (model definition) with the include filter in query.

In Model definition
```json
"scope":{
"include":"owner"
}
```

In query 
```js
Model.find({
  include:{
    posts:'comments'
  }
})
```
After merging, the resulting query will be:

```js
{
  include:{
    owner:true
    posts:'comments'
  }
}
```

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -9,6 +9,7 @@ exports.defineCachedRelations = defineCachedRelations;
 exports.sortObjectsByIds = sortObjectsByIds;
 exports.setScopeValuesFromWhere = setScopeValuesFromWhere;
 exports.mergeQuery = mergeQuery;
+exports.mergeIncludes = mergeIncludes;
 exports.createPromiseCallback = createPromiseCallback
 
 var traverse = require('traverse');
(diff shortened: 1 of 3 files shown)


slnode:

ningsuhen:
Fix for https://github.com/strongloop/loopback/issues/893

lib/utils.js
```diff
@@ -53,6 +54,52 @@ function setScopeValuesFromWhere(data, where, targetModel) {
...
+function convertToObject(param) {
+  if (!param) {
+    return {};
+  }
+  if (typeof param === 'string') {
+    var obj = {};
+    obj[param] = true
+    return obj;
+  }
+  if (param instanceof Array) {
```
raymondfeng: We typically use Array.isArray()

raymondfeng:
Added a minor comment. The fix LGTM.

raymondfeng:
ok to test

raymondfeng:
Please investigate the test failures. You can run `npm test` to see.

lib/utils.js
```diff
@@ -81,9 +128,8 @@ function mergeQuery(base, update, spec) {
     if (!base.include) {
       base.include = update.include;
     } else {
-      var saved = base.include;
-      base.include = {};
-      base.include[update.include] = saved;
+      //merge inclusion
+      base.include = mergeIncludes(base.include, update.include);
```
raymondfeng: Please double check to make sure the `include` in default scope should have precedence.

ningsuhen: Thanks @raymondfeng  for pointing out. I'll push it along with the test fix.
However, it seems mergeQuery is used for inclusion nesting.
```js
physician.patients.getAsync({include: 'address'})
```
here - scope applied query should be 
```js
{
  collect:'patient',
  include:{
    patient:'address'
  }
}
```
but because of the fix, it is
```js
{
  collect:'patient',
  include:{
    patient:true,
    address:true
  }
}
```
But it works if I don't use getAsync as shown below.
```js
physician.patients({include: 'address'})
```

Any pointers?

ningsuhen: @raymondfeng, Latest push has fixed the failed test case.
https://travis-ci.org/walkonsocial/loopback-datasource-juggler/jobs/59677367

raymondfeng:
@ningsuhen Can you squash the git commits into one?

Issue #571, opened 2015-04-24
HasManyThrough on same model can be inconsistent


ningsuhen:
I have a HasManyThrough relation from user model to user model itself via a through model "Follower". 
```json
"relations": {
    "follower": {
      "type": "belongsTo",
      "model": "user",
      "foreignKey": "followerId"
    },
    "followee": {
      "type": "belongsTo",
      "model": "user",
      "foreignKey": "userId"
    }
  },
```
On calling someUser.followers.add(anotherUser), the behavior is inconsistent. Depending upon the order of relations defined in Follower, it creates a new Follower object but with wrong values of followerId & userId. 
E.g. after calling someUser.followers.add(anotherUser), the correct Follower Object should be
```js
{
followerId:[anotherUser's id]
userId:[someUser's id]
}
```
However, it can be filled incorrectly like this:
```js
{
followerId:[someUser's id]
userId:[anotherUser's id]
}
```

This is because of :
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L868
findBelongsTo returns an array of foreign keys with no specific order but as defined in the model json. But the result is consumed with the expectation that first value will be the foreign key for base model.

P.S. I have a fix for this. Will send pull request soon. this is an issue for tracking purposes in case other users find the same problem.

ningsuhen:
Fixed by https://github.com/strongloop/loopback-datasource-juggler/commit/ef0a10dbb658f76f84df8c9aec2cb4ab63272804

PR #572, opened 2015-04-24
Fix for https://github.com/strongloop/loopback-datasource-juggler/issues/571

ningsuhen:
HasManyThrough is inconsistent because of arbitrary order of foreign keys returned by findBelongsTo. This fix sorts the result so that the first key is the one that is mapped to the base (modelFrom) model.


slnode:

raymondfeng:
@ningsuhen Thank you for the patch. Can you add a test case?

ningsuhen:
Thanks @raymondfeng , Test case has been added.

raymondfeng:
ok to test

PR #573, opened 2015-04-24
Allow leading slash for `path` in model settings

raymondfeng:
/to @bajtos


bajtos:
:+1: 

please wait for the CI build to finish and pass, otherwise LGTM

PR #576, opened 2015-05-02
Don't cache static scope method results

### LINKED ISSUE #575: Model scopes should not cache results
The current behavior has a bug because `__cachedRelations` is set on the Class-level (static) as well, which should never happen. Caching is only allowed on the instance (relations) level.

To reproduce:

```
Thing.scope('active', { where: { active: true } });

var widget;

Thing.create({ name: 'Widget', active: true }, function(err, inst) {
   widget = inst;
});

Model.active(function(err, instances) { ... }); // will include 'Widget', as expected

widget.updateAttribute('active', false);

Model.active(function(err, instances) { ... }); // should NOT include 'Widget', but does anyway

Model.__cachedRelations // has been set, but should be undefined!
```

/cc @raymondfeng @bajtos

fabien:
Fixes #575 /cc @raymondfeng

### DIFF
--- a/test/scope.test.js
+++ b/test/scope.test.js
@@ -75,6 +75,28 @@ describe('scope', function () {
       });
     });
   });
+  
+  it('should not cache any results', function (done) {
+    Station.scope('active', {where: {isActive: true}});
+    Station.active.create(function (err, s) {
+      if (err) return done(err);
+      s.isActive.should.be.true;
+      Station.active(function(err, ss) {
+        if (err) return done(err);
+        ss.should.have.lengthOf(1);
+        ss[0].id.should.eql(s.id);
+        s.updateAttribute('isActive', false, function(err, s) {
+          if (err) return done(err);
+          s.isActive.should.be.false;
+          Station.active(function(err, ss) {
+            if (err) return done(err);
+            ss.should.have.lengthOf(0);
+            done();
+          });
+        });
+      });
+    });
+  });
 
 });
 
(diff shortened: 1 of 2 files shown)


fabien:
@raymondfeng @bajtos can we please move ahead with this? I've been bitten by this again today. Thanks.

raymondfeng:
LGTM

test/scope.test.js
```diff
@@ -76,6 +76,24 @@ describe('scope', function () {
     });
   });
   
+  it('should not cache any results', function (done) {
+    Station.scope('active', {where: {isActive: true}});
+    Station.active.create(function (err, s) {
+      s.isActive.should.be.true;
```
bajtos: Please handle the error case:

```
if (err) return done(err);
```

Without that, the error details are lost and the test fails with rather unhelpful error "TypeError: Cannot read property 'isActive' of undefined".

test/scope.test.js
```diff
@@ -76,6 +76,24 @@ describe('scope', function () {
...
+      Station.active(function(err, ss) {
+        ss.should.have.lengthOf(1);
+        ss[0].id.should.eql(s.id);
+        s.updateAttribute('isActive', false, function(err, s) {
```
bajtos: ditto.

test/scope.test.js
```diff
@@ -76,6 +76,24 @@ describe('scope', function () {
...
+          s.isActive.should.be.false;
+          Station.active(function(err, ss) {
+            ss.should.have.lengthOf(0);
```
bajtos: ditto.

raymondfeng:
@fabien ping

fabien:
@raymondfeng I'll have a look later, sorry for the late reply.

fabien:
@bajtos rebased & fixed the test error handling.

bajtos:
Landed, thank you!

PR #578, opened 2015-05-05
dao: support validateUpsert:false

### LINKED ISSUE #570: isValid on model when using mongodb native commands
When using mongodb commands, instance is considered as non-valid, ie:

```js
    DeviceSDKRelation.upsert({
      id: 'compositeid1',
      '$set': {
        sdkVersion: '1.0.0',
        enabled: true,
        updatedAt: new Date()
      },
      '$setOnInsert': {
        deviceId: 'deviceid1',
        applicationId: 'appid1',
        createdAt: new Date()
      }
    });
```

```deviceId``` and ```applicationId``` are required, and so validation reject the upsert

The only way to make it work now is to remove the required attribute on the models

Related commit: https://github.com/strongloop/loopback-datasource-juggler/commit/b642d52fbefa386897d4f0da8db1e4b223a20ac9

bajtos:
- `validateUpsert:true` reports validation errors back to the callback
 - `validateUpsert:false` does not call `isValid()` at all
 - any other value report validation errors via `console.warn`

Fix #570

This is a quick fix to support existing applications that are using MongoDB's `$setOnInsert` together with `updateOrCreate`, see #570 for more details. In the future, I'd like to implement `$setOnInsert` as a feature, see https://github.com/strongloop/loopback/issues/1360.

/cc @PuKoren Is this a good solution for your problem described in #570?

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -403,22 +403,28 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
         Model.applyProperties(update, inst);
         Model = Model.lookupModel(update);
 
-        inst.isValid(function(valid) {
-          if (!valid) {
-            // TODO(bajtos) Remove validateUpsert option in v3.0
-            if (Model.settings.validateUpsert) {
-              return cb(new ValidationError(inst), inst);
-            } else {
-              console.warn('Ignoring validation errors in updateOrCreate():');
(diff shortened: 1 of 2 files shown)


PuKoren:
@bajtos this seems fine for me thanks for the fix

raymondfeng:
LGTM

PR #579, opened 2015-05-07
Conditionally pass options to connector CRUD methods

raymondfeng:
The PR allows the options argument to pass into connectors that choose to accept it. 

/to @bajtos or @ritch

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -124,6 +124,14 @@ DataAccessObject.lookupModel = function(data) {
   return this;
 };
 
+/**
+ * Get the connector instance for the given model class
+ * @returns {Connector} The connector instance
+ */
+DataAccessObject.getConnector = function() {
+  return this.getDataSource().connector;
+}
+
 // Empty callback function
 function noCallback(err, result) {
   // NOOP
@@ -154,6 +162,10 @@ DataAccessObject.create = function (data, options, cb) {
   }
 
   var Model = this;
+  var connector = Model.getConnector();
+  assert(typeof connector.create === 'function',
+    'create() must be implemented by the connector');
+
   var self = this;
 
   if (options === undefined && cb === undefined) {
@@ -260,7 +272,7 @@ DataAccessObject.create = function (data, options, cb) {
         var _idName = idName(Model);
         var modelName = Model.modelName;
         var val = removeUndefined(obj.toObject(true));
-        this._adapter().create(modelName, this.constructor._forDB(val), function (err, id, rev) {
+        function createCallback(err, id, rev) {
           if (id) {
             obj.__data[_idName] = id;
             defineReadonlyProp(obj, _idName, id);
@@ -290,7 +302,13 @@ DataAccessObject.create = function (data, options, cb) {
               });
             });
           });
-        }, obj);
+        }
+
+        if (connector.create.length === 4) {
+          connector.create(modelName, this.constructor._forDB(val), options, createCallback);
+        } else {
+          connector.create(modelName, this.constructor._forDB(val), createCallback);
+        }
       }, obj, cb);
     }, obj, cb);
   }
@@ -363,6 +381,7 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
 
   var self = this;
   var Model = this;
+  var connector = Model.getConnector();
 
   var id = getIdValue(this, data);
   if (id === undefined || id === null) {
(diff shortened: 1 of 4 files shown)


lib/dao.js
```diff
@@ -124,6 +124,10 @@ DataAccessObject.lookupModel = function(data) {
   return this;
 };
 
+DataAccessObject.getConnector = function() {
```
ritch: Docs?

lib/dao.js
```diff
@@ -1602,10 +1669,17 @@ DataAccessObject.count = function (where, options, cb) {
...
   this.notifyObserversOf('access', context, function(err, ctx) {
-      if (err) return cb(err);
-      where = ctx.query.where;
-      Model.getDataSource().connector.count(Model.modelName, cb, where);
-    });
+    if (err) return cb(err);
+    where = ctx.query.where;
+
+    if (connector.count.length <= 3) {
+      // Old signature, please note where is the last
```
bajtos: Would you mind rewording the comment to make it more clear that the word "where" is referring to an argument name, and it is not a part of the sentence? E.g.

```js
// Old signature, please note "where" is the last argument
```

I had to read the current comment twice to understand the meaning. Although this may be just me.

bajtos:
It's a nice trick to use `Function.prototype.length` to distinguish between the old and the new version. Could you please check that the `length` property works in non-V8 browsers too?

I am also missing unit tests that would verify that the `options` argument is correctly forwarded to the connector. Perhaps you could implement these tests by creating a custom instance of the memory connector that would mock individual methods to check the arguments provided. See [test/memory.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/d8ecea61115e24fa7df61e826595a402bbc469b1/test/memory.test.js#L487-L510) for inspiration.

bajtos:
I like the trick of using `Function.prototype.length` to distinguish between the old and the new version. Could you please check that the `length` property works in non-V8 browsers too?

I am missing unit tests that would verify that the `options` argument is correctly forwarded to the connector. Perhaps you could implement these tests by creating a custom instance of the memory connector that would mock individual methods to check the arguments provided. See [test/memory.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/d8ecea61115e24fa7df61e826595a402bbc469b1/test/memory.test.js#L487-L510) for inspiration.I like the trick of using `Function.prototype.length` to distinguish between the old and the new version. Could you please check that the `length` property works in non-V8 browsers too?

I am missing unit tests that would verify that the `options` argument is correctly forwarded to the connector. Perhaps you could implement these tests by creating a custom instance of the memory connector that would mock individual methods to check the arguments provided. See [test/memory.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/d8ecea61115e24fa7df61e826595a402bbc469b1/test/memory.test.js#L487-L510) for inspiration.

bajtos:
I like the trick of using `Function.prototype.length` to distinguish between the old and the new version. Could you please check that the `length` property works in non-V8 browsers too?

I am missing unit tests that would verify that the `options` argument is correctly forwarded to the connector. Perhaps you could implement these tests by creating a custom instance of the memory connector that would mock individual methods to check the arguments provided. See [test/memory.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/d8ecea61115e24fa7df61e826595a402bbc469b1/test/memory.test.js#L487-L510) for inspiration. Unless this is implicitly tested by the existing unit-tests for the memory connector?

raymondfeng:
Function.prototype.length is a standard feature and compatible with major browsers, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length.

bajtos:
> `Function.prototype.length` is a standard feature and compatible with major browsers, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length.

Awesome :+1:

raymondfeng:
@bajtos Tests added

bajtos:
> Tests added

Cool. How about [this comment](#discussion_r30131342)?

I noticed that you have merged in master branch instead of rebasing this patch on top of master. I believe it's better to always rebase feature branches to keep the history as linear as possible. Please consider applying the rebase to get rid of the merge commit d770840.

raymondfeng:
Fixed the comment and rebased the PR

bajtos:
Thanks! I still see the old unchanged comment in the diff displayed by GitHub, but let's not dwell on this. 

LGTM.

Issue #559, opened 2015-04-08
New hook "persist" called after validation before calling the connector


bajtos:
Create a new hook that is called just before handing the data over to the connector. 

This new hook can be used to implement e.g. encrypted properties as discussed in https://github.com/strongloop/loopback/issues/1226, together with "loaded" hook discussed in https://github.com/strongloop/loopback-datasource-juggler/issues/441.

bajtos:
Implemented in #586.

PR #586, opened 2015-05-12
Add new hook `persist`

PradnyaBaviskar:
Connect to #559

Points to note -
 - During `create` the updates applied through `persist` hook are reflected into the database, but the same updates are NOT reflected in the `instance` object obtained in callback of `create`.
 - Secondly, `findOrCreate`, creates a new instance of the object every time. So, 
  - Both `ctx.data.id` and `ctx.currentInstance.id` are set to new id. 
  - And, `ctx.isNewInstance` is `true`
  
 
@bajtos, can you please review?

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -284,6 +284,7 @@ DataAccessObject.create = function (data, options, cb) {
             return cb(err, obj);
           }
           obj.__persisted = true;
+
           saveDone.call(obj, function () {
             createDone.call(obj, function () {
               if (err) {
@@ -304,11 +305,23 @@ DataAccessObject.create = function (data, options, cb) {
           });
         }
 
-        if (connector.create.length === 4) {
-          connector.create(modelName, this.constructor._forDB(val), options, createCallback);
-        } else {
-          connector.create(modelName, this.constructor._forDB(val), createCallback);
-        }
+        context = {
+          Model: Model,
+          data: val,
+          isNewInstance: true,
+          currentInstance: obj,
+          hookState: hookState,
+          options: options
+        };
+        Model.notifyObserversOf('persist', context, function(err) {
+          if (err) return cb(err);
+
+          if (connector.create.length === 4) {
+            connector.create(modelName, obj.constructor._forDB(context.data), options, createCallback);
+          } else {
+            connector.create(modelName, obj.constructor._forDB(context.data), createCallback);
+          }
+        });
       }, obj, cb);
     }, obj, cb);
   }
@@ -425,12 +438,7 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
         var connector = self.getConnector();
 
         if (Model.settings.validateUpsert === false) {
-          update = removeUndefined(update);
-          if (connector.updateOrCreate.length === 4) {
-            connector.updateOrCreate(Model.modelName, update, options, done);
-          } else {
-            connector.updateOrCreate(Model.modelName, update, done);
-          }
+          callConnector();
         } else {
           inst.isValid(function(valid) {
             if (!valid) {
@@ -443,16 +451,29 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
                 // continue with updateOrCreate
               }
             }
+            callConnector();
+          }, update);
+        }
 
-            update = removeUndefined(update);
+        function callConnector() {
+          update = removeUndefined(update);
+          context = {
+            Model: Model,
+            where: ctx.where,
+            data: update,
+            currentInstance: inst,
+            hookState: ctx.hookState,
+            options: options
+          };
+          Model.notifyObserversOf('persist', context, function(err) {
+            if (err) return done(err);
             if (connector.updateOrCreate.length === 4) {
               connector.updateOrCreate(Model.modelName, update, options, done);
             } else {
               connector.updateOrCreate(Model.modelName, update, done);
             }
-          }, update);
+          });
         }
-
         function done(err, data, info) {
           var obj;
           if (data && !(data instanceof Model)) {
@@ -562,9 +583,8 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
   var self = this;
   var connector = Model.getConnector();
 
-  function _findOrCreate(query, data) {
+  function _findOrCreate(query, data, currentInstance) {
     var modelName = self.modelName;
-    data = removeUndefined(data);
     function findOrCreateCallback(err, data, created) {
       var obj, Model = self.lookupModel(data);
 
@@ -598,11 +618,26 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
       }
     }
 
-    if (connector.findOrCreate.length === 5) {
-      connector.findOrCreate(modelName, query, self._forDB(data), options, findOrCreateCallback);
-    } else {
-      connector.findOrCreate(modelName, query, self._forDB(data), findOrCreateCallback);
-    }
+    data = removeUndefined(data);
+    var context = {
+      Model: Model,
+      where: query.where,
+      data: data,
+      isNewInstance: true,
+      currentInstance : currentInstance,
+      hookState: hookState,
+      options: options
+    };
+
+    Model.notifyObserversOf('persist', context, function(err) {
+      if (err) return cb(err);
+
+      if (connector.findOrCreate.length === 5) {
+        connector.findOrCreate(modelName, query, self._forDB(context.data), options, findOrCreateCallback);
+      } else {
+        connector.findOrCreate(modelName, query, self._forDB(context.data), findOrCreateCallback);
+      }
+    });
   }
 
   if (connector.findOrCreate) {
@@ -653,7 +688,7 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
         // validation required
         obj.isValid(function (valid) {
           if (valid) {
-            _findOrCreate(query, data);
+            _findOrCreate(query, data, obj);
           } else {
             cb(new ValidationError(obj), obj);
           }
@@ -1789,11 +1824,25 @@ DataAccessObject.prototype.save = function (options, cb) {
             });
           }
 
-          if (connector.save.length === 4) {
-            connector.save(modelName, inst.constructor._forDB(data), options, saveCallback);
-          } else {
-            connector.save(modelName, inst.constructor._forDB(data), saveCallback);
-          }
+          context = {
+            Model: Model,
+            data: data,
+            where: byIdQuery(Model, getIdValue(Model, inst)).where,
+            currentInstance: inst,
+            hookState: hookState,
+            options: options
+          };
+
+          Model.notifyObserversOf('persist', context, function(err) {
+            if (err) return cb(err);
+
+            if (connector.save.length === 4) {
+              connector.save(modelName, inst.constructor._forDB(data), options, saveCallback);
+            } else {
+              connector.save(modelName, inst.constructor._forDB(data), saveCallback);
+            }
+          });
+
         }, data, cb);
       }, data, cb);
     }
@@ -1919,11 +1968,22 @@ DataAccessObject.updateAll = function (where, data, options, cb) {
       });
     }
 
-    if (connector.update.length === 5) {
-      connector.update(Model.modelName, where, data, options, updateCallback);
-    } else {
-      connector.update(Model.modelName, where, data, updateCallback);
-    }
+    var context = {
+      Model: Model,
+      where: where,
+      data: data,
+      hookState: hookState,
+      options: options
+    };
+    Model.notifyObserversOf('persist', context, function(err, ctx) {
+      if (err) return cb (err);
+
+      if (connector.update.length === 5) {
+        connector.update(Model.modelName, where, data, options, updateCallback);
+      } else {
+        connector.update(Model.modelName, where, data, updateCallback);
+      }
+    });
   }
   return cb.promise;
 };
@@ -2249,13 +2309,23 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
             });
           }
 
-          if (connector.updateAttributes.length === 5) {
-            connector.updateAttributes(model, getIdValue(inst.constructor, inst),
-              inst.constructor._forDB(typedData), options, updateAttributesCallback);
-          } else {
-            connector.updateAttributes(model, getIdValue(inst.constructor, inst),
-              inst.constructor._forDB(typedData), updateAttributesCallback);
-          }
+          context = {
+            Model: Model,
+            where: byIdQuery(Model, getIdValue(Model, inst)).where,
+            data: data,
+            currentInstance: inst,
+            hookState: hookState,
+            options: options
+          };
+          Model.notifyObserversOf('persist', context, function(err) {
+            if (connector.updateAttributes.length === 5) {
+              connector.updateAttributes(model, getIdValue(inst.constructor, inst),
+                inst.constructor._forDB(typedData), options, updateAttributesCallback);
+            } else {
+              connector.updateAttributes(model, getIdValue(inst.constructor, inst),
+                inst.constructor._forDB(typedData), updateAttributesCallback);
+            }
+          });
         }, data, cb);
       }, data, cb);
     }, data);
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -186,6 +186,53 @@ module.exports = function(dataSource, should) {
         });
       });
 
+      it('triggers `persist` hook', function(done) {
+        TestModel.observe('persist', pushContextAndNext());
+
+        TestModel.create(
+          { id: 'new-id', name: 'a name' },
+          function(err, instance) {
+            if (err) return done(err);
+
+            observedContexts.should.eql(aTestModelCtx({
+              data: { id: 'new-id', name: 'a name' },
+              isNewInstance: true,
+              currentInstance: { extra: null, id: 'new-id', name: 'a name' }
+            }));
+
+            done();
+          });
+      });
+
+      it('applies updates from `persist` hook', function(done) {
+        TestModel.observe('persist', pushContextAndNext(function(ctx){
+          ctx.data.extra = 'hook data';
+        }));
+
+        TestModel.create(
+          { id: 'new-id', name: 'a name' },
+          function(err, instance) {
+            if (err) return done(err);
+
+            // the, instance returned by `create` context does not have the
+            // values updated from `persist` hook
+            instance.should.not.have.property('extra', 'hook data');
+
+            // So, we must query the database here because on `create` 
+            // updates from `persist` hook are reflected into database
+            TestModel.findById('new-id', function(err, dbInstance) {
+              if (err) return done(err);
+              dbInstance.toObject(true).should.eql({
+                id: 'new-id',
+                name: 'a name',
+                extra: 'hook data'
+              });
+            });
+
+            done();
+          });
+      });
+
       it('triggers `after save` hook', function(done) {
         TestModel.observe('after save', pushContextAndNext());
 
@@ -372,6 +419,7 @@ module.exports = function(dataSource, should) {
             triggered.should.eql([
               'access',
               'before save',
+              'persist',
               'after save'
             ]);
             done();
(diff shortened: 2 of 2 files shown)


raymondfeng:
@PradnyaBaviskar Please rebase the PR to master.

bajtos:
@raymondfeng I have discussed this PR with @PradnyaBaviskar in a Hangout call. I suggested that the "persist" hook's context should be modelled in the same way as the "before save" context for "updateAttributes":

 - `data` contains the key/value map of data as sent to the connector method (i.e. after applying `model.toJSON` and any other transformations)
 - `where` and optionally also `currentInstance` can be used to detect which model instances are affected
 - plus other usual properties like `options` and `hookState`

See https://github.com/strongloop/loopback/issues/1226#issuecomment-88376078 for an example of usage for this new hook. I would like to use this new hook for hashing passwords in LoopBack's `User` model too, so that we can fix the validation of passwords.

raymondfeng:
The code LGTM. Do we have unit tests for the new hook?

lib/dao.js
```diff
@@ -1748,7 +1763,14 @@ DataAccessObject.prototype.save = function (options, cb) {
 
     inst.isValid(function (valid) {
       if (valid) {
-        save();
+        Model.notifyObserversOf('persist', context, function(err) {
+          if (err) return cb(err);
+
+          data = inst.toObject(true);
+          Model.applyProperties(data, inst);
+          inst.setAttributes(data);
```
PradnyaBaviskar: I was required to add the following block of code in the `persist` hook, so that the new test `"applies updates from 'persist' hook"` passes successfully.

```
data = inst.toObject(true);
Model.applyProperties(data, inst);
inst.setAttributes(data);
```

But, after this block was added, another test `'should preserve properties with dynamic setters'` fails. The test `'should preserve properties with dynamic setters'` expects password to be `'bar-BAR'`, but owing to above block it gets `bar-BAR-BAR-BAR`

If we need to encrypt password within the `persist` hook, then, my block of code would be essential. But, I'm not sure, how to tackle the other test i.e. test `'should preserve properties with dynamic setters'`, in this case!

fabien:
@PradnyaBaviskar @raymondfeng +1 looking forward to this hook! However, we might also need a new hook type for handling cases where the operation is bi-directional, eq. from plaintext to encrypted, and from encrypted to plaintext for example.

lib/dao.js
bajtos: Without studying the source code extensively, I would expect that save performs the following steps (see also https://github.com/strongloop/loopback-datasource-juggler/pull/599#issuecomment-105183289):

```
 1. call "before save" hook with "ctx.instance"
 2. validate the instance
 3. convert the instance to a data object to be passed to the connector 
    - inst.toObject() + removeUndefined()
 4. call "persist" hook with the data from step 3
 5. call connector method with the data from step 3
 --- (async invocation of callback arg "saveCallback")--
 6. call "loaded" hook with data returned by the connector
 7. call inst._initProperties(data, { persisted: true });
 8. cal "after save" hook with "ctx.instance"
```
 
I think the reason why you need to add these extra three lines of code is that you are passing the instance to the hook, whereas my vision is to pass the raw data instead.

Here are few reasons why I think it's better to pass the data:
 - Changes made by the "persist" hook don't affect the original instance. If the DAO method does not update the instance with the data returned by the database, then the persist hook modifies only database representation, which is IMO exactly the intent of this hook.
 - My belief is that a model instance is always valid in an ideal world. The hook "persist" may modify some property values in such way that they no longer pass validation (for example hashed passwords may not meet rules like "must contain a lowercase and an uppercase character"). In that case the hook should not modify (and invalidate) the instance, but merely alter the data passed to the connector.

bajtos:
@fabien 
> However, we might also need a new hook type for handling cases where the operation is bi-directional, eq. from plaintext to encrypted, and from encrypted to plaintext for example.

My intention was to implement a "loaded" hook for the second leg (encrypted-to-plaintext), see #441 and the pending patch #599 from @PradnyaBaviskar. Would the "loaded" hook work for you?

A mockup code with a simplified implementation of an encrypted property "secret", as I was envisioning it:

```js
MyModel.observe('persist', function(ctx, next) {
  encrypt(cx.data.secret, function(err, value) {
    ctx.data.secret = value;
    next();
  });
});

MyModel.observe('loaded', function(ctx, next) {
  decrypt(cx.data.secret, function(err, value) {
    ctx.data.secret = value;
    next();
  });
});
```

PradnyaBaviskar:
@bajtos, Corrected the context objects passed to ```persist``` hook. Also added few more test cases. 

Can you review?

CHANGES.md
```diff
@@ -1,3 +1,15 @@
...
+
+ * Fix the target id resolution (Raymond Feng)
+
+ * DB Call Optimization in relation includes - Fixes #408 & #166 (ningsuhen)
+
+ * Conditionally pass options to connector CRUD methods (Raymond Feng)
+
+ * Pass-through options from save to create (Fabien Franzen)
+
+
```
bajtos: This is most likely an unintended side-effect of rebase/merge, please revert.

lib/dao.js
```diff
@@ -304,11 +304,22 @@ DataAccessObject.create = function (data, options, cb) {
           });
         }
 
-        if (connector.create.length === 4) {
-          connector.create(modelName, this.constructor._forDB(val), options, createCallback);
-        } else {
-          connector.create(modelName, this.constructor._forDB(val), createCallback);
-        }
+        var inst = this;
```
bajtos: I believe `this === obj`, in which case I am proposing to drop `inst` and use `obj` instead. Thoughts?

lib/dao.js
```diff
@@ -304,11 +304,22 @@ DataAccessObject.create = function (data, options, cb) {
...
+        context = {
+          Model: Model,
+          data: val,
```
bajtos: Please add `where` and `currentInstance` fields to the context.

lib/dao.js
```diff
@@ -424,13 +435,26 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
...
+          Model.notifyObserversOf('persist', context, function(err) {
+            if (err) return cb(err);
+
+            update = removeUndefined(update);
+            if (connector.updateOrCreate.length === 4) {
+              connector.updateOrCreate(Model.modelName, update, options, done);
+            } else {
+              connector.updateOrCreate(Model.modelName, update, done);
+            }
+          });
```
bajtos: The original code contained some duplication. With your changes in place, I feel the amount of duplicated code became unmanageable. Could you please refactor (extract) a new method with the shared code and use it in both places?

Example:

```js
function callConnector() {
  update = removeUndefined(update);
  context = {
    Model: Model,
    where: ctx.where,
    data: update,
    currentInstance: inst,
    hookState: ctx.hookState,
    options: options
  };
  Model.notifyObserversOf('persist', context, function(err) {
    if (err) return done(err);
    if (connector.updateOrCreate.length === 4) {
      connector.updateOrCreate(Model.modelName, update, options, done);
    } else {
      connector.updateOrCreate(Model.modelName, update, done);
    }
  });
}

File: lib/dao.js [hunk 6]
```diff
@@ -424,13 +435,26 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
...
+
         if (Model.settings.validateUpsert === false) {
-          update = removeUndefined(update);
-          if (connector.updateOrCreate.length === 4) {
-            connector.updateOrCreate(Model.modelName, update, options, done);
-          } else {
-            connector.updateOrCreate(Model.modelName, update, done);
-          }
+          Model.notifyObserversOf('persist', context, function(err) {
+            if (err) return cb(err);
```
bajtos: There is a subtle difference between calling `cb(err)` and `done(err)`. I think we should call `done`.

lib/dao.js
```diff
@@ -598,11 +626,21 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb)
...
     }
 
-    if (connector.findOrCreate.length === 5) {
-      connector.findOrCreate(modelName, query, self._forDB(data), options, findOrCreateCallback);
-    } else {
-      connector.findOrCreate(modelName, query, self._forDB(data), findOrCreateCallback);
-    }
+    context = {
+      Model: Model,
+      data: data,
```
bajtos: Please add `where` and `currentInstance`.

lib/dao.js
```diff
@@ -1789,11 +1827,24 @@ DataAccessObject.prototype.save = function (options, cb) {
...
 
-          if (connector.save.length === 4) {
-            connector.save(modelName, inst.constructor._forDB(data), options, saveCallback);
-          } else {
-            connector.save(modelName, inst.constructor._forDB(data), saveCallback);
-          }
+          context = {
+            Model: Model,
+            data: data,
+            currentInstance: inst,
```
bajtos: Please add `where`.

test/persistence-hooks.suite.js
```diff
@@ -575,6 +576,204 @@ module.exports = function(dataSource, should) {
...
+        it('applies updates from `persist` hook - on create', function(done) {
+          TestModel.observe('persist', pushContextAndNext(function(ctx){
+            ctx.data.extra = 'hook data';
+          }));
+
+          TestModel.create(
+            { id: 'new-id', name: 'a name' },
+            function(err, instance) {
+              if (err) return done(err);
+              observedContexts.data.should.have.property('extra', 'hook data');
```
bajtos: Please check that the created `instance` has the extra data. The fact that `observerdContexts.data.extra` is set is not important IMO and it should be already verified by the previous test

test/persistence-hooks.suite.js
```diff
@@ -575,6 +576,204 @@ module.exports = function(dataSource, should) {
           done();
         });
       });
+
+      describe('persist hook', function() {
```
bajtos: The test file structures tests by DAO method (create, update), not by hooks. Please preserve that structure and move the tests from `describe('persist hook')` to appropriate per-method suites.

test/persistence-hooks.suite.js
```diff
@@ -575,6 +576,204 @@ module.exports = function(dataSource, should) {
...
+        });
+
+        it('applies updates from `persist` hook - on save', function(done) {
+          TestModel.observe('persist', pushContextAndNext(function(ctx){
+            ctx.data.extra = 'hook data';
+          }));
+
+          existingInstance.save(function(err, instance) {
+            if (err) return done(err);
+            observedContexts.data.should.have.property('extra', 'hook data');
```
bajtos: Same comment as above, please check that the created `instance` has the extra data set. You can look at "applies updates from `before save` hook" for inspiration.

test/persistence-hooks.suite.js
```diff
@@ -575,6 +576,204 @@ module.exports = function(dataSource, should) {
...
+                data: {
+                  id: getLastGeneratedUid(),
+                  name: existingInstance.name
+                }
+              }));
+              done();
+            });
+        });
+
+        it.skip('applies updates from `persist` hook - on findOrCreate(record found)', function(done) {
```
bajtos: Please add a comment explaining why this test is skipped.

test/persistence-hooks.suite.js
```diff
@@ -575,6 +576,204 @@ module.exports = function(dataSource, should) {
...
+          TestModel.observe('persist', pushContextAndNext(function(ctx){
+            ctx.data.extra = 'hook data';
+          }));
+
+          TestModel.findOrCreate(
+            { where: { name: 'new-record' } },
+            { name: 'new-record' },
+            function(err, instance) {
+              if (err) return done(err);
+              observedContexts.data.should.have.property('extra', 'hook data');
```
bajtos: Please check that `instance` has the extra data set.

bajtos:
@PradnyaBaviskar cool, you are on the right track now. I left a few comments to address, see above.

Please rebase your code on top of the current master, check why the CI build is failing and make it pass.

test/persistence-hooks.suite.js
```diff
@@ -575,6 +576,233 @@ module.exports = function(dataSource, should) {
...
+        });
+
+        /* TODO (pradnya) - skipped this because findOrCreate (record found) gives following error for unoptimized connector
+         * expected 'hook not called' to equal {
+            Model: '[ModelCtor TestModel]',
+            data: { id: '2', name: 'first' },
+            hookState: { test: true },
+            options: {}
+          }
+         */
```
bajtos: See https://github.com/strongloop/loopback-datasource-juggler/blob/c0fa60b42c49ea17c0065528dc8c734f2b4719f3/test/persistence-hooks.suite.js#L303-L324 for an example how to run or skip a test depending on whether a connector provides an optimized version of findOrCreate.

It is expected that that the unoptimized connector does not trigger the "persist" hook, it's the same situation as with the "before save" hook, which is not triggered either.

test/persistence-hooks.suite.js
```diff
@@ -575,6 +576,233 @@ module.exports = function(dataSource, should) {
...
+            });
+        });
+
+        /* TODO (pradnya) - skipped this because findOrCreate (record found) gives following error for unoptimized connector
+           expected 'hook not called' to equal {
+            Model: '[ModelCtor TestModel]',
+            data: { id: '2', name: 'first' },
+            hookState: { test: true },
+            options: {}
+          } */
```
bajtos: I think this is same as above, the test should be skipped for unoptimized connector.

test/persistence-hooks.suite.js
```diff
@@ -575,6 +576,233 @@ module.exports = function(dataSource, should) {
...
+                where: { where: { name: existingInstance.name } }
+              }));
+
+              done();
+            });
+        });
+
+        /* TODO (pradnya) - this gives arrestion error -
+         * expected { count: 0 } to have property extra. verify this.
+         */
```
bajtos: Ah, this is a mistake in the existing `updateAll` tests. The second callback argument of `update` (aka `updateAll`) is an `info` object, not an instance.

You should run `findTestModels` to load the updated models and verify the changes were applied. See https://github.com/strongloop/loopback-datasource-juggler/blob/c0fa60b42c49ea17c0065528dc8c734f2b4719f3/test/persistence-hooks.suite.js#L1270-L1288 for an example.

test/persistence-hooks.suite.js
```diff
@@ -575,6 +576,233 @@ module.exports = function(dataSource, should) {
...
+        it.skip('applies updates from `persist` hook - on update', function(done) {
+          TestModel.observe('persist', pushContextAndNext(function(ctx){
+            ctx.data.extra = 'hook data';
+          }));
+
+          TestModel.update(
```
bajtos: Please use `updateAll` instead of `update`. Both names are aliases for the same method, but I find the name `update` a bit confusing. When I was reading this code for the first time, I thought it was updating a single instance only, while it is updating all instances matching the query instead.

lib/dao.js
```diff
@@ -291,6 +292,11 @@ DataAccessObject.create = function (data, options, cb) {
               }
               var context = {
                 Model: Model,
+                /* TODO (pradnya), here obj, does NOT have the attributes, 
+                 * newly added/updated through persist hook. how should 
+                 * `obj` be updated, so that the following test succeeds
+                 * 'applies updates from `persist` hook - on create'
+                 */
```
bajtos: I think you should make two changes:
1) In the test, add another check to verify what has been stored in the database. I believe this check should be already passing with the current implementation.
2) To fix the particular issue you have described, you will need to copy properties from `data` back to the instance. I think `obj.setAttributes(context.data);` should do the job, see the implementation of `updateAttributes` which calls that method too.

test/persistence-hooks.suite.js
```diff
@@ -575,6 +576,232 @@ module.exports = function(dataSource, should) {
...
+              loadTestModel(existingInstance.id, function(err, instance) {
+                instance.should.have.property('extra', 'hook data');
+                done();
+              });
+            });
+        });
+
+        if (dataSource.connector.findOrCreate) {
+          /* TODO (pradnya), skipped because, currentInstance is not getting set
+           * correctly through findOrCreate() in dao.js */
```
bajtos: As I see it, `findOrCreate` cannot report the current instance because it may not be created yet, is my understanding correct? In that case it's ok to leave it out of the context. Just make sure you always provide `where` in the context.

PradnyaBaviskar:
@bajtos,  Updated the PR with changes suggested above. 

Please review.

bajtos:
@slnode test please

lib/dao.js
```diff
@@ -598,11 +617,23 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb)
...
-      connector.findOrCreate(modelName, query, self._forDB(data), options, findOrCreateCallback);
-    } else {
-      connector.findOrCreate(modelName, query, self._forDB(data), findOrCreateCallback);
-    }
+    data = removeUndefined(data);
+    var context = {
+      Model: Model,
+      data: data,
+      hookState: hookState,
+      options: options
```
bajtos: Please add a `where` property, I think the value should be `query.where`.

Notice that when the connector provides `connector.findOrCreate`, then we do create a new Model instance at https://github.com/strongloop/loopback-datasource-juggler/pull/586/files#diff-d717a01e40d8164976e4468af1bce663R666, thus `context.currentInstance` should be set too. Sorry for saying otherwise before.

lib/dao.js
```diff
@@ -1897,6 +1941,7 @@ DataAccessObject.updateAll = function (where, data, options, cb) {
     try {
       where = removeUndefined(where);
       where = Model._coerce(where);
+      var currentInstance = data;
```
bajtos: `updateAll` is possibly modifying multiple instances, therefore `context.currentInstance` should not be set.

test/persistence-hooks.suite.js
```diff
@@ -137,7 +137,7 @@ module.exports = function(dataSource, should) {
         });
       });
 
-      it('applies updates from `before save` hook', function(done) {
+      it('applies updates from `before save` hook for create', function(done) {
```
bajtos: In my opinion, the test suite name "PersistedModel.create" already provides the fact that these tests are testing "create" operation, therefore I think the suffix " for created" is redundant. Could you please remove?

test/persistence-hooks.suite.js
```diff
@@ -186,6 +186,52 @@ module.exports = function(dataSource, should) {
         });
       });
 
+      it('triggers `persist` hook on create', function(done) {
```
bajtos: Ditto, please remove " on create".

test/persistence-hooks.suite.js
```diff
@@ -186,6 +186,52 @@ module.exports = function(dataSource, should) {
...
+            observedContexts.should.eql(aTestModelCtx({
+              data: { id: 'new-id', name: 'a name' },
+              currentInstance: { extra: null, id: 'new-id', name: 'a name' }
+            }));
+
+            done();
+          });
+      });
+
+      it('applies updates from `persist` hook - on create', function(done) {
```
bajtos: Ditto, please remove " on create".

test/persistence-hooks.suite.js
```diff
@@ -402,6 +449,112 @@ module.exports = function(dataSource, should) {
           });
       });
 
+      if (dataSource.connector.findOrCreate) {
+        it('triggers `persist` hook on findOrCreate when found', function(done) {
```
bajtos: Ditto, please remove " on findOrCreate".

I'll stop commenting on this, could you please fix names of all newly added tests?

bajtos:
@PradnyaBaviskar AFAICT, `updateAttributes` does not trigger `persist` hook,please implement the hook there too.

Both "before save" and "after save" hook report `ctx.isNewInstance` when this information is available, I'd like the new "persist" hook to report this flag too:

- True for all CREATE operations
- False for all UPDATE operations
- Undefined for updateOrCreate, prototype.save,  prototype.updateAttributes, and updateAll operations.

(See [Operation Hooks](http://docs.strongloop.com/display/LB/Operation+hooks#Operationhooks-beforesave).)

test/persistence-hooks.suite.js
```diff
@@ -402,6 +449,112 @@ module.exports = function(dataSource, should) {
...
+          TestModel.findOrCreate(
+            { where: { name: existingInstance.name } },
+            { name: existingInstance.name },
+            function(err, record, created) {
+              if (err) return done(err);
+
+              record.id.should.eql(existingInstance.id);
+              observedContexts.should.eql(aTestModelCtx({
+                data: {
+                  id: getLastGeneratedUid(),
```
bajtos: If I understand this code correctly, then `getLastGeneratedUid()` is the same as `existingInstance.id`, right? I am proposing to  change the assertion here and assert against `existingInstance.id`, as that makes the test easier to understand, at least to me. The point is to use the same expected value for `record.id` and `context.data.id`.

test/persistence-hooks.suite.js
```diff
@@ -402,6 +449,112 @@ module.exports = function(dataSource, should) {
...
+          { where: { name: 'new-record' } },
+          { name: 'new-record' },
+          function(err, instance) {
+            if (err) return done(err);
+
+            if (dataSource.connector.findOrCreate) {
+              instance.should.have.property('extra', 'hook data');
+            } else {
+              instance.should.not.have.property('extra', 'hook data');
+            }
```
bajtos: Could you please add a comment explaining why different connectors behave differently?

Also please add a check that the record stored in the database always contains updated data.

test/persistence-hooks.suite.js
```diff
@@ -575,6 +764,45 @@ module.exports = function(dataSource, should) {
           done();
         });
       });
+
+      describe('persist hook', function() {
+        
+
+        it('triggers `persist` hook on updateOrCreate', function(done) {
```
bajtos: Please move this test to the existing test suite `describe('PersistedModel.updateOrCreate')` per my earlier [comment](https://github.com/strongloop/loopback-datasource-juggler/pull/586/files#r31209708).

test/persistence-hooks.suite.js
```diff
@@ -1287,6 +1515,42 @@ module.exports = function(dataSource, should) {
...
+
+        TestModel.updateAll(
+          { where: { name: existingInstance.name } },
+          { name: 'changed' },
+          function(err, instance) {
+            if (err) return done(err);
+
+            observedContexts.should.eql(aTestModelCtx({
+              data: { name: 'changed' },
+              currentInstance: { name: 'changed' },
```
bajtos: As I mentioned earlier, there should be no `currentInstance` in the context provided for `updateAll`.

bajtos:
@PradnyaBaviskar thank you for the update, I have reviewed the code - see the comments above. From now on, could you stop amending the single commit and start submitting changes in new commits, so that I could review only what's changed since the last review?

PradnyaBaviskar:
@bajtos - Addressed your comments above. Please review.

lib/dao.js
bajtos: > Please add `where`.

I think this comment fell under the radar, could you please fix it too?

bajtos:
Awesome, almost there! :+1: 

As I am looking at the context objects as observed by the unit-tests, I am little bit worried how easy/difficult they will be to use. I'll try to give this some more thoughts tomorrow and see if I can come up with a way how to make the context objects more consistent and/or describe a set of rules that would help users to remember and understand what to expect in the context in different operations.

lib/dao.js
PradnyaBaviskar: But for `save()`, the `context` object being passed to both `before save` and `after save` hooks did not have `where`. So, I had skipped adding it here.

PradnyaBaviskar:
Pushed the changes per the recent comments - please review.

bajtos:

PR #588, opened 2015-05-13
fix issue #587

mamboer:
fix issue #587 
also add the proper unit test

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -162,15 +162,12 @@ function defineScope(cls, targetClass, name, params, methods, options) {
           if (f._scope.collect &&
             condOrRefresh !== null && typeof condOrRefresh === 'object') {
             //extract the paging filters to the through model
-            if( typeof(condOrRefresh.limit) !== 'undefined' ){
-                f._scope.limit = condOrRefresh.limit;    
-            }
-            if( typeof(condOrRefresh.skip) !== 'undefined' ){
-                f._scope.skip = condOrRefresh.skip;    
-            }
-            delete condOrRefresh.limit;
-            delete condOrRefresh.skip;
- 
+            ['limit','offset','skip','order'].forEach(function( pagerFilter ){
+                if( typeof( condOrRefresh[ pagerFilter ] ) !== 'undefined' ){
+                    f._scope[ pagerFilter ] = condOrRefresh[ pagerFilter ];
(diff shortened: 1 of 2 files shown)


slnode:

lib/scope.js
```diff
@@ -161,6 +161,16 @@ function defineScope(cls, targetClass, name, params, methods, options) {
...
             condOrRefresh !== null && typeof condOrRefresh === 'object') {
+            //extract the paging filters to the through model
+            if( typeof(condOrRefresh.limit) !== 'undefined' ){
+                f._scope.limit = condOrRefresh.limit;    
+            }
+            if( typeof(condOrRefresh.skip) !== 'undefined' ){
+                f._scope.skip = condOrRefresh.skip;    
+            }
+            delete condOrRefresh.limit;
+            delete condOrRefresh.skip;
```
raymondfeng: Please note that `skip` is an alias to `offset`. We'll need to check that too.

mamboer: @raymondfeng 
How about the `order` filter, it should be checked too, right?

mamboer:
hi,
guys, i've updated the patch, pls take a review.

lib/scope.js
```diff
@@ -161,6 +161,13 @@ function defineScope(cls, targetClass, name, params, methods, options) {
           // see https://github.com/strongloop/loopback/issues/1076
           if (f._scope.collect &&
             condOrRefresh !== null && typeof condOrRefresh === 'object') {
+            //extract the paging filters to the through model
+            ['limit','offset','skip','order'].forEach(function( pagerFilter ){
+                if( typeof( condOrRefresh[ pagerFilter ] ) !== 'undefined' ){
+                    f._scope[ pagerFilter ] = condOrRefresh[ pagerFilter ];
+                    delete condOrRefresh[ pagerFilter ];
+                }
```
bajtos: Could you please follow the coding & formatting style of the surrounding code?

- `function(pageFilter)` instead of `function( pageFilter )`
- `if (typeof(condOrRefresh[pageFilter]))`

etc.

slnode:

lib/scope.js
mamboer: coding & formatting style fixed

raymondfeng:
ok to test

Issue #166, opened 2014-07-08
hasMany through - filter/where/order not working


fabien:
```
var Space = app.loopback.getModel('Space');
var User = app.loopback.getModel('User');
var Account = app.loopback.getModel('Account');

Space.hasAndBelongsToMany('users', { through: Account });
Space.hasMany('accounts');

User.hasAndBelongsToMany(Space, { as: 'spaces', through: Account });
User.hasMany('accounts');
```

Example: `/api/spaces/client-a/users?filter[where][username]=toby` // result: []
Example: `/api/spaces/client-a/users?filter[order]=username` // result: unordered array

Regarding sorting, since models are fetched individually (byId vs. find), ordering is not applied either. This might very well be a limitation of how juggler handles relations at the moment (very flexible regarding mixed datasources, but not optimal performance wise).

I think the best approach would be to gather the related ids from the through-model first, then issue a second query on the actual target model, scoped to the previously retrieved ids - this is not only more efficient, but should enable the filtering capabilities as expected.

superkhau:
Possibly related to https://github.com/strongloop/loopback-datasource-juggler/pull/476

superkhau:
@raymondfeng Is this related to strongloop/loopback/issues/1076

raymondfeng:
@superkhau I think so. Can you verify?

superkhau:
>/api/spaces/client-a/users?filter[where][username]=toby // result: []

@fabien What is `client-a` in this example?

fabien:
@superkhau it's just an `:id` value

superkhau:
>Example: /api/spaces/client-a/users?filter[where][username]=toby // result: []
This issue has been resolved in strongloop/loopback#1076.

>Example: /api/spaces/client-a/users?filter[order]=username // result: unordered array
This issue is still unresolved due to the way we query related models. The current implementation performs a query on the join table and sorts each returned row individually instead of sorting all the resulting rows together as a whole.

Reassigned to @raymondfeng

maelena87:
+1 @superkhau Example: /api/spaces/client-a/users?filter[order]=username // result: unordered array 
still not solved, right? still getting an unordered array.

jannyHou:
I am working on this issue and create a repo to test the filters:
https://github.com/jannyHou/loopback-sandbox/tree/has-and-belongs-to-many

I tried 6 filters:
- fields
- include
- limit
- order
- skip
-where

[...]

PR #590, opened 2015-05-13
DB Call Optimization in relation includes - Fixes #408 & #166

### LINKED ISSUE #408: Spike: LB performance issue with relational database (SELECT N+1)
http://stackoverflow.com/questions/24187582/loopback-performance-issue-with-relational-databases

ningsuhen:
The number of DB calls was to high especially in case of nested includes. New relic was showing ~100x db calls for includes. I had to fix it for our project and use our custom branch. But I have written few test cases and verified the tests. Please merge this if you think this looks good. 

Performance Improvement:
================
Tool: gobench (https://github.com/cmpxchg16/gobench)

Before Optimization:
================
Concurrent Clients which can be connected w/o error per single process (on my mac): ~300 but average response time was ~20 secs
Maximum Requests/sec per single process: ~47 req/s with 500 clients
Maximum Requests/sec with 3 cluster processes: ~78 req/s with 500 clients

After Optimization:
================
Note: Because of increase in processing, there's a negative impact
Concurrent Users which can be connected per single process (on my mac): ~200 but average response time was ~3 secs
Maximum Requests/sec per single process: ~77 req/s with 500 clients
Maximum Requests/sec with 3 cluster processes: ~147 req/s with 500 clients

Conclusion
================
Although there's a performance impact on nodejs/web layer, it is much easier to scale web application layer than Database. So, I think this is a good solution for now. Please merge & optimize it more if possible.

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -2,6 +2,7 @@ var async = require('async');
 var utils = require('./utils');
 var isPlainObject = utils.isPlainObject;
 var defineCachedRelations = utils.defineCachedRelations;
+var debug = require('debug')('loopback:include');
 
 /*!
  * Normalize the include to be an array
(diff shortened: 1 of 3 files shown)


raymondfeng:
@ningsuhen Great work! 

What's your take on the DB limit on the number of values for the `IN` cause?  For example, if the `through` table has many matching ids and it goes beyond the limit of `IN`, should we consider to do the queries in multiple calls and merge the results (a special sort of pagination) ? That will complicate the sorting through.

raymondfeng:
I fixed a few issues and landed the PR. Thanks!

We'll continue to discuss the IN limit issue.

ningsuhen:
Thanks @raymondfeng. 
Assuming IN array limit will be a problem majorly in HasManyThrough relations. Using loop for this is also a problem especially when we have to sort & limit results. Optimizing client side is ok only upto a certain limit. With large number of records, it won't work. 

I'm not sure of the implications but another way is to denormalize indexed data (of target model) onto the through model. That way, we can sort and limit the results in the through Model itself. This is of course assuming each through model has a valid target model object.

PR #591, opened 2015-05-14
Add transaction support

raymondfeng:
/to @ritch @bajtos 

This PR adds local transaction apis for relational DBs.

Connect to strongloop/loopback#109

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -1417,7 +1417,7 @@ BelongsTo.prototype.related = function (condOrRefresh, options, cb) {
     cb = condOrRefresh;
     condOrRefresh = false;
   } else if (typeof options === 'function' && cb === undefined) {
-    // order.customer(options, cb)
+    // order.customer(condOrRefresh, cb)
     cb = options;
     options = {};
   }
@@ -1518,7 +1518,7 @@ BelongsTo.prototype.related = function (condOrRefresh, options, cb) {
  */
 BelongsTo.prototype.getAsync = function (options, cb) {
   if (typeof options === 'function' && cb === undefined) {
-    // order.customer(options, cb)
+    // order.customer.getAsync(cb)
     cb = options;
     options = {};
   }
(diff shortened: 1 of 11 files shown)


violet-day:
nice~

raymondfeng:
@ritch See the usage example at https://github.com/strongloop/loopback-connector/blob/feature/transaction/test/transaction.test.js

lib/relation-definition.js
```diff
@@ -1606,7 +1682,12 @@ RelationDefinition.hasOne = function (modelFrom, modelTo, params) {
  * @param {String|Object} err Error string or object
  * @param {Object} The newly created target model instance
  */
-HasOne.prototype.create = function (targetModelData, cb) {
+HasOne.prototype.create = function (targetModelData, options, cb) {
+  if (typeof options === 'function' && cb === undefined) {
+    // order.customer(options, cb)
```
ritch: is this comment correct?

ritch:
LGTM

lib/transaction.js
```diff
@@ -0,0 +1,181 @@
...
+ * or rollback an inactive transaction will receive an error from the callback.
+ *
+ * Please also note that the transaction is only honored with the same data
+ * source/connector instance. CRUD methods will not join the current transaction
+ * if its model is not attached the same data source.
+ *
+ */
+TransactionMixin.beginTransaction = function(options, cb) {
+  cb = cb || utils.createPromiseCallback();
+  if (Transaction) {
```
bajtos: FWIW, if I am reading this code correctly, then `Transaction` is always set, because it is loaded from `loopback-connector` which is shipped as juggler's dependency. Should we check whether the backing connector (something like `Model.dataSource.connector`) supports transactions instead?

bajtos:
Cool stuff :+1:

violet-day:
Hi,is it possible to promiseify transaction?

bajtos:
@violet-day AFAICT, tall transaction methods are already promisified. If you don't provide a callback, then they return a promise.

raymondfeng:
Transaction apis support promise. See https://github.com/strongloop/loopback-connector-mysql/blob/master/test/transaction.promise.test.js

violet-day:
@bajtos @raymondfeng  Cool :+1:

PR #597, opened 2015-05-20
Enhance fieldsToArray to consider strict mode

raymondfeng:
/to @ritch 
/cc @Joesimonsen 

Connect to https://github.com/strongloop-internal/scrum-cs/issues/211


ritch:
Isn't this a breaking change?

ritch:
Well I guess we aren't changing any tests or documentation. LGTM

Issue #464, opened 2015-02-23
Add before/after hooks for connector native operations


raymondfeng:
The connector translates model methods into native operations implemented using a driver to its backend. It's often desirable to inject some logic before/after a native operation is invoked. For example, connectors for databases should allow hooks to intercept DB commands. REST/SOAP/REMOTE connectors should allow hooks to access HTTP req/res objects.

- [x] REST connector
- [x] SOAP connector
- [x] SQL connectors
- [x] MongoDB connector

chandadharap:
PRs are not done with review, moving it back to review.

fabien:
@raymondfeng interesting stuff +1

PR #598, opened 2015-05-20
Mix in observer apis to the connector

raymondfeng:
This PR does a few things:

1. Enhance the observer mixin so that it allows around behavior
2. Enable connectors with observer apis

Connect to #464

### DIFF
--- a/lib/observer.js
+++ b/lib/observer.js
@@ -88,11 +88,59 @@ ObserverMixin.notifyObserversOf = function(operation, context, callback) {
     );
   });
   return callback.promise;
-}
+};
 
 ObserverMixin._notifyBaseObservers = function(operation, context, callback) {
   if (this.base && this.base.notifyObserversOf)
     this.base.notifyObserversOf(operation, context, callback);
   else
     callback();
-}
+};
+
+/**
+ * Run the given function with before/after observers. It's done in three serial
+ * steps asynchronously:
+ *
+ * - Notify the registered observers under 'before ' + operation
+ * - Execute the function
+ * - Notify the registered observers under 'after ' + operation
+ *
+ * If an error happens, it fails fast and calls the callback with err.
+ *
+ * @param {String} operation The operation name
+ * @param {Context} context The context object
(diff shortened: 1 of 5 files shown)


lib/observer.js
```diff
@@ -88,11 +88,51 @@ ObserverMixin.notifyObserversOf = function(operation, context, callback) {
...
 ObserverMixin._notifyBaseObservers = function(operation, context, callback) {
   if (this.base && this.base.notifyObserversOf)
     this.base.notifyObserversOf(operation, context, callback);
   else
     callback();
-}
+};
+
+/**
+ * Create an around behavior by notifying before/after observers
```
ritch: The wording of this is confusing... Not really sure what this does by reading it.

raymondfeng: Fixed the comment

lib/observer.js
```diff
@@ -88,11 +88,59 @@ ObserverMixin.notifyObserversOf = function(operation, context, callback) {
...
+ *
+ * If an error happens, it fails fast and calls the callback with err.
+ *
+ * @param {String} operation The operation name
+ * @param {Context} context The context object
+ * @param {Function} fn The task to be invoked as fn(done) or fn(context, done)
+ * @param {Function} callback The callback function
+ * @returns {*}
+ */
+ObserverMixin.around = function(operation, context, fn, callback) {
```
ritch: shouldn't this method name conform to the `notifyObserversOf` convention?

Consider this idea (in the context of your usage):

```js
function work(done) {
  self.connector.commit(self.connection, done);
}

self.notifyObserversBeforeAndAfter('commit', context, work, function(err) {
  // Deference the connection to mark the transaction is not active
  // The connection should have been released back the pool
  self.connection = null;
  cb(err);
});
```

test/memory.test.js
```diff
@@ -584,4 +584,48 @@ describe('Memory connector with options', function() {
 
 });
 
+describe('Memory connector with hookd', function() {
```
ritch: hookd => typo?

ritch:
I think we should rename `around` otherwise LGTM

PR #600, opened 2015-05-21
Fix for https://github.com/strongloop/loopback/issues/1401

raymondfeng:
/to @ritch 

Connect to https://github.com/strongloop/loopback/issues/1401

### DIFF
--- a/test/include.test.js
+++ b/test/include.test.js
@@ -1,6 +1,9 @@
 // This test written in mocha+should.js
 var should = require('./init.js');
 var async = require('async');
+var assert = require('assert');
+
+var DataSource = require('../').DataSource;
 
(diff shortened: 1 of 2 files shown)


raymondfeng:
@ritch ping

test/include.test.js
```diff
@@ -690,3 +693,96 @@ function clearAndCreate(model, data, callback) {
...
+        }
+      }
+    });
+
+    async.waterfall([
+      createChallengers,
+      createGameParticipations,
+      createResults,
+      function() {
+        done();
```
ritch: ignoring errors here on purpose?

ritch:
Minor comment otherwise LGTM

test/include.test.js
raymondfeng: Good catch, fixed.

PR #601, opened 2015-05-21
Fix ReferenceError: definition is not defined

slonoed:
occure when try to delete object in has one relation, but object not created yet


slnode:

raymondfeng:
@slonoed Thanks for the patch. Please sign the CLA - https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler

PR #602, opened 2015-05-22
Enhance the apis and add more tests

raymondfeng:
/to @ritch or @bajtos 

This PR enhances the observer APIs and add more tests.

### DIFF
--- a/lib/observer.js
+++ b/lib/observer.js
@@ -58,16 +58,35 @@ ObserverMixin.clearObservers = function(operation) {
 };
 
 /**
- * Invoke all async observers for the given operation.
- * @param {String} operation The operation name.
+ * Invoke all async observers for the given operation(s).
+ * @param {String|String[]} operation The operation name(s).
  * @param {Object} context Operation-specific context.
  * @param {function(Error=)} callback The callback to call when all observers
  *   has finished.
  */
 ObserverMixin.notifyObserversOf = function(operation, context, callback) {
-  var observers = this._observers && this._observers[operation];
-
+  var self = this;
   if (!callback) callback = utils.createPromiseCallback();
+
+  function createNotifier(op) {
+    return function(ctx, done) {
+      if (typeof ctx === 'function' && done === undefined) {
+        done = ctx;
+        ctx = context;
+      }
+      self.notifyObserversOf(op, context, done);
+    };
+  }
+
+  if (Array.isArray(operation)) {
+    var tasks = [];
+    for (var i = 0, n = operation.length; i < n; i++) {
+      tasks.push(createNotifier(operation[i]));
+    }
+    return async.waterfall(tasks, callback);
+  }
+
+  var observers = this._observers && this._observers[operation];
 
   this._notifyBaseObservers(operation, context, function doNotify(err) {
     if (err) return callback(err, context);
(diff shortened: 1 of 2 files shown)


test/async-observer.test.js
```diff
@@ -194,6 +206,57 @@ describe('async observer', function() {
             done();
           });
       });
+
+    it('should allow observers to skip other ones',
+      function(done) {
+        TestModel.observe('before invoke',
+          function(context, next) {
+            notifications.push('before invoke');
+            context.callback(null, 0);
```
ritch: Looking at this in context the name `callback` is pretty misleading since we are controlling the flow of the entire chain.

lib/observer.js
```diff
@@ -58,14 +58,25 @@ ObserverMixin.clearObservers = function(operation) {
...
+  var observers = [];
+
+  if (Array.isArray(operation)) {
+    for (var i = 0, n = operation.length; i < n; i++) {
+      var items = this._observers && this._observers[operation[i]];
+      if (items) {
+        observers = observers.concat(items);
+      }
+    }
+  } else {
```
ritch: Shouldn't this just be `else if(this._observers)`

test/async-observer.test.js
raymondfeng: What's a better name?

ritch: `done` or `end`

raymondfeng: rename to `end`

bajtos:
I have few concerns about the changes introduced in the PR.

1) If I am reading the code correctly, then the current implementation of `notifyObserversOf` for an array of hook names is not equivalent to calling `notifyObserversOf` multiple times, the difference is in the order in which the observers from the base class are called.

```
Car.notifyObserversOf(['hookA', 'hookB'])
  1. Car hookA
  2. Car hookB
  3. PersistedModel hookA
  4. PersistedModel hookB

Car.notifyObserversOf('hookA')
  1. Car hookA
  2. PersistedModel hookA
Car.notifyObserversOf('hookB')
  3. Car hookB
  4. PersistedModel hookB
```

IMO, this is a subtle change in behaviour that may be very difficult to troubleshoot, and therefore I think it should be rather fixed.

2) `notifyObserversOf` with an array of hooks to invoke assumes that all hooks expect the same context object. This feels like a bad design to me - if two hooks expect the same context and are invoked at the same time, then perhaps they should be merged to a single hook instead?

3) The behaviour of calling `context.end()` depends on the order in which the observers are registered. We don't have any good API for controlling that order when registering a new observer, the situation reminds me of express middleware registration. I am worrying that we may end up with a flood of confused users.

bajtos:
To clarify my last comment, 2) and 3) are quite subjective, it's ok to leave them out of scope of this pull request. 

I think 1) is pretty important and I would like to see it fixed before landing this PR. Please add a unit-test to verify (and document) the behaviour.

test/async-observer.test.js
bajtos: :+1:  for `end()`

raymondfeng:
@bajtos Good catch. Both 1 and 2 are fixed now. PTAL.

lib/observer.js
```diff
@@ -58,17 +58,36 @@ ObserverMixin.clearObservers = function(operation) {
...
+      self.notifyObserversOf(op, context, done);
+    };
+  }
+
+  if (Array.isArray(operation)) {
+    var tasks = [];
+    for (var i = 0, n = operation.length; i < n; i++) {
+      tasks.push(createNotifier(operation[i]));
+    }
+    return async.waterfall(tasks, callback);
```
bajtos: I am proposing to simplify this using `Array.prototype.map`:

```js
return async.waterfall(operation.map(createNotifier), callback);
```

bajtos:
FWIW, I think 2) is not addressed, but that's ok.

The code LGTM. Please fix the build before landing the patch.

PR #604, opened 2015-05-22
addressing #603

### LINKED ISSUE #603: HasMany.prototype.destroyById not removing instance from cache
In [`relation-definition.js:872`](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L872), `HasMany.prototype.destroyById()`

```js
// ...
self.removeFromCache(inst[fkId]);
inst.destroy(options, cb);
// ...
```

`removeFromCache()` expects `fkId` and instead receives `undefined` (as an instance of a model object, subscripting by a numerical ID is meaningless).

This causes the removal from cache to fail.

pcperini:
Resolves #603

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -869,7 +869,7 @@ HasMany.prototype.destroyById = function (fkId, options, cb) {
     if (err) {
       return cb(err);
     }
-    self.removeFromCache(inst[fkId]);
+    self.removeFromCache(fkId);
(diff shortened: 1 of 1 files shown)


slnode:

raymondfeng:
Please sign the CLA - https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler

raymondfeng:
ok to test

PR #616, opened 2015-06-05
default values are being overwritten in upsert

clarkbw:
clarkbw/loopback-ds-timestamp-mixin#11 was filed and it appears to be an issue with the upsert coming from juggler.

This PR creates a test that fails and I believe I’ve tracked down the area of the code that causes the issue, wrapped in `console.log` statements currently.

When you run the test it will print the following information:
```
not instance { id: 1000, name: 'updated name' }
inst { id: 1000,
  name: 'updated name',
  createdAt: Fri Jun 05 2015 09:59:03 GMT-0700 (PDT) }
      1) should preserve default property values
```

As you can see in the debug output the `createdAt` field is being generated and then overwriting the existing `createdAt` field.  We need to prevent the default fields from being generated or passed to the update because default functions that are not idempotent will always be storing new values.

I’m investigating a solution, any comments are appreciated.

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -238,6 +238,7 @@ Memory.prototype.updateOrCreate = function (model, data, options, callback) {
 };
 
 Memory.prototype.save = function save(model, data, options, callback) {
+  var self = this;
   var id = this.getIdValue(model, data);
   var cachedModels = this.collection(model);
   var modelData = cachedModels && this.collection(model)[id];
@@ -247,7 +248,7 @@ Memory.prototype.save = function save(model, data, options, callback) {
   }
   this.collection(model)[id] = serialize(data);
   this.saveToFile(data, function(err) {
-    callback(err, data, { isNewInstance: !modelData });
+    callback(err, self.fromDb(model, data), { isNewInstance: !modelData });
   });
 };
 
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -445,8 +445,8 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
         data = ctx.data;
         var update = data;
         var inst = data;
-        if(!(data instanceof Model)) {
-          inst = new Model(data);
+        if (!(data instanceof Model)) {
+          inst = new Model(data, { applyDefaultValues: false });
         }
         update = inst.toObject(false);
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -43,6 +43,9 @@ function ModelBaseClass(data, options) {
     // Default to true
     options.applySetters = true;
   }
+  if (!('applyDefaultValues' in options)) {
+    options.applyDefaultValues = true;
+  }
   this._initProperties(data, options);
 }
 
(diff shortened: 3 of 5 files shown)


lib/dao.js
```diff
@@ -415,7 +415,9 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
         var update = data;
         var inst = data;
         if(!(data instanceof Model)) {
+          console.log('not instance', data);
           inst = new Model(data);
```
raymondfeng: Maybe the constructor should have an option to skip default value generation in this case.

clarkbw: Ok, I'll look into adding another option here: https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model.js#L42
With that value turned on we skip the default value and defaultFn generation blocks.

clarkbw:
> What problem do you try to solve? Is it related to the default value override?

The memory connector was returning incorrect values such that my tests were working with the proposed changes but failing because of the memory connector issue.

In this test I'm trying to compare a Date object to what was being returned as a date represented by a string.
https://github.com/clarkbw/loopback-datasource-juggler/commit/5fa09c9fda6bad481ab418d2fe5c752b787f2b8b#diff-404450f6fd92e7f3f2031232213c07c5R637

Again, I can pull the memory connector issue out into a separate PR because it should be fixed and is separate from the issue related to this PR but was in the way of fixing it.

raymondfeng:
> I think you should call fromDb inside the save method.

@clarkbw What's your take on the suggestion above?

clarkbw:
> What's your take on the suggestion above?

Just ran it through the tests, its the right call.  Likely fixes an issue in the `updateAttributes` method as well.

clarkbw:
Need to finish with more tests specifically for the `setDefaultValues` option.

clarkbw:
Changed the option name to `applyDefaultValues` to fit in better with other option naming and usage.

clarkbw:
Did another pass over this change and I don't think there's a need for more individual tests on the `applyDefaultValues` option.  Let me know if you feel otherwise.

@raymondfeng I think this is ready for your review

lib/connectors/memory.js
```diff
@@ -247,8 +247,8 @@ Memory.prototype.save = function save(model, data, options, callback) {
   }
   this.collection(model)[id] = serialize(data);
   this.saveToFile(data, function(err) {
-    callback(err, data, { isNewInstance: !modelData });
-  });
+    callback(err, this.fromDb(model, data), { isNewInstance: !modelData });
+  }.bind(this));
```
raymondfeng: Can we use `self` instead of `bind`? `bind` is expensive.

clarkbw: Good point, sorry.  Fixed.

clarkbw:
@raymondfeng I updated [the memory connector](https://github.com/clarkbw/loopback-datasource-juggler/commit/1066313f3f3788e12fc0b57ffcb6d2cd487edaca#diff-98aa884e786e8c02d53648b6f0afaa82R241) as you requested to use `self = this` instead of `bind`.  The code has been rebased to the master branch and passes all tests.  Let me know if there's anything else you'd like to see.

PR #621, opened 2015-06-10
Create a script to describe operation hooks

bajtos:
/cc @crandmck You can find an example output in [slack archive](https://strongloop.slack.com/files/bajtos/F065RJ2JE/hooks.html). Perhaps we can include the table in [Operation Hooks](docs.strongloop.com/display/LB/Operation+hooks)? Ideally in a way that will make it easy to update the table content in the future, e.g. by copy-n-paste from the HTML file into raw confluence source markup.

### DIFF
--- a/support/describe-operation-hooks.js
+++ b/support/describe-operation-hooks.js
@@ -1,0 +1,220 @@
+/*
+ * Describe context objects of operation hooks in comprehensive HTML table.
+ * Usage:
+ *  $ node support/describe-operation-hooks.js > hooks.html
+ *  $ open hooks.hml
+ *
+ */
+var Promise = global.Promise = require('bluebird');
+var DataSource = require('../').DataSource;
+var Memory = require('../lib/connectors/memory').Memory;
+
+var HOOK_NAMES = [
+  'access',
+  'before save', 'persist', 'after save',
+  'before delete', 'after delete'
+];
+
+var dataSources = [
+  createOptimizedDataSource(),
+  createUnoptimizedDataSource()
+];
+
+var observedContexts = [];
+var lastId = 0;
+
+Promise.onPossiblyUnhandledRejection(function(err) {
+  console.error('POSSIBLY UNHANDLED REJECTION', err.stack);
+});
+
+var operations = [
+  function find(ds) {
+    return ds.TestModel.find({ where: { id: '1' } });
+  },
+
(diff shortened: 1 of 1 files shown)


raymondfeng:
Cool stuff! LGTM.

crandmck:
Can you explain to me: 
- What this script does, and why one would use it? Is it used on a per-app basis?
- How someone uses it?  The comment says to do 
```
$ node support/describe-operation-hooks.js > hooks.html
```
But I believe that will only work if you are running it from the `loopback-datasource-juggler` module root dir, but normally a user won't be there since it's in `/usr/local/lib/node_modules`.

bajtos:
@crandmck sure, let me explain in more detail.

This script is an internal tool for people working on and maintaining loopback-datasource-juggler. It is not intended for general use by LoopBack users.

The file "hooks.html" produced by this script contains a comprehensive overview of context properties for different CRUD methods and their hooks. @raymondfeng and I think this overview may be useful to LoopBack users and therefore it would be good to include it in our documentation.

The script produces very basic HTML, I used the same structure as Confluence does. I think we may be able to paste this HTML code (excluding the style header) directly into Confluence in raw edit mode. That way it will be easy to performs updates in the future - just run the script again and replace the content in Confluence with the new HTML code. That assumes we don't need to edit/polish the table in Confluence after it was imported.

Perhaps we can setup a short Hangout to discuss this live?

bajtos:
For posterity

> How you use it? The comment says to do
```
$ node support/describe-operation-hooks.js > hooks.html
```
> But I believe that will only work if you are running it from the loopback-datasource-juggler module root dir.

Yes, the command above should be run from loopback-datasource-juggler module root dir. I think it's ok because the script is for juggler developers, not for module users.

crandmck:
Ah, thanks... much clearer now!

The suggestion to add hooks.html to a Confluence page sounds reasonable.  We can discuss, but if you send me the hooks.html file (eg. via email or however you wish), then I can just try it out. We might not even need a meeting.

crandmck:
OK, here is the page: http://docs.strongloop.com/display/LB/Operation+hooks+summary
It's currently hidden, so you have to login to see it.

I do have a number of questions about how to interpret this table, so perhaps we should chat about it.  @bajtos I'll ping you next week.

Issue #623, opened 2015-06-10
Include includePolymorphic is broken


fabien:
The same targetIds are fed to `processPolymorphicType`, which have been clobbered here: 

https://github.com/strongloop/loopback-datasource-juggler/blob/d009557a76cc8fe39314c2c1f1b4cdfe2520b4e2/lib/include.js#L540

@raymondfeng @ningsuhen

fabien:
Hotfix while I'm investigating the other include issues: add `var targetIds = targetIdsByType[modelType];`

at: https://github.com/strongloop/loopback-datasource-juggler/blob/d009557a76cc8fe39314c2c1f1b4cdfe2520b4e2/lib/include.js#L563

fabien:
Furthermore `includePolymorphic` does not work with `belongsTo` at all. While I applaud the refactoring of this aspect of Juggler, I'm afraid it introduces quite some breaking changes.

ningsuhen:
@fabien , thanks for reporting. I thought the existing tests must have covered it. I'll fix the issues along with tests asap. Can you also add the case includePolymorphic belongsTo? I remember it working for me.

fabien:
@ningsuhen thanks for your prompt reply! Do you want me to add a test-case for the polymorphic belongsTo, or can you give it a shot first?

Unfortunately the existing tests apparently don't cover include-related functionality for all relation setups. Before it was no big deal, because it would work off the existing code, but your (much welcomed!) refactoring goes much deeper in implementing it independently.

Can you look at the other issue regarding referencesMany, too? Thanks!

ningsuhen:
@fabien , I'll start working on the reported issues tonight IST. It would be great if you can add a test-case for polymorphic belongsTo before that. Else, I'll add the test-case and let you know to check if matches your case.

fabien:
@ningsuhen great, I'll see if I can add such a test to a feature branch of mine. I'll update here with a link.

fabien:
@ningsuhen I've added two failing test cases (marked // Fixme @ningsuhen) here:

https://github.com/fabien/loopback-datasource-juggler/tree/fix/include-refactor

Please review, and see if you can identify any other includes that were not covered by tests. I think the 2 additional tests cover hasOne and belongsTo (as inverse relations) now.

Issue #622, opened 2015-06-10
Include of referencesMany issues (after include refactoring)


fabien:
One of the key aspects of a referencesMany relation is that it should keep the order of the ids when inclusion fetches them. Effectively `findByIds` (here https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L2966) makes sure that this is the case.

However, the refactoring here https://github.com/strongloop/loopback-datasource-juggler/commit/d009557a76cc8fe39314c2c1f1b4cdfe2520b4e2 does not take this into account.

I'm currently investigating other possibly include-related issues, as I now have many failing tests after this particular commit. 

/cc @raymondfeng

fabien:
There's another issue hidden in there, take the following setup:

```
Image.referencesMany('users');
User.hasMany('images', { foreignKey: 'userIds' });

User.find({ include: 'images' }, function() { ... });

Uncaught TypeError: Cannot read property '__cachedRelations' of undefined
      at linkManyToOne (loopback-datasource-juggler/lib/include.js:520:16)
```

This happens because the `target[relation.keyTo]` will be an array with ids, not just a single id value in this case.

fabien:
Turns out this was a very specific case, which worked thanks to MongDB's transparant handling of arrays - e.g. it doesn't matter if a property is an array or not when querying. In other words, it used to work, even with includes, but it has surfaced as a deeper issue. However, I do want to see this resolved. I'm working on it right now.

fabien:
@ningsuhen please see #625 - perhaps we can come to a solution like that; the tests now pass. Even if we don't implement a `contains` operator, the new `linkManyToOne` implementation should safely handle it automatically in case of MongoDB. So it would be great to consider this change anyway.

ningsuhen:
Thanks @fabien , I'll use #625 's linkManyToOne and fix #623 together. I'll leave the `contains` operator implementation for @raymondfeng to look into. Btw, feel free to send pull requests (with tests) if you have fixed the issue.

ningsuhen:
This has been addressed in https://github.com/strongloop/loopback-datasource-juggler/pull/626 but the case of "hasMany -> referencesMany" doesn't have any documentation or test cases. @fabien, your test case change was breaking other tests and therefore, I haven't added the tests for referencesMany with hasMany.

fabien:
[...]

PR #626, opened 2015-06-11
Fix for issues #622 & #623

ningsuhen:
polymorphic hasOne needs separate handling and hasMany->referencesMany has a unique case which is fixed for MongoDB as suggested by @fabien.

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -272,7 +272,10 @@ Inclusion.include = function (objects, include, options, cb) {
     }
     else {
       if (polymorphic) {
-        return includePolymorphic(cb);
+        if (relation.type === 'hasOne') {
+          return includePolymorphicHasOne(cb);
+        }
+        return includePolymorphicBelongsTo(cb);
       }
       if (relation.type === 'embedsOne') {
         return includeEmbeds(cb);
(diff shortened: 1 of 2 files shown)


raymondfeng:
LGTM. @fabien Please review.

lib/include.js
```diff
@@ -623,6 +631,66 @@ Inclusion.include = function (objects, include, options, cb) {
...
+            relation.modelTo.include(targets, subInclude, options, next);
+          });
+        }
+        //process each target object
+        tasks.push(targetLinkingTask);
+        function targetLinkingTask(next) {
+          async.each(targets, linkOneToOne, next);
+          function linkOneToOne(target, next) {
+            var sourceId = target[relation.keyTo];
+            var obj = objIdMap[sourceId.toString()];
```
fabien: I wonder if we should check that `obj` is actually defined at this stage (defensive coding ...)? There are probably other parts of this code where this might be appropriate.

fabien:
@raymondfeng other than the comment, LGTM

lib/include.js
ningsuhen: I think that it makes sense only for the referencesMany -> hasMany case because it is possible to have no value for a map key. In other cases, it should always have a map entry.

raymondfeng:
@ningsuhen Can you address the comment from @fabien so that we can merge the PR?

ningsuhen:
@raymondfeng , I had replied to the comment. I think it is applicable only in case of referencesMany <-> hasMany because `target[relation.keyTo]` can be an Array of ObjectIds instead of an ObjectId. Even if there are other cases where it can be an array, we need to handle it as an array apart from just `null` checking. Is there any other case where referencesMany can be used in conjunction with another relation?

fabien:
@raymondfeng LGTM

ningsuhen:
@fabien, @raymondfeng  , I have added the checks anyway :).

PR #630, opened 2015-06-15
Promisify 'automigrate'

PradnyaBaviskar:

This PL, provides `promise` support to `automigrate` method of juggler's datasource.

@bajtos - can you review?

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -775,9 +775,12 @@ DataSource.prototype.automigrate = function (models, cb) {
     models = undefined;
   }
 
+  cb = cb || utils.createPromiseCallback();
+
   if (!this.connector.automigrate) {
     // NOOP
-    return cb && process.nextTick(cb);
+    process.nextTick(cb);
+    return cb.promise;
   }
 
   // First argument is a model name
@@ -791,7 +794,8 @@ DataSource.prototype.automigrate = function (models, cb) {
     models = models || Object.keys(attachedModels);
 
     if (models.length === 0) {
-      return cb && process.nextTick(cb);
+      process.nextTick(cb);
+      return cb.promise;
     }
 
     var invalidModels = models.filter(function (m) {
@@ -799,16 +803,16 @@ DataSource.prototype.automigrate = function (models, cb) {
     });
 
     if (invalidModels.length) {
-      return process.nextTick(function () {
-        if (cb) {
+      process.nextTick(function () {
           cb(new Error('Cannot migrate models not attached to this datasource: ' +
           invalidModels.join(' ')));
-        }
       });
+      return cb.promise;
     }
   }
 
   this.connector.automigrate(models, cb);
+  return cb.promise;
 };
 
 /**
--- a/test/memory.test.js
+++ b/test/memory.test.js
@@ -463,10 +463,30 @@ describe('Memory connector', function() {
       });
     });
 
+    it('automigrate all models - promise variant', function(done) {
+      ds.automigrate()
+        .then(function(result) {
+          done();
+        })
+        .catch(function(err){
+          done(err);
+        });
+    });
+
     it('automigrate one model', function(done) {
       ds.automigrate('m1', function(err) {
         done(err);
       });
+    });
+
+    it('automigrate one model - promise variant', function(done) {
+      ds.automigrate('m1')
+        .then(function(result) {
+          done();
+        })
+        .catch(function(err){
+          done(err);
+        });
     });
 
     it('automigrate one or more models in an array', function(done) {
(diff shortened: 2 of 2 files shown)


lib/datasource.js
```diff
@@ -775,8 +775,14 @@ DataSource.prototype.automigrate = function (models, cb) {
...
 
+  cb = cb || utils.createPromiseCallback();
+
   if (!this.connector.automigrate) {
     // NOOP
+    if(cb.promise) {
+      process.nextTick(cb);
+      return cb.promise;
+    }
     return cb && process.nextTick(cb);
```
bajtos: Please simplify this code, `cb` is always defined thanks to L778.

I'd personally write this block as follows:

```js
if (!this.connector.automigrate) {
  process.nextTick(cb);
  return cb.promise;
}
```

lib/datasource.js
```diff
@@ -791,6 +797,10 @@ DataSource.prototype.automigrate = function (models, cb) {
     models = models || Object.keys(attachedModels);
 
     if (models.length === 0) {
+      if(cb.promise) {
+        process.nextTick(cb);
+        return cb.promise;
+      }
       return cb && process.nextTick(cb);
```
bajtos: Ditto.

lib/datasource.js
```diff
@@ -799,16 +809,24 @@ DataSource.prototype.automigrate = function (models, cb) {
     });
 
     if (invalidModels.length) {
-      return process.nextTick(function () {
+      var error = process.nextTick(function () {
```
bajtos: AFAIK, `process.nextTick` does not return any error object, I think it returns `undefined`. Please simplify this block.

lib/datasource.js
```diff
@@ -799,16 +809,24 @@ DataSource.prototype.automigrate = function (models, cb) {
...
+        return cb.promise;
+      }
+      return error;
     }
   }
 
   this.connector.automigrate(models, cb);
+  if (cb.promise) {
+    return cb.promise;
+  }
```
bajtos: It's not necessary to check if `cb.promise` is truthy, you can always `return cb.promise`. The result will be the same(*):
 - When `cb.promise` is not defined, the return value will be `undefined`.
 - When a function does not return anything (does not call `return`), then the return value is `undefined `too.

(*) Exception: if `cb.promise` is defined and falsy (e.g. `0`, `null` or `''`), your version returns `undefined` while my version returns the falsy value. I think this difference does not really matter.

test/memory.test.js
```diff
@@ -463,25 +463,88 @@ describe('Memory connector', function() {
...
     });
 
+    it('automigrate all models - promise variant', function(done) {
+      ds.automigrate()
+        .then(function(result) {
+          done();
+        })
+        .catch(function(err){
+          done(err);
+        });
```
bajtos: FYI: this can be shortened to `.catch(done)`. For the purpose of these tests, I think it's ok to keep the longer version. It's more explicit and thus it may be easier to understand.

test/memory.test.js
```diff
@@ -463,25 +463,88 @@ describe('Memory connector', function() {
...
     it('automigrate reports errors for models not attached', function(done) {
       ds.automigrate(['m1', 'm2'], function(err) {
         err.should.be.an.instanceOf(Error);
         done();
       });
     });
+
+    it('automigrate reports errors for models not attached - promise variant', function(done) {
+      ds.automigrate(['m1', 'm2'])
+        .then(done)
```
bajtos: This will make the test pass even if there is no error reported. You need to provide a throwing `then` handler.

```js
.then(function() {
  done(new Error('automigrate() should have failed'));
})
```

test/memory.test.js
```diff
@@ -463,25 +463,88 @@ describe('Memory connector', function() {
...
+        });
+    });
+
+  });
+
+  it('automigrate does NOT report error when NO models are attached', function(done) {
+    var ds;
+    ds = new DataSource({
+        connector: 'memory'
+      });
```
bajtos: I am proposing to create a `describe` block and move the initialisation of an empty datasource to a `beforeEach` hook, in order to reduce the amount of duplicated code.

```js
describe('automigrate when NO models are attached', function() {\
  var ds;
  beforeEach(/*setup ds */);
  it('does NOT report error', /*...*/);
  it('does NOT report error - promise variant');
});
```

Thoughts?

bajtos:
@PradnyaBaviskar nice! I have reviewed the patch. The implementation is good in principle, I'd like you to simplify it a bit - see my comments above.

bajtos:
@slnode test please

PradnyaBaviskar:
Addressed the comments above.

@bajtos , Please review.

bajtos:
Landed, thank you!

Issue #631, opened 2015-06-16
loopback.getCurrentContext() is null with 2.29+ in Operational Hooks


kblcuk:
I have a fairly basic operation hook that stopped working with newer datasource-juggler, because loopback.getCurrentContext() always returns null.
```
  Model.observe('before save', function addFields(ctx, next) {
  if (ctx.isNewInstance) {
    var accessCtx = loopback.getCurrentContext(); // Always null with datasource-juggler 2.29+ 
    if (accessCtx && accessCtx.active && accessCtx.active.accessToken) {
      ctx.instance.authorId = accessCtx.active.accessToken.userId;
    }
    next();
  });
```

Is this supposed to be that way? If yes, then what would be the recommended way to get authenticated user that created model?

Thanks!

Might be relevant: https://github.com/strongloop/loopback/issues/1442

mrfelton:
I didn't have a problem with 2.29, however in 2.30 I have this problem - null returned from getCurrentContext when used in operation hooks

mrfelton:
Above pull request resolves for me which pins async to version 1.0.0 as per the suggestion in #1442

raymondfeng:
Can you share your test case? I cannot reproduce it with the following one:

/common/models/notes.js
```js
var loopback = require('loopback');
module.exports = function(Note) {
  Note.observe('before save', function(ctx, next) {
    var cc = loopback.getCurrentContext();
    console.log('context', cc);
    next();
  });
};
```

mrfelton:
Looks like this made it into the latest release already. https://github.com/strongloop/loopback-datasource-juggler/commit/8302b2480ddc24c30b650df50424b53699b635c8

barocsi:
Unfortunately this is not that trivial, even if you pin down the version, the installation order of deps will bring back the issue. I mean:

#### Test Case 1
1. I have forked installed the loopback-sandbock project
2. added some async test
3. added "async": "1.1.0" to the dependencies
4. node_modules dir was deleted/empty
3. ran npm install --- so async was installed altogether with other modules
4. After async.series callback the process.context was broken.

#### Test Case 2
1. I have forked installed the loopback-sandbock project
2. added some async test
3. added "async": "1.0.0" to the dependencies
4. node_modules dir was deleted/empty
3. ran npm install --- so async was installed altogether with other modules
4. After async.series callback the process.context was *NOT* broken.

#### Test Case 3
1. I have forked installed the loopback-sandbock project
[...]

PR #632, opened 2015-06-16
Pin async to version 1.0.0

mrfelton:
Fix for #631 by pinning async module to version 1.0.0


slnode:

bajtos:
@mrfelton @raymondfeng is this still relevant, considering https://github.com/strongloop/loopback-datasource-juggler/issues/631#issuecomment-117812328?

PR #634, opened 2015-06-16
Fixes #633 - Pass through case insenstive search params.

### LINKED ISSUE #633: Pass through ilike and nilike parameters so they can be handled by connectors
Every application needs to be able to search without case sensitivity. Other than regular expressions (not always supported) there isn't a good way to do this. See PR for fixing this in Postgres connector (https://github.com/strongloop/loopback-connector-postgresql/pull/93). This patch needs to be applied to the juggler as well.

jasonaden:
See https://github.com/strongloop/loopback-connector-postgresql/pull/93 for corresponding PR in Postgres.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -898,7 +898,9 @@ var operators = {
   nin: 'NOT IN',
   neq: '!=',
   like: 'LIKE',
-  nlike: 'NOT LIKE'
+  nlike: 'NOT LIKE',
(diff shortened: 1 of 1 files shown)


slnode:

bajtos:
@slnode ok to test

superkhau:
Closing in favour of https://github.com/strongloop/loopback-datasource-juggler/pull/1091

PR #636, opened 2015-06-22
add 'connecting' event

BerkeleyTrue:
useful for debugging slow dataBase connections


BerkeleyTrue:
Actually this is useless.

PR #637, opened 2015-06-22
Fix promise chaining in case of error

rus0000:


slnode:

PR #638, opened 2015-06-23
Promisify 'autoupdate'

PradnyaBaviskar:

@bajtos,
Please review.

### DIFF
--- a/test/memory.test.js
+++ b/test/memory.test.js
@@ -548,6 +548,95 @@ describe('Memory connector', function() {
     });
   });
 
+  describe('With mocked autoupdate', function() {
+    var ds, model;
+    beforeEach(function() {
+      ds = new DataSource({
+        connector: 'memory'
+      });
+
+      ds.connector.autoupdate = function(models, cb) {
+        process.nextTick(cb);
+      };
+
+      model = ds.createModel('m1', {
+        name: String
+      });
+
+      ds.automigrate();
+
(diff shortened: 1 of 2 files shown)


test/memory.test.js
```diff
@@ -548,6 +548,42 @@ describe('Memory connector', function() {
...
+      ds.autoupdate()
+        .then(function(result) {
+            done();
+          })
+        .catch(function(err){
+          done(err);
+        });
+    });
+  });
+
```
PradnyaBaviskar: I was planning to add test cases for `autoupdate`, using -

```
ds = new DataSource({
  connector: 'memory'
});
```
But then realized that memory connector does not have `autoupdate` method. Hence skipped both the test cases above. How should I be verifying `promise` support for `autoupdate`?

PradnyaBaviskar: Oh got it, in `beforeEach`, I can add a mock `autoupdate`. Something like this -

```
ds.autoupdate = function(models, cb) {
  process.nextTick(cb);
};
```

Will update my test suite, for this.

bajtos: :+1:

PradnyaBaviskar:
Done with adding test cases for autoupdate.
Can you please review?

bajtos:
@slnode test please

bajtos:
Landed, thank you!

PR #642, opened 2015-06-23
Fix updateOrCreate transaction propagation

rus0000:


slnode:

lib/dao.js
```diff
@@ -505,7 +505,11 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
         }
       });
     } else {
-      Model.findOne({ where: ctx.query.where }, { notify: false }, function (err, inst) {
+      var opts = {notify: false};
+      if (ctx.options && ctx.options.transaction) {
+        opts.transaction = ctx.options.transaction;
```
raymondfeng: Should we pass other options too?

rus0000:
I checked findOne call chain, and only `transaction` and `notify` options are used. So, currently there is no sane need for other options, but it may change in future. Wouldn't we have side effects of passing other `options` to this auxiliary query?

PR #643, opened 2015-06-23
Includes with transaction support

rus0000:
Passing `options` to all find calls


slnode:

PR #647, opened 2015-06-25
Promisify all 'discover' methods

PradnyaBaviskar:

Promisify following methods -
  1. discoverModelDefinitions
  2. discoverModelProperties
  3. discoverPrimaryKeys
  4. discoverForeignKeys
  5. discoverExportedForeignKeys
  6. discoverSchema
  7. discoverSchemas

@bajtos - can you please review?

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -886,11 +886,20 @@ DataSource.prototype.autoupdate = function (models, cb) {
  */
 DataSource.prototype.discoverModelDefinitions = function (options, cb) {
   this.freeze();
+
+  if (cb === undefined && typeof options === 'function') {
+    cb = options;
+    options = {};
+  }
+  options = options || {};
+  cb = cb || utils.createPromiseCallback();
+
   if (this.connector.discoverModelDefinitions) {
     this.connector.discoverModelDefinitions(options, cb);
   } else if (cb) {
     process.nextTick(cb);
   }
+  return cb.promise;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


bajtos:
Nice, LGTM.

@raymondfeng You are more familiar with autodiscovery, could you please take a look too? Especially on the unit tests.

raymondfeng:
The `options` argument is optional. Most of the connectors have the following code:
```
if (cb === undefined && typeof options === 'function') {
      cb = options;
      options = {};
    }
    options = options || {};
```

We need to normalize the args before your changes. 

@PradnyaBaviskar Can you fix that?

PradnyaBaviskar:
@raymondfeng,

> We need to normalize the args before your changes.

Just so that I understand this better, before we do `cb = cb || utils.createPromiseCallback();`, in each `discover*()` method, I should add

```
if (cb === undefined && typeof options === 'function') {
      cb = options;
      options = {};
    }
    options = options || {};
```

Is this correct understanding?

raymondfeng:
@PradnyaBaviskar Yes.

PradnyaBaviskar:
Done with normalizing the args.

raymondfeng:
@PradnyaBaviskar Thanks. Can you also add a test to cover the case where `options` is not present?

PradnyaBaviskar:
Done - added tests to cover the case of missing `options` parameter.

raymondfeng:
ok to test

PR #648, opened 2015-06-30
Fix #623 - use actual id order

### LINKED ISSUE #623: Include includePolymorphic is broken

fabien:
In order to fix @623 completely, the array of a referencesMany ids should be used for the order of returned items.

For this, `sortObjectsByIds` is appropriate. Somehow the Juggler tests did not uncover this pending issue, despite checking the order in reverse as well (see: https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/relations.test.js#L4895)

I only was able to find this after I finally updated my app (running with MongoDB) and running test there.

/cc @ningsuhen @bajtos

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -407,6 +407,7 @@ Inclusion.include = function (objects, include, options, cb) {
      * @param callback
      */
     function includeReferencesMany(callback) {
+      var modelToIdName = idName(relation.modelTo);
       var allTargetIds = [];
       //Map for Indexing objects by their id for faster retrieval
       var targetObjsMap = {};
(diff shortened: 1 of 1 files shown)


raymondfeng:
LGTM

bajtos:
> somehow the Juggler tests did not uncover this pending issue

Could you please add a new test to get this issue covered for the future?

fabien:
@bajtos I was wondering about this - I already used an array of ids with a non-insertion order (reversed) and it worked before. I don't really know how to write another test for this - it appears that I was only able to uncover the issue while running MongoDB (and a model with a default sort order/scope). Instead of the expected array order, items were now ordered by `name`, which is the global order for that model.

I'd suggest to go ahead with this right now, unless you have a way to test this differently.

bajtos:
I guess it's ok to go ahead then. @raymondfeng has already approved the patch, feel free to merge it yourself.

Issue #649, opened 2015-06-30
Hook 'persist' not using context.data


fabien:
It appears that context.data is not always used when actually persisting the data through the connector, for example: https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L2394

I believe the best way would have been: `inst.constructor._forDB(context.data)` - in other places it will probably work, however, explicitly using `context.data` everywhere ensures that replacing the data within a `persist` hook works as expected.

/cc @bajtos @raymondfeng

bajtos:
@PradnyaBaviskar could you please take a look? IIUC, @fabien says that `prototype.updateAttributes` is not using `ctx.data` provided by the "persist" hook, which looks like an unintended bug to me.

I think we should make one more change and set `ctx.data` initially to `typedData`, as that's closer to what is being persisted in the database. The difference will be visible for nested model instances, e.g. if you have

```
var p = new Product();
p.updateAttributes({ categories: [new Category("Cars"), new Category("Vehicles")] });
```

then `data` contains Category instances while `typedData` contains plain objects instead (if I am reading the code correctly).

fabien:
Actually, I would find it much clearer if `ctx.data` would be used whenever `persist` is called, instead of relying on `data` still pointing to the same object as `ctx.data`. This might not always be the case.

bajtos:
> Actually, I would find it much clearer if ctx.data would be used whenever persist is called, instead of relying on data still pointing to the same object as ctx.data. This might not always be the case.

+1 Let's move this discussion to #650 please

0candy:
Closing this issue as PR was merged.

PR #650, opened 2015-07-01
Fix 'persist' hook for updateAttributes()

PradnyaBaviskar:
Connect https://github.com/strongloop/loopback-datasource-juggler/issues/649

@bajtos, can you review?

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2358,14 +2358,16 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
             }
           }
 
+          context.data = typedData;
+
           function updateAttributesCallback(err) {
-            var context = {
+            var ctx = {
               Model: Model,
-              data: data,
+              data: context.data,
               hookState: hookState,
               options: options
             };
-            Model.notifyObserversOf('loaded', context, function(err) {
+            Model.notifyObserversOf('loaded', ctx, function(err) {
               if (!err) inst.__persisted = true;
 
               // By default, the instance passed to updateAttributes callback is NOT updated
@@ -2373,7 +2375,7 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
               // backwards compatibility, we introduced a new setting updateOnLoad,
               // which if set, will apply these changes to the model instance too.
               if(Model.settings.updateOnLoad) {
-                inst.setAttributes(context.data);
+                inst.setAttributes(ctx.data);
               }
               done.call(inst, function () {
                 saveDone.call(inst, function () {
@@ -2395,21 +2397,21 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
             });
           }
 
-          context = {
+          var ctx = {
             Model: Model,
             where: byIdQuery(Model, getIdValue(Model, inst)).where,
-            data: data,
+            data: context.data,
             currentInstance: inst,
             hookState: hookState,
             options: options
           };
-          Model.notifyObserversOf('persist', context, function(err) {
+          Model.notifyObserversOf('persist', ctx, function(err) {
             if (connector.updateAttributes.length === 5) {
               connector.updateAttributes(model, getIdValue(inst.constructor, inst),
-                inst.constructor._forDB(typedData), options, updateAttributesCallback);
+                inst.constructor._forDB(context.data), options, updateAttributesCallback);
             } else {
               connector.updateAttributes(model, getIdValue(inst.constructor, inst),
-                inst.constructor._forDB(typedData), updateAttributesCallback);
+                inst.constructor._forDB(context.data), updateAttributesCallback);
             }
           });
         }, data, cb);
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -1154,6 +1154,58 @@ module.exports = function(dataSource, should) {
         });
       });
 
+      it('applies updates from `persist` hook - for nested model instance', function(done) {
+        var Address = dataSource.createModel('Address', {
+          id: { type: String, id: true, default: 1 },
+          city: { type: String, required: true },
+          country: { type: String, required: true }
+        });
+
+        var User = dataSource.createModel('User', {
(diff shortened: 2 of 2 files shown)


test/persistence-hooks.suite.js
```diff
@@ -1150,6 +1150,41 @@ module.exports = function(dataSource, should) {
...
+          User.observe('persist', pushContextAndNext(function(ctx){
+            ctx.data.extra = 'hook data';
+          }));
+
+          // By default, the instance passed to updateAttributes callback is NOT updated 
+          // with the changes made through persist/loaded hooks. To preserve 
+          // backwards compatibility, we introduced a new setting updateOnLoad, 
+          // which if set, will apply these changes to the model instance too.
+          User.settings.updateOnLoad = true;
+          existingUser.updateAttributes({ address: new Address({city: 'Springfield', country: 'USA'}) }, function(err, inst) {
```
bajtos: Please wrap at 80 characters.

test/persistence-hooks.suite.js
```diff
@@ -1150,6 +1150,41 @@ module.exports = function(dataSource, should) {
...
+          address: {type: Address, required: false}
+        });
+
+        User.create({name: 'Joe'}, function(err, instance) {
+          if (err) return done(err);
+
+          var existingUser = instance;
+
+          User.observe('persist', pushContextAndNext(function(ctx){
+            ctx.data.extra = 'hook data';
```
bajtos: Please add an assertion that `ctx.data.address` is set and a plain object (not an instance of Address model).

lib/dao.js
```diff
@@ -2350,6 +2350,7 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
               typedData[key] = typedData[key].toObject();
             }
           }
+          data = typedData;
```
bajtos: This is IMO rather evil. Please rework the code below to use `typedData` and `context.data` as we discussed earlier and keep the `data` variable unmodified.

PradnyaBaviskar:
Addressed the comments above. Please review.

test/persistence-hooks.suite.js
```diff
@@ -1150,6 +1150,46 @@ module.exports = function(dataSource, should) {
...
+        });
+
+        User.create({name: 'Joe'}, function(err, instance) {
+          if (err) return done(err);
+
+          var existingUser = instance;
+
+          User.observe('persist', pushContextAndNext(function(ctx){
+            should.exist(ctx.data.address)
+            ctx.data.address.should.be.type('object');
```
bajtos: A model instances passes this check too. I think you need to check `ctx.data.address.should.not.be.instanceOf(Address)`.

lib/dao.js
```diff
@@ -2388,15 +2390,15 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
...
+            data: context.data,
             currentInstance: inst,
             hookState: hookState,
             options: options
           };
-          Model.notifyObserversOf('persist', context, function(err) {
+          Model.notifyObserversOf('persist', ctx, function(err) {
             if (connector.updateAttributes.length === 5) {
               connector.updateAttributes(model, getIdValue(inst.constructor, inst),
                 inst.constructor._forDB(typedData), options, updateAttributesCallback);
```
bajtos: I think you need to change `typedData` to `ctx.data` here.

A test case:

```js
User.observe('persist', pushContextAndNext(function(ctx){
  ctx.data = { extra: 'hook data' };
  next();
});

existingUser.updateAttributes(
  { address: new Address({city: 'Springfield', country: 'USA'}) }, 
  // verify that user.extra is set, user.address is not set
```

bajtos: The same change should be applied on L2404 (L2407) below.

bajtos: > verify that user.extra is set, user.address is not set

Check the data in the database too. If I am reading the code correctly, it will apply hook updates to the instance returned to callback, but it won't apply them in the database.

PradnyaBaviskar:
Addressed your comments above.

raymondfeng:
ok to test

test/persistence-hooks.suite.js
```diff
@@ -1150,6 +1150,59 @@ module.exports = function(dataSource, should) {
...
+                if (err) return done(err);
+                dbInstance.toObject(true).should.eql({
+                  id: existingUser.id,
+                  name: existingUser.name,
+                  address: {id: '1', city: 'Springfield', country: 'USA'},
+                  extra: 'hook data'
+                })
+              });
+
+              done();
```
raymondfeng: `done()` should be moved into the callback closure of findById().

test/persistence-hooks.suite.js
```diff
@@ -1150,6 +1150,59 @@ module.exports = function(dataSource, should) {
...
+              inst.should.have.property('extra', 'hook data');
+
+              User.findById(existingUser.id, function(err, dbInstance) {
+                if (err) return done(err);
+                dbInstance.toObject(true).should.eql({
+                  id: existingUser.id,
+                  name: existingUser.name,
+                  address: {id: '1', city: 'Springfield', country: 'USA'},
+                  extra: 'hook data'
+                })
```
raymondfeng: Missing ;

PradnyaBaviskar:
@raymondfeng, addressed your comments above.

test/persistence-hooks.suite.js
```diff
@@ -1154,6 +1154,58 @@ module.exports = function(dataSource, should) {
...
 
+      it('applies updates from `persist` hook - for nested model instance', function(done) {
+        var Address = dataSource.createModel('Address', {
+          id: { type: String, id: true, default: 1 },
+          city: { type: String, required: true },
+          country: { type: String, required: true }
+        });
+
+        var User = dataSource.createModel('User', {
+          id: { type: String, id: true, default: uid },
```
raymondfeng: Do you mean `default: uid()` here?

PradnyaBaviskar: Oh, yes!

I've corrected it to be uid() now.

PR #652, opened 2015-07-01
Fix coercion from ObjectID to String

raymondfeng:
/to @ritch 

This patch is result of https://groups.google.com/forum/#!topic/loopbackjs/zUQY1aQ5oGA

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -294,9 +294,9 @@ DataAccessObject.create = function (data, options, cb) {
           };
           Model.notifyObserversOf('loaded', context, function(err) {
 
-            // By default, the instance passed to create callback is NOT updated 
-            // with the changes made through persist/loaded hooks. To preserve 
-            // backwards compatibility, we introduced a new setting updateOnLoad, 
+            // By default, the instance passed to create callback is NOT updated
+            // with the changes made through persist/loaded hooks. To preserve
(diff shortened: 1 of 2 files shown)


pulkitsinghal:
@raymondfeng - wanted to let you know that it worked

I added `"loopback-datasource-juggler": "git://github.com/strongloop/loopback-datasource-juggler.git#eb6fa41"` to `package.json` and nuked the db to start over and my boot script finished without any issues.

lib/dao.js
```diff
@@ -1135,12 +1135,7 @@ DataAccessObject._coerce = function (where) {
     }
     // Check there is an operator
     var operator = null;
-    if ('object' === typeof val) {
-      if (Object.keys(val).length !== 1) {
-        // Skip if there are not only one properties
-        // as the assumption for operators is not true here
-        continue;
-      }
+    if (val.constructor === Object) {
```
ritch: This seems a bit more strict than we need... Isn't `typeof val === 'object' && !Array.isArray(val)` what we are looking for?

ritch:
LGTM

lib/dao.js
raymondfeng: One case that we want to skip if mongodb.ObjectID.

PR #660, opened 2015-07-10
Make sure base property definitions are cloned

raymondfeng:
/to @fabien 
/cc @ritch 

Sub models sometimes need to customize the properties from the base model.
This change allows each sub model has its own copy of the base property
definition to avoid potential conflicts across multiple sub models of the
same base.


fabien:
@raymondfeng I can confirm that this fixes my issue, thanks!

Issue #552, opened 2015-03-31
Support regular expressions in SQL connectors.


carragom:
Hello,

I would like to start a discussion about adding a new filter for regular expressions in where clauses. AFAIK the only supported database without native regex support is MSSQL. All others have the feature. Here are some links:

https://dev.mysql.com/doc/refman/5.5/en/regexp.html
http://www.postgresql.org/docs/9.0/static/functions-matching.html#FUNCTIONS-POSIX-REGEXP
http://docs.oracle.com/cd/B19306_01/server.102/b14200/conditions007.htm

I have a working patch for the postgresql connector. I would like to submit it for review but first some questions:

* What would be an acceptable keyword for this ?. rlike, regex ?.
* What would be the correct way to handle this in a connector where this functionality is not available?. Throw and error?.
* Should it be case insensitive by default?.

Thanks.

superkhau:
Bumping up to 3 points because its taking longer than the expected 2 points.

/cc @chandadharap

superkhau:
Bumping up to a 5 as it's taken 2 days now.

//cc @chandadharap

PR #665, opened 2015-07-16
Add support for RegExp operator

superkhau:
Connect to #552

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -350,9 +350,8 @@ Memory.prototype.all = function all(model, filter, options, callback) {
     }
 
     // do we need some filtration?
-    if (filter.where) {
-      nodes = nodes ? nodes.filter(applyFilter(filter)) : nodes;
-    }
+    if (filter.where && nodes)
+      nodes = nodes.filter(applyFilter(filter));
 
     // field selection
     if (filter.fields) {
@@ -401,7 +400,7 @@ function applyFilter(filter) {
   var keys = Object.keys(where);
   return function (obj) {
     var pass = true;
-    keys.forEach(function (key) {
+    keys.forEach(function(key) {
       if(key === 'and' || key === 'or') {
         if(Array.isArray(where[key])) {
           if(key === 'and') {
@@ -461,6 +460,10 @@ function applyFilter(filter) {
     if (typeof value === 'string' && (example instanceof RegExp)) {
       return value.match(example);
     }
+
+    if (example.regexp)
+      return value.match(example.regexp);
+
     if (example === undefined) {
       return undefined;
     }
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -939,7 +939,8 @@ var operators = {
   nin: 'NOT IN',
   neq: '!=',
   like: 'LIKE',
-  nlike: 'NOT LIKE'
+  nlike: 'NOT LIKE',
+  regexp: 'REGEXP'
 };
 
 /*
@@ -1169,6 +1170,13 @@ DataAccessObject._coerce = function (where) {
                 throw err;
               }
               break;
+            case 'regexp':
+              val = utils.toRegExp(val);
+              if (val instanceof Error) {
+                result.statusCode = 400;
+                throw err;
+              }
+              break;
           }
           break;
         }
@@ -1183,8 +1191,9 @@ DataAccessObject._coerce = function (where) {
       }
     } else {
       if (val != null) {
-        if (!((operator === 'like' || operator === 'nlike') &&
-          val instanceof RegExp)) {
+        if (operator === 'regexp' && val instanceof RegExp) {
+          // do not coerce regex literals/objects
+        } else if (!((operator === 'like' || operator === 'nlike') && val instanceof RegExp)) {
           val = DataType(val);
         }
       }
@@ -1237,6 +1246,7 @@ DataAccessObject._coerce = function (where) {
  * - neq: !=
  * - like: LIKE
  * - nlike: NOT LIKE
+ * - regexp: REGEXP
  *
  * You can also use `and` and `or` operations.  See [Querying models](http://docs.strongloop.com/display/DOC/Querying+models) for more information.
  * @property {String|Object|Array} include Allows you to load relations of several objects and optimize numbers of requests.
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -12,6 +12,9 @@ exports.mergeQuery = mergeQuery;
 exports.mergeIncludes = mergeIncludes;
 exports.createPromiseCallback = createPromiseCallback;
 exports.uniq = uniq;
+exports.toRegExp = toRegExp;
+exports.hasRegExpFlags = hasRegExpFlags;
+exports.getRegExpExpression = getRegExpExpression;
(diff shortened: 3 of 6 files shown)


lib/dao.js
```diff
@@ -1169,6 +1170,12 @@ DataAccessObject._coerce = function (where) {
                 throw err;
               }
               break;
+            case 'regexp':
+              if (typeof val !== 'string') {
```
raymondfeng: The val can be a RegExp too.

lib/dao.js
```diff
@@ -1169,6 +1170,12 @@ DataAccessObject._coerce = function (where) {
                 throw err;
               }
               break;
+            case 'regexp':
+              if (typeof val !== 'string' || val instanceof RegExp) {
```
superkhau: @raymondfeng Like this?

raymondfeng: It should be `if (!(typeof val === 'string' || (val instanceof RegExp))) {`.. We complains if the value is not string or RegExp.

lib/dao.js
```diff
@@ -1169,6 +1170,12 @@ DataAccessObject._coerce = function (where) {
                 throw err;
               }
               break;
+            case 'regexp':
+              if (!(typeof val === 'string' || val instanceof RegExp)) {
```
superkhau: @raymondfeng Fixed.

lib/dao.js
```diff
@@ -1169,6 +1170,12 @@ DataAccessObject._coerce = function (where) {
...
+                err = new Error(util.format('The %s property has invalid clause %j', p, where[p]));
+                err.statusCode = 400;
+                throw err;
+              }
```
raymondfeng: As we discussed, the operand of regexp should be normalized to RegExp here.

lib/connectors/memory.js
```diff
@@ -461,6 +460,13 @@ function applyFilter(filter) {
     if (typeof value === 'string' && (example instanceof RegExp)) {
       return value.match(example);
     }
+
+    if (example.regexp && typeof value !== 'undefined') {
```
raymondfeng: This logic should be refactored to dao.js to normalize the operand of `regexp` into an instance of `RegExp`

superkhau:
@raymondfeng Added utility function that normalizes regexes. Please review again.

lib/dao.js
```diff
@@ -1169,6 +1170,14 @@ DataAccessObject._coerce = function (where) {
                 throw err;
               }
               break;
+            case 'regexp':
+              try {
+                val = utils.toRegExp(val);
+              } catch (err) {
+                err.statusCode = 400;
+                throw err;
```
raymondfeng: Let's try to avoid try/catch which will prevents V8 from optimizing the function. Can we return the Error from `utils.toRegExp(val);`?

raymondfeng: Throw err is not good idea.

superkhau: Yes we can return err and have the caller check return values for instance of Error before proceeding instead if you prefer.

raymondfeng: +1

superkhau:
@raymondfeng Review again please.

superkhau:
@raymondfeng Please review again.

lib/connectors/memory.js
```diff
@@ -461,6 +460,15 @@ function applyFilter(filter) {
     if (typeof value === 'string' && (example instanceof RegExp)) {
       return value.match(example);
     }
+
+    if (example.regexp) {
+      var normalizedRegExp = utils.toRegExp(example.regexp);
+      if (normalizedRegExp instanceof Error)
```
raymondfeng: This should not happen in the connector. If the regexp operand is invalid, the juggler should have reported the error.

lib/dao.js
```diff
@@ -1169,6 +1170,13 @@ DataAccessObject._coerce = function (where) {
                 throw err;
               }
               break;
+            case 'regexp':
+              val = utils.toRegExp(val);
+              if (val instanceof Error) {
+                result.statusCode = 400;
+                throw err;
```
raymondfeng: We should open an issue to fix this error handling style.

raymondfeng:
Beyond the minor comments, LGTM.

lib/dao.js
superkhau: https://github.com/strongloop-internal/scrum-loopback/issues/397

lib/connectors/memory.js
superkhau: I'm only using toRegExp here to normalize the regex because somewhere above the chain my regex is converted to a string.

raymondfeng: That's not good. You need to dig into the root cause. The idea is that `dao` will normalize the operand of `regexp` to an instance of `RegExp` so that connectors don't have to deal with conversions.

superkhau: Makes sense. I will dig into the root cause and make sure dao provides the correct data type to the connector.

lib/dao.js
```diff
@@ -1183,8 +1191,9 @@ DataAccessObject._coerce = function (where) {
       }
     } else {
       if (val != null) {
-        if (!((operator === 'like' || operator === 'nlike') &&
-          val instanceof RegExp)) {
+        if (val instanceof RegExp) {
+          // do not coerce regex literals/objects
+        } else if (!((operator === 'like' || operator === 'nlike') && val instanceof RegExp)) {
```
superkhau: @raymondfeng This is the culprit. How do you recommend I clean this up without breaking old code? My implementation with the empty if check for RegExp works, but I'm not sure if this is the best solution. The other possibility is to fix `!((operator === 'like' || operator === 'nlike') && val instanceof RegExp))`.

superkhau: Does `like` and `nlike` use regex strings? Are we trying to say:

- if operator !== like
- or operator !== nlike
- and val is a regex object
- then coerce the value based on its data type in the model definition?

raymondfeng: Yes, add check to `regexp` too.

superkhau: What do you mean add check to `regexp` too?

superkhau: I see 3 cases:

1. if operator === 'regexp' and val is a regex literal/object, do nothing because we need the object to remain intact
2. if operator === like/nlike and val is regex object, convert to string because like/nlike uses string values
3. do nothing for any other data types

Is this correct?

raymondfeng: Yes.

lib/dao.js
```diff
@@ -1183,8 +1191,9 @@ DataAccessObject._coerce = function (where) {
       }
     } else {
       if (val != null) {
-        if (!((operator === 'like' || operator === 'nlike') &&
-          val instanceof RegExp)) {
+        if (operator === 'regexp' && val instanceof RegExp) {
+          // do not coerce regex literals/objects
```
superkhau: @raymondfeng Is this minor change good enough? or should I change the entire block?

raymondfeng: Sure. With it in place, you don't have to coerce regexp in memory connector, right?

superkhau: Correct.

raymondfeng:
LGTM

PR #666, opened 2015-07-20
Filter attributes when strict: true

fabien:
This fixes https://github.com/strongloop/loopback/issues/742 in the following way:

- with strict: true, only known properties should be written to the dataSource
- from my findings, the data never made it to the dataSource, even before this fix (unlike what the issue above states)
- what **did** happen though, is that the in-memory representation was altered and returned, which was unaware of what was filtered or not (eg. actually written to the dataSource)

This distinction was clear from the test I added. Even before this fix, the last `should.not.exist(p.foo)` would pass correctly, however, the former `should.not.exist(p.foo)` call would not. With this fix, both pass.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2297,6 +2297,7 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
   assert(typeof connector.updateAttributes === 'function',
     'updateAttributes() must be implemented by the connector');
 
+  var strict = this.__strict;
   var model = Model.modelName;
   var hookState = {};
 
@@ -2333,6 +2334,12 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
   Model.notifyObserversOf('before save', context, function(err, ctx) {
     if (err) return cb(err);
     data = ctx.data;
+
+    if (strict) {
+      var keys = Object.keys(Model.definition.properties);
+      data = utils.selectFields(keys)(data);
+      data = removeUndefined(data);
+    }
 
     // update instance's properties
     inst.setAttributes(data);
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2336,9 +2336,21 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
     data = ctx.data;
 
     if (strict) {
-      var keys = Object.keys(Model.definition.properties);
-      data = utils.selectFields(keys)(data);
-      data = removeUndefined(data);
+      var props = Model.definition.properties;
+      var keys = Object.keys(data);
+      var result = {};
+      for (var i = 0; i < keys.length; i++) {
+        key = keys[i];
+        if (props[key]) {
+          result[key] = data[key];
+        } else if (strict === 'throw') {
+          cb(new Error('Unknown property: ' + key));
+          return;
+        } else if (strict === 'validate') {
+          inst.__unknownProperties.push(key);
+        } 
+      }
+      data = removeUndefined(result);
     }
 
     // update instance's properties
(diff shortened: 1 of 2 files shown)


lib/dao.js
```diff
@@ -2334,6 +2335,12 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
     if (err) return cb(err);
     data = ctx.data;
 
+    if (strict) {
+      var keys = Object.keys(Model.definition.properties);
```
raymondfeng: `strict` also supports other setting such as `throw` or `validate`. Do we want to call back with such errors for those cases?

fabien: We might ... what do you think @raymondfeng ? I haven't used `throw` or `validate` yet, can you explain? I don't believe these other settings are documented right now.

raymondfeng: It's documented at http://docs.strongloop.com/display/LB/Model+definition+JSON+file#ModeldefinitionJSONfile-Top-levelproperties. 

- `throw` Throw an error if unknown properties are present
- `validate` Report as an validation error (I prefer this)

See https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model.js#L227-L230

fabien:
I see - so effectively, it can be implemented like the code you linked to? Do you want me to add this?

raymondfeng:
Yes. Please.

fabien:
@raymondfeng I added both options. However, I chose not to throw but return the error to the callback instead.

lib/dao.js
```diff
@@ -2334,6 +2335,24 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
...
+        } else if (strict === 'throw') {
+          cb(new Error('Unknown property: ' + key));
+          return;
+        } else if (strict === 'validate') {
+          inst.__unknownProperties.push(key);
+        } 
+      }
+      data = removeUndefined(result);
+    }
+
```
raymondfeng: Is it going to break https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L741? MongoDB allows extended `$` operators for the update.

fabien: Hmmm... that's tricky... the only way I can think of right now would be to pass the raw data as options to the connector as well, and then let the connector handle it (like `parseUpdateData`).

I must admit, having this raw power is already a lot less safe - this is why it needs to be explicitly set `allowExtendedOperators: true`. So I guess any form of `strict` being enabled wouldn't probably help much anyway.

What would you suggest?

raymondfeng: The good news is that models attached to mongodb connector by default have `strict` set to false. For now, maybe we can test `allowExtendedOperators === true` to decide if we should relax the check.

fabien:
@raymondfeng I added this check to relax strictness. 

However, I would have expected `allowExtendedOperators` to be a setting on `Model.definition.settings` instead (or even better, alongside) `connector.settings`. That way you can be selective about which models are allowed to use extended operators.

Perhaps open a ticket on lb-connector-mongodb for this?

raymondfeng:
Yes, it should be very easy to honor `Model.definition.settings.allowExtendedOperators`.

cookejames:
I've spotted a small problem with this when using the mongo connecter. The default mongo behaviour is to define an id on the model with the name 'id' however when this is saved it is turned into '_id' in the mongo database. 

Using the angular sdk I typically send back the whole model as I received it when getting an update which results in the database having fields of both 'id' and '_id'.

raymondfeng:
Is it a regression introduced by this PR?

cookejames:
No, previously it would allow 'id' plus everything else now it just allows 'id'. I imagine this doesn't happen with a mysql back end just mongo due to the way it renames id

PR #671, opened 2015-07-30
Fix regexp error for the memory connector

superkhau:
- Check for undefined/null values when accessing `regexp` key

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -461,14 +461,14 @@ function applyFilter(filter) {
       return value.match(example);
     }
 
-    if (example.regexp)
-      return value.match(example.regexp);
-
     if (example === undefined) {
       return undefined;
     }
 
     if (typeof example === 'object' && example !== null) {
+      if (example.regexp)
+        return value.match(example.regexp);
+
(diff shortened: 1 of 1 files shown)


superkhau:
@raymondfeng Please review.

raymondfeng:
LGTM

lib/connectors/memory.js
```diff
@@ -461,14 +461,14 @@ function applyFilter(filter) {
...
 
-    if (example.regexp)
-      return value.match(example.regexp);
-
     if (example === undefined) {
       return undefined;
     }
 
     if (typeof example === 'object' && example !== null) {
+      if (example.regexp)
```
sam-github: Drive by commenting here... but maybe line 464 above would be clearer as `if(example == undefined || typeof example !== 'object')`, because if example is `null`, then `undefined` will be returned anyhow when the function falls off the end, but right now you have to scroll down 100 lines to see that. And if it was like that, the entire conditional at line 468 could be removed and the body outdented.

superkhau: There can definitely be improvements to the conditionals here. ;)

PR #679, opened 2015-08-04
Indicate result of destroyById/protototype.destroy

bajtos:
Return `info.count` to the callback to indicate whether the model instance was deleted or not. When `options.strict` is true, return 404 error when the model instance was not found.

See #521 for the initial solution.

/to @raymondfeng @fabien please review

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -265,8 +265,9 @@ Memory.prototype.find = function find(model, id, options, callback) {
 };
 
 Memory.prototype.destroy = function destroy(model, id, options, callback) {
+  var exists = this.collection(model)[id];
   delete this.collection(model)[id];
-  this.saveToFile(null, callback);
+  this.saveToFile({ count: exists ? 1 : 0 }, callback);
 };
 
 Memory.prototype.fromDb = function (model, data) {
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1697,11 +1697,18 @@ DataAccessObject.removeById = DataAccessObject.destroyById = DataAccessObject.de
 
   var Model = this;
 
-  this.remove(byIdQuery(this, id).where, options, function(err) {
-    if ('function' === typeof cb) {
-      cb(err);
+  this.remove(byIdQuery(this, id).where, options, function(err, info) {
+    if (err) return cb(err);
+    var deleted = info && info.count > 0;
+    if (Model.settings.strictDelete && !deleted) {
+      err = new Error('No instance with id ' + id + ' found for ' + Model.modelName);
+      err.code = 'NOT_FOUND';
+      err.statusCode = 404;
+      return cb(err);
     }
-    if(!err) Model.emit('deleted', id);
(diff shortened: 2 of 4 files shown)


raymondfeng:
1. What's the implication of REST mapping? At the moment, the `deleteById` reports http status code 204. Do we need to introduce a flag to keep it backward compatible?

2. What's the response from `deleteAll()`?

bajtos:
@raymondfeng thank you for thoughtful comments.

> What's the implication of REST mapping? At the moment, the deleteById reports http status code 204. Do we need to introduce a flag to keep it backward compatible?

There is already a new flag `options.strict` for that. 404 is returned only when this flag is set.

Note that the `options` object is not yet exposed via REST, thus REST clients cannot enable this behaviour yet. To fix this, we need to modify remoting metadata in the main loopback module.

> What's the response from deleteAll()?

The callback receives `(err, info)` from the underlying connector. `deleteAll` is not exposed via REST by default, therefore the HTTP response of `deleteAll` is not a part of loopback API contract IMO.

LGTY now? Is there anything else to address/discuss?

fabien:
> Note that the options object is not yet exposed via REST, thus REST clients cannot enable this behaviour yet. To fix this, we need to modify remoting metadata in the main loopback module.

@bajtos @raymondfeng if I gather correctly, the plan would be to expose the optional `options` of methods like `find` or `create` to the REST API as well? If so, then this would probably be a bad idea. I can think of a few ways of how this could be a security risk, even (flag to bypass validation etc.).

Instead I would propose to allow one to hard-code or otherwise explicitly set which option keys to accept, or to even declare this options object with fixed values. In other words, prevent them from being set through REST (or any other remoting adapter). For example, pseudo code:

```Widget.setOptionsFor('delete', { strict: true })```

bajtos:
> if I gather correctly, the plan would be to expose the optional options of methods like find or create to the REST API as well? If so, then this would probably be a bad idea. I can think of a few ways of how this could be a security risk, even (flag to bypass validation etc.).

+1 for NOT exposing `options` object as it is.

I am wondering if I should perhaps rework this PR and move per-method `options.strict` to a model-level setting `strictDelete`?

fabien:
@bajtos having a `strictDelete` model-level settings sounds great

raymondfeng:
> The callback receives (err, info) from the underlying connector. deleteAll is not exposed via REST by default, therefore the HTTP response of deleteAll is not a part of loopback API contract IMO.

I would like to have a way to expose the deleted `count` over REST. Maybe under a flag for LB 2.x.

bajtos:
> having a strictDelete model-level settings sounds great

Reworked, PTAL again.

> I would like to have a way to expose the deleted count over REST. Maybe under a flag for LB 2.x.

Sounds good to me. Such change must be implemented in loopback, not juggler, therefore it's out of scope of this patch.

 @fabien @raymondfeng  LGTY now?

raymondfeng:
LGTM

Issue #670, opened 2015-07-28
Relax the id (primary key) requirement for certain operations, such as find or updateAll


raymondfeng:
https://github.com/strongloop-internal/scrum-cs/issues/309

superkhau:
As discussed with @raymondfeng, the first task is to come up with a list of areas that need relaxing:

**Memory connector**
- https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/connectors/memory.js#L324

raymondfeng:
> https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/connectors/memory.js#L324

This is fine as we check idNames.length

> https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L2334-L2336

Is this link out of date?

> https://github.com/strongloop/loopback-connector/blob/master/lib/sql.js#L986-L991

This is fine as it checks idNames.length

> https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L603-L608

This is fine as it checks idNames.length

> https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L670-L675

This is fine as it checks idNames.length

> https://github.com/strongloop/loopback-connector-mssql/blob/master/lib/mssql.js#L349-L354

This is fine as it checks idNames.length

> https://github.com/strongloop/loopback-connector-oracle/blob/master/lib/oracle.js#L469-L474

This is fine as it checks idNames.length

PR #689, opened 2015-08-12
Relax id requirement for basic query operations

superkhau:
Connect to #670

### DIFF
--- a/lib/model-definition.js
+++ b/lib/model-definition.js
@@ -302,3 +302,25 @@ ModelDefinition.prototype.toJSON = function (forceRebuild) {
   this.json = json;
   return json;
 };
+
+/**
+ * Checks if the id property has been defined in the model definition. Returns
+ * true if the id is found in the model definition properties OR in at least one
+ * id object from the _id property.
+ *
+ * @returns {Boolean} True if id is defined, false otherwise
+ */
+ModelDefinition.prototype.hasIdDefined = function() {
+  var isDefined = false;
+
+  if (this.properties.id)
+    isDefined = true;
+
+  // check if id is defined by a relation
+  this.ids().forEach(function(item) {
+    if (item.hasOwnProperty('id'))
+      isDefined = true;
+  });
+
+  return isDefined;
+};
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -620,6 +620,172 @@ describe('basic-querying', function () {
   });
 });
 
+describe('queries', function() {
+  var Todo;
+
+  before(function setupDb(done) {
+    var db = getSchema();
+    Todo = db.define('Todo', {
+      id: false,
+      content: {type: 'string'}
+    }, {
+      idInjection: false
+    });
+    db.automigrate(done);
+  });
+  beforeEach(function resetFixtures(done) {
+    Todo.destroyAll(function() {
+      Todo.create([
+        {content: 'Buy eggs'},
+        {content: 'Buy milk'},
+        {content: 'Buy sausages'}
+      ], done);
+    });
+  });
+
+  context('that do not require an id', function() {
+    it('should work for create', function(done) {
+      Todo.create({content: 'Buy ham'}, function(err) {
+        should.not.exist(err);
+        done();
+      });
+    });
+
+    it('should work for updateOrCreate/upsert', function(done) {
+      var aliases = ['updateOrCreate', 'upsert'];
+      async.each(aliases, function(alias, cb) {
+        Todo[alias]({content: 'Buy ham'}, function(err) {
+          should.not.exist(err);
+          cb();
+        });
+      }, done);
+    });
+
+    it('should work for findOrCreate', function(done) {
+      Todo.findOrCreate({content: 'Buy ham'}, function(err) {
+        should.not.exist(err);
+        done();
+      });
+    });
+
+    it('should work for exists', function(done) {
+      Todo.exists({content: 'Buy ham'}, function(err) {
+        should.not.exist(err);
+        done();
+      });
+    });
+
+    it('should work for find', function(done) {
+      Todo.find(function(err) {
+        should.not.exist(err);
+        done();
+      });
+    });
+
+    it('should work for findOne', function(done) {
+      Todo.findOne(function(err) {
+        should.not.exist(err);
+        done();
+      });
+    });
+
+    it('should work for deleteAll/destroyAll/remove', function(done) {
+      // FIXME: We should add a DAO.delete static method alias for consistency
+      // (DAO.prototype.delete instance method already exists)
+      var aliases = ['deleteAll', 'destroyAll', 'remove'];
+      async.each(aliases, function(alias, cb) {
+        Todo[alias](function(err) {
+          should.not.exist(err);
+          cb();
+        });
+      }, done);
+    });
+
+    it('should work for update/updateAll', function(done) {
+      Todo.update({content: 'Buy ham'}, function(err) {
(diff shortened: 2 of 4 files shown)


test/basic-querying.test.js
```diff
@@ -620,6 +620,61 @@ describe('basic-querying', function () {
   });
 });
 
+describe('id field', function() {
+  var Todo = getSchema().define('Todo', {
+    id: {type: 'number', id: false},
```
superkhau: @raymondfeng Does this look correct? I wasn't sure how to set the `id: false` here.

superkhau:
@raymondfeng Review please. Can you think of any operations I should add or check? I just did the basic CRUD operations.

test/basic-querying.test.js
```diff
@@ -620,6 +620,67 @@ describe('basic-querying', function () {
   });
 });
 
+describe('id field', function() {
+  var Todo;
+
+  before(function setupDb(done) {
+    var db = getSchema();
+    Todo = db.define('Todo', {
+      id: {type: 'number', id: false},
```
raymondfeng: It should be `id: false`.

test/basic-querying.test.js
```diff
@@ -620,6 +620,67 @@ describe('basic-querying', function () {
...
+
+
+  it('should not be required for update', function(done) {
+    Todo.update({content: 'Buy ham'}, function(err) {
+      should.not.exist(err);
+      done();
+    });
+  });
+});
+
```
raymondfeng: Please also add `count()`.

superkhau:
Raymond Feng [3:23 PM]

```
look for all methods from persisted model
All of the static methods that don’t take an `id` argument should work without PK (edited)
```

test/basic-querying.test.js
```diff
@@ -620,6 +620,73 @@ describe('basic-querying', function () {
   });
 });
 
+describe('id field', function() {
+  var Todo;
+
+  before(function setupDb(done) {
+    var db = getSchema();
+    Todo = db.define('Todo', {
+      id: false,
```
superkhau: @raymondfeng Like this?

test/basic-querying.test.js
```diff
@@ -620,6 +620,73 @@ describe('basic-querying', function () {
...
+    Todo.destroyAll();
+    done();
+  });
+
+  it('should not be required for count', function(done) {
+    Todo.count({content: 'Buy eggs'}, function(err) {
+      should.not.exist(err);
+      done();
+    });
+  });
```
superkhau: @raymondfeng Added count.

superkhau:
>look for all methods from persisted model

**Static**
- [ ] create
- [ ] upsert, updateOrCreate
- [ ] findOrCreate
- [ ] exists
- [ ] findById
- [ ] find
- [ ] findOne
- [ ] destroyAll, remove, deleteAll
- [ ] updateAll, update
- [ ] destroyById, deleteById, removeById
- [ ] count

- [ ] getIdName
- [ ] setupRemoting
- [ ] diff
- [ ] changes
- [ ] currentCheckpoint
- [ ] replicate
- [ ] createUpdates
- [ ] bulkUpdate
- [ ] getChangeModel
- [ ] getSourceId
- [ ] enableChangeTracking
- [ ] _defineChangeModel
- [ ] recitfyAllChanges
- [ ] handleChangeError
- [ ] rectifyChange
- [ ] findLastChange
- [ ] updateLastChange
- [ ] createChangeStream

**Instance**
- [ ] save
- [ ] isNewRecord
- [ ] destroy, remove, delete
- [ ] updateAttribute(s)
- [ ] reload
- [ ] setId
- [ ] getId
- [ ] getIdName
- [ ]

raymondfeng:
@superkhau Thanks for the list. You should mark the methods that don't require a PK.

Two more things to consider:
- For those methods that require a PK, we should check the model definition and report an error if the model doesn't have a PK
- For those methods that require a PK but the model doesn't have one, we probably should filter them out of REST APIs.

test/basic-querying.test.js
```diff
@@ -620,6 +620,131 @@ describe('basic-querying', function () {
...
+  it('should not be required for findOne', function(done) {
+    Todo.findOne(function(err) {
+      should.not.exist(err);
+      done();
+    });
+  });
+
+  it('should not be required for delete/deleteAll/destroyAll/remove', function(done) {
+    // FIXME: We should add a DAO.delete static method for consistency
+    // (DAO.prototype.delete instance method already exists)
```
superkhau: @raymondfeng Thoughts?

superkhau: @raymondfeng Thoughts?

superkhau:
@raymondfeng Review please.

@chandadharap I'm bumping this ticket 2 more points as the additional requirements are taking a bit more time to implement.

**tldr**

1. Adding 2 points to this story as scope creep added a bit more effort to the current task
2. Creating a 3 point story to track the second requirement
3. Creating a 1 point doc story to let users know about hiding REST endpoints with ID requirements

---

>For those methods that require a PK, we should check the model definition and report an error if the model doesn't have a PK

I'm creating a new story to track this task, and it will be a 3 pointer.

>For those methods that require a PK but the model doesn't have one, we probably should filter them out of REST APIs.

A third requirement discussed is we will need to add a doc story for @crandmck to let users know that the REST endpoints for functions such as `findById` will be "filtered", meaning we will hide REST endpoints that require an id, but do not have id defined in the model definition

This was all discussed with @raymondfeng over Slack, just giving you the heads up.

chandadharap:
sounds good.

superkhau:
@raymondfeng Actually 3 tests are failing. It's when the code can't access `this.definition.property.id` because it doesn't exist for models accessed through relations. I'm thinking I should validate **both** `this.definition.property.id` and all objects in the `this.definition._id` array for the `id` property since it gets automatically generated there. Thoughts?

superkhau:
@slnode test please

superkhau:
@slnode test please

superkhau:
@slnode test please

superkhau:
@slnode test please

superkhau:
@raymondfeng Ready for review.

test/basic-querying.test.js
```diff
@@ -620,6 +620,172 @@ describe('basic-querying', function () {
   });
 });
 
+describe('queries', function() {
```
ritch: This could be a bit more descriptive

lib/model-definition.js
```diff
@@ -302,3 +302,25 @@ ModelDefinition.prototype.toJSON = function (forceRebuild) {
...
+ */
+ModelDefinition.prototype.hasIdDefined = function() {
+  var isDefined = false;
+
+  if (this.properties.id)
+    isDefined = true;
+
+  // check if id is defined by a relation
+  this.ids().forEach(function(item) {
+    if (item.hasOwnProperty('id'))
```
ritch: What if `id` is `false`? I think `id` must be `=== true`.

ritch:
@superkhau - I had a couple minor comments, otherwise LGTM.

test/basic-querying.test.js
superkhau: When you run the test, it forms the BDD sentence. It is purposely left generic because this test covers a variety of different query types.

lib/model-definition.js
superkhau: The check for `false` happens before this. If the model definition has `id: false`, I believe `this.properties.id` is not set at all. Same goes for `this.ids()`.

Issue #681, opened 2015-08-05
RegExp objects in query should not be coerced to string


fabien:
See: https://github.com/strongloop/loopback-datasource-juggler/commit/b8f15987238bda922d3607203f2fdcdead4f250a#commitcomment-12537479

raymondfeng:
I'm fine to treat it as a shortcut to the newly introduced `regexp`, for example:

`{name: /John.*/i}`should be normalized as `{name: {regexp: /John.*/i}}`

PR #691, opened 2015-08-13
Do not coerce RegExp objects to strings

superkhau:
Queries like `{where: {name: /John.*/i}}` should work. Notice there is no
`regexp` operator (ie. `{where: {name: {regexp: /John.*/i}}}`).

Connect to #681

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1196,8 +1196,11 @@ DataAccessObject._coerce = function (where) {
       }
     } else {
       if (val != null) {
-        if (operator === 'regexp' && val instanceof RegExp) {
-          // do not coerce regex literals/objects
+        if (operator === null && val instanceof RegExp) {
+          // Normalize {name: /A/} to {name: {regexp: /A/}}
+          operator = 'regexp';
+        } else if (operator === 'regexp' && val instanceof RegExp) {
+          // Do not coerce regex literals/objects
         } else if (!((operator === 'like' || operator === 'nlike') && val instanceof RegExp)) {
           val = DataType(val);
         }
--- a/test/memory.test.js
+++ b/test/memory.test.js
@@ -356,6 +356,16 @@ describe('Memory connector', function() {
       });
     });
 
+    it('should work when a regex is provided without the regexp operator',
+        function(done) {
+      User.find({where: {name: /John.*/i}}, function(err, users) {
+        should.not.exist(err);
+        users.length.should.equal(1);
+        users[0].name.should.equal('John Lennon');
+        done();
+      });
+    });
+
     it('should support the regexp operator with regex strings', function(done) {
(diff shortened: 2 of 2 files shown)


superkhau:
@raymondfeng Review please.

lib/dao.js
```diff
@@ -1196,7 +1196,8 @@ DataAccessObject._coerce = function (where) {
       }
     } else {
       if (val != null) {
-        if (operator === 'regexp' && val instanceof RegExp) {
+        if (val instanceof RegExp ||
```
raymondfeng: You should check `(operator === null && val instanceof RegExp)`. If it's true, set the operator to `regexp`. This way, `{name: /A/}` will be normalized as `{name: {regexp: /A/}}` and you don't have to change any code in connectors.

lib/dao.js
```diff
@@ -1196,6 +1196,9 @@ DataAccessObject._coerce = function (where) {
       }
     } else {
       if (val != null) {
+        if (operator === null && val instanceof RegExp)
+          operator = 'regexp';
+
```
superkhau: @raymondfeng Better?

superkhau:
@slnode test please

lib/dao.js
raymondfeng: I prefer the following:

```
if (operator === null && val instanceof RegExp) {
  // Normalize {name: /A/} to {name: {regexp: /A/}}
  operator = 'regexp';
  // do not coerce regex literals/objects
} else if (!((operator === 'like' || operator === 'nlike') && val instanceof RegExp)) {
  val = DataType(val);
}
```

lib/connectors/memory.js
```diff
@@ -457,15 +457,13 @@ function applyFilter(filter) {
...
-      return value.match(example);
-    }
-
-    if (example === undefined) {
+    if (example === undefined)
       return undefined;
-    }
 
     if (typeof example === 'object' && example !== null) {
+      if (example instanceof RegExp)
```
raymondfeng: Please revert the connector change now that you normalize the shortcut to full regexp operator.

lib/dao.js
```diff
@@ -1196,8 +1196,11 @@ DataAccessObject._coerce = function (where) {
...
-        if (operator === 'regexp' && val instanceof RegExp) {
-          // do not coerce regex literals/objects
+        if (operator === null && val instanceof RegExp) {
+          // Normalize {name: /A/} to {name: {regexp: /A/}}
+          operator = 'regexp';
+        } else if (operator === 'regexp' && val instanceof RegExp) {
+          // Do not coerce regex literals/objects
         } else if (!((operator === 'like' || operator === 'nlike') && val instanceof RegExp)) {
           val = DataType(val);
         }
```
superkhau: @raymondfeng LGTY?

raymondfeng:
LGTM

PR #694, opened 2015-08-14
Look in array for value if data is array and query is not

jrschumacher:
Useful when using memory for tests to replicate mongodb

http://docs.mongodb.org/manual/reference/method/db.collection.find/#quer
y-a-field-that-contains-an-array

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -513,6 +513,19 @@ function applyFilter(filter) {
         return true;
       }
     }
+
+    // Look in array for value
+    // Useful when using memory for tests when trying to replicate mongodb
+    if (typeof example !== 'object' && typeof value === 'object'
+        && Object.prototype.toString.call( value ) === '[object Array]') {
+      if (value.indexOf(example) === -1) {
+        return false;
+      }
+      else {
+        return true;
+      }
+    }
+
     // not strict equality
     return (example !== null ? example.toString() : example)
       == (value != null ? value.toString() : value);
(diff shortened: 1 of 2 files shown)


slnode:

lib/connectors/memory.js
```diff
@@ -513,6 +513,19 @@ function applyFilter(filter) {
         return true;
       }
     }
+
+    // Look in array for value
+    // Useful when using memory for tests when trying to replicate mongodb
+    if (typeof example !== 'object' && typeof value === 'object'
+        && Object.prototype.toString.call( value ) === '[object Array]') {
```
raymondfeng: You should use `Array.isArray(value)` instead.

test/memory.test.js
```diff
@@ -430,6 +431,17 @@ describe('Memory connector', function() {
       });
     });
 
+    it('should search array for value', function(done) {
+      User.find({where: {favoriteNumbers: 1}}, function(err, users) {
```
raymondfeng: I think we probably should introduce a different operator such as `contains` instead of overloading the default operator (eq).

raymondfeng: BTW, it is not very straightforward to implements `contains` for SQL based connectors as we store the array value as stringified string in a column.

jrschumacher:
@raymondfeng then I think should just nix this patch. It's inconsistent with mongodb to use contains and it's inconsistent with sql to use eq.

Issue #695, opened 2015-08-17
embedsMany relations dont work with mongo ObjectID


mazesoul:
The embedsMany relation doesn't take into account mongo style ids. 

Here are some instances where the code uses `indexOf`and breaks with mongo object based ids:

https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L2273
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L2874
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L3014

Some working instances:

https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L3035

raymondfeng:
Just to make sure: in your case, does the array contain the `equal` but not `same` ObjectID?

mazesoul:
Exactly. The comparison occurs by identity in indexOf (`===`), and doesn't work with two ObjectID that contain the same string: 

```
var ids = [ObjectID('123')];
var id2 = ObjectID('123');

ids.indexOf(id2) !== 1
```

raymondfeng:
We already have a utility to deal with id comparison - https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L2280. It should be refactored into utils.js and use it for both places.

mazesoul:
I did saw it but the full-blown refactoring seemed a bit above my head.

PR #696, opened 2015-08-17
Fix id comparison by using strings

mazesoul:
Fix issue #695 by using string conversion before `indexOf` usage


slnode:

raymondfeng:
I just refactored the function into utils - https://github.com/strongloop/loopback-datasource-juggler/commit/f5270c39c5f3ba17ca5f15040582786f1a3541f7. Can you rebase your patch?

mazesoul:
I've used idEquals whenever possible. Tests are green. I also added `findIndexOf` in utils that accepts an isEqual predicate.

PR #697, opened 2015-08-18
Support embedded query in memory connector. Fix memory connector bug

mazesoul:
Adds support for the ability to query embedsMany models from the parent.

Fix a memory connector bug that could occur when having an "or" or "and"
clause combined with another property. In that case, the and would revert
true for 'Paul McCartney'

```
{name:'John Lennon',and: [{role:'lead'}, {vip:true}]}}
```

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -400,29 +400,39 @@ function applyFilter(filter) {
   }
   var keys = Object.keys(where);
   return function (obj) {
-    var pass = true;
-    keys.forEach(function(key) {
+    return keys.every(function(key) {
       if(key === 'and' || key === 'or') {
         if(Array.isArray(where[key])) {
           if(key === 'and') {
-            pass = where[key].every(function(cond) {
+            return where[key].every(function(cond) {
               return applyFilter({where: cond})(obj);
             });
-            return pass;
           }
           if(key === 'or') {
-            pass = where[key].some(function(cond) {
+            return where[key].some(function(cond) {
               return applyFilter({where: cond})(obj);
             });
-            return pass;
           }
         }
       }
-      if (!test(where[key], getValue(obj, key))) {
-        pass = false;
+
+      if (test(where[key], getValue(obj, key))) {
+        return true;
       }
+
+      // If we have a composed key a.b and b would resolve to a property of an object inside an array
+      // then, we attempt to emulate mongo db matching. Helps for embedded relations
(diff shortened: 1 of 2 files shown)


mazesoul:
@altsang Are you in a position to give me an ETA for the review/merge. Also, is this something that you can accept?

raymondfeng:
1. Before your PR, the `and` and `or` operator cannot be combined with regular clauses. Your case needs to be written as:
```
{and: [{name:'John Lennon'}, {role:'lead'}, {vip:true}]}
```
If we allow and/or to be used with regular clauses, we need to make sure other connectors support it too.

2. We already support nested properties for the memory connector even before this PR, see https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/memory.test.js#L397-L428. What else does the PR add?

mazesoul:
@raymondfeng The memory connector didn't support nested properties over arrays. Like this https://github.com/WyzeLink/loopback-datasource-juggler/blob/feature/support-embedded-query-in-memory/test/memory.test.js#L431-L439. Where friends is an array of object. This is important to be able to test embedsMany relations

Concerning the support for the and/or clauses, from what I can gather the following all default to loopback-connector's `_buildWhere` that does concatenate all the statements together https://github.com/strongloop/loopback-connector/blob/master/lib/sql.js#L827.

Moreover, in the memory connector's implementation, there was a bug even if it doesn't support both. If the `and/or` clause appeared before another key, it could be reverted because of the forEach iteration. An `and` that was true, could have been made false if the key was declared after. The `pass` in the closure, although only setting to false, could in certain circumstances cause confusion by failing the filtering on the secondary case.

- https://github.com/strongloop/loopback-connector-mysql
- https://github.com/strongloop/loopback-connector-postgresql
- https://github.com/strongloop/loopback-connector-oracle/blob/master/lib/oracle.js#L484
- https://github.com/strongloop/loopback-connector-mssql/blob/master/lib/mssql.js#L368

As for mongo, its where build is additive: https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L553

mazesoul:
And finally, a bug which I forgot to mention in the PR's comment: supposing there are two objects, only one of which having a value for a string key, the regex match would throw because there was no verification: https://github.com/strongloop/loopback-datasource-juggler/pull/697/files#diff-98aa884e786e8c02d53648b6f0afaa82R481

mazesoul:
Thanks :)

PR #699, opened 2015-08-19
Fixed ReferencesMany .findById to check the given id in the modelInstance[fk]

satyadeepk:
### Issue
`findById` in `ReferencesMany` was comparing the `fkId` with itself. Hence `findById` method always returns the `modelTo` instance whether the `modelFrom` instance contains it or not.

### Fix
Fixed the above issue by checking whether the `fkId` is present in `modelInstance[fk]`. 

cc: @fabien


slnode:

raymondfeng:
ok to test

PR #704, opened 2015-08-25
With simpe and fast hasmany

wertlex:
Hey guys, I'm not Loopback guru or even experienced user, but we were having some problems with performance of `hasMany` relation. 

It was affected cases when for ex. you have about 100 users (user hasMany permissions) and 100 permissions for them totally. As all join job done in app it was rather easy to trace whats going on. 

I introduced method `includeHasManySimple` which handles only simple cases of `hasMany` - without `subIncludes`. This is how it works: 
* building id -> obj1 map in object form to make by-id queries easy 
* collects ids to query from 2nd model
* builds and run query on 2nd model
* after that result of 2nd query collected to id -> [obj] map which make by-id queries easy and fast again
* for each object in 1st map cache filled and `processTargetObj` called 

This allow to do blazing fast joins. 

Second thing to pay attention is `processTargetObj` function

`var inst = (obj instanceof self) ? obj : new self(obj);`

moved down to place where it really used, and as in a lot of cases `inst` is used in `if(obj === inst)` clause `isInst` value introduced. So in most cases `inst` never used and it makes `processTargetObj` very fast. 

BTW, I still not sure, but in my case `new self(obj);` takes about 230ms which looks very poor.  I still not found the reason of such behaviour.

[Review on Reviewable](https://reviewable.io/reviews/strongloop/loopback-datasource-juggler/704)

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -845,7 +845,9 @@ Inclusion.include = function (objects, include, options, cb) {
      * @returns {*}
      */
     function processTargetObj(obj, callback) {
-      var inst = (obj instanceof self) ? obj : new self(obj);
+      
+      var isInst = obj instanceof self;
+
       // Calling the relation method on the instance
       if (relation.type === 'belongsTo') {
         // If the belongsTo relation doesn't have an owner
@@ -853,7 +855,7 @@ Inclusion.include = function (objects, include, options, cb) {
           defineCachedRelations(obj);
           // Set to null if the owner doesn't exist
           obj.__cachedRelations[relationName] = null;
-          if (obj === inst) {
+          if (isInst) {
             obj.__data[relationName] = null;
           } else {
             obj[relationName] = null;
@@ -867,7 +869,7 @@ Inclusion.include = function (objects, include, options, cb) {
        * @param cb
        */
       function setIncludeData(result, cb) {
-        if (obj === inst) {
+        if (isInst) {
           if (Array.isArray(result) && !(result instanceof List)) {
             result = new List(result, relation.modelTo);
           }
@@ -885,6 +887,9 @@ Inclusion.include = function (objects, include, options, cb) {
         return setIncludeData(obj.__cachedRelations[relationName],
           callback);
       }
+
+      var inst = (obj instanceof self) ? obj : new self(obj);
+
       //If related objects are not cached by include Handlers, directly call
       //related accessor function even though it is not very efficient
       var related; // relation accessor function
--- a/lib/include.js
+++ b/lib/include.js
@@ -1,6 +1,7 @@
 var async = require('async');
 var utils = require('./utils');
 var List = require('./list');
+var includeUtils = require('./include_utils');
 var isPlainObject = utils.isPlainObject;
 var defineCachedRelations = utils.defineCachedRelations;
 var uniq = utils.uniq;
@@ -267,6 +268,11 @@ Inclusion.include = function (objects, include, options, cb) {
       if (relation.type === 'referencesMany') {
         return includeReferencesMany(cb);
       }
+
+      //This handles exactly hasMany. Fast and straightforward. Without parallel, each and other boilerplate.
+      if(relation.type === 'hasMany' && relation.multiple && !subInclude) {
+        return includeHasManySimple(cb);
+      }
       //assuming all other relations with multiple=true as hasMany
       return includeHasMany(cb);
     }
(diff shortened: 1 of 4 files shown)


slnode:

lib/include.js
```diff
@@ -486,6 +492,37 @@ Inclusion.include = function (objects, include, options, cb) {
      * Handle inclusion of HasMany relation
      * @param callback
      */
+    function includeHasManySimple(callback) {
+      //Map for Indexing objects by their id for faster retrieval
+      var objIdMap = includeUtils.buildOneToOneIdentityMap(objs, relation.keyFrom);
+      // all ids of primary objects to use in query
+      var sourceIds = Object.keys(objIdMap);
```
raymondfeng: The source ids will be string[]. It will probably breaks for number/objectid keys. Can we keep the original id values?

raymondfeng:
@wertlex I added a comment. Can you address that? This is a good optimization.

wertlex:
Yes, sure. Will do in couple of days. A bit busy right now.

raymondfeng:
@wertlex Did you have a chance to refine the PR?

wertlex:
@raymondfeng nope, but I'm definetly very close to it.

wertlex:
@raymondfeng eventually start working on that. Looks like I've done with half and write my own simple map which stores original values too. Not sure I should proud about that, but looks like this is the only way to do that with JS.

raymondfeng:
@wertlex Thanks. Are you done now?

wertlex:
@raymondfeng looks like thats it. I made special data structure KVMap which is kind of map where keys are arbitrary values. Looks like this should solve problem with non-string keys.
Test also included. Check this out plz :)

lib/include_utils.js
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Map with arbitrary keys and values. User .set() and .get() to work with values instead of []
+ * @returns {{set: Function, get: Function, remove: Function, exist: Function, getKeys: Function}}
+ * @constructor
+ */
+function KVMap(){
+  var _originalKeyFieldName = 'originalKey';
+  var _valueKeyFieldName = 'value';
+  var _dict = {};
+  var keyToString = function(key){ return key.toString() };
```
zanemcca: This is breaking my production environment.
The key passed in does not have a toString() function. I am not sure why yet.
@wertlex @raymondfeng are all keys supposed to have a toString() function?

raymondfeng: What key type do you use?

zanemcca: I believe include is set to an empty array in the failing case.
I am pushing some changes that should allow me to know for certain.

zanemcca: I was wrong about the empty array.
Here is my include query.
```
include = [{
  relation: 'child',
  scope: {
    where: {
      username: 'someUser'
    }
  }
},
{
  relation: 'otherChild',
  scope: {
    where: {
      username: 'someUser'
    }
  }
}];
```

zanemcca: Here is the stack trace from the crash.

```
 events.js:141
       throw er; // Unhandled 'error' event
       ^
 
 TypeError: Cannot read property 'toString' of undefined
     at KVMap.keyToString (/usr/src/app/node_modules/loopback-datasource-juggler/lib/include_utils.js:64:46)
     at Object.KVMap.mapImpl.set (/usr/src/app/node_modules/loopback-datasource-juggler/lib/include_utils.js:70:13)
     at Object.buildOneToOneIdentityMapWithOrigKeys (/usr/src/app/node_modules/loopback-datasource-juggler/lib/include_utils.js:19:11)
     at includeHasManySimple (/usr/src/app/node_modules/loopback-datasource-juggler/lib/include.js:497:36)
     at processIncludeItem (/usr/src/app/node_modules/loopback-datasource-juggler/lib/include.js:274:16)
     at /usr/src/app/node_modules/loopback-datasource-juggler/lib/include.js:172:5
     at /usr/src/app/node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:157:13
     at _each (/usr/src/app/node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:57:9)
     at Object.async.each (/usr/src/app/node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:156:9)
     at Function.Inclusion.include (/usr/src/app/node_modules/loopback-datasource-juggler/lib/include.js:171:9)
```

raymondfeng: Can you try to dump out the offending keys that don't have toString()?

zanemcca: My deployment is not playing nice so I am not able to easily get into the node_modules folder to change this to give a dump. 

I am not sure what a "key" is supposed to be in this context.
Is it supposed to be the contents of the include query?

zanemcca: Ok I found a way of getting the key dump. 
The problem seems to have appeared because I use the fields query.
```
"fields": [
   "position",
   "value"
]
```
So in buildOneToOneIdentityMapWithOrigKeys(objs, idName) I ended up with 
`obj = { location: {lat: 45, lng: 45}, value: 5 }`
and
`idName = 'id'`
which resulted in the key being undefined.

raymondfeng: Interesting, I guess we should always include `id` property in the `fields` so that the id value can be used as foreign keys. Can you test if including `id` in `fields` will solve your problem?

wertlex:
Looks like `includeHasMany()` have  special handling for cases when original object doesn't have `id` property: https://github.com/wertlex/loopback-datasource-juggler/blob/d9918d526a7a59dbc610295b5727311be4ef0359/lib/include.js#L533

But I'm still not sure what exactly `defineCachedRelations(obj)` for. 

@zanemcca simplest fix right now should be just adding `id` in `fields`. 

@raymondfeng can you tell how reasonable fields queries without explicit id?

If this is the case we can add related tests and fix `includeHasManySimple`.

lib/include_utils.js
zanemcca: I can confirm that adding `id` to `fields` does indeed solve the problem.

Issue #706, opened 2015-08-26
Fix primary key checks


superkhau:
- [x] Update `hasIdDefined` to check is `ids()` empty and `idNames()` empty
- [x] Fix functions that operate on single instances (all `.prototype` functions should not work without a PK defined)

superkhau:
prototype methods:

- [ ] save
- [ ] destroy/remove/delete
- [ ] updateAttribute(s)

PR #707, opened 2015-08-26
Fix primary key checks

superkhau:
Connect to #706

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -856,8 +856,8 @@ DataAccessObject.findById = function findById(id, filter, options, cb) {
   assert(typeof options === 'object', 'The options argument must be an object');
   assert(typeof cb === 'function', 'The cb argument must be a function');
 
-  if (!this.definition.hasIdDefined()) {
-    process.nextTick(cb(getIdDefinitionError()));
+  if (isPKMissing(this, cb)) {
+    return cb.promise;
   } else if (id == null || id === '') {
     process.nextTick(function() {
       cb(new Error('Model::findById requires the id argument'));
@@ -906,8 +906,7 @@ DataAccessObject.findByIds = function(ids, query, options, cb) {
   assert(typeof options === 'object', 'The options argument must be an object');
   assert(typeof cb === 'function', 'The cb argument must be a function');
 
-  if (!this.definition.hasIdDefined()) {
-    process.nextTick(cb(getIdDefinitionError()));
+  if (isPKMissing(this, cb)) {
     return cb.promise;
   } else if (ids.length === 0) {
     process.nextTick(cb(null, []));
@@ -1697,8 +1696,7 @@ DataAccessObject.removeById = DataAccessObject.destroyById = DataAccessObject.de
   assert(typeof options === 'object', 'The options argument must be an object');
   assert(typeof cb === 'function', 'The cb argument must be a function');
 
-  if (!this.definition.hasIdDefined()) {
-    process.nextTick(cb(getIdDefinitionError()));
+  if (isPKMissing(this, cb)) {
     return cb.promise;
   } else if (id == null || id === '') {
     process.nextTick(function() {
@@ -1839,7 +1837,9 @@ DataAccessObject.prototype.save = function (options, cb) {
   assert(typeof options === 'object', 'The options argument should be an object');
   assert(typeof cb === 'function', 'The cb argument should be a function');
 
-  if (this.isNewRecord()) {
+  if (isPKMissing(Model, cb)) {
+    return cb.promise;
+  }  else if (this.isNewRecord()) {
     return Model.create(this, options, cb);
   }
 
@@ -2141,6 +2141,9 @@ DataAccessObject.prototype.remove =
       var id = getIdValue(this.constructor, this);
       var hookState = {};
 
+      if (isPKMissing(Model, cb))
+        return cb.promise;
+
       var context = {
         Model: Model,
         query: byIdQuery(Model, id),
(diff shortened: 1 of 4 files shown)


lib/model-definition.js
```diff
@@ -303,24 +303,6 @@ ModelDefinition.prototype.toJSON = function (forceRebuild) {
...
-
-  // check if id is defined by a relation
-  this.ids().forEach(function(item) {
-    if (item.hasOwnProperty('id'))
-      isDefined = true;
-  });
-
-  return isDefined;
+ModelDefinition.prototype.hasPkDefined = function() {
+  return (this.ids().length !== 0 || this.idNames().length !== 0);
```
raymondfeng: No need to check both. Either ids() or idNames() is good enough.

lib/model-definition.js
```diff
@@ -303,24 +303,6 @@ ModelDefinition.prototype.toJSON = function (forceRebuild) {
...
-
-  // check if id is defined by a relation
-  this.ids().forEach(function(item) {
-    if (item.hasOwnProperty('id'))
-      isDefined = true;
-  });
-
-  return isDefined;
+ModelDefinition.prototype.hasPrimaryKey = function() {
+  return (this.ids().length > 0 || this.idNames().length > 0);
```
superkhau: @raymondfeng Is this ok?

raymondfeng: It's good enough to have `return this.ids().length > 0;`

superkhau:
@raymondfeng Review please.

lib/dao.js
```diff
@@ -2542,13 +2550,8 @@ jutil.mixin(DataAccessObject, Relation);
...
-/**
- * Return an error object stating the `id` property must be part of the model
- * definition.
- *
- * @return {Object} The error stating the requirements
- */
-function getIdDefinitionError() {
-  return new Error('`id` property must be defined as part of the model ' +
-    'definition');
+function PrimaryKeyMissingError(msg) {
```
raymondfeng: We probably need to know which model it is.

lib/dao.js
```diff
@@ -2542,13 +2550,9 @@ jutil.mixin(DataAccessObject, Relation);
...
- * definition.
- *
- * @return {Object} The error stating the requirements
- */
-function getIdDefinitionError() {
-  return new Error('`id` property must be defined as part of the model ' +
-    'definition');
+function PrimaryKeyMissingError(modelName) {
+  this.name = 'PrimaryKeyMissingError';
+  this.message = 'Primary key missing from the ' + modelName + ' model ' +
```
superkhau: @raymondfeng Including model name in the error message now.

lib/dao.js
```diff
@@ -1839,7 +1839,9 @@ DataAccessObject.prototype.save = function (options, cb) {
   assert(typeof options === 'object', 'The options argument should be an object');
   assert(typeof cb === 'function', 'The cb argument should be a function');
 
-  if (this.isNewRecord()) {
+  if (!Model.definition.hasPrimaryKey()) {
+    process.nextTick(cb(new PrimaryKeyMissingError(Model.modelName)));
```
raymondfeng: Do you need to `return cb.promise;`?

lib/dao.js
```diff
@@ -2141,6 +2143,9 @@ DataAccessObject.prototype.remove =
       var id = getIdValue(this.constructor, this);
       var hookState = {};
 
+      if (!Model.definition.hasPrimaryKey())
+        process.nextTick(cb(new PrimaryKeyMissingError(Model.modelName)));
+
```
raymondfeng: Do you need to `return cb.promise;`?

lib/dao.js
```diff
@@ -2327,6 +2332,9 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
   assert(typeof connector.updateAttributes === 'function',
     'updateAttributes() must be implemented by the connector');
 
+  if (!Model.definition.hasPrimaryKey())
+    process.nextTick(cb(new PrimaryKeyMissingError(Model.modelName)));
+
```
raymondfeng: Do you need to `return cb.promise;`?

lib/dao.js
superkhau: Actually, yes I think so.

lib/dao.js
```diff
@@ -856,8 +856,9 @@ DataAccessObject.findById = function findById(id, filter, options, cb) {
   assert(typeof options === 'object', 'The options argument must be an object');
   assert(typeof cb === 'function', 'The cb argument must be a function');
 
-  if (!this.definition.hasIdDefined()) {
-    process.nextTick(cb(getIdDefinitionError()));
+  if (!this.definition.hasPrimaryKey()) {
+    process.nextTick(cb(new PrimaryKeyMissingError(this.modelName)));
```
raymondfeng: This is not going to work. `process.nextTick()` takes a function. The code should look like the following:
```js

   // Define a helper function
   function isPKMissing(modelClass, cb) {
     var hasPK = modelClass.definition.hasPrimaryKey();
     if (hasPK) return false;
     process.nextTick(function() {
       cb(new PrimaryKeyMissingError(modelClass.modelName);
     }
     return true;
 }

// Use the helper function
if (isPKMissing(this, cb)) return cb.promise;
```

lib/dao.js
```diff
@@ -856,8 +856,8 @@ DataAccessObject.findById = function findById(id, filter, options, cb) {
   assert(typeof options === 'object', 'The options argument must be an object');
   assert(typeof cb === 'function', 'The cb argument must be a function');
 
-  if (!this.definition.hasIdDefined()) {
-    process.nextTick(cb(getIdDefinitionError()));
+  if (isPKNotFound(this, cb)) {
```
raymondfeng: Nick-pick: double-negative is confusing.

superkhau: True, I will change it all back to `Missing` ;)

raymondfeng:
LGTM

superkhau:
@slnode test please

Issue #625, opened 2015-06-11
Implement `contains` operator to mimic MongoDB


fabien:
In order to resolve #622 I would like to propose a `contains` operator, which mimics the transparent handling of array matching in MongoDB. 

A feature branch lives here: https://github.com/fabien/loopback-datasource-juggler/tree/feature/query-contains

Currently, the logic of `contains` returns true if either case is true (based on `compare`):
- array `a` contains value `b`
- array `a` intersects with array `b` (you probably don't want to nest arrays for this anyway)

This is probably not 100% exactly like MongoDB, but I think we can make all connectors behave the same if we'd agree on adding `contains` as a new core operator.

/cc @raymondfeng @bajtos

fabien:
It's important to note that we don't strictly need `contains` to resolve the `include` issue in MongoDB.

bajtos:
I don't understand this topic enough to judge the proposed solution, I'll leave that task to @raymondfeng - I believe he is more familiar with this part of juggler. Adding a new operator "contains" looks like a good thing to me.

@fabien does you proposal include ability to nest operators inside `contains`? E.g. is it possible to say `array A contains value between B-C` or `array A contains an object value that has property B equal to C`?

fabien:
@bajtos my proposal is rather simple, it's like an inverted `inq` in a way. Let's start simple and see if we can indeed nest - it's probably possible to do so, however, having all connectors support it is a different matter I guess.

jannyHou:
@fabien do you still plan to work on this feature branch?
https://github.com/fabien/loopback-datasource-juggler/tree/feature/query-contains
Or any update thought about this feature?
Thanks.

Issue #147, opened 2014-06-20
Memory Connector: inq filter should respect Arrays


fabien:
See: https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/connectors/memory.js#L379

To make the comparison follow MongoDB for example, the value should intersect at least once (by casting any value to an array). 

Something like (needs a more fitting and well-tested solution, without dependencies):

``` javascript
if (example.inq) {
  if (!value) return false;
  var sample = [].concat(value || []);
  return _.intersect(example.inq, sample).length > 0;
}
```

loay:
Hi @fabien 
Did the merged pull request #710 Fix the issue?
Thanks.

rmmlxr:
Does not work for me; using postgres connector.

PR #710, opened 2015-08-28
Add support for matching array values à la mongo.

mazesoul:
Support mongo style matcher in where clause. 

embedsMany, referencesMany all have type array properties that can be filtered with when using loopback-mongo-connector. These don't have counterparts in relational dbs but precluding the use of the filters in the memory connectors complicates tests for mongo based apps.

Fixes the memory connector support mentionned in strongloop/loopback#953, strongloop/loopback#517.

Provides a baseline for the memory connector support of #625. At least for unit tests of a mongo based app.

Other issues related to this: #147

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -416,7 +416,20 @@ function applyFilter(filter) {
         }
       }
 
-      if (test(where[key], getValue(obj, key))) {
+      var value = getValue(obj, key);
+      // Support referencesMany and other embedded relations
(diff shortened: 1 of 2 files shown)


mazesoul:
Trying to retrigger build by close and open

mazesoul:
@raymondfeng Could you have a look at this, it's the "simple array of values" counter part of the change made in #697. This is useful when using 'referencesMany'. I know other connector don't support this but I doesn't prevent anything and is massively useful when building unit test for a mongo base app. 

Thanks.

PR #711, opened 2015-09-02
dao: should not convert the val type if object

yorkie:
If the value is an object, maybe we should not convert it to corresponding type like supporting:

```
{ '$exists': true }
```

/cc @raymondfeng

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1494,7 +1494,8 @@ DataAccessObject._coerce = function(where) {
           operator = 'regexp';
         } else if (operator === 'regexp' && val instanceof RegExp) {
           // Do not coerce regex literals/objects
-        } else if (!((operator === 'like' || operator === 'nlike') && val instanceof RegExp)) {
+        } else if (!((operator === 'like' || operator === 'nlike') &&
+          (val instanceof RegExp || val instanceof Object))) {
           val = DataType(val);
         }
       }
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -525,31 +525,31 @@ describe('basic-querying', function() {
 
     it('should find first record (default sort by id)', function(done) {
       User.all({ order: 'id' }, function(err, users) {
-        User.findOne(function(e, u) {
-          should.not.exist(e);
-          should.exist(u);
-          u.id.toString().should.equal(users[0].id.toString());
+        User.findOne(function(err, user) {
+          should.not.exist(err);
+          should.exist(user);
+          user.id.toString().should.equal(users[0].id.toString());
           done();
         });
       });
     });
 
     it('should find first record', function(done) {
-      User.findOne({ order: 'order' }, function(e, u) {
-        should.not.exist(e);
-        should.exist(u);
-        u.order.should.equal(1);
-        u.name.should.equal('Paul McCartney');
+      User.findOne({ order: 'order' }, function(err, user) {
+        should.not.exist(err);
+        should.exist(user);
+        user.order.should.equal(1);
+        user.name.should.equal('Paul McCartney');
         done();
       });
     });
 
     it('should find last record', function(done) {
-      User.findOne({ order: 'order DESC' }, function(e, u) {
-        should.not.exist(e);
-        should.exist(u);
-        u.order.should.equal(6);
-        u.name.should.equal('Ringo Starr');
+      User.findOne({ order: 'order DESC' }, function(err, user) {
+        should.not.exist(err);
+        should.exist(user);
+        user.order.should.equal(6);
+        user.name.should.equal('Ringo Starr');
         done();
       });
     });
@@ -574,6 +574,21 @@ describe('basic-querying', function() {
           should.exist(user);
           done();
         });
+      });
+    });
+
+    it('should query by using database operator', function(done) {
+      User.findOne({
+        where: {
+          role: {
+            '$exists': false,
+          },
+        },
+        order: 'order DESC',
+      }, function(err, user) {
+        should.not.exist(err);
+        should.not.exist(user);
+        done();
       });
     });
 


slnode:

raymondfeng:
@yorkie Thank you for the patch. Can you provide a test case to demonstrate what issue you are trying to solve? We'll include the test when the PR is merged.

yorkie:
Hey @raymondfeng I'm fixing the same issue at https://github.com/strongloop/loopback/issues/646, because I was occurring as well, and this patch should just provide a ability to use MongoDB's any $ operator without supporting by loopback.

If you think this is valuable, I will add the test case :)

yorkie:
Ping @raymondfeng :-)

slnode:

superkhau:
>@yorkie Thank you for the patch. Can you provide a test case to demonstrate what issue you are trying to solve? We'll include the test when the PR is merged. 

Can you add a test case please?

lib/dao.js
```diff
@@ -1206,7 +1206,7 @@ DataAccessObject._coerce = function (where) {
           operator = 'regexp';
         } else if (operator === 'regexp' && val instanceof RegExp) {
           // Do not coerce regex literals/objects
-        } else if (!((operator === 'like' || operator === 'nlike') && val instanceof RegExp)) {
+        } else if (!((operator === 'like' || operator === 'nlike') && val instanceof Object)) {
```
superkhau: I think we should add another || for your case. Why is the RegExp check being changed?

yorkie: As I have said in description, current loopback's `where` filter doesn't support complicated query which contains something mongoDB operators in an object.

superkhau: That part I understand. But why not `.. && (val instanceof RegExp || val instanceof Object)) {`?

yorkie: But I guess the `RegExp` is also an `Object`?

superkhau: Yes it is, but you hide the fact that you're checking for RegExp explicitly when you combine them together.

yorkie: Okay, that makes sense.

superkhau: Do you want to make that change and then we can get it merged in?

superkhau: Can you also write a unit test to verify your changes and prevent regressions in the future?

yorkie: Ok, I will update nits and add test case ASAP :-)

superkhau: TY, ping me again after and I will get it merged in ASAP. ;)

yorkie:
@superkhau I have added the test case and fix nits, could you please have a review again? Thank you :-)

superkhau:
Can you run the tests through the linter? I see some stylistic issues that will probably cause linting to fail.

```
npm test
```

superkhau:
Can you also squash the commits?

yorkie:
Hey @superkhau the `npm test` seems not include the lint check?

yorkie:
Done the squash :-)

superkhau:
I just looked, it does, but still uses makefiles (which we are in the process of removing). However the lint code is there, so you should be able to run `npm run lint` manually. ;) Please try that instead.

bajtos:
> ability to use MongoDB's any $ operator without supporting by loopback.

FWIW, we already support MongoDB's `$` operators in `DAO.updateAttributes`, one has to enable that feature via model-level setting `allowExtendedOperators`. Perhaps we should do the same here and enable this feature only when the flag is enabled?

yorkie:
@bajtos do you mean check `allowExtendedOperators` here?

bajtos:
> do you mean check `allowExtendedOperators` here?

Yes, that's correct, allow MongoDB-like object values in query only when `allowExtendedOperators` is turned on. Sorry for not being more clear!

However, I am not sure if it's the right thing to do. I'd rather leave it up to you, @raymondfeng and @superkhau  to decide.

yorkie:
Okay, I will fix the lint warnings pointed by @superkhau firstly :-)

yorkie:
Fixed and did rebase @superkhau :-)

superkhau:
@raymondfeng Input on the above comments?

test/basic-querying.test.js
```diff
@@ -577,6 +577,21 @@ describe('basic-querying', function() {
...
 
+    it('should query by using database operator', function(done) {
+      User.findOne({
+        where: {
+          role: {
+            '$exists': false,
+          },
+        },
+        order: 'order DESC',
+      }, function(e, u) {
```
superkhau: err, user

test/basic-querying.test.js
```diff
@@ -577,6 +577,21 @@ describe('basic-querying', function() {
...
+        should.not.exist(e);
```
superkhau: exist(err);

test/basic-querying.test.js
```diff
@@ -577,6 +577,21 @@ describe('basic-querying', function() {
...
+        should.not.exist(u);
```
superkhau: exist(user);

superkhau:
@yorkie A few more minor comments to make your code more clear. One more time please. ;)

superkhau:
And squash all changes back into one commit please.

yorkie:
Hi @superkhau I have fixed and squash done, please review again, thanks :-)

superkhau:
LGTM. Thanks for the contribution! ;)

superkhau:
>Yes, that's correct, allow MongoDB-like object values in query only when allowExtendedOperators is turned on. Sorry for not being more clear!

@raymondfeng What do you think about this comment?

raymondfeng:
LGTM

superkhau:
@yorkie Can you add the flag check to this PR? We can merge it after that

superkhau:
Note, I spoke to @raymondfeng regarding the flag, we want it implemented to work with `allowExtendedOperators`.

yorkie:
Ok, got it, I will do it soon later :-)

yorkie:
Hey @superkhau @raymondfeng Do you have something documentation about the config `allowExtendedOperators`?

superkhau:
@crandmck ^

crandmck:
> Do you have something documentation about the config allowExtendedOperators?

What we have is just for Mongo: https://docs.strongloop.com/display/APIC/MongoDB+connector#MongoDBconnector-UsingMongoDBoperatorsinupdateoperations

IIUC, this change means you can use it with other data sources?  If so, which ones?  
And is the config the same in `<modelname>.json` ?

superkhau:
>IIUC, this change means you can use it with other data sources--right? If so, which ones?

Not sure about this one (I believe is only mongodb for now), but @bajtos or @raymondfeng can confirm this.

>And is the config the same in <modelname>.json ?

Yes, the config is the same.

crandmck:
If it's only Mongo, then what does this add over what's already documented?

test/basic-querying.test.js
```diff
@@ -525,31 +525,31 @@ describe('basic-querying', function() {
...
     });
 
     it('should find first record', function(done) {
-      User.findOne({ order: 'order' }, function(e, u) {
-        should.not.exist(e);
-        should.exist(u);
-        u.order.should.equal(1);
-        u.name.should.equal('Paul McCartney');
+      User.findOne({ order: 'order' }, function(err, user) {
+        should.not.exist(err);
```
superkhau: We should change this to `if (err) return done(err)` for better error output.

test/basic-querying.test.js
```diff
@@ -525,31 +525,31 @@ describe('basic-querying', function() {
...
     });
 
     it('should find last record', function(done) {
-      User.findOne({ order: 'order DESC' }, function(e, u) {
-        should.not.exist(e);
-        should.exist(u);
-        u.order.should.equal(6);
-        u.name.should.equal('Ringo Starr');
+      User.findOne({ order: 'order DESC' }, function(err, user) {
+        should.not.exist(err);
```
superkhau: same here

test/basic-querying.test.js
```diff
@@ -577,6 +577,21 @@ describe('basic-querying', function() {
...
+    it('should query by using database operator', function(done) {
+      User.findOne({
+        where: {
+          role: {
+            '$exists': false,
+          },
+        },
+        order: 'order DESC',
+      }, function(err, user) {
+        should.not.exist(err);
```
superkhau: here too

test/basic-querying.test.js
```diff
@@ -577,6 +577,21 @@ describe('basic-querying', function() {
...
+        should.not.exist(user);
```
superkhau: We should check all the user properties (to make sure we get the correct user -- ie. user.name, etc).

superkhau:
@yorkie is adding a feature that will only turn on if `allowExtendedOperators` is set to true in the config, so it doesn't really depend on the connector.

>If it's only Mongo, then what does this add over what's already documented?

It's a change to DAO so it will affect all connectors.

superkhau:
@bajtos Reassigning this one to you as I'll be gone for a few weeks. It's pretty close to done.

test/basic-querying.test.js
```diff
@@ -525,31 +525,31 @@ describe('basic-querying', function() {
...
     it('should find first record (default sort by id)', function(done) {
       User.all({ order: 'id' }, function(err, users) {
-        User.findOne(function(e, u) {
-          should.not.exist(e);
-          should.exist(u);
-          u.id.toString().should.equal(users[0].id.toString());
+        User.findOne(function(err, user) {
+          should.not.exist(err);
+          should.exist(user);
+          user.id.toString().should.equal(users[0].id.toString());
```
bajtos: IIUC, this change is cleaning up the code by renaming `e, u` to `err, user`. While I see a lot of value in clean code, I see much more value in git commit history that makes code archaeology easy. Changes like these makes examination of history more difficult, as they add noise.

If you think this is still a change worth making, then please put it in a standalone commit, possibly in a new pull request.

test/basic-querying.test.js
```diff
@@ -525,31 +525,31 @@ describe('basic-querying', function() {
...
-      User.findOne({ order: 'order' }, function(e, u) {
-        should.not.exist(e);
-        should.exist(u);
-        u.order.should.equal(1);
-        u.name.should.equal('Paul McCartney');
+      User.findOne({ order: 'order' }, function(err, user) {
+        should.not.exist(err);
+        should.exist(user);
+        user.order.should.equal(1);
+        user.name.should.equal('Paul McCartney');
```
bajtos: ditto

test/basic-querying.test.js
```diff
@@ -525,31 +525,31 @@ describe('basic-querying', function() {
...
-      User.findOne({ order: 'order DESC' }, function(e, u) {
-        should.not.exist(e);
-        should.exist(u);
-        u.order.should.equal(6);
-        u.name.should.equal('Ringo Starr');
+      User.findOne({ order: 'order DESC' }, function(err, user) {
+        should.not.exist(err);
+        should.exist(user);
+        user.order.should.equal(6);
+        user.name.should.equal('Ringo Starr');
```
bajtos: ditto

test/basic-querying.test.js
bajtos: IMO, it's enough to check for user's id, possibly one more property. No need to check all user properties.

test/basic-querying.test.js
superkhau: I asked @yorkie to change it to `err/user` earlier because I didn't realize it was already named `e/u`. My bad! @yorkie Can you change it back to so these irrelevant lines aren't part of this commit?

bajtos:
Closing in favour of https://github.com/strongloop/loopback-datasource-juggler/pull/1180

slnode:

slnode:

slnode:

PR #715, opened 2015-09-03
foreignKey dataLength fix

nennad:
If the PK has a definition like
```
      "mysql": {
        "dataType": "CHAR",
        "dataLength":36,
        "nullable":"N"
      }
```

dataLength is ignored when the PK is used as a foreignKey in other models


slnode:

Issue #592, opened 2015-05-15
Date attribute exists


ekussberg:
I have declared a date field in the model:

```
    "availableToDate": {
        "type": "date"
    },
```

and then tried to make a find with an or condition:

```
or: [{
          availableToDate: {gte: today}
        }, {
          availableToDate: {exists: false}
        }, {
          availableToDate: null
        }],
```

 and i get following error:

```
Error: Invalid date: [object Object]
    at DateType (/Users/edgarkuskov/Documents/Foodo Workspace/foodo-api-server/node_modules/loopback-datasource-juggler/lib/dao.js:929:11)
    at Function.DataAccessObject._coerce (/Users/edgarkuskov/Documents/Foodo Workspace/foodo-api-server/node_modules/loopback-datasource-juggler/lib/dao.js:1082:15)
    at Function.DataAccessObject._coerce (/Users/edgarkuskov/Documents/Foodo Workspace/foodo-api-server/node_modules/loopback-datasource-juggler/lib/dao.js:982:16)
    at Function.DataAccessObject._normalize (/Users/edgarkuskov/Documents/Foodo Workspace/foodo-api-server/node_modules/loopback-datasource-juggler/lib/dao.js:922:8)
    at Function.find (/Users/edgarkuskov/Documents/Foodo Workspace/foodo-api-server/node_modules/loopback-datasource-juggler/lib/dao.js:1184:10)
    at Object.eval [as forward] (eval at recompile (/usr/local/lib/node_modules/strongloop/node_modules/strong-agent/lib/proxy.js:237:15), <anonymous>:4:32)
    at Function.find (eval at wrap (/usr/local/lib/node_modules/strongloop/node_modules/strong-agent/lib/proxy.js:193:20), <anonymous>:3:21)
    at Function.Product.findByCategory (/Users/edgarkuskov/Documents/Foodo Workspace/foodo-api-server/common/models/product.js:31:13)
    at SharedMethod.invoke (/Users/edgarkuskov/Documents/Foodo Workspace/foodo-api-server/node_modules/loopback/node_modules/strong-remoting/lib/shared-method.js:207:25)
    at HttpContext.invoke (/Users/edgarkuskov/Documents/Foodo Workspace/foodo-api-server/node_modules/loopback/node_modules/strong-remoting/lib/http-context.js:335:12)
```

Is this a bug in the validation?

raymondfeng:
[...]

PR #718, opened 2015-09-14
`Validatable.validatesDateOf()` replacing exception thrown for invalid dates.

hotaru355:
Please review this PR that fixes issue #592, but also takes a broader approach on how loopback handles invalid dates. 
- _Current behavior_ - Making a `POST` request or instantiating a model with the following data throws an exception:

```
{
  "createdAt": "2015-13-13"
}
```

I would like to propose changing this behavior since it allows any client to crash the server with invalid data. Unfortunately, trying to catch the error on the server is not possible by adding custom validation methods, since parsing the data into a model (and throwing the exception) happens before any validation method is called. The only feasible solution I see for this problem is to implement a `before save` hook on _every_ model that contains a date field or maybe implement some middleware to handle date field errors (? not tested).
- _Proposed behavior_ - Providing a `validatesDateOf()` method which is added to any date field by default could be the cleaner solution. Making a request to the REST API with an invalid date field triggers a proper validation error as a client would expect it. Instantiating a model with an invalid date would not throw an exception, but calling `isValid()` on it would return false, without having to add the validation method explicitly. Removing the default validation method can easily be done if so desired.

### DIFF
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -425,6 +425,9 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
       var requiredOptions = typeof prop.required === 'object' ? prop.required : undefined;
       ModelClass.validatesPresenceOf(propertyName, requiredOptions);
     }
+    if (DataType === Date) {
+      ModelClass.validatesDateOf(propertyName);
+    }
 
     Object.defineProperty(ModelClass.prototype, propertyName, {
       get: function () {
@@ -526,10 +529,10 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
 
 // DataType for Date
 function DateType(arg) {
-  var d = new Date(arg);
-  if (isNaN(d.getTime())) {
-    throw new Error('Invalid date: ' + arg);
+  if (arg === null) {
+    return null;
   }
+  var d = new Date(arg);
   return d;
 }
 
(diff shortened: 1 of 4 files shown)


raymondfeng:
@hotaru355 Thanks for the patch. Enforcing date value check as part of the validation is nice. What I'm wondering is what expectation a developer has against the following code:

```js
p1 = new Person({name: 'John', dob: 'X'});
```

Should we throw from the constructor as dob property has an invalid value?

dhmlau:
@hotaru355 Thanks for the PR.  Sorry that it's been a while we last looked at it.  It's good to include your changes.  Could you please rebase?  Thanks!

slnode:

slnode:

slnode:

dhmlau:
@hotaru355, I actually rebase it for you.  Could you please grant me access to your fork?  Thanks! 
https://github.com/blog/2247-improving-collaboration-with-forks

dhmlau:
@hotaru355, your changes look good to us after rebasing and changing some test cases.  Could you please grant me access to your fork (see instructions in my previous post)?  Please respond by this Friday (April 21), otherwise we'd have to unfortunately close the PR.   Thanks!

dhmlau:
I've created another PR for the changes suggested here and fixed the test cases in #1339.
Closing this.

PR #721, opened 2015-09-23
findByIds would fail when an array of 0 length was passed as its first argument

cookejames:
This is a regression introduced in 2.39.0.
process.nextTick expects a function as its argument however in this case it was being given undefined with the result of a warning 'TypeError: callback is not a function'.


slnode:

superkhau:
LGTM

PR #725, opened 2015-09-30
Ability to define normalization of undefined query

jrschumacher:
Add setting to datasource `normalizeUndefinedInQuery` to determine how
it will handle undefined values. Options:

- setNull : converts undefined to null
- throwError : throw an error on undefined value
- strip : strip the key where undefined value is found

Defualt operation is to strip the key.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -934,6 +934,27 @@ DataAccessObject.all = function () {
   return DataAccessObject.find.apply(this, arguments);
 };
 
+/**
+ * Get settings via hiarchical determiniation
+ *
+ * @param {String} key The setting key
+ */
+DataAccessObject._getSetting = function(key) {
+  // Check for settings in model
+  var m = this.definition;
+  if(m && m.settings && m.settings[key]) {
+    return m.settings[key]
+  }
+
+  // Check for settings in connector
+  var ds = this.getDataSource();
+  if(ds && ds.settings && ds.settings[key]) {
+    return ds.settings[key];
+  }
+
+  return;
+};
+
 var operators = {
   gt: '>',
   gte: '>=',
@@ -955,6 +976,7 @@ var operators = {
  * @private
  */
 DataAccessObject._normalize = function (filter) {
+
   if (!filter) {
     return undefined;
   }
@@ -1032,9 +1054,8 @@ DataAccessObject._normalize = function (filter) {
       Object.keys(this.definition.properties), this.settings.strict);
   }
 
-  var handleUndefined = this.getConnector().settings.normalizeUndefinedInQuery;
+  var handleUndefined =  this._getSetting('normalizeUndefinedInQuery');
   // alter configuration of how removeUndefined handles undefined values
-  // @TODO should refactor `removeUndefined` to something like `handleUndefined`
   filter = removeUndefined(filter, handleUndefined);
   this._coerce(filter.where);
   return filter;
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1032,7 +1032,10 @@ DataAccessObject._normalize = function (filter) {
       Object.keys(this.definition.properties), this.settings.strict);
   }
 
-  filter = removeUndefined(filter);
+  var handleUndefined = this.getConnector().settings.normalizeUndefinedInQuery;
+  // alter configuration of how removeUndefined handles undefined values
+  // @TODO should refactor `removeUndefined` to something like `handleUndefined`
+  filter = removeUndefined(filter, handleUndefined);
   this._coerce(filter.where);
   return filter;
 };
--- a/test/util.test.js
+++ b/test/util.test.js
@@ -68,15 +68,10 @@ describe('util.removeUndefined', function () {
 
     // test handling of undefined
     var q5 = {where: {x: 1, y: undefined}};
-    should.deepEqual(removeUndefined(q5, 'setNull'), {where: {x: 1, y: null}});
+    should.deepEqual(removeUndefined(q5, 'nullify'), {where: {x: 1, y: null}});
 
     var q6 = {where: {x: 1, y: undefined}};
-    try {
-      removeUndefined(q6, 'throwError');
-    }
-    catch(err) {
-      err.should.be.instanceof(Error);
-    }
+    (function(){ removeUndefined(q6, 'throw') }).should.throw(/`undefined` in query/);
 
   });
 });
--- a/test/util.test.js
+++ b/test/util.test.js
@@ -66,6 +66,18 @@ describe('util.removeUndefined', function () {
     var q4 = {where: {x: 1, y: date}};
     should.deepEqual(removeUndefined(q4), {where: {x: 1, y: date}});
 
+    // test handling of undefined
+    var q5 = {where: {x: 1, y: undefined}};
+    should.deepEqual(removeUndefined(q5, 'setNull'), {where: {x: 1, y: null}});
+
+    var q6 = {where: {x: 1, y: undefined}};
+    try {
+      removeUndefined(q6, 'throwError');
+    }
+    catch(err) {
+      err.should.be.instanceof(Error);
+    }
+
   });
 });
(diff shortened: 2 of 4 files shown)


slnode:

bajtos:
@slnode ok to test

bajtos:
Hi @jrschumacher, thank you for the patch!

I would personally prefer different option names.

1. IMO we should rename `throwError` to `throw` to make it consistent with the existing model option `strict` which allows the value `throw` - see https://github.com/strongloop/loopback-datasource-juggler/blob/c4de830bfe4e4a4aa704b467c4d6c037665bb091/lib/model.js#L227-L228.
2. I'd prefer `nullify` over `setNull`, but I admit this is subjective.
3. As for `strip`, I'd rather use `remove` or `delete`, but again, this is subjective too.

Few more ideas to consider:
 - Should we make this option configurable at Model level too? 
 - Should we take into account Model option `persistUndefinedAsNull` (#542) and set `handleUndefined='setNull'` when `normalizeUndefinedInQuery` is not explicitly set?

I'm fine to leave those two out of scope of this PR though.

bajtos:
@raymondfeng since you are more familiar with the issue being fixed, I'd like to leave review & approval for you to make.

jrschumacher:
@bajtos I'm happy to change the naming. Will fix later today also I'll look into exploring the model level settings. Questions regarding that: if you set data source level should model level overwrite? Is there an order of precedence?

jrschumacher:
@bajtos regarding the persistUndefinedToNull would it be better to change the datasource setting name to be similar in naming? Or does persist refer more to the longevity of the data in the datasource rather than the transient query?

raymondfeng:
Here is my take:

1. Make the configuration available at multiple levels (preferably app & datasource level)
  1. app (server/config.json, global to all models within the app)
  2. datasource (connector settings, global to all models within the same ds. I'm not sure if it's necessary)
  3. model (model settings, local to the model)
  4. `options` for model methods (local to the request. I'm not sure if it's necessary)

2. Name/value of the config property:
  - normalizeUndefinedInQuery
    - throw
    - nullify
    - skip (or ignore)

jrschumacher:
@raymondfeng got it

jrschumacher:
@raymondfeng I think the connector level will be valuable to MongoDB users. So maybe if you have a RDBM or redis datasource you will want the stripping effect but with MongoDB the stripping created some serious issues.

bajtos:
:+1:  for @raymondfeng proposal.

> if you set data source level should model level overwrite? Is there an order of precedence?

Yes, more-specific level should override less-specific level. I.e. model level overrides datasource level.

> regarding the persistUndefinedToNull would it be better to change the datasource setting name to be similar in naming? Or does persist refer more to the longevity of the data in the datasource rather than the transient query?

You name is fine. `persistUndefinedAsNull` is a different thing - it controls how well-known fields with no value are persisted.

jrschumacher:
@raymondfeng @bajtos could I get some help on how to fetch the app settings? I've created a method to get settings based on priority but need to add app settings: https://github.com/strongloop/loopback-datasource-juggler/pull/725/files#diff-d717a01e40d8164976e4468af1bce663R942

raymondfeng:
this.app.get('propName') should work

jrschumacher:
@raymondfeng I've done this, but I don't know how to add the tests for this. There doesn't seem to be a global `app` object.

bajtos:
I'd rather not use `app` in the juggler, it's a concept from loopback. Since loopback-datasource-juggler is a dependee of loopback, it should not access loopback-specific properties, as that would break encapsulation and isolation of this component. Instead, loopback should inject `normalizeUndefinedInQuery` when attaching a model to an app - see https://github.com/strongloop/loopback/blob/e68cd8fc87730404a78e1475b828886e3582bd6f/lib/application.js#L156-L159 and https://github.com/strongloop/loopback/blob/e68cd8fc87730404a78e1475b828886e3582bd6f/lib/application.js#L156-L159

bajtos:
Another approach is to override datasource settings when creating a new datasource in an app - see https://github.com/strongloop/loopback/blob/e68cd8fc87730404a78e1475b828886e3582bd6f/lib/application.js#L221-L227

jrschumacher:
@bajtos then is there anything else that's needed from me here? I'm not sure why the above tests fail. Locally all the tests pass.

bajtos:
Hmm, the tests are failing with this error:

```
  1) util.removeUndefined Remove undefined values from the query object:

      AssertionError: expected { where: { x: 1 } } deepEqual { where: { x: 1, y: null } }
      + expected - actual

       {
         "where": {
           "x": 1
      +    "y": [null]
         }
       }
      
      at Context.<anonymous> (test/util.test.js:71:12)
```

Could you please rebase your feature branch on top of the current master and run the tests locally again?

lib/dao.js
```diff
@@ -934,6 +934,29 @@ DataAccessObject.all = function () {
...
+    return m.settings[key]
+  }
+
+  // Check for settings in connector
+  var ds = this.getDataSource();
+  if(ds && ds.settings && ds.settings[key]) {
+    return ds.settings[key];
+  }
+
+  // @TODO: Check for settings in app
```
bajtos: Please remove per the conversation above.

test/util.test.js
```diff
@@ -66,6 +66,18 @@ describe('util.removeUndefined', function () {
...
 
+    // test handling of undefined
+    var q5 = {where: {x: 1, y: undefined}};
+    should.deepEqual(removeUndefined(q5, 'setNull'), {where: {x: 1, y: null}});
+
+    var q6 = {where: {x: 1, y: undefined}};
+    try {
+      removeUndefined(q6, 'throwError');
+    }
+    catch(err) {
```
bajtos: Coding style: please put catch on the same line as the previous closing brace

```
try {
  // ...
} catch {
  // ...
}
```

jrschumacher:
@bajtos yep I'll rebase. 

Regarding the settings. Shall I still get model and datasource but do it in the normalize method?

test/util.test.js
bajtos: BTW the canonical way of testing this is to use `should.throw`, see [should.js docs](https://github.com/tj/should.js/#throw-and-throwerror). It's also better to assert that the error message thrown loosely contains the expected error message.

```js
(function(){ removeUndefined(q6, 'throwError') }).should.throw(/`undefined` in query/);
```

bajtos:
> Shall I still get model and datasource but do it in the normalize method?

Your current approach LGTM, I left few mostly stylistic comments to address.

lib/dao.js
jrschumacher: @bajtos am I just removing the TODO or all of it?

bajtos: Just the TODO. It would be nice if you could send a patch to LoopBack which will apply app settings to newly created datasources, but that's not necessary to get this patch landed.

jrschumacher:
@bajtos okay I think I handled all the issues you mentioned.

jrschumacher:
> It would be nice if you could send a patch to LoopBack which will apply app settings to newly created datasources, but that's not necessary to get this patch landed.

I'll consider this, but its not going to be anytime soon. Also I took note of the references you mentioned, but I have to say I'm a bit at a loss exactly what you're asking me to do. Is this specific to this setting or is it for all settings?

bajtos:
Landed via 98fb3c6, thank you for the contribution!

> Also I took note of the references you mentioned, but I have to say I'm a bit at a loss exactly what you're asking me to do. Is this specific to this setting or is it for all settings?

I'd modify [app.datasource](https://github.com/strongloop/loopback/blob/e68cd8fc87730404a78e1475b828886e3582bd6f/lib/application.js#L221-L227) to apply this specific setting only.

bajtos:
Released as `loopback-datasource-juggler@2.41.0`

jrschumacher:
Glad to help!

PR #732, opened 2015-10-08
Fix coercion

gausie:
The coercion function seemed to have some behaviours that didn't make sense:
- The way that the reference is used mutably is confusing in some situations.
- Coercion was abandoned after the use of a logical operator such as 'and', due to a return statement being used erroneously.

As a result of these inconsistencies, when I used a filter with an "and" in it on a query that also contained a foreign key, the `_coerce` function would be abandoned before the foreign key could be coerced to an ObjectId. This resulted in a query with a string, returning an empty result and breaking our queries!

This solution fixes that edge case, while also passing all tests. I'm not extremely happy with my efforts in 4f53ccf to meet tests, but I am at least content!

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1061,7 +1061,7 @@ DataAccessObject._normalize = function (filter) {
   var handleUndefined =  this._getSetting('normalizeUndefinedInQuery');
   // alter configuration of how removeUndefined handles undefined values
   filter = removeUndefined(filter, handleUndefined);
-  this._coerce(filter.where);
+  filter.where = this._coerce(filter.where);
   return filter;
 };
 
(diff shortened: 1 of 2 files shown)


slnode:

raymondfeng:
@gausie Thanks for the patch. I believe that coercion is done in place. As a result, we don't have to use the return value. Do you have a test that fails before and passes with your PR?

gausie:
@raymondfeng No problem! The test added in commit 89209da failed before because the string '10' is not cast to an integer.

gausie:
@raymondfeng Have you managed to take a look at this?

gausie:
@raymondfeng Any luck looking at this?

slnode:

gausie:
@raymondfeng It's all here with tests to indicate...

slnode:

gausie:
Well never mind then I guess

Issue #720, opened 2015-09-23
Filtering relations of a model with an order doesn't seem to work


mdartic:
Let's have a collection 'state' which references many 'country'.

If I want to order countries by name, for one state, I will make a request like this :

    api/states/:id/countries?filter[order]=name

But, I'm going to have a list ordered by country's id, not by name.

I've searched a little in the code, and, when I do this request, I go through :

    ReferencesMany.prototype.related 

https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/relation-definition.js#L2934-L2968

then

    DataAccessObject.findByIds 

https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L885-L924

And, if we observe line 921, we do a sort if err is undefined... by ids. Well.

    this.find(filter, options, function(err, results) {
        cb(err, err ? results : utils.sortObjectsByIds(pk, ids, results));
    });

I don't really understand why, but the fact is if I try to remove the sort with this line :

        cb(err, results);

well, it seems to work.

We could simplify more by just passing 'cb' to this.find like this :

    this.find(filter, options, cb);

I've fixed locally the code like this, it seems to work for me.

But not sure it's good for everyone.

Could you help me and say if my implementation seems right ? If so, I'll try to make a PR.

Thanks

(probably related to https://github.com/strongloop/loopback-datasource-juggler/issues/166)

superkhau:
@raymondfeng Any suggestion here?

mdartic:
@raymondfeng @superkhau Do you want me I make a PR ?

superkhau:
@mdartic Yes please.

mdartic:
I've tried to write the code to fix it.
Implement 2 tests, and modify dao.js to not filter if already specified in the filter (query exactly) parameter.
Hope it will help.

jannyHou:
Thank you @mdartic I am working on issue https://github.com/strongloop/loopback-datasource-juggler/issues/166 to fix it.
To keep issues easy to track let's move discussion under that issue, I also put a workaround in https://github.com/strongloop/loopback-datasource-juggler/issues/166#issuecomment-213163882

Closing it.

mdartic:
thanks !

PR #739, opened 2015-10-16
Fix filtering relations of a model with an order specified

mdartic:
PR for #720


slnode:

PR #747, opened 2015-10-31
Fix search statements becomes "[object Object]"

yshing:
fix issues nlike,inq,nin,between,regexp VALUE becomes "[object object]"

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1179,7 +1179,7 @@ DataAccessObject._coerce = function (where) {
     // Check there is an operator
     var operator = null;
     var exp = val;
-    if (val.constructor === Object) {
+    if (typeof val==='object') {
       for (var op in operators) {
         if (op in val) {
           val = val[op];


slnode:

superkhau:
@yshing Can you add a test to verify your change?

yshing:
I encounter this issue when testing queries for Model.find() using ```slc debug``` without `deactivate breakpoints`. 
image

When `deactivate breakpoint` is ON, the query works fine..

There are one cause of this issue: 
1. Constructor of the where cause is not Object.
Which happen in 2 cases:
1. Default settings when using ```slc debug```  **My case (should be quite common...)
2. Some uses a custom constructor function to make the filter query sth like following or messing up with the `Object` object(very edge case):
```js
var likeFilter = new LikeQueryConstructor( someConditions );
Model.find({
    where:{
        property: likeFilter
    }
}, callback)
```
The code will run without any warning or something, but

PR #752, opened 2015-11-04
make automatic validation optional

### LINKED ISSUE #744: Is it possible to determine when to use validation, and not automatically call it from create or update?
I'm having a model that has to be validated only when it's getting published (via custom remote method).
When I use loopbacks built-in validation, isValid also get's executed on create and save.
I just want it to be executed in my remote method only

I tried first with validateUpsert-option set to false, and then use upsert for every creation, update-operation. That worked fine, but now I have to set my ACL's and in an upsert operation I cannot detect the owner of the model instance, so I have to use the other methods again. 

I think it would be better if the creation of the validation rules could be seperated from the execution.

Is there a quick solution for this?

ernie58:
make automatic validation optional via CRUD-options or via setting in model definition

@bajtos , @superkhau What should I do with the validateUpsert-option? Which one needs precedence? Now, when explicitly set to false, it has precedence on the new flags, which makes sense IMO

Close #744

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -257,6 +257,16 @@ DataAccessObject.create = function (data, options, cb) {
 
     data = obj.toObject(true);
 
+    // options has precedence on model-setting
+    if (options.validate === false) {
+      return create();
+    }
+
+    // only when options.validate is not set, take model-setting into consideration
+    if (options.validate === undefined && Model.settings.automaticValidation === false) {
+      return create();
+    }
+    
     // validation required
     obj.isValid(function (valid) {
       if (valid) {
@@ -457,12 +467,25 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
 
         var connector = self.getConnector();
 
-        if (Model.settings.validateUpsert === false) {
+        var doValidate = undefined;
+        if (options.validate === undefined) {
+          if (Model.settings.validateUpsert === undefined) {
+            if (Model.settings.automaticValidation !== undefined) {
+              doValidate = Model.settings.automaticValidation;
+            }
+          } else {
+            doValidate = Model.settings.validateUpsert
+          }
+        } else {
+          doValidate = options.validate;
+        }
+
+        if (doValidate === false) {
           callConnector();
         } else {
           inst.isValid(function(valid) {
             if (!valid) {
-              if (Model.settings.validateUpsert) {
+              if (doValidate) { // backwards compatibility with validateUpsert:undefined
                 return cb(new ValidationError(inst), inst);
               } else {
                 // TODO(bajtos) Remove validateUpsert:undefined in v3.0
@@ -731,6 +754,16 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
 
         var obj = ctx.instance;
         var data = obj.toObject(true);
+
+        // options has precedence on model-setting
+        if (options.validate === false) {
+          _findOrCreate(query, data, obj);
+        }
+
+        // only when options.validate is not set, take model-setting into consideration
+        if (options.validate === undefined && Model.settings.automaticValidation === false) {
+          _findOrCreate(query, data, obj);
+        }
 
         // validation required
         obj.isValid(function (valid) {
@@ -1876,8 +1909,13 @@ DataAccessObject.prototype.save = function (options, cb) {
   var hookState = {};
 
   if (options.validate === undefined) {
-    options.validate = true;
+    if (Model.settings.automaticValidation === undefined) {
+      options.validate = true;
+    } else {
+      options.validate = Model.settings.automaticValidation;
+    }
   }
+
   if (options.throws === undefined) {
     options.throws = false;
   }
@@ -2422,15 +2460,32 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
       data = removeUndefined(result);
     }
 
+    var doValidate = true;
+    if (options.validate === undefined) {
+      if (Model.settings.automaticValidation !== undefined) {
+        doValidate = Model.settings.automaticValidation;
+      }
+    } else {
+      doValidate = options.validate;
+    }
+
     // update instance's properties
     inst.setAttributes(data);
 
-    inst.isValid(function (valid) {
-      if (!valid) {
-        cb(new ValidationError(inst), inst);
-        return;
-      }
+    if (doValidate){
+      inst.isValid(function (valid) {
+        if (!valid) {
+          cb(new ValidationError(inst), inst);
+          return;
+        }
 
+        triggerSave();
+      }, data);
+    } else {
+        triggerSave();
+    }
+
+    function triggerSave(){
       inst.trigger('save', function (saveDone) {
         inst.trigger('update', function (done) {
           var typedData = {};
@@ -2506,7 +2561,7 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
           });
         }, data, cb);
       }, data, cb);
-    }, data);
+    }
   });
 return cb.promise;
 };
--- a/test/optional-validation.test.js
+++ b/test/optional-validation.test.js
@@ -1,0 +1,514 @@
+// This test written in mocha+should.js
+var should = require('./init.js');
+var db, User, options, whereCount = 0;
+var j = require('../');
+var ValidationError = j.ValidationError;
+
+var INITIAL_NAME = 'Bert';
+var NEW_NAME = 'Ernie';
+var INVALID_DATA = {name: null};
+var VALID_DATA = {name: INITIAL_NAME};
+
+describe('optional-validation', function () {
+
+  before(function (done) {
+    db = getSchema();
+    User = db.define('User', {
+      seq: {type: Number, index: true},
+      name: {type: String, index: true, sort: true},
+      email: {type: String, index: true},
+      birthday: {type: Date, index: true},
+      role: {type: String, index: true},
+      order: {type: Number, index: true, sort: true},
+      vip: {type: Boolean}
+    }, { forceId: true, strict: true });
+
+    db.automigrate(['User'], done);
+
+  });
+
+  beforeEach(function (done) {
+    User.destroyAll(function () {
+      delete User.validations;
+      User.validatesPresenceOf('name');
+      done();
+    });
+  });
+
+  function expectValidationError(done) {
+    return function (err, result) {
+      should.exist(err);
+      err.should.be.instanceOf(Error);
+      err.should.be.instanceOf(ValidationError);
+      done();
+    };
+  }
+
+  function expectCreateSuccess(data, done) {
+    if (done === undefined && typeof data === 'function') {
+      done = data;
+      data = { name: INITIAL_NAME };
+    }
+    return function(err, instance) {
+      if (err) return done(err);
+      instance.should.be.instanceOf(User);
+      if (data.name) {
+        instance.name.should.eql(data.name || INITIAL_NAME);
+      } else {
+        should.not.exist(instance.name);
+      }
+      done();
+    };
+  }
+
+  function expectChangeSuccess(data, done) {
+    if (done === undefined && typeof data === 'function') {
+      done = data;
+      data = { name: NEW_NAME };
+    }
+    return function(err, instance) {
+      if (err) return done(err);
+      instance.should.be.instanceOf(User);
+      if (data.name) {
+        instance.name.should.eql(data.name || NEW_NAME);
+      } else {
+        should.not.exist(instance.name);
+      }
+      done();
+    };
+  }
+
+  function createUserAndChangeName(name, cb) {
+    User.create(VALID_DATA, {validate: true}, function (err, d) {
+      d.name = name;
+      cb(err, d);
+    });
+  }
+
+  function createUser(cb) {
+    User.create(VALID_DATA, {validate: true}, cb);
+  }
+
+  function callUpdateOrCreateWithExistingUserId(name, options, cb){
+    User.create({'name': 'Groover'}, function(err, user){
+      if (err) return cb(err);
+      var data = {name: name};
+      data.id = user.id;
+      User.updateOrCreate(data, options, cb);
+    });
+  }
+
+  function getNewWhere() {
+    return {name: 'DoesNotExist' + (whereCount++)};
+  }
+
+  describe('no model setting', function () {
+
+    describe('method create', function() {
+      it('should throw on create with validate:true with invalid data', function (done) {
+        User.create(INVALID_DATA, {validate: true}, expectValidationError(done));
+      });
+
+      it('should NOT throw on create with validate:false with invalid data', function (done) {
+        User.create(INVALID_DATA, {validate: false}, expectCreateSuccess(INVALID_DATA, done));
+      });
+
+      it('should NOT throw on create with validate:true with valid data', function (done) {
+        User.create(VALID_DATA, {validate: true}, expectCreateSuccess(done));
+      });
+
+      it('should NOT throw on create with validate:false with valid data', function (done) {
+        User.create(VALID_DATA, {validate: false}, expectCreateSuccess(done));
+      });
+
+      it('should throw on create with invalid data', function (done) {
+        User.create(INVALID_DATA, expectValidationError(done));
+      });
+
+      it('should NOT throw on create with valid data', function (done) {
+        User.create(VALID_DATA, expectCreateSuccess(done));
+      });
+    });
+
+    describe('method findOrCreate', function() {
+      it('should throw on findOrCreate with validate:true with invalid data', function (done) {
+        User.findOrCreate(getNewWhere(), INVALID_DATA, {validate: true}, expectValidationError(done));
+      });
+
+      it('should NOT throw on findOrCreate with validate:false with invalid data', function (done) {
+        User.findOrCreate(getNewWhere(), INVALID_DATA, {validate: false}, expectCreateSuccess(INVALID_DATA, done));
+      });
+
+      it('should NOT throw on findOrCreate with validate:true with valid data', function (done) {
+        User.findOrCreate(getNewWhere(), VALID_DATA, {validate: true}, expectCreateSuccess(done));
+      });
+
+      it('should NOT throw on findOrCreate with validate:false with valid data', function (done) {
+        User.findOrCreate(getNewWhere(), VALID_DATA, {validate: false}, expectCreateSuccess(done));
+      });
+
+      it('should throw on findOrCreate with invalid data', function (done) {
+        User.findOrCreate(getNewWhere(), INVALID_DATA, expectValidationError(done));
+      });
+
+      it('should NOT throw on findOrCreate with valid data', function (done) {
+        User.findOrCreate(getNewWhere(), VALID_DATA, expectCreateSuccess(done));
+      });
+    });
+
+    describe('method updateOrCreate on existing data', function() {
+      it('should throw on updateOrCreate(id) with validate:true with invalid data', function (done) {
+        callUpdateOrCreateWithExistingUserId(null, {validate: true}, expectValidationError(done));
+      });
+
+      it('should NOT throw on updateOrCreate(id) with validate:false with invalid data', function (done) {
+        callUpdateOrCreateWithExistingUserId(null, {validate: false}, expectChangeSuccess(INVALID_DATA, done));
+      });
+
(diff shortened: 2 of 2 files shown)


slnode:

ernie58:
pull request for #744

lib/dao.js
```diff
@@ -257,6 +257,16 @@ DataAccessObject.create = function (data, options, cb) {
 
     data = obj.toObject(true);
 
+    //options has precedence on model-setting
+    if(options.disableValidation === true){
```
bajtos: Coding style:

```js
if (options.disableValidation === true) {
```

Please apply this style to the rest of your changes too.

bajtos:
I am proposing to rename `options.disableValidation` to `options.validation` and `disableAutomaticValidation` to `automaticValidation`. [Double negative](https://en.wikipedia.org/wiki/Double_negative) like `disableValidation:false` tends to be confusing for readers.

bajtos:
> What should I do with the validateUpsert-option? Which one needs precedence? Now, when explicitly set to false, it has precedence on the new flags, which makes sense IMO.

IMO, it's ok for `validateUpsert` to take precedence over `automaticValidation`. However, `options.validate` should have always the highest precedence.

bajtos:
@raymondfeng would you like to take a look too?

test/optional-validation.test.js
```diff
@@ -0,0 +1,504 @@
...
+  beforeEach(function (done) {
+    User.destroyAll(function () {
+      delete User.validations;
+      done();
+    });
+  });
+
+  describe('save with validation errors and options.disableValidation, no model setting', function () {
+
+    it('should throw validation-errors on create(data, {disableValidation: false}, cb) with invalid data', function (done) {
```
bajtos: Please shorten test case names, e.g. `it('should throw on create when options.disableValidation=false`)`

test/optional-validation.test.js
```diff
@@ -0,0 +1,504 @@
...
+  });
+
+  beforeEach(function (done) {
+    User.destroyAll(function () {
+      delete User.validations;
+      done();
+    });
+  });
+
+  describe('save with validation errors and options.disableValidation, no model setting', function () {
```
bajtos: This description is incorrect, one of the tests are actually passing a valid instance:

```
should NOT throw validation-errors on create(data, {disableValidation: true}, cb) with valid dat
```

bajtos:
@slnode ok to test

lib/dao.js
ernie58: I thought I was doing this? Could you tell me where I don't use this style? Thx

ernie58:
I will take out the double negation, and change and shorten the descriptions in the test cases

ernie58:
@bajtos I'm confused about the `options.validate` - flag. I thought it was set by `validateUpsert`, but in your comment you state that `options.validate`has precedence on `validateUpsert`. So it's used elsewhere then? Why can't I use directly `options.validate` instead of creating another option `options.validation`

slnode:

ernie58:
Sorry, closed it by accident

lib/dao.js
bajtos: Check the spaces - add a space after `if` and another in between `) {`.

bajtos:
> I'm confused about the options.validate - flag. I thought it was set by validateUpsert, but in your comment you state that options.validate has precedence on validateUpsert. So it's used elsewhere then? Why can't I use directly options.validate instead of creating another option options.validation

Actualy, AFAICT, `prototype.save` already supports `options.validate` flag - see [here](https://github.com/strongloop/loopback-datasource-juggler/blob/5ca9c47f440ab999c4dfe4a323a3d72991fd9c9f/lib/dao.js#L1878-L1880) and [here](https://github.com/strongloop/loopback-datasource-juggler/blob/5ca9c47f440ab999c4dfe4a323a3d72991fd9c9f/lib/dao.js#L1902-L1905). Which is yet another argument why to use this name instead of adding `options.disableValidation`.

Yes, you can and should use `options.validate` directly.

I believe `options.validate` is unrelated to model-level setting `validateUpsert` at the moment. The changes you are proposing will change that and I think that any option passed to a method call should always overwrite model-level settings.

ernie58:
@bajtos I've added a commit, could you check again? Thx!

ernie58:
@bajtos did you have a chance to look into the changes?

lib/dao.js
```diff
@@ -1899,7 +1918,6 @@ DataAccessObject.prototype.save = function (options, cb) {
     Model.applyProperties(data, inst);
     inst.setAttributes(data);
 
-    // validate first
```
bajtos: I don't see why this comment is no longer relevant? Would you mind putting it back again?

test/optional-validation.test.js
```diff
@@ -0,0 +1,504 @@
...
+        done();
+      });
+    });
+
+    it('should NOT throw on create with validate:false with invalid data', function (done) {
+      var options = {validate: false};
+      User.validatesPresenceOf('name');
+
+      User.create({name: null}, options, function (e, d) {
+        should.not.exist(e);
```
bajtos: Please use `if (err) return done(err)` - it gives more helpful stack traces on error. Also please apply the same changes to other places where you call `should.not.exist(e)`.

lib/dao.js
```diff
@@ -1876,8 +1886,17 @@ DataAccessObject.prototype.save = function (options, cb) {
   var hookState = {};
 
   if (options.validate === undefined) {
-    options.validate = true;
+    if (Model.settings.validateUpsert === undefined) {
```
bajtos: I believe `PersistedModel.prototype.save` is not affected by `validateUpsert` setting. I think this block of code should belong to `PersistedModel.prototype.updateAttributes`, which was not changed to support `options.validate` yet?

bajtos:
AFAICT, this patch adds the new option/setting only to `create()` and `prototype.save()` methods. How about other CRUD methods like `updateOrCreate()`, `findOrCreate()`, `prototype.updateAttributes()`?

I don't mind landing a change that adds `options.validate` to a subset of CRUD methods.

However, I am worried that if we land model-level `settings.automaticValidation` that supports only a subset of CRUD methods, then users will be confused.

@ernie58 thoughts?

test/optional-validation.test.js
```diff
@@ -0,0 +1,504 @@
...
+      delete User.validations;
+      done();
+    });
+  });
+
+  describe('no model setting', function () {
+
+    it('should throw on create with validate:true with invalid data', function (done) {
+      var options = {validate: true};
+      User.validatesPresenceOf('name');
```
bajtos: Please move this line to `beforeEach` and/or top-level model setup code, so that it is not duplicated in every test. Unless I missed something?

bajtos:
BTW I appreciate the extensive test coverage! 

I'm wondering how to refactor the code to reduce duplication. For example:

```js
    it('should throw on create with validate:true with invalid data', function (done) {
      var options = {validate: true};
      User.create(INVALID_DATA, options, expectValidationError(done));
    });

    it('should NOT throw on create with validate:true with valid data', function (done) {
      var options = {validate: true};
      User.create(VALID_DATA, options, expectSuccess(done));
    });

   it('should NOT throw on create with validate:false with valid data', function (done) {
      var options = {validate: false};
      User.create(VALID_DATA, options, expectSuccess(done));
    });

   // etc ...

function expectSuccess(done) {
  return function(err, result) {
    // make the assertions
  };
}
```

ernie58:
@bajtos I thought those extra methods used save internally, because in #744 you only mentioned create and save. I'll add the other methods, that will be better, it's not a big effort

ernie58:
@bajtos I agree on the code duplication, i'll look into it. It started small but with the double negation and the model setting, it expanded quick. I'll clean it up

bajtos:
>  I thought those extra methods used save internally, because in #744 you only mentioned create and save. I'll add the other methods, that will be better, it's not a big effort

Sorry for the confusion. It's great to hear that you don't mind adding support to other methods too! :+1: 

> I agree on the code duplication, i'll look into it. It started small but with the double negation and the model setting, it expanded quick. I'll clean it up

Awesome :)

ernie58:
@bajtos I refactored the testing and added the optional validation to `updateAttributes`, `updateOrCreate` and `findOrCreate`. I still have some question, but I'll try to post them inline

lib/dao.js
```diff
@@ -457,12 +467,27 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
...
+          doValidate = options.validate;
+        }
+
+        if (doValidate === false) {
           callConnector();
         } else {
           inst.isValid(function(valid) {
             if (!valid) {
-              if (Model.settings.validateUpsert) {
+              if (doValidate) { //backwards compatibility with validateUpsert:undefined
```
ernie58: I've used the some logic as validateUpsert. When options and settings are undefined I default to false instead of true for automatic validation, only in updateOrCreate, to stay backwards compatible.

I also use a local variable `doValidate` instead of setting  `options.validate`, because some test required the options to stay the same during the lifecycle. I did that in `save`, but the tests completed successfully even when changing the `options`. Should it be changed there as well?

lib/dao.js
```diff
@@ -2426,7 +2470,19 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
     inst.setAttributes(data);
 
     inst.isValid(function (valid) {
-      if (!valid) {
+
```
ernie58: At first I did not call the `isValid`-function, when no validation was needed. But then some test concerning the lifecycle hooks failed, because the `beforeValidation` and `afterValidation` hooks where missing. I was not sure if I could modify those, so instead I always validate, and depending on the settings and options I throw a ValidationError or I don't

ernie58:
@bajtos I checked the docs for other CRUD functions, I don't know if any of these apply for this patch?
  * bulkUpdate
  * createUpdates
  * updateAll

bajtos:
> I checked the docs for other CRUD functions, I don't know if any of these apply for this patch?

> * bulkUpdate
> * createUpdates
> * updateAll

These don't apply for this patch.

lib/dao.js
```diff
@@ -457,12 +467,27 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
 
         var connector = self.getConnector();
 
-        if (Model.settings.validateUpsert === false) {
+        var doValidate = undefined;
+        if (options.validate === undefined) {
+          if (Model.settings.validateUpsert === undefined) {
+            if (Model.settings.automaticValidation === undefined) {
+
+            } else {
```
bajtos: Please remove the empty block.

```js
if (Model.settings.automaticValidation !== undefined) {
  doValidate = Model.settings.automaticValidation;
}
```

lib/dao.js
bajtos: > I've used the some logic as validateUpsert. When options and settings are undefined I default to false instead of true for automatic validation, only in updateOrCreate, to stay backwards compatible.

Great, that's the right thing to do here :+1:

bajtos: > I also use a local variable doValidate instead of setting options.validate, because some test required the options to stay the same during the lifecycle. I did that in save, but the tests completed successfully even when changing the options. Should it be changed there as well?

It's better to treat input arguments as immutable and use a local variable (e.g. `doValidate`) everywhere.

lib/dao.js
```diff
@@ -1876,8 +1911,13 @@ DataAccessObject.prototype.save = function (options, cb) {
   var hookState = {};
 
   if (options.validate === undefined) {
-    options.validate = true;
+    if (Model.settings.automaticValidation === undefined) {
+      options.validate = true;
+    } else {
+      options.validate = Model.settings.automaticValidation;
+    }
```
bajtos: Please rewrite this to use a local variable as we discussed above.

test/optional-validation.test.js
```diff
@@ -0,0 +1,575 @@
...
+    }
+  }
+
+  function expectCreateSuccess(done, incompleteData) {
+    if (incompleteData === undefined) incompleteData = false;
+    return function (err, d) {
+      if (err) return done(err);
+      d.should.be.instanceOf(User);
+      if (!incompleteData) {
+        d.name.should.eql(OLD_NAME);
```
bajtos: I find this suspicious. The callback receives an updated model instance, so I would expect the name to equal to the new name? Unless `OLD_NAME` is referring to the value used for `create()` call, in which case a name like `INITIAL_NAME` would be probably better?

bajtos:
@ernie58 thank you for the update, you are making a good progress here!

lib/dao.js
bajtos: > But then some tests concerning the lifecycle hooks failed, because the beforeValidation and afterValidation hooks where missing. 

Hmm, that's interesting and and rather unexpected. Could you please provide more details on the test failures? I am surprised that an existing test fails when this change should be fully backwards compatible.

lib/dao.js
ernie58: Now that I'm trying: the old code also modified the options-object. When I use a local variable, some tests fail, because they expect options==true when it's undefined

lib/dao.js
ernie58: guess i did something wrong here, changed it again, but now it works. isValid is now only called when needed

ernie58:
@bajtos I did all the changes, except one (in the save-method) see my comment above.
Could you check again? Thx

ernie58:
Hi @bajtos , could you give an estimate when you could look again at the changes, and when the pull request could be available in the main repository? Just so I can schedule some things. Thx!

bajtos:
@slnode ok to test

lib/dao.js
bajtos: I'll still prefer to not modify function arguments (`options`), but since the current code is already modifying it, this patch is not making the matter any worse. Let's leave refactoring to a local variable out of scope of this PR.

test/optional-validation.test.js
```diff
@@ -0,0 +1,595 @@
...
+        seq: {type: Number, index: true},
+        name: {type: String, index: true, sort: true},
+        email: {type: String, index: true},
+        birthday: {type: Date, index: true},
+        role: {type: String, index: true},
+        order: {type: Number, index: true, sort: true},
+        vip: {type: Boolean}
+      }, { forceId: true, strict: true });
+
+    filter = {fields: ['name', 'id']};
```
bajtos: AFAICT, this is an unused variable.

test/optional-validation.test.js
```diff
@@ -0,0 +1,595 @@
...
+      });
+
+      it('should NOT throw on findOrCreate with valid data', function (done) {
+        User.findOrCreate(getNewWhere(), VALID_DATA, expectCreateSuccess(done));
+      });
+    });
+
+    describe('method updateOrCreate with new data', function() {
+      it('should throw on updateOrCreate with validate:true with invalid data', function (done) {
+        User.updateOrCreate(INVALID_DATA, {validate: true}, expectValidationError(done));
```
bajtos: IIRC, `updateOrCreate` short-circuits to calling `create` when the data does not include any id property. As a result, these tests are checking `create` and not `updateOrCreate`. Could you please check?

test/optional-validation.test.js
bajtos: Please remove.

bajtos:
Hi @ernie58, sorry for taking so long to respond. I have one last concern about this patch - see https://github.com/strongloop/loopback-datasource-juggler/pull/752#discussion_r45731801. Other than that, I think this is good to be merged.

test/optional-validation.test.js
ernie58: That's correct, but the suite of test below does not, i've created 2 describe block, for both scenarios:
  * method updateOrCreate with new data
  * method updateOrCreate with existing data

test/optional-validation.test.js
ernie58: I've removed it

ernie58:
@bajtos I've removed the comment and responded to the other issue https://github.com/strongloop/loopback-datasource-juggler/pull/752#discussion_r45734394

test/optional-validation.test.js
bajtos: My concern is that the current suite creates a false assumption about what's being tested. We should either remove it or else rewrite to not short-circuit to `create`.

Thoughts?

ernie58:
@bajtos I've fixed most jshint errors. It would be nice to add a .jshintrc to the repository, that way I could immediately adhere to the code guidelines

bajtos:
>  I've fixed most jshint errors. It would be nice to add a .jshintrc to the repository, that way I could immediately adhere to the code guidelines

That's a great idea, would you like to contribute that patch yourself? You can start with jshint & jscs config from loopback repo.

ernie58:
There are quite a lot jshint errors now (more then 600)

ernie58:
@bajtos About the checks failing all the time: is this because something is wrong with the patch I've submitted? If yes, how can I see what has failed so I can make the changes necesary? I'm not that familiar with Jenkins

ernie58:
@bajtos Hi Bajtos, any news on this?

bajtos:
> There are quite a lot jshint errors now (more then 600)

Please move these changes to a standalone pull request, so that it's easier to tell in the future which code changes were related to automatic validation and which were purely style fixes.

> About the checks failing all the time: is this because something is wrong with the patch I've submitted? 

The failed checks reported here on GH are caused by downstream repositories. I am afraid our CI build is not exactly stable now, downstream failures are usually unrelated to the changes made :(

Don't worry about them now. Just make sure `npm test` passes on your machine.

> If yes, how can I see what has failed so I can make the changes necesary? I'm not that familiar with Jenkins

I believe our Jenkins logs are not accessible to public yet. We have a parallel Travis CI job that allows contributors to see at least some build results. But as I said before, don't worry about the failed downstream Jenkins job.

bajtos:
> There are quite a lot jshint errors now (more then 600)

When we were adding jshint to LoopBack, we have carefully tweaked the config to keep the number of errors at a reasonable level, to avoid too many white-space-only changes. Could you do the same?

ernie58:
@bajtos I've moved the config files to another branch

test/optional-validation.test.js
bajtos: @ernie58 ping ^^^^^

ernie58: IMHO it's not a false assumption, because it does what is says. It takes new data and uses updateOrCreate. If in the future the short-circuit would be changed to something else, there would be no tests here to check if updateOrCreate uses the optionalValidation in the intended way. Does that makes sense?

But if you think it's better, I'll remove it! Just let me know!

ernie58: @bajtos what will we decide?

bajtos: @ernie58 I see your point.

Let's remove the whole block `describe('method updateOrCreate with new data')` then, so that we can move this forward.

ernie58:
@bajtos removed the blocks! :)

lib/dao.js
```diff
@@ -257,6 +257,16 @@ DataAccessObject.create = function (data, options, cb) {
 
     data = obj.toObject(true);
 
+    //options has precedence on model-setting
```
bajtos: Please add a space after the slashes.

```
// options etc.
```

lib/dao.js
```diff
@@ -257,6 +257,16 @@ DataAccessObject.create = function (data, options, cb) {
...
+    if (options.validate === false) {
+      return create();
+    }
+
+    //only when options.validate is not set, take model-setting into consideration
```
bajtos: Please add a space after the slashes.

```
// only etc.
```

lib/dao.js
```diff
@@ -732,6 +755,16 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb)
         var obj = ctx.instance;
         var data = obj.toObject(true);
 
+        // options has precedence on model-setting
+        if (options.validate === false) {
+          _findOrCreate(query, data, obj);
+        }
+
+        //only when options.validate is not set, take model-setting into consideration
```
bajtos: Please add a space after the slashes.

```
// only etc.
```

bajtos:
@ernie58 thanks. I found three more style details to fix, otherwise the patch LGTM now (031fc34)

Once you have it fixed, could you please squash all commits into a single one and provide a descriptive commit message following the [50/72 rule](tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)? If you are not familiar with git history rewriting, you are looking for `git rebase -i master` and `git push -f`.

ernie58:
@bajtos done!

bajtos:

ernie58:
you're welcome

PR #764, opened 2015-11-19
Capture includeHasMany() as a find()'s caller by findCaller option

eugene-frb:
To replace incorrect location of capturing this condition in find() in
loopback-connector/lib/sql.js  proposed earlier in
https://github.com/strongloop/loopback-connector/pull/34.

Being set to 'includeHasMany', findCaller option triggers injection of
PARTITION BY clause in buildColumnNames() in
loopback-connector-mssql/lib/mssql.js only when find() function is
called to process include filter with 'has many' relation.

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -507,7 +507,7 @@ Inclusion.include = function (objects, include, options, cb) {
         inq: uniq(sourceIds)
       };
       relation.applyScope(null, filter);
-      options.findCaller = 'includeHasMany';
+      options.partitionBy = relation.keyTo;
       relation.modelTo.find(filter, options, targetFetchHandler);
       /**
        * Process fetched related objects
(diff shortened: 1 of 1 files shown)


slnode:

lib/include.js
```diff
@@ -507,6 +507,7 @@ Inclusion.include = function (objects, include, options, cb) {
         inq: uniq(sourceIds)
       };
       relation.applyScope(null, filter);
+      options.findCaller = 'includeHasMany';
```
raymondfeng: Can we have something like `partitionBy = '<fk>'` instead?

eugene-frb:
@raymondfeng: thanks for suggestion! I updated the code in both juggler and mssql repositories.

PR #767, opened 2015-11-20
change the test script for package.json

Hans-Zhang28:
In order to generate the mocha test report, our automation script will invoke "npm test" and it needs to detect that mocha is a command in the test script so we move the command in the test case in the Makefile to here.


slnode:

bajtos:
The issue has been fixed in the meantime.

PR #770, opened 2015-11-23
Correction of a regression introduced by commit 632898b

MichaelDiguet:
Correction of a regression introduced by commit 632898b: when querying an empty array ([]) with a 'neq' filter, there were no matching.

For example:

a filter like this one:

’children': {
	neq: 'barack'
}

On a set contain this object:

{
  children: [],
  id: 1
}

would not find any match.

Thanks to the correction I propose, the preceding situation leads to a match.

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -421,6 +421,11 @@ function applyFilter(filter) {
       // Also support array types. Mongo, possibly PostgreSQL
       if (Array.isArray(value)) {
         var matcher  = where[key];
+        // The following condition is for the case where we are querying with
+        // a neq filter, and when the value is an empty array ([]).
+        if (matcher.neq !== undefined && value.length <= 0) {
+          return true;
+        }
         return value.some(function (v, i) {
           var filter = {where: {}};
           filter.where[i] = matcher;
--- a/test/memory.test.js
+++ b/test/memory.test.js
@@ -338,6 +338,19 @@ describe('Memory connector', function() {
       });
     });
 
+    it('should successfully extract 5 users matching a neq filter over array values', function (done) {
+      User.find({
+        where: {
+          'children': {neq: 'Dhani'}
+        }
+      }, function (err, users) {
+        should.not.exist(err);
+        console.log(users);
+        users.length.should.be.equal(5);
+        done();
+      });
+    });
+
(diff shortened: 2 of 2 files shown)


slnode:

raymondfeng:
@MichaelDiguet Thanks for the patch. Can you sign the CLA at https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler?

MichaelDiguet:
@raymondfeng You are welcome Raymond. I signed the CLA.

lib/connectors/memory.js
```diff
@@ -421,6 +421,11 @@ function applyFilter(filter) {
       // Also support array types. Mongo, possibly PostgreSQL
       if (Array.isArray(value)) {
         var matcher  = where[key];
+        // The following condition is for the case where we are querying with
+        // a neq filter, and when the value is an empty array ([]).
+        if (matcher.neq !== undefined && value.length <= 0) {
```
raymondfeng: Do you handle `{neq: [1, 2]}`?

MichaelDiguet: In the scope of my correction, yes. My correction only deals with the empty array ([]). If you query with a 'neq' filter, the empty array should always match. I supposed that the preceding code works for cases that are not with the empty array. I can check that ...

raymondfeng: I see. `value.some(..)` should be able to handle that.

MichaelDiguet: I made the test, and {neq: [1, 2]} does not work correctly. The code written in the commit 632898b clearly doesn't handle this kind of cases since it makes a one by one analysis of the 'value' array.

raymondfeng: I interpret `{neq: [1, 2]}` as `not equal to 1 or 2`. WDYT?

MichaelDiguet: For a value, I interpret {neq: [1, 2]} as not equal to 1 or 2. 3 matches, 1 does not match, 2 does not match.
For an array, I would interpret it as does not contain 1 or 2. [3,4,5] matches, [1,3,4,5] does not match, [2,3,4,5] does not match.
Anyway, is this syntax should really be supported by loopback ? I can't find it in the documentation ...

PR #772, opened 2015-11-23
Ignored Error

Abebw:
I came across this bug when an update attribute call was violating a database restriction. In this case the api call returned 200 and didn't update the database.

Since I don't know what the Strongloop/Loopback standards are for protecting database information in error messages.  I just propagated the database error out the the frontend. 

This is my first time putting in a Pull request for opensource software so guidance about if I'm doing it right or not is appreciated.

Thanks for loopback. It's great
-Abe Barth-Werb

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2449,6 +2449,7 @@ DataAccessObject.prototype.updateAttributes = function updateAttributes(data, op
           context.data = typedData;
 
           function updateAttributesCallback(err) {
+            if (err) return cb(err);
             var ctx = {
               Model: Model,
               data: context.data,


slnode:

fabien:
@raymondfeng @bajtos I'm late to the party (only now starting to update my dependencies again), however, this PR has introduced some unexpected behavior for me. 

A lot of my tests (+70!) started to fail due to:

`MongoError: $set is empty. You must specify a field like so: {$mod: {<field>: ...}}`

Indeed, since 2014 MongoDB started to throw an error if `$set` contained no data to update, but until this patch, this particular error was silently ignored. Heading over to `loopback-connector-mongodb` to fix this (checking if the data to perform an update with was empty, and then omitting `$set` altogether) only helps half-way: the error is gone, but the current state of the document is never returned from the database either. So the tests still fail because of this.

Issue #774, opened 2015-11-24
datasource-connector documentation javascript code


pktippa:
docs\datasource-connector documentation has a small javascript code bug at line 83.

loay:
@pktippa  Thanks a lot  .. Looking into it.

loay:
commit was merged .. Issue is fixed
closes #774

PR #775, opened 2015-11-24
Fix for issue #774

pktippa:
Updated the document to fix the issue #774


slnode:

PR #776, opened 2015-11-25
Update helper scripts

superkhau:
Cleaned up the Makefile and added a few helper scripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,25 +1,33 @@
-## TESTS
-
 TESTER = ./node_modules/.bin/mocha
 OPTS = --growl
 TESTS = test/*.test.js
 
+default: help
+
+.PHONY: help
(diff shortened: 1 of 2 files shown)


superkhau:
@slnode test please

superkhau:
@raymondfeng Code review please. I want to merge this, but I think there are issues with CI.

superkhau:
@rmg Any reason why the above is failing? I'm only changing a makefile.

raymondfeng:
LGTM even though I don't see high value to have the extra helper targets.

superkhau:
@raymondfeng Its only to speed up simple repetitive tasks during dev time. I also don't want to accidently rm -rf the wrong dir on my system. ;p

PR #778, opened 2015-11-29
Add model def objectid support

superkhau:
Short term fix for ObjectId support.

Connect to strongloop-internal/scrum-loopback#625
See also https://github.com/strongloop-internal/scrum-loopback/issues/566

### DIFF
--- a/lib/objectid.js
+++ b/lib/objectid.js
@@ -1,0 +1,12 @@
+var ObjectId;
+
+try {
+  ObjectId = require('mongodb').ObjectId;
+} catch(e) {
+  if (e.code === 'MODULE_NOT_FOUND')
+    throw e;
+}
+
+module.exports = ObjectId;
+module.exports.ObjectID = ObjectId;
+module.exports.ObjectId = ObjectId;
--- a/lib/types.js
+++ b/lib/types.js
@@ -36,6 +36,7 @@ Types.Any.prototype.toObject = Types.Any.prototype.toJSON = function () {
 module.exports = function (modelTypes) {
 
   var GeoPoint = require('./geo').GeoPoint;
+  var ObjectId = require('./objectid');
 
   for(var t in Types) {
     modelTypes[t] = Types[t];
@@ -62,6 +63,7 @@ module.exports = function (modelTypes) {
   modelTypes.registerType(Array);
   modelTypes.registerType(GeoPoint);
   modelTypes.registerType(Object);
+  modelTypes.registerType(ObjectId);
 };
 
 module.exports.Types = Types;
--- a/test/datatype.test.js
+++ b/test/datatype.test.js
@@ -3,7 +3,30 @@ var should = require('./init.js');
 
 var db, Model;
 
-describe('datatypes', function () {
+describe('datatypes', function() {
+  context('ObjectId', function(done) {
+    var OidModel;
+
+    before(function setup() {
+      var db = getSchema();
+      OidModel = db.define('OidModel', {
+        id: {
+          type: 'ObjectId',
+          id: true
+        }
+      });
+      db.automigrate('OidModel', done);
+    });
+
+    it('should work', function(done) {
+      OidModel.create({id: 'hello1hello1'}, function(err, inst) {
+        if (err) return done(err);
+        inst.id.should.be.type('object');
+        inst.id.toString().should.be.type('string').and.have.length(24);
+        done();
+      });
+    });
+  });
 
   before(function (done) {
     db = getSchema();
--- a/test/e2e/model-definition.js
+++ b/test/e2e/model-definition.js
@@ -1,0 +1,59 @@
+var DataSource = require('../..').DataSource;
+var ObjectId = require('../../lib/objectid');
+var should = require('should');
+
+describe('model definition', function() {
+  context('ObjectId definition', function() {
+    var Todo;
+
+    before(function setup() {
+      var ds = new DataSource();
+      Todo = ds.define('Todo', {
+        id: {type: 'ObjectId', id: true}
+      });
+    });
+
+    it('should allow 12 byte strings', function() {
+      var todo = new Todo({id: 'hello1hello1'});
+
+      todo.id.should.be.an.instanceof(ObjectId);
+      // 12 byte strings are coerced into 24 character hex strings
+      todo.id.toString().should.have.length(24);
+    });
+
+    it('should allow 24 character hex strings', function() {
+      var hexStr = '507f191e810c19729de860ea';
+
+      var todo = new Todo({id: hexStr});
+
+      todo.id.should.be.an.instanceof(ObjectId);
+      todo.id.toString().should.equal(hexStr);
+    });
+
+    it('should allow ObjectIds', function() {
+      var objId = new ObjectId();
+
+      var todo = new Todo({id: objId});
+
+      todo.id.should.be.an.instanceof(ObjectId);
+      todo.id.toString().should.equal(objId.toString());
+    });
+  });
+
+  context('ObjectID (alias) definition', function() {
+    var Todo;
+
+    before(function setup() {
+      var ds = new DataSource();
+      Todo = ds.define('Todo', {
+        id: {type: 'ObjectID', id: true}
+      });
+    });
+
+    it('should work', function() {
+      var todo = new Todo({id: 'hello1hello1'});
+
+      todo.id.should.be.an.instanceof(ObjectId);
+    });
+  });
+});
--- a/Makefile
+++ b/Makefile
@@ -1,43 +1,55 @@
 TESTER = ./node_modules/.bin/mocha
 OPTS = --growl
 TESTS = test/*.test.js
+E2E_TEST_DIR = test/e2e/*.js
 
-default: help
+# Default target
 
-.PHONY: clean
-clean:
-	rm -rf $(CURDIR)/node_modules
-
-.PHONY: help
-help:
+.PHONY: help h
+help h:
 	@echo 'Usage: make [target]'
 	@echo 'Targets:'
-	@echo '  clean        Delete `node_modules`'
-	@echo '  help         Print help (this message)'
-	@echo '  refresh      Delete `node_modules` and run `npm install`'
-	@echo '  test         Run tests in silent mode'
-	@echo '  test-verbose Run tests in verbose mode'
-	@echo '  testing      Run tests continuously'
+	@echo '  e  e2e          Run end-to-end tests'
+	@echo '  h  help         Print help (this message)'
+	@echo '  t  test[s]      Run unit tests and e2e tests'
+	@echo '  u  unit         Run unit tests in silent mode'
+	@echo '  uv unit-verbose Run unit tests in verbose mode'
+	@echo '  uw unit-watch   Run unit tests in watch (--watch) mode'
 
-.PHONY: refresh
-refresh: clean
-	npm install
+# Targets
 
-.PHONY: test
-test:
+.PHONY: e2e e
+e2e e:
+	$(TESTER) --reporter spec $(E2E_TEST_DIR)
+
+.PHONY: test tests t
+test tests t: unit e2e
+
+.PHONY: unit u
+unit u:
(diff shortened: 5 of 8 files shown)


superkhau:
@ritch Review please.

superkhau:
@slnode test please

superkhau:
@bajtos Review please.

lib/objectid.js
```diff
@@ -0,0 +1,5 @@
+var ObjectId = require('mongodb').ObjectId;
```
ritch: I don't think we should depend on `mongodb` directly in the juggler. We should contribute `ObjectID` some other way. It should work without the `mongodb` connector (eg. in the browser).

ritch: Another option is to optionally depend on `mongodb` (or similar, eg. an `ObjectID` module). Sort of like this:

```js
```js
try {
  mongodb = require('mongodb');
} catch(e) {
  if (!isNotFound(err)) { 
    throw e;
  }
}
```

superkhau: My initial impl had our own ObjectId definition that delegated to mongo's object id, which we can move towards in the long term update. Will add it to the notes there

For the short term I will use the try/catch you suggested.

lib/objectid.js
```diff
@@ -0,0 +1,12 @@
+var ObjectId;
+
+try {
+  ObjectId = require('mongodb').ObjectId;
+} catch(e) {
+  if (e.code === 'MODULE_NOT_FOUND')
+    throw e
+}
+
```
superkhau: @ritch LGTY?

bajtos:
@superkhau please make sure the redis build failure is not related to your change

lib/objectid.js
bajtos: To be honest, I don't like adding `mongodb` dependency to juggler at all. Remember, there is no guarantee that `mongodb` module referenced by juggler will be the same module as referenced by `loopback-connector-mongodb`. In fact, this will be the most frequent case.

```
app
  node_modules
    loopback-datasource-juggler
      mongodb
    loopback-connector-mongodb
      mongodb
```

As a result, you are registering a type (constructor function) that is actually never used by the connector, because the connector has it's own local type (constructor function).

superkhau: TBH, I don't like it either, but it was the simplest way to get the full power of an ObjectId (all it's validations, unique string generation, other functions etc) in one shot. The other option is to just not use mongodb at all and come up with our own unique implementation for each function. However, this is not a 5 point job. I figure in the long term proposal task I have for this sprint, we can decide whether we want to come up with our own implementations for each function or use source code from other projects, etc.

superkhau: It would have an API similar to https://docs.mongodb.org/manual/reference/object-id/ basically.

superkhau: Another thing to consider is the option to allow user-configurable ObjectId types. Like in model-config.json, we allow `{"objectid": "../path/to/your/custom/objectid"}`, or something along those lines.

superkhau:
@bajtos Is there a way to target redis specifically? All my tests are green locally.

lib/objectid.js
bajtos: Makes me wonder if there is a way how to allow `loopback-connector-mongodb` to call `modelTypes.registerType(ObjectId)` during connector/datasource initialisation. That way we can keep the implementation in mongodb and still have ObjectID available in the app.

Of course, that would not work for isomorphic app in the browser, so it would be still a short-term fix.

Anyways, I guess you spend more time thinking about this issue than I did, and if this is the best short-term solution we have, then I can live with it.

bajtos:
> Is there a way to target redis specifically? All my tests are green locally.

Sorry for the confusion, I was referring to `dependant loopback-connector-redis`. There are more test failures reported now. They may be unrelated to this patch, just make sure to check & verify it's true.

lib/objectid.js
superkhau: >Makes me wonder if there is a way how to allow loopback-connector-mongodb to call modelTypes.registerType(ObjectId) during connector/datasource initialisation. That way we can keep the implementation in mongodb and still have ObjectID available in the app.

I have thought about this too, but figured it would be harder to implement because we don't expose that part of the lib in DAO. The long term solution I had in mind and discussed with @ritch is an app-level registry for datatypes. During boot time, we would register all basic built in datatypes, then register any custom types. Then at run time, we can dynamically register/deregister new datatypes. Something along the lines of:

`app.datatypes.add(ObjectId);
`app.datatypes.list();
`app.datatypes.remove(ObjectId);

Or use an a K/V array for datatypes, etc. We can discuss further in the long-term proposal with regards to implementations.

superkhau:
The redis tests are failing on its own.

superkhau:
@bajtos Reassigning to you for review as @ritch is probably too busy to review this. Please let me know how to move forward with this.

lib/objectid.js
```diff
@@ -0,0 +1,12 @@
+var ObjectId;
+
+try {
+  ObjectId = require('mongodb').ObjectId;
+} catch(e) {
+  if (e.code === 'MODULE_NOT_FOUND')
+    throw e
```
bajtos: Missing semicolon at EOL, please add.

test/e2e/model-definition.js
```diff
@@ -0,0 +1,46 @@
...
+      Todo = ds.define('Todo', {
+        id: {type: 'ObjectId', id: true}
+      });
+    });
+
+    it('should allow 12 byte strings', function() {
+      var todo = new Todo({id: 'hello1hello1'});
+
+      todo.id.should.be.an.instanceof(ObjectId);
+      todo.id.should.be.an.instanceof(MongoObjId);
```
bajtos: The fact that juggler's `ObjectID` type is implemented via mongodb's type is IMO an implementation detail that should not be asserted in unit-tests. If we want to check that `ObjectID` behaves like mongodb's `ObjectID`, then I'd much rather add few more tests to test that behaviour.

Thoughts?

bajtos:
@superkhauplease add few tests to the test-suite run by all connectors (see [test/common.batch.js](https://github.com/strongloop/loopback-datasource-juggler/blob/4ce02cbcc13a584390349293cfc35fb1c120b1fc/test/common.batch.js)) to ensure this new ObjectID type is correctly persisted and loaded when using e.g. relational (SQL) databases.

test/e2e/model-definition.js
bajtos: Also the tests in this file duplicates the tests in `test/objectid.test.js`. What's the point of that?

I am proposing to simplify tests in this file to a bare minimum needed in order to verify that `type: 'ObjectId'` is correctly coerced into an instance of `require('../../lib/objectid')`. Everything else should be tested in `test/objectid.test.js`.

lib/types.js
```diff
@@ -62,6 +63,7 @@ module.exports = function (modelTypes) {
   modelTypes.registerType(Array);
   modelTypes.registerType(GeoPoint);
   modelTypes.registerType(Object);
+  modelTypes.registerType(ObjectId);
```
bajtos: Is there any way how to register `ObjectID` as an alias for `ObjectId`? Or is it automatically handled?

test/e2e/model-definition.js
superkhau: >The fact that juggler's ObjectID type is implemented via mongodb's type is IMO an implementation detail that should not be asserted in unit-tests. If we want to check that ObjectID behaves like mongodb's ObjectID, then I'd much rather add few more tests to test that behaviour.

I can live with this. We shouldn't test MongoDB's impl at all since it is an impl detail of our custom ObjectId type. Will remove tests that relate to MongoObjId.

superkhau: >Also the tests in this file duplicates the tests in test/objectid.test.js. What's the point of that?

They are testing from outside in vs inside out (ie. functional vs unit). They look like dupes, but should be tested independently. The e2e tests use things like `new Todo({type: 'ObjectId', ...})` vs `new ObjectId()`. IMO, LoopBack mixes unit/functional tests together way too much and shouldn't be. The different aspects should be tested independently and in isolation.

lib/types.js
superkhau: It is already registered as an alias in the ObjectId class definition, so it should work properly. I added an e2e test to prove it.

test/e2e/model-definition.js
```diff
@@ -0,0 +1,59 @@
...
+    });
+  });
+
+  context('ObjectID (alias) definition', function() {
+    var Todo;
+
+    before(function setup() {
+      var ds = new DataSource();
+      Todo = ds.define('Todo', {
+        id: {type: 'ObjectID', id: true}
```
superkhau: Test added here.

superkhau:
@bajtos Updated with the comments from your review. Can you PTAL again?

lib/objectid.js
```diff
@@ -0,0 +1,12 @@
+var ObjectId;
+
+try {
+  ObjectId = require('mongodb').ObjectId;
+} catch(e) {
+  if (e.code === 'MODULE_NOT_FOUND')
+    throw e;
```
bajtos: @superkhau AFAICT, this silently swallows any `mongodb` parse errors, which does not look like a good thing to me. Could you please explain what is the rationale for this, why we need to give `require('mongodb').ObjectId` errors a special treatment?

bajtos:
@superkhau I am missing unit-tests to be run by connectors, per my earlier comment:

> please add few tests to the test-suite run by all connectors (see [test/common.batch.js](https://github.com/strongloop/loopback-datasource-juggler/blob/4ce02cbcc13a584390349293cfc35fb1c120b1fc/test/common.batch.js)) to ensure this new ObjectID type is correctly persisted and loaded when using e.g. relational (SQL) databases.

lib/objectid.js
superkhau: This was suggested by @ritch in a [previous comment in this issues history](https://github.com/strongloop/loopback-datasource-juggler/pull/778#discussion_r46341514). I think it was for uses cases where mongodb is not available (ie. the browser).

superkhau:
@slnode test please

test/datatype.test.js
```diff
@@ -3,7 +3,30 @@ var should = require('./init.js');
 
 var db, Model;
 
-describe('datatypes', function () {
```
superkhau: @bajtos Added data type tests here.

superkhau:
@bajtos Added data type tests here: https://github.com/strongloop/loopback-datasource-juggler/pull/778/files#r47993341

lib/objectid.js
bajtos: Fair enough. Please add a debug log to let the user know `ObjectId` will be undefined because `mongodb` was not found.

```js
debug('Cannot load mongodb, ObjectId will be undefined. %s', e.stack);
```

bajtos:
One more comment: https://github.com/strongloop/loopback-datasource-juggler/pull/778#discussion_r48249729

@superkhau after you address ^^^, then please rebase the patch on top of the current master, clean up the commit history and merge this without waiting for me. IIUC this is a backwards-compatible change targeted for LoopBack LTS, in which case you need to back-port it to `2.x` too.

bajtos:
`2.x` refers to this: https://github.com/strongloop/loopback-datasource-juggler/tree/2.x

mrbatista:
@superkhau when will this patch be merged?

Thanks.

superkhau:
@mrbatista Soon I believe, I need to write a few more tests to confirm the data with relational connectors, but after that it's all good.

bajtos:
@superkhau I am not sure what's the status of this work. If we decide to eventually land this, then I would like you to try an alternative approach: get the implementation of `ObjectID` directly from [bson/lib/bson/objectid.js](https://github.com/mongodb/js-bson/blob/66bbd418c711d8b5a7f3510933fe0840a3e12b53/lib/bson/objectid.js). A believe the `mongodb` driver we use does basically the same, so it should be safe to bypass `mongodb` and load the definition directly. If it works, then this solution will reduce the footprint of juggler in applications that don't use MongodB.

superkhau:
@bajtos This has been stale for a long time and is probably not a good solution now. Let's do what you propose instead.

bajtos:
ok, I'm closing this PR then.

Issue #610, opened 2015-05-29
Find + relation inclusion with a limit is broken


seriousben:
Since 2.8.0, this is broken.

The query that used to work:
```
Post.find({
include: [{
  relation: 'likes',
  scope: {
    where: {userId: currentUserId},
    include: [{
      'owner': 'avatar'
    }],
    limit: 1
  }
}],
limit: 10,
order: 'createdAt DESC'
}
```

The problem is that the limit gets put on the merged Relation query like so:
```
QUERY:  Like {
 "userId": "55548654af697c1d7073412d",
 "postId": {
  "$in": [
   "556847b15a0fc6014959edad",
   "556847b15a0fc6014959edac",
   "556847b05a0fc6014959edab",
   "556847b05a0fc6014959edaa",
   "556847b05a0fc6014959eda9",
   "556847b05a0fc6014959eda8",
   "556847b05a0fc6014959eda7",
   "556847b05a0fc6014959eda6",
   "556847b05a0fc6014959eda5",
   "556847b05a0fc6014959eda4",
   "556847b05a0fc6014959eda3",
   "556847b05a0fc6014959eda2",
   "556847b05a0fc6014959eda1",
   "556847b05a0fc6014959eda0",
   "556847b05a0fc6014959ed9f",
   "556847b05a0fc6014959ed9e",
   "556847b05a0fc6014959ed9d",
   "556847af5a0fc6014959ed9c",
   "556847af5a0fc6014959ed9b",
   "556847af5a0fc6014959ed98",
   "556847af5a0fc6014959ed9a",
   "556847af5a0fc6014959ed97",
   "556847af5a0fc6014959ed96",
   "556847af5a0fc6014959ed94",
   "556847af5a0fc6014959ed95",
   "556839a25a0fc6014959ed5c",
   "556839a15a0fc6014959ed5a",
   "556839a15a0fc6014959ed5b"
  ]
 }
}
====>
cursor.limit(1);
<====
```

raymondfeng:
1. There is a typo: `orer: 'createdAt DESC'` should be `order: 'createdAt DESC'`
2. I assume you want to find most recent 10 `posts` including at most 1 `like` which also includes the `owner`. What's wrong with the limit(1)?

seriousben:
Thanks @raymondfeng - Fixed the typo. Just me adding it manually to this issue.

In this example, I want to include the likes of a user limiting it to 1 like *per* post.

[...]

PR #786, opened 2015-12-08
Partition by foreign key for pagination

raymondfeng:
/to @bajtos 

See https://github.com/strongloop/loopback-datasource-juggler/issues/610

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -55,7 +55,7 @@ function IncludeScope(scope) {
   } else {
     this._include = null;
   }
-};
+}
 
 IncludeScope.prototype.conditions = function() {
   return utils.deepMerge({}, this._scope);
@@ -174,11 +174,90 @@ Inclusion.include = function(objects, include, options, cb) {
 
   include = normalizeInclude(include);
 
+  // Find the limit of items for `inq`
+  var inqLimit = 256;
+  if (self.dataSource && self.dataSource.settings &&
+    self.dataSource.settings.inqLimit) {
+    inqLimit = self.dataSource.settings.inqLimit;
+  }
+
   async.each(include, function(item, callback) {
     processIncludeItem(objects, item, options, callback);
   }, function(err) {
     cb && cb(err, objects);
   });
+
+  /**
+   * Find related items with an array of foreign keys by page
+   * @param model The model class
+   * @param filter The query filter
+   * @param fkName The name of the foreign key property
+   * @param pageSize The size of page
+   * @param options Options
+   * @param cb
+   */
+  function findWithForeignKeysByPage(model, filter, fkName, pageSize, options, cb) {
+    var foreignKeys = [];
+    if (filter.where[fkName]) {
+      foreignKeys = filter.where[fkName].inq;
+    } else if (filter.where.and) {
+      // The inq can be embedded inside 'and: []'. No or: [] is needed as
+      // include only uses and. We only deal with the generated inq for include.
+      for (var j in filter.where.and) {
+        if (filter.where.and[j][fkName] &&
+          Array.isArray(filter.where.and[j][fkName].inq)) {
+          foreignKeys = filter.where.and[j][fkName].inq;
+          break;
+        }
+      }
+    }
+    if (filter.limit || filter.skip || filter.offset) {
+      // Force the find to be performed per FK to honor the pagination
+      pageSize = 1;
+    }
+    var size = foreignKeys.length;
+    if (size > inqLimit && pageSize <= 0) {
+      pageSize = inqLimit;
+    }
+    if (pageSize <= 0) {
+      return model.find(filter, options, cb);
+    }
+
+    var listOfFKs = [];
+
+    for (var i = 0; i < size; i += pageSize) {
+      var end = i + pageSize;
+      if (end > size) {
+        end = size;
+      }
+      listOfFKs.push(foreignKeys.slice(i, end));
+    }
+
+    var items = [];
+    async.each(listOfFKs, function(foreignKeys, done) {
+      var newFilter = {};
+      for (var f in filter) {
+        newFilter[f] = filter[f];
+      }
+      if (filter.where) {
+        newFilter.where = {};
+        for (var w in filter.where) {
+          newFilter.where[w] = filter.where[w];
+        }
+      }
+      newFilter.where[fkName] = {
+        inq: foreignKeys,
+      };
+      model.find(newFilter, options, function(err, results) {
+        if (err) return done(err);
+        items = items.concat(results);
+        done();
+      });
+    }, function(err) {
+      if (err) return cb(err);
+      cb(null, items);
+    });
+  }
 
   function processIncludeItem(objs, include, options, cb) {
     var relations = self.relations;
@@ -332,7 +411,9 @@ Inclusion.include = function(objects, include, options, cb) {
       /**
        * 1st DB Call of 2 step process. Get through model objects first
        */
-      relation.modelThrough.find(throughFilter, options, throughFetchHandler);
+      findWithForeignKeysByPage(relation.modelThrough, throughFilter,
+        relation.keyTo, 0, options, throughFetchHandler);
+
       /**
        * Handle the results of Through model objects and fetch the modelTo items
        * @param err
@@ -377,7 +458,10 @@ Inclusion.include = function(objects, include, options, cb) {
         /**
          * 2nd DB Call of 2 step process. Get modelTo (target) objects
          */
-        relation.modelTo.find(filter, options, targetsFetchHandler);
+        findWithForeignKeysByPage(relation.modelTo, filter,
+          modelToIdName, 0, options, targetsFetchHandler);
+
+        // relation.modelTo.find(filter, options, targetsFetchHandler);
         function targetsFetchHandler(err, targets) {
           if (err) {
             return callback(err);
@@ -452,7 +536,8 @@ Inclusion.include = function(objects, include, options, cb) {
       /**
        * Make the DB Call, fetch all target objects
        */
-      relation.modelTo.find(filter, options, targetFetchHandler);
+      findWithForeignKeysByPage(relation.modelTo, filter,
+        relation.keyTo, 0, options, targetFetchHandler);
       /**
        * Handle the fetched target objects
        * @param err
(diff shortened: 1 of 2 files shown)


lib/include.js
```diff
@@ -175,6 +175,83 @@ Inclusion.include = function (objects, include, options, cb) {
     cb && cb(err, objects);
   });
 
+  // Find the limit of items for `inq`
+  var inqLimit = 256;
+  if(self.dataSource && self.dataSource.settings &&
+    self.dataSource.settings.inqLimit) {
+    inqLimit = self.dataSource.settings.inqLimit;
+  }
+  /**
```
bajtos: Nitpick: please add an empty line as a visual delimiter between the code block and the unrelated jsdoc comment.

lib/include.js
```diff
@@ -327,10 +404,12 @@ Inclusion.include = function (objects, include, options, cb) {
         throughFilter.where[polymorphic.discriminator] =
           relation.modelFrom.definition.name;
       }
+      findWithForeignKeysByPage(relation.modelThrough, throughFilter,
+        relation.keyTo, 0, options, throughFetchHandler);
       /**
        * 1st DB Call of 2 step process. Get through model objects first
        */
-      relation.modelThrough.find(throughFilter, options, throughFetchHandler);
+      // relation.modelThrough.find(throughFilter, options, throughFetchHandler);
```
bajtos: Commented-out code: if it's an intentional change, then please remove it completely.

lib/include.js
```diff
@@ -175,6 +175,83 @@ Inclusion.include = function (objects, include, options, cb) {
...
+   * @param options Options
+   * @param cb
+   */
+  function findWithForeignKeysByPage(model, filter, fkName, pageSize, options, cb) {
+    var foreignKeys = [];
+    if (filter.where[fkName]) {
+      foreignKeys = filter.where[fkName].inq;
+    } else if (filter.where.and) {
+      // The inq can be embedded inside 'and: []'
+      for (var j in filter.where.and) {
```
bajtos: How about `or: []` - should it be supported too?

lib/include.js
```diff
@@ -175,6 +175,83 @@ Inclusion.include = function (objects, include, options, cb) {
...
+      for (var f in filter) {
+        newFilter[f] = filter[f];
+      }
+      if (filter.where) {
+        newFilter.where = {};
+        for (var w in filter.where) {
+          newFilter.where[w] = filter.where[w];
+        }
+      }
+      newFilter.where[fkName] = {
```
bajtos: Will this support the case of `where:{ and: [...] }`? I mean is `where: {inq:foreignKeys, and:[...]}` a valid one?

bajtos:
@raymondfeng I did a quick review for obvious issues, see my comments above. My biggest concern is about insufficient code coverage. I'd like to see new test cases covering all of the major scenarios affected by this change (different where filters, different page sizes, etc.)

lib/include.js
raymondfeng: No `or: []` is needed as `include` only use `and`. We only deal with the generated `inq` for `include`.

raymondfeng:
@bajtos Is it possible to have somebody from LB team to take it over?

bajtos:
> Is it possible to have somebody from LB team to take it over?

Sure, I'll reassign #610 to @hacksparrow. Note that it's unlikely to get this task into the next sprint.

superkhau:
Connect to https://github.com/strongloop/loopback/issues/1744

superkhau:
Connect to https://github.com/strongloop/loopback-datasource-juggler/issues/610

superkhau:
@bajtos It turns out this fix is related the issue I'm working on this week. Merging this in will also fix https://github.com/strongloop/loopback/issues/1744. I will PM you with regards to how to move forward on my issue.

lib/include.js
bajtos: > No `or: []` is needed as `include` only use `and`. We only deal with the generated `inq` for `include`.

I see. Could you please add a comment with this information?

bajtos:
@superkhau I am reassigning to you, @hacksparrow is busy working on other things.

As for tests, check if all major branches are covered. For example:

- `if (filter.where[fkName]) {`
- `if (filter.where.and) {`
- `if (filter.limit || filter.skip || filter.offset) {` + the case when none of that is set (4 unit tests in total, perhaps more?)
- `if(size > inqLimit && pageSize <= 0) `
- `if (pageSize <= 0) {`

Run those tests for one type of relations to verify the inside implementation works correctly

Then add one simple test for each relation type, to ensure we are calling the new implementation from all relation methods.

lib/include.js
bajtos: > > No `or: []` is needed as `include` only use `and`. We only deal with the generated `inq` for `include`.

> I see. Could you please add a comment with this information?

@superkhau ^^^ could you please take care of that?

lib/include.js
```diff
@@ -169,12 +169,91 @@ Inclusion.include = function (objects, include, options, cb) {
...
+   * @param options Options
+   * @param cb
+   */
+  function findWithForeignKeysByPage(model, filter, fkName, pageSize, options, cb) {
+    var foreignKeys = [];
+    if (filter.where[fkName]) {
+      foreignKeys = filter.where[fkName].inq;
+    } else if (filter.where.and) {
+      // The inq can be embedded inside 'and: []'. No or: [] is needed as
+      // include only use and. We only deal with the generated inq for include.
```
superkhau: @bajtos Included the comment here.

superkhau:
@bajtos Review please.

test/include.test.js
```diff
@@ -190,19 +190,322 @@ describe('include', function () {
...
+    });
+
+    it('should support include by pagination', function(done) {
+      Passport.find({
+        include: {
+          owner: {
+            relation: 'posts',
+            scope: {
+              fields: ['title'], include: ['author'],
+              order: 'title ASC'
```
bajtos: The test name says "include by pagination", but I don't see any pagination defined in the query. Am I reading this code incorrectly?

test/include.test.js
```diff
@@ -190,19 +190,322 @@ describe('include', function () {
...
+
+      it('works when using `where` without `limit`, `skip` or `offset`',
+          function(done) {
+        User.findOne({include: {relation: 'posts'}}, function(err, user) {
+          if (err) return done(err);
+
+          var posts = user.posts();
+          posts[0].id.should.equal(1);
+          posts[1].id.should.equal(2);
+          posts[2].id.should.equal(3);
```
bajtos: These three asserts can be simplified to the following, it will also give you more helpful message when e.g. the ids are sorted differently:

```js
var ids = posts.map(function(p) { return p.id; }); // or use _.property('id') from lodash
ids.should.eql([1,2,3]);
```

The same comment applies to other tests below.

bajtos: Ehm, I meant other test above.

test/include.test.js
```diff
@@ -190,19 +190,322 @@ describe('include', function () {
...
+          users.length.should.equal(5);
+
+          delete Post.dataSource.settings.inqLimit;
+
+          done();
+        });
+      });
+
+      it('works when page size is set to 0', function(done) {
+        Post.dataSource.settings.inqLimit = 0;
```
bajtos: Test names says "page size is set to 0", but here you are changing `inqLimit` instead. I find it confusing.

bajtos:
@superkhau Few minor comments. Also please check why CI build are failing.

test/include.test.js
superkhau: `inqLimit` changes `pageSize` (see https://github.com/strongloop/loopback-datasource-juggler/blob/feature/fix-610/lib/include.js#L215).

I'll leave a comment in the test stating this.

>// inqLimit modifies page size in the implementation

test/include.test.js
superkhau: Nice. I used the way you posted without lodash as I didn't want to introduce another dep.

test/include.test.js
superkhau: That test was already there before I took over the PR, but the `pagination` it's refering to is inside the impl and set by `inqLimit = 2` in the before block. I added a comment above to make it a bit more clear.

superkhau:
- angular-live-set - unrelated
- strong-agent - unrelated 
- strong-arc - unrelated
- strong-mesh-client - unrelated
- loopback-datasource-juggler.dependents - failing because the above 4 are failing (all juggler related tests are passing)

That said, we should really remove those dependents or fix it since they've been failing forever (and causing unnecessary checks whenever anything runs on CI).
- angular-live-set - should fix this one, it was passing 4 builds ago (passing yesterday), linter errors it looks like
- strong-agent - failing since march 31st, failing due to some tap error, didn't dig deeper
- strong-arc - full page of failures on CI, never passing, didn't dig deeper
- strong-mesh-client - npm install failure, can't find jshint causing failure, didn't dig deeper

@bajtos Thoughts?

superkhau:
@bajtos Please review again too, I updated with your feedback.

bajtos:
LGTM. @superkhau please rebase on top of the current master and make sure the code passes the new linting. No more review is needed.

> angular-live-set - should fix this one, it was passing 4 builds ago (passing yesterday), linter errors it looks like

:+1: 

@superkhau could you please take care of that?

> strong-agent - failing since march 31st, failing due to some tap error, didn't dig deeper

Can you check with @sam-github please?

> strong-arc - full page of failures on CI, never passing, didn't dig deeper

@rmg what would you advise here? Can we modify the code triggering dependant builds to exclude repositories that are known to fail often on CI?

http://ci.strongloop.com/job/strong-arc/buildTimeTrend

> strong-mesh-client - npm install failure, can't find jshint causing failure, didn't dig deeper

@rmg the story here is the same as in strong-arc:

http://ci.strongloop.com/job/strong-mesh-client/buildTimeTrend

superkhau:
@slnode test please

superkhau:
@slnode test please

superkhau:
@slnode test please

rmg:
@slnode test please

rmg:
@slnode test please

rmg:
@slnode test please

superkhau:
@bajtos I'm merging this as the connector errors are not related to this PR. The other errors are the same as before.

superkhau:
>angular live set

https://github.com/strongloop/angular-live-set/blob/master/modules/change-stream/test/change-stream.test.js#L5

Created issue here: https://github.com/strongloop-internal/scrum-loopback/issues/829

>strong-agent

Waiting for reply.

rmg:
@superkhau waiting for a reply from whom? And with what information?

superkhau:
@rmg I messaged @sam-github, he replied on Slack:

>looks like ci is failing on OX X with node 5.x, should be reproducible. possibly one of the dependents doesn't work with 5.x on OS X

Do you have any idea? Maybe you have some insight here.

PR #789, opened 2015-12-09
Fix failing test for MongoDB connector

superkhau:
Connect to https://github.com/strongloop/loopback-connector-mongodb/pull/204

See the connector PR for discussions.

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -492,7 +492,10 @@ describe('manipulation', function () {
     });
 
     it('should ignore unknown attributes when strict: true', function(done) {
-      person.updateAttributes({foo:'bar'},
+      // Using {foo: 'bar'} only causes dependent test failures due to the
+      // stripping of object properties when in strict mode (ie. {foo: 'bar'}
+      // changes to '{}' and breaks other tests
+      person.updateAttributes({name: 'John', foo:'bar'},
         function(err, p) {
           if (err) return done(err);
           should.not.exist(p.foo);


superkhau:
@raymondfeng The fix is here.

I hate the way we don't test in isolation, imported tests is such a hack IMO. ie. Juggler should make sure it works with the mongo connector, not the mongo connector testing it works with juggler.

test/manipulation.test.js
```diff
@@ -492,7 +492,7 @@ describe('manipulation', function () {
     });
 
     it('should ignore unknown attributes when strict: true', function(done) {
-      person.updateAttributes({foo:'bar'},
+      person.updateAttributes({name: 'John', foo:'bar'},
```
raymondfeng: Can you add a comment on why you need to change the test?

raymondfeng:
@superkhau I'm not sure if I agree. Here is the requirement:

1. Juggler is the ORM that expects certain behaviors from the connector
2. There are a set of tests in Juggler to make sure connectors implement the expected functions

When you add a new connector to LB, it needs to run the common tests to ensure it's complying with Juggler. The juggler defines the extension point and interface and connectors provide the implementation and contribute extensions to the extension point.

test/manipulation.test.js
```diff
@@ -492,7 +492,10 @@ describe('manipulation', function () {
     });
 
     it('should ignore unknown attributes when strict: true', function(done) {
-      person.updateAttributes({foo:'bar'},
+      // Using {foo: 'bar'} only causes dependent test failures due to the
+      // stripping of object properties when in strict mode (ie. {foo: 'bar'}
+      // changes to '{}' and breaks other tests
```
superkhau: @raymondfeng Comment added.

superkhau:
>1. Juggler is the ORM that expects certain behaviors from the connector

It defines the expected interface, this makes sense.

>There are a set of tests in Juggler to make sure connectors implement the expected functions

This I agree with, but it doesn't need to test the behavior of the expected functions. The extension points should be stubbed in the unit tests of Juggler. Just make sure the function (extension point) exists and can be called.

The connectors then should have their own set of tests that ensure conformity to the juggler's interface and a set of functional tests to ensure it works with the juggler in the expected way. That is of course in addition to the usual unit tests it should have itself to ensure it gets expected results even without Juggler.

This way, the Juggler can ensure the connectors implement the correct interface and the connectors are themselves responsible for its own behaviors (which vary depending on the connector and may vary depending on when it is working in unison with the Juggler).

This means juggler can work in isolation without connectors, and connectors are responsible for their own behaviors. If a test breaks in a the MongoDB connector for example, we won't need to make a fix in Juggler's hacky "imported" test (because Juggler doesn't test MongoDB behavior, only cares about the interface).

superkhau:
>When you add a new connector to LB, it needs to run the common tests to ensure it's complying with Juggler...

The key word is "it", which should be the responsibility of the connector to make sure its complying with the contract specified by Juggler.

>The juggler defines the extension point and interface and connectors provide the implementation and contribute extensions to the extension point. 

Yes, the Juggler should define the extension point and interface, but not behaviors. The connectors providing the impl should test the behaviors.

superkhau:
@slnode test please

superkhau:
@rmg How can I tell if this is safe to merge?

superkhau:
Actualy, the other dependents are unrelated. Only the redis connector is but those test failures are unrelated to this PR.

@raymondfeng LGTY?

superkhau:
@bajtos LGTY?

raymondfeng:
LGTM

superkhau:
TY

PR #790, opened 2015-12-14
[SEMVER-MAJOR] Always use bluebird as the Promise implementation

jannyHou:
This pr is created for updating files to use `bluebird` as promise library.
Details see here:
Connect to strongloop-internal/scrum-loopback#615

Related: https://github.com/strongloop/loopback/pull/1896

### DIFF
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -5,3 +5,17 @@ always describe the impact on users and instructions for upgrading
 applications from 2.x to 3.0.
 
 See also https://github.com/strongloop/loopback/blob/master/3.0-DEVELOPING.md
+
+## Always use bluebird as promise library
+
+In version 3.0, we always use bluebird as our promise library
+instead of `global.Promise`.
+We consider Bluebird API a part of LoopBack API from now on,
+you are welcome to use any Bluebird-specific methods in your applications.
+
+If you are using LoopBack with a custom promise implementation provided
+via `global.Promise`,
+you will have to check all places where you are using non-standard promise API
+and update them to use Bluebird API instead.
+
+Please see [Related code change](https://github.com/strongloop/loopback-datasource-juggler/pull/790) here.
--- a/package.json
+++ b/package.json
@@ -32,12 +32,12 @@
     "node >= 0.6"
   ],
   "devDependencies": {
-    "bluebird": "^2.9.9",
     "mocha": "^2.1.0",
     "should": "^8.0.2"
   },
   "dependencies": {
     "async": "~1.0.0",
+    "bluebird": "^3.1.1",
     "debug": "^2.1.1",
     "depd": "^1.0.0",
     "inflection": "^1.6.0",
--- a/test/init.js
+++ b/test/init.js
@@ -26,7 +26,3 @@ if (!('getModelBuilder' in global)) {
     return new ModelBuilder();
   };
 }
-
-if (!('Promise' in global)) {
-  global.Promise = require('bluebird');
-}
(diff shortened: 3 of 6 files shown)


bajtos:
Hi @jannyHou, I guess this is a good start. Now you need to clean up `createPromiseCallback`. We no longer depend `global.Promise`, thus there is no need to check if it's defined. Also please move `bluebird` from `devDependencies` to `dependencies`.

bajtos:
Also remove these three lines and make sure tests are still passing on Node v0.10: https://github.com/strongloop/loopback-datasource-juggler/blob/1e9bbd27873077ef374ea5487121a70173757364/test/init.js#L30-L32

jannyHou:
Hi miroslav, I fixed 2 of 3 failures, actually the third one is an unhandled error msg instead of a failure(at least mocha doesn't count it), and I get the same error when I run the code cloned from master branch. I guess it's not caused by my change but I'll see how to fix it(because it seems related to promise).

And I created a list of promisified apis whose doc might need update:
https://github.com/strongloop-internal/scrum-loopback/issues/615#issuecomment-165314984

Thanks!

package.json
```diff
@@ -29,12 +29,12 @@
     "node >= 0.6"
   ],
   "devDependencies": {
-    "bluebird": "^2.9.9",
     "mocha": "^2.1.0",
     "should": "^5.0.0"
   },
   "dependencies": {
     "async": "~1.0.0",
+    "bluebird": "^2.9.9",
```
bajtos: While you are making this change, could you please upgrade Bluebird to 3.x?

bajtos:
> I fixed 2 of 3 failures, and now tests pass on node v.10. Actually the third one is an unhandled error msg instead of a failure(at least mocha doesn't count it), and I get the same error when I run the code cloned from master branch. I guess it's not caused by my change but I'll see how to fix it(because it seems related to promise).

The new code looks right.

> I commented if (!global.Promise) { //DoCheck } instead of delete, since I am not sure do we still need some check for promise library, I mean bluebird.

No need to keep that code in, we will always have the promise library as long as Node handles dependencies correctly (which is safe to assume).

> And I created a list of promisified apis whose doc might need update:
strongloop-internal/scrum-loopback#615 (comment)

:+1

test/init.js
```diff
@@ -27,6 +27,3 @@ if (!('getModelBuilder' in global)) {
   };
 }
 
-if (!('Promise' in global)) {
```
bajtos: Please remove the trailing empty line (L29) too.

bajtos:
@jannyHou code changes LGTM. Please rebase your patch on top of the current master and add an entry to https://github.com/strongloop/loopback-datasource-juggler/blob/master/3.0-RELEASE-NOTES.md

The idea is write a short text explaining loopback users (developers) what this change means for their applications and provide instructions how to upgrade.

@superkhau could you please help here?

bajtos:
> The idea is write a short text explaining loopback users (developers) what this change means for their applications and provide instructions how to upgrade.

Also include a link to this pull request to make it easy to look up more details if needed in the future.

3.0-RELEASE-NOTES.md
```diff
@@ -5,3 +5,10 @@ always describe the impact on users and instructions for upgrading
 applications from 2.x to 3.0.
 
 See also https://github.com/strongloop/loopback/blob/master/3.0-DEVELOPING.md
+
+## Always use bluebird as promise library
+
+In version 3.0, we always use bluebird as our promise library instead of `global.Promise`. We consider Bluebird API a part of LoopBack API from now on, you are welcome to use any Bluebird-specific methods in your applications.
```
bajtos: Please keep line width at 80 characters max, it makes it easier to review diffs.

bajtos:
One more minor comment to address. Once you address it, please go ahead and land the PR without waiting for another review.

PR #796, opened 2015-12-18
Various fixes in operation hooks

bajtos:
- `support/describe-operation-hooks.js` - add "loaded" hook to the list of described hooks
 - "loaded" hook in DAO.find - set `ctx.data`, not `ctx.instance`
 - "persisted" hook in DAO.updateAttributes - report `isNewInstance:false`
 - "persist" hook in DAO.updateOrCreate - report `isNewInstance` if the connector provides this info


raymondfeng:
LGTM

PR #797, opened 2015-12-19
Update should.js to 8.0.2

superkhau:


raymondfeng:
You probably need to upgrade connectors too as they share some common tests from juggler.

superkhau:
Yeah, but redis connector is failing. So CI is not running the other connector tests. ;( Do you think its safe to merge this and fix each other connector individually?

bajtos:
FWIW, some time ago, I submitted a patch adding an install script to juggler that could be used by connectors to pull in the right version of dependencies needed by the tests, see https://github.com/strongloop/loopback-datasource-juggler/pull/456. Perhaps it's time to resurrect it?

raymondfeng:
I am out of the office until 01/04/2016.

Happy holidays!


Note: This is an automated response to your message  "Re:
[loopback-datasource-juggler] Update should.js to 8.0.2 (#797)" sent on
12/21/15 2:46:10 PM.

This is the only notification you will receive while this person is away.

raymondfeng:
I am out of the office until 01/04/2016.

Happy holidays!


Note: This is an automated response to your message  "Re:
[loopback-datasource-juggler] Update should.js to 8.0.2 (#797)" sent on
12/21/15 2:46:06 PM.

This is the only notification you will receive while this person is away.

superkhau:
>FWIW, some time ago, I submitted a patch adding an install script to juggler that could be used by connectors to pull in the right version of dependencies needed by the tests, see #456. Perhaps it's time to resurrect it?

I don't see how this is related to the upgrade. The connector tests must be fixed individually to support 8.0.2 syntax. IIUC, the devDeps are a separate issue altogether.

@raymondfeng I'm going to run tests on individually connectors to see how broken they are before merging this. If they are relatively small and simple to fix, I will merge it in (I don't see any issues unless I run into platform specific issues in the tests). Then commmit the fixes for each connector individually (so that they pass individually first), then rerun the juggler tests to make sure it passes after the upgrade too.

@bajtos I'm going to move ahead with this plan soon and since @raymondfeng is out until 2016/01/04, please let me know your thoughts.

bajtos:
> I don't see how this is related to the upgrade. The connector tests must be fixed individually to support 8.0.2 syntax. IIUC, the devDeps are a separate issue altogether.

Connectors are running a test suite that is defined here in juggler, see e.g. [test/common.batch.js](https://github.com/strongloop/loopback-datasource-juggler/blob/dfda177ff7a2f3f6dce924e00c479b546f68f420/test/common.batch.js). Now if you change those tests to depend on should.js@8 API and connectors keep installing let's say should.js@5, then the shared tests may start failing. Which will happen immediately after you land this PR. On the other hand, if you upgrade the connectors to use should.js@8 first, then they will start failing too, because the tests in juggler are still using should.js@5 API. Catch-22!

Having said that, updating all connectors to use the latest should.js version is always a good thing to do, I don't have any objections.

The only important thing is to eventually end up with all connectors and juggler passing CI tests. I don't mind much how do you get there. @superkhau

superkhau:
>Catch-22!

Exactly this. ;) That's why I was asking if it was safe to upgrade juggler first, but from your explanation, it doesn't matter which route.

Now, regarding https://github.com/strongloop/loopback-datasource-juggler/pull/456, I see what you're saying now. It's an interesting idea, but messing with the package.json at runtime does seem kind of evil versus just upgrading package.json. I think we need less coupling/magic between juggler and connectors, so I will just upgrade package.json manually and the connectors as needed. On the other hand I like the automation idea, so I think ponder over the 456 PR a bit more. ;)

superkhau:
Closing this in favour of https://github.com/strongloop/loopback-datasource-juggler/pull/804 which is rebased.

PR #798, opened 2015-12-21
Add parameter caseInsensitive to validatesUniquenessOf validator

### LINKED ISSUE #607: Validate unique field case insesitive
Hello,

we have tried today to add a unique validation to a model of countries on the isoCode. I think there is missing a flag boolean "caseSensitive". Right now we can have 2 countries with iso codes: de, DE an unique validation doesn't trigger error.

beeman:
This patch adds an option to check the uniqueness of a value in a case insensitive way. 

Should fix #607, though I'm not sure if I overlooked any edge cases.


raymondfeng:
I am out of the office until 01/04/2016.

Happy holidays!


Note: This is an automated response to your message
"[loopback-datasource-juggler] Add parameter caseInsensitive to
validatesUniquenessOf validator (#798)" sent on 12/21/15 12:41:05 PM.

This is the only notification you will receive while this person is away.

Issue #731, opened 2015-10-07
Unload a element on Operation hook loaded


FelipeAguayo:
I wish I could unload an item in 'loaded' hook, example:

result from datasource: [{id:1, name: 'example'}, {id:2, name: 'example2'}]

``` js
Model.observe('loaded', function (ctx, next, unload){
   if(ctx.instance.id === 2){
     unload();
   }else{
     next();
   }
});
```

result for request: [{id:1, name: 'example'}]

is possible?

superkhau:
@bajtos ^

bajtos:
Hi @FelipeAguayo, thank you for reporting this use case. Could you please provide more details on the filtering criteria you would like to apply?

If these criteria are static, then you can use the [access](https://docs.strongloop.com/display/LB/Operation+hooks#Operationhooks-access) hook to add extra filter conditions to all requests.

```js
// mock-up, not tested
Model.observe('access', function(ctx, next) {
  var extra = { id: { neq: 2 } };
  if (ctx.query.where) {
    ctx.query.where = { and: [ctx.query.where, extra] };
  } else {
    ctx.query.where = extra;
  }
  next();
});
```

FelipeAguayo:
the criteria are dynamic, depend on variables that are external to the loaded model, but this can't be validate in Operation Hook 'access'. An example of this would be:

datasource: mysql.

Model:
```json
{
  "name":"Model",
  "plural":"Models",
  "idInjection":true,
  "properties":{
     "name":"string"
  },
  "relations":{
    "relOtherModel":{
      "type": "belongsTo",
      "model": "OtherModel",
      "foreignKey": "idOtherModel"
    }
  }
}
```
values Model: 
```js
[ {id:1, idOtherModel:1, name:'one'}
, {id:2, idOtherModel:1, name:'two'}
, {id:3, idOtherModel:2, name:'three'}
, {id:4, idOtherModel:3, name:'four'}]
```
OtherModel:
```json
{
  "name":"OtherModel",
  "plural":"OtherModels",
  "idInjection":true,
  "properties":{
     "name":"string",
     "state":"string"
  }
}
```
values OtherModel:
```js
[ {id:1, name:'otherOne', state: 'created'}
, {id:2, name:'otherTwo', state: 'deleted'}
, {id:3, name:'otherThree', state: 'created'}]
```
after load:

```js
Model.observe('loaded', function (ctx, next, unload){
[...]

PR #812, opened 2016-01-05
Feature/all loaded hook + access callback in context

Juneil:
Hi, 

I would like to have a new operation hook triggered after a find query with the full results.

I did an enhancement for the access operation hook, it's providing the end callback in the context to be able to stop the flow without an error.

One test failed for this change, I didn't found a way to fix the Optimize or the Unoptimize test of findOrCreate.
Do you have any idea to fix them both ?

The purpose the these changes is to be able to plug a cache system between the 'access' and 'allLoaded' operation hooks.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -266,7 +266,7 @@ DataAccessObject.create = function (data, options, cb) {
     if (options.validate === undefined && Model.settings.automaticValidation === false) {
       return create();
     }
-    
+
     // validation required
     obj.isValid(function (valid) {
       if (valid) {
@@ -1516,7 +1516,18 @@ DataAccessObject.find = function find(query, options, cb) {
           results = geo.filter(results, near);
         }
 
-        cb(err, results);
+        if (options.notify === false || !(results && results.length > 0)) return cb(err, results);
+
+        context = {
+          Model: self,
+          hookState: hookState,
+          results: results,
+          options: options
+        };
+
+        self.notifyObserversOf('allLoaded', context, function(err) {
+          cb(err, results);
+        });
       });
     }
     else {
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -435,7 +435,8 @@ DataAccessObject.updateOrCreate = DataAccessObject.upsert = function upsert(data
     Model: Model,
     query: byIdQuery(Model, id),
     hookState: hookState,
-    options: options
+    options: options,
+    end: cb
   };
   Model.notifyObserversOf('access', context, doUpdateOrCreate);
 
@@ -1406,7 +1407,8 @@ DataAccessObject.find = function find(query, options, cb) {
         Model: self,
         query: query,
         hookState: hookState,
-        options: options
+        options: options,
+        end: cb
       };
       self.notifyObserversOf('access', context, function(err, ctx) {
         if (err) return cb(err);
@@ -1546,7 +1548,8 @@ DataAccessObject.find = function find(query, options, cb) {
       Model: this,
       query: query,
       hookState: hookState,
-      options: options
+      options: options,
+      end: cb
     };
     this.notifyObserversOf('access', context, function(err, ctx) {
       if (err) return cb(err);
@@ -1668,7 +1671,8 @@ DataAccessObject.remove = DataAccessObject.deleteAll = DataAccessObject.destroyA
       Model: Model,
       query: query,
       hookState: hookState,
-      options: options
+      options: options,
+      end: cb
     };
     Model.notifyObserversOf('access', context, function(err, ctx) {
         if (err) return cb(err);
@@ -1866,7 +1870,8 @@ DataAccessObject.count = function (where, options, cb) {
     Model: Model,
     query: { where: where },
     hookState: hookState,
-    options: options
+    options: options,
+    end: cb
   };
   this.notifyObserversOf('access', context, function(err, ctx) {
     if (err) return cb(err);
@@ -2109,7 +2114,8 @@ DataAccessObject.updateAll = function (where, data, options, cb) {
     Model: Model,
     query: { where: where },
     hookState: hookState,
-    options: options
+    options: options,
+    end: cb
   };
   Model.notifyObserversOf('access', context, function(err, ctx) {
     if (err) return cb(err);
@@ -2227,7 +2233,8 @@ DataAccessObject.prototype.remove =
         Model: Model,
         query: byIdQuery(Model, id),
         hookState: hookState,
-        options: options
+        options: options,
+        end: cb
       };
 
       Model.notifyObserversOf('access', context, function(err, ctx) {
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -62,7 +62,8 @@ module.exports = function(dataSource, should) {
 
             triggered.should.eql([
               'access',
-              'loaded'
+              'loaded',
+              'allLoaded'
             ]);
             done();
           });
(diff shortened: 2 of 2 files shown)


slnode:

Juneil:
Hi, 
Someone can help us ?
I made a schema to explain what we want to do. [Here](https://drive.google.com/open?id=0B2s4BXQtmGTAVkNSRDh6UkFMdFk)

Thanks,

bajtos:
Hi @Juneil, I am confused about this pull request. What are you trying to achieve? Please paste some code snippets showing the intended usage. We will also need you to add good test coverage for your proposed changes, see [test/persistence-hooks.suite.js](https://github.com/strongloop/loopback-datasource-juggler/blob/54ca06761063e6cab6bb3cac9b2b0c1a790e69bf/test/persistence-hooks.suite.js).

Juneil:
@bajtos Hi,
We want to implement a cache system in our API.
To do that, we want to use the operation hooks (Example):

    Model.observe('access', function (context, cb) {
       var name = context.Model.definition.name; // Get model name
       var query = JSON.stringify(context.query); // Stringify query
       var key = crypt.hash(name + query); // Concat and hash
       var cacheResult = cache.get(key); // Get cached result
       var cached = context.options.cached = !!cacheResult;
       context.options.keyCache = key;
       if (cached) return context.end(null, cacheResult); // If cached then bypass querying and return directly the result
       return cb();
    });

*New operation hook:

    Model.observe('all_loaded', function (context, cb) {
        if (context.options.cached) return cb(); // If already cached, just return callback
        cache.set(context.options.keyCache, context.results);
        return cb();
    });

Our API is starting to have huge traffic and to gain perfomance, we need to implement a cache system. This is the cleanest way to do it without change to many things.

slnode:

bajtos:
Hi @Juneil, thank you for providing context, now I better understand what you are trying to achieve. Your use case makes perfect sense.

However, I think there is a major flaw in your first code snippet using "access" hook: this hooks is triggered for all operations (read and write), not only for queries. For example, when you call `Customer.updateAttributes` (`PUT /api/customers/:id`), it triggers the "access" hook too.

Let's discuss your requirements. Would you like to cache all read operations (including requests for a single entity - either via `findById`/`findOne`, or in operations like `updateOrCreate` that call `find*` under the hood), or only cache the result when the user calls `find` via REST API?

slnode:

bajtos:
Sorry for accidentally closing the PR, I have reopened it again.

Juneil:
HI @bajtos, Mostly `find`, `findById` and `findOne`, we don't need the others.
Is it possible to identify the method ? I didn't see a SharedMethod instance in the context.

bajtos:
@Juneil I see, thank you for the confirmation. The thing is, `find*` methods can be invoked internally by other operations, e.g. `findOrCreate` may call `find` and then `create`.

If it's ok to limit the caching mechanism to requests made via REST API (but not when calling `find*` methods directly from javascript), then I would advise you to use remote hooks instead.

```js
Model.beforeRemote('find', function (context, cb) {
   var name = context.method.ctor.definition.name; // Get model name
   var query = JSON.stringify(ctx.args.filter.where); // Stringify query
   var key = crypt.hash(name + query); // Concat and hash
   var cacheResult = cache.get(key); // Get cached result
   var cached = context.cached = !!cacheResult;
   context.keyCache = key;
   if (!cached) cb();

   context.result = cacheResult; // If cached then bypass querying and return directly the result
   context.done(function(err) {
     if (err) return cb(err);
     // else don't call cb() - the request is already handled
   });
});

Model.afterRemote('find', function(context, cb) {
  cache.set(context.keyCache, context.result);
  cb();
});
```

Note the code above is a mock-up, I didn't test it.

It has few down-sides: 
 - no "after remote" hooks are triggered
 - for "before remote" hooks registered after this caching hook was registered, they won't be triggered either

I think your use case is a good example where [Per-method hooks](https://github.com/strongloop/loopback-datasource-juggler/issues/397) would be very helpful (but they are not implemented yet).

Juneil:
Hi @bajtos, We use remote methods that use `find` internally. So we want to handle all READ operations.
Per-method hooks could be a solution, if it's called at any case. Not just REST calls.
Is it planned ? Because we really need that cache system feature.

Juneil:
Hi @bajtos ,
The flaw with access is not really an issue. We can identify with the `query` object.
The purpose of this PR is to add two features:
- Add end callback in the context
- Add new operation hook allLoaded

So is it possible to know if these features will be released ?
Thanks

bajtos:
> Add end callback in the context

I am personally not comfortable to land this change.

1. I think it can create confusion and situations that are hard to troubleshoot. When there are several observers calling `ctx.end`, then only the first observer will be executed. Now it's not clear which observer is the first one, because there are many places that can register observers: model definition files (`common/models/my-model.js`), mixins, boot scripts (`server/boot/cache.js`), components (`node_modules/loopback-component-cache/index.js`).

2. When an observer short-circuits the operation, hooks down the line will be skipped. This breaks an important quarantee that operation observers are always called, regardless of the operation invoked (unless the operation fails).

Here is a concrete example: the change replication feature, which powers offline sync, uses "after save" and "after delete" hooks to detect changes. If another observer calls `ctx.end` before change-replication observers were called, the change replication will stop working.

> Add new operation hook allLoaded

I think this feature is a valid one, a similar hook was discussed in https://github.com/strongloop/loopback-datasource-juggler/issues/731#issuecomment-148298889

If possible, I would like to come up with a solution that supports both use cases.

> We use remote methods that use find internally. So we want to handle all READ operations.

Operations hooks are not silver bullet, I think this can be handled reasonably well without hooks. I'd like to describe few more approaches that may work well for you.

The idea is to either create a new base model or implement a mixin that will provide a cached version of the `find` method. This new method can either override the `find` method provided by datasource, or it can be a new method e.g. `findCached`.

Here is a sample code for illustration:

```js
// common/models/cached-model.json
{
  "name": "CachedModel",
  "base": "PersistedModel"
}

// common/models/cached-model.js
module.exports = function(CachedModel) {
  /*** A) overwrite find() for everybody ***/
  // wait until the datasource defined `find` method
  CachedModel.on('dataAccessConfigured', function() {
    var Model = this;
    Model._findLive = CachedModel.find;
    Model.find = function(filter,...,cb) {
      var cached = // check the cache
      if (cached) 
        cb(null, cached);
      else
        Model._findLive(filter,...,cb);
    };
  });

  /*** or B) create new method ***/
  CachedModel.setup = function() {
    var Model = this;
    Model.findCached = function(filter,...,cb) {
      var cached = // check the cache
      if (cached) 
        cb(null, cached);
      else
        this.find(filter,...,cb);
    };
    // remap "GET /" from find() to findCached()
    Model.disableRemoteMethod('find', true);
    Model.remoteMethod('findCached', {
      isStatic: true,
      description: 'Find all instances of the model matched by filter from the data source.',
      accessType: 'READ',
      accepts: {arg: 'filter', type: 'object', description: 'Filter defining fields, where, include, order, offset, and limit'},
      returns: {arg: 'data', type: [typeName], root: true},
      http: {verb: 'get', path: '/'}
  });
};
```

Let's move the discussion about how to implement a caching layer into a new GH issue: https://github.com/strongloop/loopback/issues/2076

bajtos:
> The idea is to either create a new base model or implement a mixin that will provide a cached version of the find method. This new method can either override the find method provided by datasource, or it can be a new method e.g. findCached.

In my comment above, I forgot to mention how to use the new `CachedModel`. The idea is that all application models (e.g. `Customer`, `Product`) will use `CachedModel` instead of `PersistedModel` as the base class.

bajtos:
I'm closing this pull request in favour of strongloop/loopback#2076

PR #788, opened 2015-12-08
Implementation of replaceOrCreate and replace

Amir-61:
See also https://github.com/strongloop-internal/scrum-loopback/issues/621

##### Added functionality:

* Initial implemetation for replaceAttributes method
* Initial implemetation for replaceOrCreate

##### Other notes:

* These changes are related to the changes in the pull requests under `loopback-connector` and `loopback-connector-mysql`
* It seems test cases cannot be run on windows machines; I need to dig into it.

Connect to a spike task: https://github.com/strongloop-internal/scrum-loopback/issues/621

PR #815, opened 2016-01-05
Implement `findOrCreate` for memory connector

Amir-61:
##### Related task:
Connect to: strongloop/loopback/issues/1864

##### Side Notes: 
`Loopback-connector-redis` is one of the connectors, which is not compatible with tests on `loopback-datasource-juggler` and it is expected to fail! For instance, there is no `Null` return value from `Loopback-connector-Redis` and instead it returns empty string and many other incompatibilities,...

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -189,7 +189,7 @@ Memory.prototype.define = function defineModel(definition) {
   if(!this.collection(m)) this.initCollection(m);
 };
 
-Memory.prototype.create = function create(model, data, options, callback) {
+Memory.prototype._createSync = function(model, data, fn) {
   // FIXME: [rfeng] We need to generate unique ids based on the id type
   // FIXME: [rfeng] We don't support composite ids yet
   var currentId = this.collectionSeq(model);
@@ -207,18 +207,27 @@ Memory.prototype.create = function create(model, data, options, callback) {
   var idName = this.idName(model);
   id = (props[idName] && props[idName].type && props[idName].type(id)) || id;
   this.setIdValue(model, data, id);
-  if(!this.collection(model)) {
+  if (!this.collection(model)) {
     this.collection(model, {});
   }
 
-  if (this.collection(model)[id]) {
-    return process.nextTick(function() {
-      callback(new Error('Duplicate entry for ' + model + '.' + idName));
-    });
-  }
+  if (this.collection(model)[id])
+    return fn(new Error('Duplicate entry for ' + model + '.' + idName));
 
   this.collection(model)[id] = serialize(data);
-  this.saveToFile(id, callback);
+  fn(null, id);
+};
+
+Memory.prototype.create = function create(model, data, options, callback) {
+  var self = this;
+  this._createSync(model, data, function(err, id) {
+    if (err) {
+      return process.nextTick(function() {
+        callback(err);
+      });
+    };
+    self.saveToFile(id, callback);
+  });
 };
 
 Memory.prototype.updateOrCreate = function (model, data, options, callback) {
@@ -234,6 +243,36 @@ Memory.prototype.updateOrCreate = function (model, data, options, callback) {
         callback(err, data, { isNewInstance: true });
       });
     }
+  });
+};
+
+Memory.prototype.findOrCreate = function(model, filter, data, callback) {
+  var self = this;
+  var nodes = self._findAllSkippingIncludes(model, filter);
+  var found = nodes[0];
+
+  if(!found) {
+    // Calling _createSync to update the collection in a sync way and to guarantee to create it in the same turn of even loop
+    return self._createSync(model, data, function(err, id) {
+      if (err) return callback(err);
+      self.saveToFile(id, function(err, id) {
+        self.setIdValue(model, data, id);
+        callback(err, data, true);
+      });
+    });
+  }
+
+  if (!filter || !filter.include) {
+    return process.nextTick(function() {
+      callback(null, found, false);
+    });
+  }
+   
+  self._models[model].model.include(nodes[0], filter.include, {}, function(err, nodes) {
+    process.nextTick(function() {
+      if (err) return callback(err);
+      callback(null, nodes[0], false);
+    });
   });
 };
 
@@ -312,8 +351,7 @@ function getValue(obj, path) {
   return val;
 }
 
-Memory.prototype.all = function all(model, filter, options, callback) {
-  var self = this;
+Memory.prototype._findAllSkippingIncludes = function(model, filter) {
   var nodes = Object.keys(this.collection(model)).map(function (key) {
     return this.fromDb(model, this.collection(model)[key]);
   }.bind(this));
@@ -364,15 +402,8 @@ Memory.prototype.all = function all(model, filter, options, callback) {
     var limit = filter.limit || nodes.length;
     nodes = nodes.slice(skip, skip + limit);
   }
-
-  process.nextTick(function () {
-    if (filter && filter.include) {
-      self._models[model].model.include(nodes, filter.include, options, callback);
-    } else {
-      callback(null, nodes);
-    }
-  });
-
+  return nodes;
+  
   function sorting(a, b) {
     var undefinedA, undefinedB;
 
@@ -391,6 +422,19 @@ Memory.prototype.all = function all(model, filter, options, callback) {
 
     return 0;
   }
+};
+
+Memory.prototype.all = function all(model, filter, options, callback) {
+  var self = this;
+  var nodes = self._findAllSkippingIncludes(model, filter);
+
+  process.nextTick(function() {
+    if (filter && filter.include) {
+      self._models[model].model.include(nodes, filter.include, options, callback);
+    } else {
+      callback(null, nodes);
+    }
+  });
 };
 
 function applyFilter(filter) {
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -758,12 +758,12 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
 
         // options has precedence on model-setting
         if (options.validate === false) {
-          _findOrCreate(query, data, obj);
+          return _findOrCreate(query, data, obj);
         }
 
         // only when options.validate is not set, take model-setting into consideration
         if (options.validate === undefined && Model.settings.automaticValidation === false) {
-          _findOrCreate(query, data, obj);
+          return _findOrCreate(query, data, obj);
         }
 
         // validation required
--- a/test/memory.test.js
+++ b/test/memory.test.js
@@ -345,7 +345,6 @@ describe('Memory connector', function() {
         }
       }, function (err, users) {
         should.not.exist(err);
-        console.log(users);
         users.length.should.be.equal(5);
         done();
       });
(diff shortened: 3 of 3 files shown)


lib/dao.js
```diff
@@ -757,12 +757,12 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb)
 
         // options has precedence on model-setting
         if (options.validate === false) {
-          _findOrCreate(query, data, obj);
+          return _findOrCreate(query, data, obj);
```
Amir-61: I believe this was a bug; irrespective of whether `validate` is false or ... it was calling `_findOrCreate` after this block if the `obj` is valid; here was the block of code which was getting called:

```
        // validation required
        obj.isValid(function (valid) {
          if (valid) {
            _findOrCreate(query, data, obj);
          } else {
            cb(new ValidationError(obj), obj);
          }
        }, data);
```

Amir-61:
Hi @bajtos,

Could you please review this patch as well?

Thanks again :-)

lib/dao.js
bajtos: It's a bad practice to mix unrelated changes (new features and bug fixes) in the same commit/pull request. Imagine yourself several months later, trying to figure out why exactly this line was changed, and `git blame` pointing you to a commit adding `replaceOrCreate` functionality. I would be confused.

So, could you please submit this fix as a new pull request and include a unit-test to reproduce the problem? We should land this fix ASAP and back-port it to 2.x too.

bajtos: > trying to figure out why exactly this line was changed, and git blame pointing you to a commit adding replaceOrCreate functionality.

Edit: I meant `findOrCreate` functionality.

lib/connectors/memory.js
```diff
@@ -207,17 +207,24 @@ Memory.prototype.create = function create(model, data, options, callback) {
...
   this.collection(model)[id] = serialize(data);
+  return id;
+}
+
+Memory.prototype.create = function create(model, data, options, callback) {
+  try {
+    var id = this._create(model, data);
+  } catch (e) {
+    return callback(e);
+  }
```
bajtos: Mixing callback-based error handling with try/catch is not a good idea. Since most Node.js code never throws, it's not a common practice to wrap calls in try/catch blocks, thus we may forget to handle errors in future changes and cause crashes.

There is also (micro)optimization concern: IIRC, V8 does not optimize functions containing try/catch blocks.

I am proposing to rework the `_create` method to accept the usual error-first callback function.


```js
Memory.prototype.create = function(...) {
  var self = this;
  this._create(model, data, function(err, id) {
    if (err) return callback(err);
    // ...
  });
}
```

lib/connectors/memory.js
```diff
@@ -312,8 +351,7 @@ function getValue(obj, path) {
   return val;
 }
 
-Memory.prototype.all = function all(model, filter, options, callback) {
-  var self = this;
+Memory.prototype._findAll = function(model, filter) {
```
bajtos: IIRC, this new method ignores `filter.include`, which is not clear from the name. Please rename, e.g. to something like `_findAllSkippingIncludes`

lib/connectors/memory.js
```diff
@@ -237,6 +244,38 @@ Memory.prototype.updateOrCreate = function (model, data, options, callback) {
   });
 };
 
+Memory.prototype.findOrCreate = function(model, filter, data, callback) {
+  var self = this;
+  var nodes = self._findAll(model, filter);
+
+  if (nodes.length > 0) {
+    if (filter && filter.include) {
+      self._models[model].model.include(nodes[0], filter.include, {}, function(err, nodes) {
```
bajtos: @raymondfeng is it expected for `findOrCreate` to honour `filter.include`? Do other connectors (e.g. MongoDB or MySQL) process this filter property? If not, then I am proposing to simplify this code and ignore `filter.include`.

lib/connectors/memory.js
```diff
@@ -237,6 +244,38 @@ Memory.prototype.updateOrCreate = function (model, data, options, callback) {
...
+        process.nextTick(function() {
+          if (err)  callback(err);
+          else
+            callback(null, nodes[0], false);
```
bajtos: Style: too many spaces after `if (err)`, plus inconsistency in which statements are on the same line vs. on a new line.

We usually use one of these two ways:

```js
if (err) return callback(err);
callback(null, nodes[0], false);

// or

if (err)
  callback(err);
else
  callback(null, nodes[0], false);
```

lib/dao.js
bajtos: Hmm, I got confused here. Since this pull request is changing findOrCreate and we need to back-port it to 2.x in whole, it's ok to include related fixes. Just make sure the fix is in a standalone commit and includes unit-tests.

bajtos:
@Amir-61 reviewed, see the comments above. I am reassigning this PR back to you.

lib/connectors/memory.js
Amir-61: It seems Mongo is using `filter.include` for  Please see: https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L1134-L1137

lib/dao.js
Amir-61: In fact if we don't have this fix, 16 test cases fail; for instance, [this](https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/optional-validation.test.js#L435-L437) is one of the test cases which fails; in other words, as long as we have the implementation for 'findOrCreate' in the connector and the code does not run the [workaround](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L779-L797) we have these failures; so I believe test cases cover this. Please let me know if you think otherwise?

PS: I commit this fix separately...

Amir-61:
Hey @bajtos,

I applied your feedback and left a few comments; I'm assigning this PR to you now. Please review.

Thanks!

lib/connectors/memory.js
```diff
@@ -207,18 +207,23 @@ Memory.prototype.create = function create(model, data, options, callback) {
...
 
   this.collection(model)[id] = serialize(data);
-  this.saveToFile(id, callback);
+  callback(null, id);
+}
+
+Memory.prototype.create = function create(model, data, options, callback) {
+  var self = this;
+  this._create(model, data, function(err, id) {
+    if (err) return callback(err);
```
bajtos: I think you need to call `process.nextTick` here, because `_create` calls the callback immediately (sync).

lib/connectors/memory.js
```diff
@@ -237,6 +242,34 @@ Memory.prototype.updateOrCreate = function (model, data, options, callback) {
...
+  } else {
+    this._create(model, data, function(err, id) {
+      if (err) return callback(err);
+      self.saveToFile(id, function(err, id) {
+        self.setIdValue(model, data, id);
+        callback(err, data, true);
+      });
+    });
+  }
+};
```
bajtos: Nitpick, feel free to ignore: long if/else branches are difficult to read, using an early returns (possibly with inverse logic in `if`) is usually better. In this case, I would personally structure the code as follows - notice how the last block has 2-levels of indentation less:

```js
Memory.prototype.findOrCreate = function(model, filter, data, callback) {
  var self = this;
  var nodes = self._findAllSkippingIncludes(model, filter);
  var found = nodes[0];

  if (!found) {
    return this._create(model, data, function(err, id) {
      if (err) return callback(err);
      self.saveToFile(id, function(err, id) {
        self.setIdValue(model, data, id);
        callback(err, data, true);
      });
    });
  }

  if (!filter || !filter.include) {
    return process.nextTick(function() {
      callback(null, found, false);
    });
  }

  self._models[model].model.include(found, filter.include, {}, function(err, nodes) {
    process.nextTick(function() {
      if (err) return callback(err);
      callback(null, found, false);
    });
  });
};
```

lib/dao.js
bajtos: Oh, I see! In that case it's ok to not include any additional tests. I think we should strive to have all commits leaving the repo in a state that passes all unit-tests. In this case, it means reordering the two commits - first fix the issue, then add the new changes that would fail without the fix. You should be able to easily reorder commits via `git rebase -i`

bajtos:
Few more comments to address, see above.

@raymondfeng could you please take over the review of this patch and give it the final approval?

lib/dao.js
Amir-61: :+1:

lib/connectors/memory.js
Amir-61: That looks easier to ready! Thanks!

Amir-61:
Hi @raymondfeng,

I applied Miroslav's feedback; Could you please review this patch.

I appreciate it!

lib/connectors/memory.js
```diff
@@ -207,18 +207,27 @@ Memory.prototype.create = function create(model, data, options, callback) {
...
     this.collection(model, {});
   }
 
-  if (this.collection(model)[id]) {
-    return process.nextTick(function() {
-      callback(new Error('Duplicate entry for ' + model + '.' + idName));
-    });
-  }
+  if (this.collection(model)[id])
+    return callback(new Error('Duplicate entry for ' + model + '.' + idName));
```
raymondfeng: Why did you remove the process.nextTick wrapper which ensures the callback to invoked after this function returns?

Amir-61: _create is `synchronous` (I took out `saveToFile` which was async to handle race codition); do I still need to have a process.nextTick wrapper here?

raymondfeng: It's very confusing to see a synchronous function that takes a callback. Why do we need to make `_create` synchronous?

Amir-61: Hey Raymond,

This is the reason why we decided to have this refactoring for create and use synchronous _create:

Miroslav: "it may be better to not use `this.create`, because it's not clear when exactly it creates the record (whether it guarantees to create it in the same turn of even loop)
and do what Ryan suggested - refactor the creation into a private method that updates the collection in a sync way (no file writing!) and then call it from both `this.create` and `this.findOrCreate"

>It's very confusing to see a synchronous function that takes a callback. Why do we need to make _create synchronous?

I was returning or throwing an exception for this function but I understood that is not a good practice based on [this comment](https://github.com/strongloop/loopback-datasource-juggler/pull/815#discussion_r48939890)

test/memory.test.js
```diff
@@ -695,6 +695,38 @@ describe('Memory connector', function() {
...
+          data.length.should.equal(1);
+          data[0].color.should.equal('white');
+          creationNum.should.equal(1);
+          done();
+        });
+      });
+    });
+  });
+
+
```
Amir-61: This test was failing before this PR and now it is passing; however, I'm not sure whether I defined this test in the right place or not,...

Amir-61:
Hi @raymondfeng,

I added a test case, which was failing before this patch and now it is passing as a result of atomic implementation of `findOrCreate` in memory connector; could you please review it again?

Thanks a lot :-)

raymondfeng:
Please remove `console.log` at https://github.com/strongloop/loopback-datasource-juggler/blob/findOrCreate/test/memory.test.js#L348 too.

Amir-61:
@raymondfeng 

>Please remove console.log at https://github.com/strongloop/loopback-datasource-juggler/blob/findOrCreate/test/memory.test.js#L348 too.

I removed that extra log as well; could you please review?

[PR#788](https://github.com/strongloop/loopback-datasource-juggler/pull/788) is dependent to this PR#815 because of [this comment](https://github.com/strongloop/loopback-datasource-juggler/pull/788/files#r49378429) in PR#788.

Question: What is your suggestion based on [this comment](https://github.com/strongloop/loopback-datasource-juggler/pull/788/files#r49378429) in PR#788; once this patch (PR#815) gets reviewed and merged to master, should I merge master to [my PR#788](https://github.com/strongloop/loopback-datasource-juggler/pull/788)? In other words for implementing atomic `replaceOrCreate` in memory I need to use synchronous `_findAllSkippingIncludes` and synchronous `_create` which are implemented in this PR#815

Thanks!

raymondfeng:
LGTM

Amir-61:
@slnode test please

Amir-61:
@slnode test please.

Amir-61:
@slnode test please.

Amir-61:
@slnode test ok

superkhau:
@raymondfeng @Amir-61 I believe this PR is causing CI to fail. See CI #2806 for the LoopBack repository.

superkhau:
Sorry wrong post. I meant http://ci.strongloop.com/job/loopback/2806/

PR #823, opened 2016-01-19
Prevent constructor to be property name

jannyHou:
Connect to strongloop-internal/scrum-loopback#642

Hey @bajtos this pr is created for handling `constructor` in two places:
1) Throw error in `/lib/model-builder.js` if model has a property named `constructor`.
2) Give warning in `/lib/model.js` if user assigns property named `constructor` in instance.

Could you review it? Thanks!
Move to this pr since the previous one keep failing even I commented all of my changes. Thanks for your understanding.

### DIFF
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -257,6 +257,12 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
       deprecated('Property names containing a dot are not supported. ' +
         'Model: ' + className + ', property: ' + p);
     }
+    
+    // Warn if property name is 'constructor'
+    if (p === "constructor") {
+      deprecated('Property name should not be "constructor" in Model: ' +
+        className);
+    }
   }
 
   var modelDefinition = new ModelDefinition(this, className, properties, settings);
--- a/lib/model.js
+++ b/lib/model.js
@@ -63,6 +63,11 @@ ModelBaseClass.prototype._initProperties = function (data, options) {
   var self = this;
   var ctor = this.constructor;
 
+  if (typeof data !== 'undefined' && 
+      typeof (data.constructor) !== 'function') {
+    throw new Error('Property name "constructor" is not allowed in ' + ctor.modelName +' data');
+  }
+
   if(data instanceof ctor) {
     // Convert the data to be plain object to avoid pollutions
     data = data.toObject(false);
--- a/test/model-definition.test.js
+++ b/test/model-definition.test.js
@@ -379,6 +379,15 @@ describe('ModelDefinition class', function () {
     message.should.match(/Dotted.*dot\.name/);
   });
 
+  it('should report deprecation warning for property named constructor', function() {
+    var message = 'deprecation not reported';
+    process.once('deprecation', function(err) { message = err.message; });
+
+    memory.createModel('Ctor', { 'constructor': String });
+
+    message.should.match(/Property name should not be "constructor" in Model: Ctor/);
+  });
+
   it('should report deprecation warning for dynamic property names containing dot', function(done) {
     var message = 'deprecation not reported';
     process.once('deprecation', function(err) { message = err.message; });
@@ -387,6 +396,14 @@ describe('ModelDefinition class', function () {
     Model.create({ 'dot.name': 'dot.value' }, function(err) {
       if (err) return done(err);
       message.should.match(/Dotted.*dot\.name/);
+      done();
+    });
+  });
+
+  it('should throw error for dynamic property named constructor', function(done) {
+    var Model = memory.createModel('DynamicCtor');
+    Model.create({ 'constructor': 'myCtor' }, function(err) {
+      assert.equal(err.message, 'Property name "constructor" is not allowed in DynamicCtor data');
       done();
     });
   });


lib/model.js
```diff
@@ -61,7 +61,11 @@ function ModelBaseClass(data, options) {
  */
 ModelBaseClass.prototype._initProperties = function (data, options) {
   var self = this;
-  var ctor = this.constructor;
+  if (typeof this.constructor !== 'function') {
```
bajtos: Shouldn't you check `data.constructor` instead?

bajtos:
@jannyHou please add a unit-test to verify your changes. I am not sure if we can intercept `deprecated` calls, take a look at other code using it to see if there are any existing examples.

lib/model.js
jannyHou: sure, I rewrite my code.

jannyHou:
hey @bajtos I got the idea of `deprecated` from here:
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model.js#L223
That's why I consider it as a possible way to handle this error.

jannyHou:
@slnode test please

package.json
```diff
@@ -26,7 +26,7 @@
     "clean": "make clean",
     "help": "make help",
     "refresh": "make refresh",
-    "test": "make test"
+    "test": "mocha -R list"
```
bajtos: Please don't change the reported used by `npm test` and revert this line. If it helps you to use a different reporter during development, you can always invoke mocha manually instead of calling `npm test`.

test/model-definition.test.js
```diff
@@ -391,6 +400,15 @@ describe('ModelDefinition class', function () {
     });
   });
 
+  it('should throw error for dynamic property named constructor', function(done) {
+
```
bajtos: Extra empty line, please remove.

test/model-definition.test.js
```diff
@@ -391,6 +400,15 @@ describe('ModelDefinition class', function () {
...
+    var Model = memory.createModel('DynamicCtor');
+    Model.create({ 'constructor': 'myCtor' }, function(err) {
+      should.exist(err);
```
bajtos: I think this test would pass even without your change, because we were throwing an error when trying to use `this.constructor` which was not a function. Please extend the check and verify that the messages matches the custom error.

lib/model.js
```diff
@@ -63,6 +63,11 @@ ModelBaseClass.prototype._initProperties = function (data, options) {
   var self = this;
   var ctor = this.constructor;
 
+  if (typeof data !== 'undefined' && 
+     typeof (data.constructor) !== 'function') {
+    throw new Error('Property names could not be constructor ');
```
bajtos: Please remove the trailing space from the message.

I am proposing to reword the message a bit:

```js
throw new Error('Property name "constructor" is not allowed in ' + ctor.modelName +' data');
```

bajtos:
@jannyHou your current solution is good. Please address my minor comments above and squash all commits into a single one.

package.json
superkhau: It was actually suggested to me by @rmg before we should actually not use make for npm test and use mocha directly for CI reasons. @rmg Can you comment here?

rmg: @superkhau is correct in that I recommend running mocha directly from `npm test`.

@bajtos is correct in that it should not be specifying a reporter.

I recommend the test script in package.json simply be `"mocha test/*.test.js"` and that the `Makefile` be used for developer oriented actions so that `npm test` can be reliably used by CI. This will allow CI to better report on the results.

superkhau: @rmg TY, I didn't notice the `-R` was for reporter, thought it was `-r` for recursive. Checked the docs anyways and yes @bajtos is correct that it should not specify a reporter.

jannyHou:
Thank you @bajtos I applied your feedback, and I will test Ryan's suggestion `mocha test/*.test.js` tomorrow( don't have a laptop with dev env at hand when I saw his comment, sorry!)

package.json
bajtos: I usually use the simplest form, which is `"test": "mocha"`. Mocha picks `test/*` by default, which is good enough in most cases.

Note that some of our repositories (most notably juggler and loopback) abuse the fact that mocha picks non-test files too, put helpers to `test/support.js` and expose them on global scope.  In that case, `mocha test/*.test.js*` would cause test failures.

lib/model-builder.js
```diff
@@ -257,6 +257,12 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
       deprecated('Property names containing a dot are not supported. ' +
         'Model: ' + className + ', property: ' + p);
     }
+    
+    // Warn if property name is 'constructor'
+    if (p === "constructor") {
+      deprecated('Property name should not be constructor in Model: ' +
```
bajtos: Could you please put "constructor" in quotes too? `...should not be "constructor" in Model:...`

bajtos:
One last nitpick to address, the patch LGTM otherwise.

@jannyHou when done, please squash the commits into a single one and land the PR without waiting for further reviews.

jannyHou:
@slnode test please

jannyHou:
@slnode test please

lib/model.js
```diff
@@ -63,6 +63,11 @@ ModelBaseClass.prototype._initProperties = function (data, options) {
   var self = this;
   var ctor = this.constructor;
 
+  if (typeof data !== 'undefined' && 
+     typeof (data.constructor) !== 'function') {
```
superkhau: one more space in front of typeof

PR #826, opened 2016-01-21
Refactor `updateAttributes`

Amir-61:
Re-factore `updateAttributes` to have code re-usability for newer methods

PR #839, opened 2016-02-02
Fix conversion for `updateAttributes`

Amir-61:
Hi Miroslav,

I'm so sorry; [my earlier PR](https://github.com/strongloop/loopback-datasource-juggler/pull/826) was causing one of the test failures in loopback-connector-mongodb.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -58,11 +58,11 @@ function copyData(from, to) {
   }
 }
 
-function convertDataPropertiesByType(data) {
+function convertSubsetOfPropertiesByType(inst, data) {
   var typedData = {};
   for (var key in data) {
     // Convert the properties by type
-    typedData[key] = data[key];
+    typedData[key] = inst[key];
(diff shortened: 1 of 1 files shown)


Amir-61:
@slnode test please

lib/dao.js
```diff
@@ -58,11 +58,12 @@ function copyData(from, to) {
   }
 }
 
-function convertDataPropertiesByType(data) {
+function convertDataPropertiesByType(inst, data) {
   var typedData = {};
   for (var key in data) {
     // Convert the properties by type
-    typedData[key] = data[key];
+    inst[key] = data[key];
```
bajtos: I am confused. I though `inst` was updated by calling `copyData` just before calling `convertDataPropertiesByType`? If we are updating the instance as part of this method, then 1) do we need `copyData` at all? 2) the name of this method should be changed to better communicate the intent.

bajtos:
LGTM, :ship:

PR #847, opened 2016-02-06
Update describe-operation-hooks

bajtos:
- drop custom findOrCreate implementation - memory connector provides
   an atomic implementation out of the box now (see https://github.com/strongloop/loopback-datasource-juggler/pull/815)
 - add new methods `replaceOrCreate` and `replaceByID` (see https://github.com/strongloop/loopback-datasource-juggler/pull/788) 
 - fix error reporting to include the stack trace in the console output

/to @Amir-61 please review


Amir-61:
LGTM, thanks :-)

PR #856, opened 2016-02-14
Removed DataType condition that breaks $exists lookups in mongoDB

jsheely:
I'm not entirely sure what this DataType(val) check is even needed for.

But in a query like this: 

```
http://localhost:4001/api/products?filter[where][tags][exists]=true
```

This method will return: "[object, Object]" instead of using the actual object {exists: true} which will get converted later in the buildWhere in mongoDB to {$exists: true}

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1207,7 +1207,9 @@ DataAccessObject._coerce = function (where) {
         } else if (operator === 'regexp' && val instanceof RegExp) {
           // Do not coerce regex literals/objects
         } else if (!((operator === 'like' || operator === 'nlike') && val instanceof RegExp)) {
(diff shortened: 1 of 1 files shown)


slnode:

jsheely:
This is related to this issue: https://github.com/strongloop/loopback/issues/1009 for mongoDB

jfurie:
+1

ssh24:
@jsheely Looks like you had this branch on your old fork which you deleted. If you look at the top it says:
**jsheely  wants to merge 2 commits into strongloop:master from unknown repository**

Closing this PR for that reason. Please create a new one and assign me to it so that I can take a look.

PR #859, opened 2016-02-26
Improve error message on connector init error

bajtos:
Before this change, after adding a Cloudant-backed datasource to an app, `node .` will fail with
```
AssertionError: url is not valid
  (...stack trace...)
```

Now:

```
AssertionError: Cannot initialize connector "cloudant": url is not valid
  (...stack trace...)
```

/to @ritch please review

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -327,13 +327,22 @@ DataSource.prototype.setup = function (name, settings) {
 
     }.bind(this);
 
-    if ('function' === typeof connector.initialize) {
-      // Call the async initialize method
-      connector.initialize(this, postInit);
-    } else if ('function' === typeof connector) {
-      // Use the connector constructor directly
-      this.connector = new connector(this.settings);
-      postInit();
+    try {
+      if ('function' === typeof connector.initialize) {
+        // Call the async initialize method
+        connector.initialize(this, postInit);
+      } else if ('function' === typeof connector) {
+        // Use the connector constructor directly
+        this.connector = new connector(this.settings);
+        postInit();
+      }
+    } catch(err) {
+      if (err.message) {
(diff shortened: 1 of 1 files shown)


lib/datasource.js
```diff
@@ -327,13 +327,22 @@ DataSource.prototype.setup = function (name, settings) {
...
-      // Call the async initialize method
-      connector.initialize(this, postInit);
-    } else if ('function' === typeof connector) {
-      // Use the connector constructor directly
-      this.connector = new connector(this.settings);
-      postInit();
+    try {
+      if ('function' === typeof connector.initialize) {
+        // Call the async initialize method
+        connector.initialize(this, postInit);
```
ritch: ugh... we should remove the stupid `initialize` faux constructor at some point

lib/datasource.js
```diff
@@ -327,13 +327,22 @@ DataSource.prototype.setup = function (name, settings) {
...
+      } else if ('function' === typeof connector) {
+        // Use the connector constructor directly
+        this.connector = new connector(this.settings);
+        postInit();
+      }
+    } catch(err) {
+      if (err.message) {
+        err.message = 'Cannot initialize connector ' +
```
ritch: :+1: nice

ritch:
LGTM

lib/datasource.js
bajtos: > we should remove the stupid initialize faux constructor at some point

IIRC, all our connectors use `connector.initialize` instead of a class constructor :(

/cc @superkhau perhaps this is something you can include in your refactoring of connectors?

superkhau: Makes sense to refactor this. Created waffle task https://github.com/strongloop-internal/scrum-loopback/issues/797 to track this.

PR #860, opened 2016-02-27
fix nin support for in memory datasource

horiaradu:


slnode:

Multiply:
We have serveral of PRs for this. I wish for just one of them to come into the core. :P

Issue #800, opened 2015-12-22
Model.find returns items with incorrect order


trongnd:
I'm using loopback for mongodb and facing a problem when trying **Model.find** with **order**:

```javascript
User.find({
    order: 'createdAt DESC',
    limit: 10,
    include: ['publisher']
});
```
The problem here is the order of returned items is incorrect. I have inspect some code I find out the problem caused by this code:

```javascript
 // master/lib/dao.js#L1455
async.each(data, function(item, callback) {
....
// master/lib/dao.js#L1502
results.push(obj);
```
Because **async.each** run in parallel so **results.push(obj)** is not guaranteed to be run in correct order. That why I get back a list of items in incorrect order.

I'm currently using with the code below to temporarily fix the problem in own project:
```javascript
 // master/lib/dao.js#L1455
async.forEachOf(data, function(item, index, callback) {
....
// master/lib/dao.js#L1502
results[index] = obj;
```
It's working and also keeps the advantage of run in  parallel of async.each.

Own project is going to release soon so we really hope this problem will be fixed asap.
Thank you so much!

maximeblondeau:
Same issue for me, I used the mongo connector to solve this.

bajtos:
The pull request #864 contains a unit-test exposing the problem:

```js
    it('should query sorted desc by order integer field even though there' +
       'is an async model loaded hook', function(done) {
      User.observe('loaded', function(ctx, next) {
        var randomTimeoutTrigger = Math.floor(Math.random() * 100);

        setTimeout(function() {
          process.nextTick(next);
        }, randomTimeoutTrigger);
      });

      User.find({ order: 'order DESC' }, function(err, users) {
        should.exists(users);
        should.not.exists(err);
        users.pop().order.should.equal(1);
        users.pop().order.should.equal(2);
        users.shift().order.should.equal(6);
        done();
      });
    });
```

I am labeling this issue as a major bug and moving to Top of Backlog to prioritise higher. I think the fix should be reasonably easy, hence `beginner-friendly`.

bajtos:
Fixed in `loopback-datasource-juggler@2.47.0`

PR #864, opened 2016-03-03
Use eachSeries to make sure that the order is still the same

MichaelKohler:
(fixes issue #800)

### DIFF
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -127,7 +127,23 @@ describe('basic-querying', function() {
 
   describe('find', function() {
 
-    before(seed);
+    before(function() {
+      seed(function setupDelayingLoadedHook() {
+        User.observe('loaded', function(ctx, next) {
+          var randomTimeoutTrigger = Math.floor(Math.random() * 100);
+
+          setTimeout(function() {
+            process.nextTick(next);
+          }, randomTimeoutTrigger);
+        });
+      });
+    });
+
+    after(function removeDelayingLoadHook() {
+      User.observe('loaded', function(ctx, next) {
+        next();
+      });
+    });
 
     it('should query collection', function(done) {
       User.find(function(err, users) {
@@ -216,6 +232,20 @@ describe('basic-querying', function() {
         users.pop().name.should.equal('George Harrison');
         users.pop().name.should.equal('John Lennon');
         users.shift().name.should.equal('Stuart Sutcliffe');
+        done();
+      });
+    });
+
+    it('should query sorted desc by order integer field even though there' +
+       'is an async model loaded hook', function(done) {
+      User.find({ order: 'order DESC' }, function(err, users) {
+        if (err) return done(err);
+
+        should.exists(users);
+        should.not.exists(err);
+        users.pop().order.should.equal(1);
+        users.pop().order.should.equal(2);
+        users.shift().order.should.equal(6);
         done();
       });
     });
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1691,7 +1691,7 @@ DataAccessObject.find = function find(query, options, cb) {
   var allCb = function(err, data) {
     var results = [];
     if (!err && Array.isArray(data)) {
-      async.each(data, function(item, callback) {
+      async.eachSeries(data, function(item, callback) {
         var d = item;//data[i];
         var Model = self.lookupModel(d);
         if (options.notify === false) {


slnode:

Amir-61:
@MichaelKohler Thanks for the contribution. Please add test cases which fail with the previous implementation and it passes with your fixes.

Thanks

slnode:

MichaelKohler:
This is not reproducible with the current build, looks like something else interfered for me here. Therefore closing this.

MichaelKohler:
After further investigation on our side: actually, no, it needs to be possible to do async code within the function that is called in case you have a "loaded" observer here: https://github.com/strongloop/loopback-datasource-juggler/blob/7247b6637e098047daf8d73d4d864455d80a9566/lib/dao.js#L1761

I'll try to come up with a test case.

slnode:

MichaelKohler:
Turns out there are even more problems once something in the "loaded" observer is async. I've updated this PR with a test case. @bajtos could you please review? Thanks!

test/basic-querying.test.js
```diff
@@ -220,6 +220,26 @@ describe('basic-querying', function() {
       });
     });
 
+    it('should query sorted desc by order integer field even though there' +
+       'is an async model loaded hook', function(done) {
+      User.observe('loaded', function(ctx, next) {
```
bajtos: The `User` object is global and shared by all other tests. As can be seen from [Travis CI logs](https://travis-ci.org/strongloop/loopback-datasource-juggler/builds/124841135), other tests start failing with this change in place.

Now I think it is actually a good idea to run all `find` tests with this `loaded` hook, but we need to do it slightly differently:

1) Setup the observer in `before` hook
2) Remove the observer in `after` hook

```js
describe('find', function() {
  before(function setupDelayingLoadedHook() {
    // ...
  });

  after(function removeDelayingLoadHook() {
    // ...
  });

  // individual test cases come here
  it(...);
  // ...
});

**Comment by bajtos** [comment]
@MichaelKohler the current code version contains only the failing unit test without the fix itself. I'd like you to improve the change in tests, see my comment above.

**Comment by Amir-61** [comment]
@MichaelDiguet Please ping me here when you are done with modifications and I'll do my best to review it again...

Thanks :-)

**Comment by MichaelKohler** [comment]
@Amir-61 I've updated the patch, though I couldn't find any better way to unobserve 'loaded'. Is there a better way? I'd gladly use that one, but as I said, couldn't find one.

File: test/basic-querying.test.js [hunk 2]
```diff
@@ -220,6 +236,18 @@ describe('basic-querying', function() {
       });
     });
 
+    it('should query sorted desc by order integer field even though there' +
+       'is an async model loaded hook', function(done) {
+      User.find({ order: 'order DESC' }, function(err, users) {
+        should.exists(users);
+        should.not.exists(err);
```
Amir-61: Loopback team prefers this style, which provides more info:
`if (err) return done(err);`

And move it before `should.exists(users);`

```
if (err) return done(err);
should.exists(users);
...
```

Amir-61:
@slnode teats please

MichaelKohler:
@Amir-61 done

test/basic-querying.test.js
```diff
@@ -220,6 +236,20 @@ describe('basic-querying', function() {
       });
     });
 
+    it('should query sorted desc by order integer field even though there' +
+       'is an async model loaded hook', function(done) {
+      User.find({ order: 'order DESC' }, function(err, users) {
+        if (err) return done(err);
+
+        should.exists(users);
+        should.not.exists(err);
```
Amir-61: You are already reporting error on line 242 by the callback. You don't need this line anymore...

test/basic-querying.test.js
```diff
@@ -127,7 +127,23 @@ describe('basic-querying', function() {
...
+
+          setTimeout(function() {
+            process.nextTick(next);
+          }, randomTimeoutTrigger);
+        });
+      });
+    });
+
+    after(function removeDelayingLoadHook() {
+      User.observe('loaded', function(ctx, next) {
```
bajtos: It is possible to register multiple hook observers, this code does not remove the hook registered in `before` hook. You should call `removeObserver` API instead. It looks we don't include the observer API in the apidocs at apidocs.strongloop.com, here is the source code: https://github.com/strongloop/loopback-datasource-juggler/blob/2189b9a746071b599d6a129313b2cdd11052e36b/lib/observer.js#L45

I think the usage be like this:

```js
// put this function somewhere at the bottom of this file, where other helper methods are
function nextAfterDelay(ctx, next) {
  var randomTimeoutTrigger = Math.floor(Math.random() * 100);
  setTimeout(function() { process.nextTick(next); }, randomTimeoutTrigger);
}

before(seed);
before(function setupDelayingLoadedHook() {
  User.observe('loaded', nextAfterDelay);
});

after(function removeDelayingLoadedHook) {
  User.removeObserver('loaded', nextAfterDelay);
});
```

test/basic-querying.test.js
```diff
@@ -127,7 +127,23 @@ describe('basic-querying', function() {
 
   describe('find', function() {
 
-    before(seed);
+    before(function() {
```
bajtos: Note that your implementation has another problem - because your `setupDelayingLoadedHook` does not accept callback arg, mocha will think it's a sync function and won't wait for completion. As a result, some of the tests may run before the delaying observer is set up. Also you

test/basic-querying.test.js
```diff
@@ -127,7 +127,23 @@ describe('basic-querying', function() {
...
+      seed(function setupDelayingLoadedHook() {
```
bajtos: If `seed` fails, the error is silently ignored.

You can ignore this and the previous comment above if you use the approach outlined in https://github.com/strongloop/loopback-datasource-juggler/pull/864/files#r62497435.

test/basic-querying.test.js
```diff
@@ -220,6 +236,20 @@ describe('basic-querying', function() {
...
+    it('should query sorted desc by order integer field even though there' +
+       'is an async model loaded hook', function(done) {
+      User.find({ order: 'order DESC' }, function(err, users) {
+        if (err) return done(err);
+
+        should.exists(users);
+        should.not.exists(err);
+        users.pop().order.should.equal(1);
+        users.pop().order.should.equal(2);
+        users.shift().order.should.equal(6);
```
bajtos: When one of these three assertions fail, it's difficult to tell what exactly happened, as the error message shows only one item from the array.

I think this may be a better solution:

```js
var order = users.map(function(u) { return u.order; });
order.should.eql([1,2,3,4,5,6]);
```

test/basic-querying.test.js
bajtos: Edit: the observer API is included in our API docs, see http://apidocs.strongloop.com/loopback-datasource-juggler/#observermixin-removeobserver

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

Amir-61:
@MichaelKohler 
There some lint issues; could you please fix them:
```
  240:28  error  A space is required after ','  comma-spacing
  240:30  error  A space is required after ','  comma-spacing
  240:32  error  A space is required after ','  comma-spacing
  240:34  error  A space is required after ','  comma-spacing
  240:36  error  A space is required after ','  comma-spacing
```

Also please make sure you sign [CLA](https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler); last but not least you can trigger our CI tests by `@slnode ok to test` as @bajtos did in the previous comment.

Thanks

Amir-61:
Also the commit message does not comply with our linter (this [failure](https://cis-jenkins.swg-devops.com/job/strongloop-commit-linter-jobs/job/loopback-datasource-juggler/161//console)):
`https://cis-jenkins.swg-devops.com/job/strongloop-commit-linter-jobs/job/loopback-datasource-juggler/161//console`

Could you please fix the failures?

bajtos:
On the second thought, I think switching to serial execution is suboptimal, especially performance wise. The order of items can be preserved by using "async.map" and then filtering out extra "undefined" items, as I am proposing in #996. However, that patch is not ready yet, so it may make sense to land this fix first and then improve the performance later. I'll leave the decision up to you, @Amir-61.

MichaelKohler:
I could fix the linter failures tonight, but one commit is from @bennyt2 who hasn't signed the CLA, so he might need to do this if we want to land this one here. (see https://github.com/MichaelKohler/loopback-datasource-juggler/pull/1)

I'd say let's wait on @Amir-61's decision which PR we want to go for and then we can either close this one here or fix it. Is that okay?

test/basic-querying.test.js
```diff
@@ -129,6 +129,16 @@ describe('basic-querying', function() {
 
     before(seed);
 
+    before(function setupDelayingLoadedHook(cb) {
+      User.observe('loaded', nextAfterDelay);
+      cb();
```
Amir-61: `cb()` is not needed since `nextAfterDelay ` calls the callback.

test/basic-querying.test.js
```diff
@@ -129,6 +129,16 @@ describe('basic-querying', function() {
...
+    });
+
+    after(function removeDelayingLoadHook(cb) {
+      User.removeObserver('loaded', nextAfterDelay);
+      cb();
```
Amir-61: `cb()` not needed

test/basic-querying.test.js
```diff
@@ -129,6 +129,16 @@ describe('basic-querying', function() {
 
     before(seed);
 
+    before(function setupDelayingLoadedHook(cb) {
```
Amir-61: Nitpick: I would prefer adding a description of what you are doing here as opposed to have it as a function name:
```
before('setup delaying loaded hook', function(done) {
...
```

Feel free to ignore it.

test/basic-querying.test.js
```diff
@@ -220,6 +230,18 @@ describe('basic-querying', function() {
...
     });
 
+    it('should query sorted desc by order integer field even though there' +
+        'is an async model loaded hook', function(done) {
+      User.find({ order: 'order DESC' }, function(err, users) {
+        if (err) return done(err);
+
+        should.exists(users);
+        var order = users.map(function(u) { return u.order; });
+        order.should.eql([6,5,4,3,2,1]);
```
Amir-61: There is a linter issue here; please fix them; it should be:
```
order.should.eql([6, 5, 4, 3, 2, 1]);
```

Amir-61:
>it may make sense to land this fix first and then improve the performance later.

Sure lets land this PR and later we can optimize it.

@MichaelKohler I left a few comments on the code; other than that the patch looks mostly good to me, apart from the performance issue which we can solve it later on. Also please squash your commits into one single commit; the commit message should be less than 50 characters; it can be something like: `Preserve order of items in DAO.find()` in DAO.find()` or `Ensure stable order of items in DAO.find()` or ...

>one commit is from @bennyt2 who hasn't signed the CLA, so he might need to do this if we want to land this one here. (see MichaelKohler#1)

I'm not sure whether @bennyt2 needs to sign [CLA](https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler) or not; if he does, it will be perfect; @rmg or @superkhau may know better.

Thanks!

bennyt2:
I signed the CLA. I am glad we'll be able to get this issue addressed. Thanks everyone!

test/basic-querying.test.js
superkhau: Do not ignore it, let's do it this way for consistency. Please add this to the style guide @Amir-61.

/cc @bajtos

Amir-61:
Hi @bennyt2 @MichaelKohler,

Thanks to @bajtos the other patch is ready to land; we will close this PR once [PR#996](https://github.com/strongloop/loopback-datasource-juggler/pull/996) lands.

Thanks again for taking time to dig into the issue :-)

slnode:

slnode:

slnode:

Amir-61:
@MichaelKohler @bennyt2 
Fixed in `loopback-datasource-juggler@2.47.0`. Closing this PR now. Thanks!

PR #866, opened 2016-03-04
Fix Mongo compatibility issue

Amir-61:
##### Description: 
This patch takes care of compatibility issues for Mongodb v2.4 tests for PR: https://github.com/strongloop/loopback-connector-mongodb/pull/202

Please see the next comment for my conversation with Raymond.

@bajtos could you please review this patch.

Thanks  :-)

### DIFF
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -1,7 +1,11 @@
 var ValidationError = require('../').ValidationError;
 var traverse = require('traverse');
 
-module.exports = function(dataSource, should) {
+module.exports = function(dataSource, should, connectorCapabilities) {
+  if (!connectorCapabilities) connectorCapabilities = {};
+  if (connectorCapabilities.replaceOrCreateReportsNewInstance === undefined) {
+    console.warn('The connector does not support a recently added feature: replaceOrCreateReportsNewInstance');
+  }
   describe('Persistence hooks', function() {
     var observedContexts, expectedError, observersCalled;
     var TestModel, existingInstance;
@@ -2403,13 +2407,19 @@ module.exports = function(dataSource, should) {
             { id: 'new-id', name: 'a name' },
             function(err, instance) {
               if (err) return done(err);
-              observedContexts.should.eql(aTestModelCtx({
+              
+              var expected = {
                 data: {
                   id: 'new-id',
                   name: 'a name'
-                },
-                isNewInstance: true
-              }));     
+                }
+              };
+              
+              expected.isNewInstance = 
+                connectorCapabilities.replaceOrCreateReportsNewInstance ?
+                true : undefined;            
+              
+              observedContexts.should.eql(aTestModelCtx(expected));     
               done();
             });
         });
@@ -2422,14 +2432,19 @@ module.exports = function(dataSource, should) {
             function(err, instance) {
               if (err) return done(err);
 
+              var expected = {
+                data: {
+                  id: existingInstance.id,
+                  name: 'replaced name'
+                }
+              };
+              
+              expected.isNewInstance = 
+                connectorCapabilities.replaceOrCreateReportsNewInstance ?
+                false : undefined;
+
               if (dataSource.connector.replaceOrCreate) {
-                observedContexts.should.eql(aTestModelCtx({
-                  data: {
-                    id: existingInstance.id,
-                    name: 'replaced name'
-                  },
-                  isNewInstance: false
-                }));
+                observedContexts.should.eql(aTestModelCtx(expected));
               } else {
                 // TODO: Please see loopback-datasource-juggler/issues#836
                 // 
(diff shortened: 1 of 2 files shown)


Amir-61:
>Amirali Jafarian [12:21 PM] 
>Hi Raymond,


>I hope you are doing well :simple_smile:

>I’m messaging you to discuss one of my (old) PRs, which I have not landed yet in loopback-connector-mongodb `PR#202`. We noticed that the patch is not compatible for mongoldb version `2.4.x`. In that patch I’m using `update` method of mongodb, where in versions of `2.6` and higher it returns an object that contains the status of the operation and that is used for populating `isNewInstance`. however, that is missing in mongoldb version 2.4.x. Hence, Miroslav and I decided to return `undefined` for `isNewInstance` if this info is missing from mongoldb. However we need to handle these scenarios in test cases in loopback-datasource-juggler based on the version of mongoldb used accordingly and we decided to follow the same approach you followed in other places, where you use `process.env.MONGODB_VERSION` which is set on Jenkins. Miroslav asked me to keep you in loop to make sure you do not object with this approach. Could you please confirm you are good with this approach?

>I appreciate it in advance and sorry to bother you :simple_smile:

>>Raymond Feng [12:22 PM] 
>>That's fine

bajtos:
Hi @Amir-61, thank you for the pull request. I am not comfortable coupling juggler tests directly with MongoDB version. In my mind, juggler should stay database-independent and provide a generic mechanism that will allow any connector to let the test suite know that it the connector under test does not support `isNewInstance` flag.

I think the simplest way is to add a third parameter to the function exported by `test/persistence-hooks.test.js` and get the flag from that new parameter.

The flag can have three values:
 - `undefined` - the connector setting up the test suite is not aware of the new feature. This happens when we add a new feature to juggler and a connector was not updated to support it yet. In order to prevent test failures in connectors whenever we add new features to juggler, we should assume "false" in this case, but print a warning to let the user know that they are expected to improve their connector.
 - `true` - the connector supports the feature.
 - `false` - the connector does not support the feature.

```js
juggler/test/persistence-hooks.suite.js
module.exports = function(dataSource, should, connectorCapabilities) {
  // at the top - report a warning when a connector does not provide a feature
  // (print the warning only once)
  if (connectorCapabilities.replaceOrCreateReportsNewInstance === undefined) {
    console.warn('The connector does not support a recently added feature: isNewInstance reported by replaceOrCreate()');
  }

  // inside tests
  describe('...', function() {
    it('...', function() {
      if (connectorCapabilities.replaceOrCreateReportsNewInstance) {
        expected.isNewInstance = true; // or false, depending on the test
      } else {
        expected.isNewInstance = undefined;
      }
    });
  });
};

// in loopback-connector-mongodb test/persistence-hooks.test.js
suite(global.getDataSource(customConfig), should, {
 replaceOrCreateReportsNewInstance: process.env.MONGODB_VERSION && !semver.satisfies(process.env.MONGODB_VERSION, '>=2.6.0')
})
// note that '~2.6.0' is not good enough, as that won't match never versions like 2.8 or 3.0
```

Amir-61:
>In my mind, juggler should stay database-independent 
:+1: Thanks for the suggestion!

Amir-61:
@bajtos Could you please review this patch? Thanks :pray:

test/persistence-hooks.suite.js
```diff
@@ -1,7 +1,10 @@
 var ValidationError = require('../').ValidationError;
 var traverse = require('traverse');
 
-module.exports = function(dataSource, should) {
+module.exports = function(dataSource, should, connectorCapabilities) {
+  if (!connectorCapabilities || connectorCapabilities.replaceOrCreateReportsNewInstance === undefined) {
+    console.warn('The connector does not support a recently added feature: isNewInstance reported by replaceOrCreate()');
```
bajtos: On the second thought, perhaps it's better to report the exact capability name here?

```js
console.warn('The connector does not support a recently added feature: replaceOrCreateReportsNewInstance');
```

test/persistence-hooks.suite.js
```diff
@@ -1,7 +1,10 @@
 var ValidationError = require('../').ValidationError;
 var traverse = require('traverse');
 
-module.exports = function(dataSource, should) {
+module.exports = function(dataSource, should, connectorCapabilities) {
+  if (!connectorCapabilities || connectorCapabilities.replaceOrCreateReportsNewInstance === undefined) {
```
bajtos: I think it will be better to "normalize" `connectorCapabilities` so that we don't have to check in all places whether it's defined.


```js
module.exports = function(dataSource, should, connectorCapabilities) {
  if (!connectorCapabilities) connectorCapabilities = {};
  // ...
}
```

Thoughts?

bajtos:
The patch looks much better now, thank you @Amir-61. I left two minor comments to consider, otherwise the change LGTM.

Amir-61:
@bajtos Please note: we need to land this PR as well: https://github.com/strongloop/loopback-connector-mysql/pull/157

Amir-61:
@slnode test please

Amir-61:
@slnode test please

bajtos:
LGTM.

bajtos:
I think we should back-port this fix to 2.x too, thoughts?

PR #869, opened 2016-03-09
polymorphic discriminator type is hard coded as 'string'

mediapolis:
MySQL throws `ER_TOO_LONG_KEY: Specified key was too long; max key length is 767 bytes`

I needed to hack this with a reasonable length to get MySql CREATE TABLE working.

Probably need to make this configurable (Something like idType) for both type and length. 

```javascript
polymorphic: {
    discriminatorType: { "type": "string", length: 30, "index": true },
    foreignKey: 'imageableId',
    discriminator: 'imageableType'
}
```

Anyone else had this issue? Is there another solution? Thoughts?

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -602,7 +602,7 @@ RelationDefinition.hasMany = function hasMany(modelFrom, modelTo, params) {
       fk = polymorphic.foreignKey;
     }
     if (!params.through) {
-      modelTo.dataSource.defineProperty(modelTo.modelName, discriminator, { type: 'string', index: true });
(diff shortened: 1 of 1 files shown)


slnode:

Amir-61:
@mediapolis I'm afraid I do not think if hacking with `length:30` makes that much sense to me... It would probably be better if you start writing failing tests first.

Thanks!

slnode:

slnode:

slnode:

Amir-61:
Closing this PR due to inactivity and wrong direction.

PR #875, opened 2016-03-16
support custom field settings under the connector's namespace

bitmage:
Some database connectors need to have custom field settings.  E.g. Hbase has column families which could be different on a per field basis.  Reading/writing to Hbase is not possible without knowledge of the column families, so this is an important detail to be captured in schema discovery.

Right now, there is a namespace reserved for the connector, but it is hard coded to only populate certain properties.  I'd like to extend that to support any additional properties the connector may need.

A sample discoverModelProperties result for the 'custom' connector might look like this:

```javascript
[
    {
        owner: 'public',
        columnName: 'name',
        type: 'String',
        required: false,
        custom: {storage: 'quantum'}
    }
]
```

And discoverSchemas:

```javascript
{
    "public.person": {
        "name": "Person",
        "options": {
            "idInjection": false,
            "custom": {
                "schema": "public",
                "table": "person"
            }
        },
        "properties": {
            "name": {
                "type": "String",
                "required": false,
                "custom": {
                    "columnName": "name",
                    "storage": "quantum"
                }
            }
        }
    }
}
```

Code is implemented and tested.  I tried to support this in a way that seems idiomatic to the Loopback design.  Feedback welcome.

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -1343,8 +1343,6 @@ DataSource.prototype.discoverSchemas = function (modelName, options, cb) {
     };
 
     columns.forEach(function (item) {
-      var i = item;
-
       var propName = nameMapper('column', item.columnName);
       schema.properties[propName] = {
         type: item.type,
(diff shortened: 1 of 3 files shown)


bajtos:
@raymondfeng AFAICT, this was not back-ported to the 2.x branch and thus it will be released only in the 3.0-alpha series. Is that intentional?

PR #876, opened 2016-03-17
"Near" fix for Where clauses that contain and/or

cmoore4:
The near filter on geo searches currently only searches a first level object: eg 
```
{ "where": { "location": {"near": "30,-90"}}}.  
```

For clauses that contain an "and" or an "or" clause, the geo filter does not pick up the operator, and you tend to get a SQL error back.  For instance, Postgres forms this abomination:  
```
WHERE ("location"point($1,$2)) AND ("name" ~* $3).
```

This fix allows for queries with a format such as: 
```
{"where":{"and":[{"location":{"near":"29,-90"}},{"name":{"regexp":"/sean/i"}}]}}
```

It only handles one level of depth right now, as that's all we need, though I'm sure this could be made recursive to handle an infinite depth of and/or parameters.


slnode:

Amir-61:
@cmoore4 Could you please also add unit tests which fail without your implementation and pass with your implementation?

cmoore4:
@Amir-61 Modified the model and seed data, as well as created a test in ```basic-querying.test.js```

The test is in my branch, so you can see them working with the fix.  Do you want me to create a new branch to show them failing against master?

lib/geo.js
```diff
@@ -10,7 +10,6 @@ exports.nearFilter = function nearFilter(where) {
   if (where && typeof where === 'object') {
     Object.keys(where).forEach(function (key) {
       var ex = where[key];
-
```
Amir-61: Nit pick: Could you please un-touch this line?

test/basic-querying.test.js
```diff
@@ -441,6 +442,17 @@ describe('basic-querying', function () {
         done();
       });
     });
+    
+    it('should support nested GeoPoint near queries', function(done){
+      User.find({
+        where: {and: [{addressLoc: {near: "29.9,-90.07"}}, {vip: true}]}
+      }, function(err, users){
+        should.not.exist(err);
```
Amir-61: Please call the callback with the error:
```
if (err) return done(err);
```

Amir-61:
>Do you want me to create a new branch to show it failing against master?

No, thats perfect! Just left a few minor suggestions. could you please solve them and rebase and squash your commits as one single commit following these [standards](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).

Thanks for the contribution!

lib/geo.js
```diff
@@ -18,6 +17,20 @@ exports.nearFilter = function nearFilter(where) {
           unit: ex.unit,
           key: key
         };
+      } else if (Array.isArray(ex)){
+        ex.forEach(function (condition){
+          Object.keys(condition).forEach(function(condKey){
```
Amir-61: Nitpick: Please add a space between `)` and `{`
It should be:
```
Object.keys(condition).forEach(function(condKey) {
```

lib/geo.js
```diff
@@ -18,6 +17,20 @@ exports.nearFilter = function nearFilter(where) {
...
+            var condSubEx = condition[condKey];
+            if(condSubEx.near){
```
Amir-61: Please add a space before `{`; the same for other places...

Amir-61:
@bajtos PTAL

cmoore4:
Rebased and squashed, formatting fixes, and now throws an error if more than one near clause is specified, as that is totally unsupported (see function doc).

@Amir-61 Can you re-trigger a CI build for this branch?  Github doesn't seem to trigger them on forced rebase commits.

Git is really confused about how to diff the blank line at 13/16, and my git-fu isn't strong enough to tell it to ignore that specific change, so I wasn't able to remove that, apologies.

lib/geo.js
```diff
@@ -10,7 +13,7 @@ exports.nearFilter = function nearFilter(where) {
   if (where && typeof where === 'object') {
     Object.keys(where).forEach(function (key) {
       var ex = where[key];
-
+
```
bajtos: There is a white-space change here if I am reading this correctly. The original file contains an empty line, the new file contains a line with about 4 spaces. @cmoore4 could you PTAL?

bajtos:
@slnode ok to test

bajtos:
I am not familiar with the implementation of geo queries myself, but the change looks reasonable to me. Let's give a day or two to @raymondfeng and/or @ritch to catch up and if they don't respond, then get this landed. @Amir-61 I think we should back-port this patch to 2.x too.

bajtos:
> Can you re-trigger a CI build for this branch? Github doesn't seem to trigger them on forced rebase commits.

@Amir-61 the trick is to leave a comment containing "ok to test", as @slnode commented at the top. That tells our CI that it's ok (safe) to run CI build of this particular patch.

The build is running now, let's see what the results will be.

cmoore4:
Don't merge yet, I unlazied myself and made it work recursively to find "near" operator at any depth.  I'm going to push some commits for testing and then rebase.

cmoore4:
@Amir-61 Ok, this now does a recursive search for "near" operators, and additionally it ... mostly supports multiple near operations.  Quoting my function doc block in geo.js:exports.filter


/*!
 * Filter a set of objects using the given `nearFilter`.  Can support multiple
 * locations, but will include results from all of them. 
 * 
 * WARNING: "or" operator with GeoPoint does not work as expected, eg:
 * {where: {or: [{location: {near: (29,-90)}},{name:'Sean'}]}}
 * Will actually work as if you had used "and".  This is because geo filtering
 * takes place outside of the SQL query, so the result set of "name = Sean" is
 * returned by the database, and then the location filtering happens in the app
 * logic.  So the "near" operator is always an "and" of the SQL filters, and "or"
 * of other GeoPoint filters.
 * 
 * Additionally, since this step occurs after the SQL result set is returned,
 * if using GeoPoints with pagination the result set may be smaller than the 
 * page size.  The page size is enforced at the DB level, and then we may 
 * remove results at the Geo-app level.  If we "limit: 25", but 4 of those results
 * do not have a matching geopoint field, the request will only return 21 results.
 * This may make it erroneously look like a given page is the end of the result set.
 */


Basically, this is kinda the best we can do given how GeoSearch is implemented (app-side, as opposed to DB-side).  It's nice because it allows the use of GeoPoint against databases that don't support it, eg Sqlite.  But in Postgres, it would be about 1000x better to implement this natively in SQL.  That's way beyond the scope of my original Pull Request, though, so I'll leave it to rest here for now.

test/basic-querying.test.js
```diff
@@ -441,6 +442,48 @@ describe('basic-querying', function () {
         done();
       });
     });
+    
+    it('should support nested GeoPoint near queries', function(done){
```
Amir-61: Please add a space between `)` and `{`

It should be:

```
it('should support nested GeoPoint near queries', function(done) {
```

test/basic-querying.test.js
```diff
@@ -441,6 +442,48 @@ describe('basic-querying', function () {
...
+        where: {and: [{addressLoc: {near: "29.9,-90.07"}}, {vip: true}]}
+      }, function(err, users){
+        if (err) return done(err);
+        users.should.have.property('length', 2);
+        users[0].addressLoc.should.not.equal(null);
+        done();
+      });
+    });
+    
+    it('should support very nested GeoPoint near queries', function(done){
```
Amir-61: Please add a space between ) and {

test/basic-querying.test.js
```diff
@@ -441,6 +442,48 @@ describe('basic-querying', function () {
...
+        ]}
+      }, function(err, users){
+        if (err) return done(err);
+        users.should.have.property('length', 1);
+        users[0].addressLoc.should.not.equal(null);
+        done();
+      });
+    });
+    
+    it('should support multiple GeoPoint near queries', function(done){
```
Amir-61: Please add a space between ) and {

lib/geo.js
```diff
@@ -2,80 +2,117 @@ var assert = require('assert');
...
-      return da > db ? 1 : -1;
-    } else {
-      return 0;
-    }
+      if (max && d > max) {
+        // dont add
+      } else {
+        distances[obj.id] = d;
+        result.push(obj);
+      }
```
Amir-61: So shouldn't this be simplified to the following?

```
if (!max || d <= max) {
    distances[obj.id] = d;
    result.push(obj);
}
```

lib/geo.js
```diff
@@ -2,80 +2,117 @@ var assert = require('assert');
...
   var result = [];
+  
+  filters.forEach(function (filter) {
+    var origin = filter.near;
+    var max = filter.maxDistance > 0 ? filter.maxDistance : false;
+    var unit = filter.unit;
+    var key = filter.key;
+
+    // create distance index
+    arr.forEach(function (obj) {
```
Amir-61: should be `arr.forEach(function(obj) {`

No space between `function` and `(`

lib/geo.js
```diff
@@ -2,80 +2,117 @@ var assert = require('assert');
...
-          key: key
-        };
-      }
+  function nearSearch(clause){
+    
+    if (typeof clause !== 'object') { 
+      return false; 
+    }
+    
+    Object.keys(clause).forEach(function (clauseKey) {
```
Amir-61: No space between `function` and `(`

lib/geo.js
```diff
@@ -2,80 +2,117 @@ var assert = require('assert');
...
+      if (Array.isArray(clause[clauseKey])){
+        clause[clauseKey].forEach(function (el) {
```
Amir-61: ditto

lib/geo.js
```diff
@@ -2,80 +2,117 @@ var assert = require('assert');
...
  */
 
-exports.filter = function (arr, filter) {
-  var origin = filter.near;
-  var max = filter.maxDistance > 0 ? filter.maxDistance : false;
-  var unit = filter.unit;
-  var key = filter.key;
-
-  // create distance index
+exports.filter = function (arr, filters) {
```
Amir-61: Please remove space between `function` and `(` in other places as well

Amir-61:
>this now does a recursive search for "near" operators
:+1: 

@raymondfeng Ping

Amir-61:
Also please squash your commit, so by the time @raymondfeng reviews he can give the final LGTM

cmoore4:
@Amir-61 So, as we're rolling out a new service and using a lot of pieces of Loopback, I might be submitting a fair amount of bug-fix style PRs in the future, so I'd like to get clarification on the code style enforcement so that I can adhere to it as we move forward.  

Many of these files do not have cross-project consistency.  For instance, in geo.js, the majority of existing function declarations are of the type:

```
function (var, var2) {
```

All of the lines you called out above were already in the source, I did not write them, and the rest of the file source is split almost 50/50 between function-space and function-parens.

Should I just commit stylistic additions to all function declaration in the entire file?  Seems like it'd add a lot of noise to this PR, but better for both you and me instead of us having to go back and forth over changing styles for stuff I didn't write over the course of several check-ins?  What's your team's preference?

Amir-61:
>so I'd like to get clarification on the code style enforcement so that I can adhere to it as we move forward.

Coding style check should be automated as a part of CI test by linters and caught as part of CI build, but we are not there yet; we currently use `jslint` and `jscs` for lots of our modules; however, we plan to switch to `Eslint`; please see [this task](https://github.com/strongloop/loopback/issues/1749). I believe there is a work in progress to move to `eslint`.

>Many of these files do not have cross-project consistency. 

Unfortunately thats true and they need to be consistent. But you don't need to take care of them here.

Also I just noticed @raymondfeng is on vacation this week; lets wait for him for final review.

Thanks!

Amir-61:
@slnode test please

bajtos:
> Many of these files do not have cross-project consistency. 
> Should I just commit stylistic additions to all function declaration in the entire file? Seems like it'd add a lot of noise to this PR, but better for both you and me instead of us having to go back and forth over changing styles for stuff I didn't write over the course of several check-ins? What's your team's preference?

In general, we try to avoid making stylistic/formatting/white-space changes on lines that are not otherwise affected by the patch (pull request) in question.

We are working on improving our linter setup to catch more style violations and also to unify the coding style a bit. I have just landed #886, could you please rebase your patch on top of the new master?

slnode:

slnode:

slnode:

PR #879, opened 2016-03-28
Add polymorphic discriminator customization (modelName)

dotpack:


slnode:

PR #880, opened 2016-03-28
Add polymorphic discriminator name customization

dotpack:


slnode:

raymondfeng:
@dotpack Thank you for the patch. Would you please add a simple test case?

alexpls:
I'm running into an issue with a relation definition which would be solved by the patch in this PR. Is the only reason this hasn't been merged in yet lack of tests? If yes, I'm happy to add some so this can get merged.

PR #882, opened 2016-03-29
Remove next tag (2.x)

Amir-61:
Remove next tag for `2.x` branch

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
   "name": "loopback-datasource-juggler",
   "version": "2.45.2",
   "publishConfig": {
-    "tag": "next",
     "export-tests": true
   },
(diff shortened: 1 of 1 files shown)


bajtos:
+1, the "next" tag should not have been added at all.

> [--tag <tag>] Registers the published package with the given tag, such that npm install <name>@<tag> will install this version. By default, npm publish updates and npm install installs the latest tag. See npm-dist-tag for details about tags.

2.x is published with the default `latest` tag, while 3.0-alpha is published with `next` tag. That way `npm install loopback-datasource-juggler` installs the stable 2.x series, but at the same time, one can install the upcoming 3.0 series via `npm install loopback-datasource-juggler@next`

:shipit:

PR #887, opened 2016-04-01
Relicense as MIT only

rmg:
Also update all JS files to include a copyright notice.


superkhau:
LGTM

Will this be required in all LoopBack repos or are you picking out specific ones to update? I'm talking about the IBM headers.

PR #891, opened 2016-04-06
Relicense as MIT only [2.x]

bajtos:
Also update all JS files to include a copyright notice.

This is a back-port of #887

@rmg @superkhau PTAL


superkhau:
LGTM

PR #893, opened 2016-04-06
Test coverages for hashed password

Amir-61:
* Test coverages for hashed password for replaceAttributes
* Test coverages for hashed password for updateAttribute

#### Notes: 
Connect to strongloop/loopback-datasource-juggler#896

https://github.com/strongloop/loopback-datasource-juggler/pull/896 needs to be landed before this patch (PR#893)

@bajtos: Please review... Thanks!

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -37,6 +37,15 @@ describe('manipulation', function() {
   before(function setupStubUserModel(done) {
     StubUser = db.createModel('StubUser', { password: String }, { forceId: true });
     StubUser.setter.password = function(plain) {
+      var hashed = false;
+      if (!plain) return;
+      var pos = plain.indexOf('-');
+      if (pos !== -1) {
+        var head = plain.substr(0, pos);
+        var tail = plain.substr(pos + 1, plain.length);
+        hashed = head.toUpperCase() === tail;
+      }
+      if (hashed) return;
       this.$password = plain + '-' + plain.toUpperCase();
     };
     db.automigrate('StubUser', done);
@@ -462,6 +471,22 @@ describe('manipulation', function() {
     it('has an alias "patchAttributes"', function(done) {
       person.updateAttributes.should.equal(person.patchAttributes);
       done();
+    });
+
+    it('should have updated password hashed with updateAttribute',
+    function(done) {
+      StubUser.create({ password: 'foo' }, function(err, created) {
+        if (err) return done(err);
+        created.updateAttribute('password', 'test', function(err, created) {
+          if (err) return done(err);
+          created.password.should.equal('test-TEST');
+          StubUser.findById(created.id, function(err, found) {
+            if (err) return done(err);
+            found.password.should.equal('test-TEST');
+            done();
+          });
+        });
+      });
     });
 
     it('should update one attribute', function(done) {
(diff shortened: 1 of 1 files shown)


test/manipulation.test.js
```diff
@@ -37,7 +37,13 @@ describe('manipulation', function() {
...
     StubUser = db.createModel('StubUser', { password: String }, { forceId: true });
     StubUser.setter.password = function(plain) {
-      this.$password = plain + '-' + plain.toUpperCase();
+      if (plain === null)
+        return;
+      var pos = plain.indexOf('-');
+      // Check that it has not been already hashed
+      if (pos === -1 ||
+      (plain.substr(0, pos).toUpperCase() !== plain.substr(pos + 1, plain.length)))
+        this.$password = plain + '-' + plain.toUpperCase();
```
bajtos: Uff, this is rather difficult to understand. Could you please split the code into smaller steps and save intermediate results into named variables? Something like this:

```js
var hashed = false;
var pos = plain.indexOf('-');
if (pos !== -1) {
  var head = plain.substr(0, pos);
  var tail = plain.substr(pos + 1, plain.length);
  hashed = head.toUpperCase() === tail;
}

if (hashed) return;

this.$password = plain + '-' + plain.toUpperCase();
```

test/manipulation.test.js
```diff
@@ -934,6 +956,23 @@ describe('manipulation', function() {
         });
       });
 
+      it('should have updated password hashed with replaceAttributes',
+      function(done) {
+        StubUser.create({ password: 'foo' }, function(err, created) {
+          if (err) return done(err);
+          created.replaceAttributes({ password: 'test' }, { validate: false },
```
bajtos: Why is `{validate:false}` needed here?

bajtos:
Hi @Amir-61, the tests look good in general, I left two minor comments to address. Have you seen the tests fail? I'd like you to check that these new tests would catch the regression which was fixed by https://github.com/strongloop/loopback-datasource-juggler/commit/dbdf9153346ed3aabc925eddf6280e19df519cd3

test/manipulation.test.js
Amir-61: Please see the fix: https://github.com/strongloop/loopback-datasource-juggler/pull/896

I belive https://github.com/strongloop/loopback-datasource-juggler/pull/896 needs to be landed before this patch#893

Amir-61:
Thanks @bajtos ,

In case if my comment gets lost: https://github.com/strongloop/loopback-datasource-juggler/pull/893#discussion_r58956495

I believe https://github.com/strongloop/loopback-datasource-juggler/pull/896 needs to be landed first.

> Have you seen the tests fail?
Yes without the fix the added test was failing. :+1: 

Thanks!

test/manipulation.test.js
bajtos: Let's fix the other problem first then and keep this patch clean of tweaks/hacks that are difficult to understand.

bajtos:
LGTM. Let's get #896 landed first and the re-run the CI builds before landing this one, to make sure they pass.

Amir-61:
@slnode test please

Amir-61:
Also back-ported the commit to `2.x`; I'm so sorry I forgot to add the PR# for this back-porting; I knew I could do `git commit --amend` and do `git push -f`, but I preferred not to do it since it rewrites the git history... I'll make sure I'll add the PR# in the upcoming PRs when back-porting...

PR #896, opened 2016-04-07
Fix `forceId` check for `replaceById`

Amir-61:
This patch fixes `forceId` check for `replaceById` method:

##### Problem:
Validation fails if you pass `{ forceId: true }` for `replaceById` even though you don't pass the primary key in data object.

#### Notes: 
Relate to https://github.com/strongloop/loopback-datasource-juggler/pull/893
This ptach (PR#896) needs to be landed before https://github.com/strongloop/loopback-datasource-juggler/pull/893

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2657,7 +2657,7 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
   if (!data[pkName]) data[pkName] = id;
 
   var Model = this;
-  var inst = new Model(data);
+  var inst = new Model(data, { persisted: true });
   var enforced = {};
   this.applyProperties(enforced, inst);
   inst.setAttributes(enforced);
--- a/test/optional-validation.test.js
+++ b/test/optional-validation.test.js
@@ -4,8 +4,9 @@
 // License text available at https://opensource.org/licenses/MIT
 
 // This test written in mocha+should.js
+var async = require('async');
 var should = require('./init.js');
-var db, User, options, whereCount = 0;
+var db, User, options, ModelWithForceId, whereCount = 0;
 var j = require('../');
 var ValidationError = j.ValidationError;
 
@@ -18,6 +19,10 @@ describe('optional-validation', function() {
 
   before(function(done) {
     db = getSchema();
+    ModelWithForceId = db.createModel(
+    'ModelWithForceId',
+    { name: String },
+    { forceId: true });
     User = db.define('User', {
       seq: { type: Number, index: true },
       name: { type: String, index: true, sort: true },
@@ -27,9 +32,7 @@ describe('optional-validation', function() {
       order: { type: Number, index: true, sort: true },
       vip: { type: Boolean },
     }, { forceId: true, strict: true });
-
-    db.automigrate(['User'], done);
-
+    db.automigrate(['ModelWithForceId', 'User'], done);
   });
 
   beforeEach(function(done) {
@@ -106,6 +109,33 @@ describe('optional-validation', function() {
   function getNewWhere() {
     return { name: 'DoesNotExist' + (whereCount++) };
   }
+
+  describe('forceId', function() {
+    context('replaceAttributes', function() {
+      it('should not fail if you do not pass the Primary key in data object',
+      function(done) {
+        ModelWithForceId.create({ name: 'foo' }, function(err, created) {
+          if (err) return done(err);
+          created.replaceAttributes({ name: 'bar' }, function(err, data) {
+            done(err);
+          });
+        });
+      });
+
+      it('should fail if you pass the Primary key in data object',
+      function(done) {
+        ModelWithForceId.create({ name: 'foo' }, function(err, created) {
+          if (err) return done(err);
+          created.replaceAttributes({ name: 'bar', id: 999 },
+          function(err, data) {
+            should.exist(err);
+            done();
+          });
+        });
+      });
+    });
+  });
+
 
   describe('no model setting', function() {
 


Amir-61:
@slnode test please

bajtos:
@Amir-61 I am confused, what is `forceId` option? Please add a unit-test that demonstrates the problem by failing with the current implementation

bajtos:
AFAICT, `lib/dao.js` does not contain any `forceId` string. I am really confused.

Amir-61:
@slnode test please

Amir-61:
@slnode test please

rmg:
@slnode test please

rmg:
@slnode test please

rmg:
Found and fixed the problem strongloop/ephemeral-npm@1c4ba10 :blush:

Amir-61:
>I am confused, what is forceId option?
 
My understanding is If you want to prevent someone to set the `auto-id` value manually, you can set `{ forceId: true }`. so if you set the the PK in your `data` object and `forceId` is set to true the validation should fail... Please correct me if I'm wrong...

>AFAICT, lib/dao.js does not contain any forceId string. I am really confused.

Yes thats true, but we [call `isValid` method](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L2695) so again if forceId is set and you pass the PrimaryKey's value in `data` object, validation should fail...

Again please correct me if I'm wrong... Thanks!

test/optional-validation.test.js
```diff
@@ -27,9 +30,10 @@ describe('optional-validation', function() {
       order: { type: Number, index: true, sort: true },
       vip: { type: Boolean },
     }, { forceId: true, strict: true });
-
-    db.automigrate(['User'], done);
-
+    var models = ['modelWithForceId', 'User'];
+    async.each(models, function(model, cb) {
+      db.automigrate(model, cb);
+    }, done);
```
bajtos: I believe you can pass multiple model names to automigrate function, no need to use `async.each`:

```js
db.automigrate(['modelWithForceId', 'User'], done);
```

test/optional-validation.test.js
```diff
@@ -18,6 +19,8 @@ describe('optional-validation', function() {
 
   before(function(done) {
     db = getSchema();
+    modelWithForceId = db.createModel('modelWithForceId',
```
bajtos: `db.createModel` returns a class constructor, therefore we use names starting with a capital letter - `ModelWithForceId`.

bajtos:
>  > I am confused, what is forceId option?

> My understanding is If you want to prevent someone to set the auto generated id's value manually, you can set { forceId: true }. So if you set the the PK in your data object and forceId is set to true the validation should fail... Please correct me if I'm wrong...

Ah, thank you, I found it: https://github.com/strongloop/loopback-datasource-juggler/blob/1e8fe6a0d7b5c84a3f67ad4d6a05ddaa8d1229ba/lib/datasource.js#L516-L518

Your conclusions about this feature look reasonable to me.

lib/dao.js
```diff
@@ -2632,6 +2632,8 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
           'replaceById() must be implemented by the connector');
 
   var pkName = idName(this);
+  var initialData = {};
+  copyData(data, initialData);
```
bajtos: Hmm, creating a copy of all data is rather expensive operation. Could you please take another look and check if there is a way how to avoid that?

For example, take a look a the following line - is it necessary to always fill in primary key?

```js
if (!data[pkName]) data[pkName] = id;
```

bajtos: Also consider the data passed to hooks, e.g. "before save" hook. I think the data should not include primary key when the user is not including it.

It makes me wonder, perhaps we are hard-setting the primary key to prevent people from _changing_ primary key via `updateAttributes`? I am sure there are other ways how to implement that, without side effects like the one you are trying to fix...

bajtos:
I think it would be nice to add tests to `test/persistence-hooks.suite.js` that would check what `ctx.data` is reported by `updateAttributes` when `forceId:true`. Thoughts?

lib/dao.js
Amir-61: >Hmm, creating a copy of all data is rather expensive operation. Could you please take another look and check if there is a way how to avoid that?

Right, now I use a flag...

>Also consider the data passed to hooks, e.g. "before save" hook. I think the data should not include primary key when the user is not including it.

Sorry I'm a little bit confused here; for `before save` we are not passing `data` in context; please see [here](https://github.com/strongloop/loopback-datasource-juggler/blob/063983ccc9257db2e500eb273a585e6a9dad6652/lib/dao.js#L2662-L2670)

>It makes me wonder, perhaps we are hard-setting the primary key to prevent people from changing primary key via updateAttributes?

No for `updateAttributes` you can still pass the value of the PK in data object, but it should be the same the PK value for `instance` (it should pass `idEquals(data[idName], inst[idName])`); please see [here](https://github.com/strongloop/loopback-datasource-juggler/blob/063983ccc9257db2e500eb273a585e6a9dad6652/lib/dao.js#L2832-L2844)

Amir-61:
>I think it would be nice to add tests to `test/persistence-hooks.suite.js` that would check what ctx.data is reported by updateAttributes when forceId:true. Thoughts?

Sorry which hook are you talking about in `updateAttributes`? I guess it is better if this PR/commit focuses on just `replaceById `?

Amir-61:
@slnode test please

bajtos:
> Sorry which hook are you talking about in updateAttributes? I assume before save? I guess it is better if this PR/commit focuses on just fixing forceId for replaceById?

Sorry for the confusion. I was thinking about "before save" for "updateAttributes", which does not make sense in the context of this pull request.

I agree that it's better to focus on fixing just forceId for replaceById in this pull request. Having said that, I think it still makes sense to add tests to verify how "before save" hook handles "forceId" flag for replaceById. For example, verify that when the hook changes `ctx.instance.id`, the implementation of replaceById rejects such request.

Amir-61:
>For example, verify that when the hook changes ctx.instance.id, the implementation of replaceById rejects such request. It's fine to do that in a follow-up pull request though.

I'll create a task and will take care of it ASAP.

So your suggestion about using `var inst = new Model(data, { persisted: true });` works and simplified the solution...

Thanks!

test/optional-validation.test.js
```diff
@@ -18,6 +19,8 @@ describe('optional-validation', function() {
 
   before(function(done) {
     db = getSchema();
+    ModelWithForceId = db.createModel('ModelWithForceId',
+    { name: String }, { forceId: true });
```
bajtos: See https://github.com/strongloop-internal/loopback-knowledge-base/blob/style-guide/style-guide.md

```js
ModelWithForceId = db.createModel(
  'ModelWithForceId',
  { name: String },
  { forceId: true });
```

bajtos:
Nice! One more nitpick to take a look at, the rest LGTM. No further reviews are needed.

Amir-61:
Backported to `2.x`,  [commit](https://github.com/strongloop/loopback-datasource-juggler/commit/1eeacd02554a29779fb75b101ce0af81d2a91c36)

Issue #836, opened 2016-01-31
'find' should not trigger hooks if options.notify is false


Amir-61:
`find` method irrespective of whether `options.notify`is false or not it calls triggers some hooks.

PR #898, opened 2016-04-09
Fix notify for find method

Amir-61:
##### Bug description:

`find` should not trigger hooks if options.notify is false

##### Connect to https://github.com/strongloop/loopback-datasource-juggler/issues/836

@bajtos Please review :pray:

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1631,16 +1631,26 @@ DataAccessObject.find = function find(query, options, cb) {
       // do in memory query
       // using all documents
       // TODO [fabien] use default scope here?
+      if (options.notify === false) {
+        queryGeo(query);
+      } else {
+        withNotifyGeo();
+      }
 
-      var context = {
-        Model: self,
-        query: query,
-        hookState: hookState,
-        options: options,
-      };
-      self.notifyObserversOf('access', context, function(err, ctx) {
-        if (err) return cb(err);
+      function withNotifyGeo() {
+        var context = {
+          Model: self,
+          query: query,
+          hookState: hookState,
+          options: options,
+        };
+        self.notifyObserversOf('access', context, function(err, ctx) {
+          if (err) return cb(err);
+          queryGeo(ctx.query);
+        });
+      }
 
+      function queryGeo(query) {
         function geoCallback(err, data) {
           var memory = new Memory();
           var modelName = self.modelName;
@@ -1661,7 +1671,7 @@ DataAccessObject.find = function find(query, options, cb) {
             });
 
             // FIXME: apply "includes" and other transforms - see allCb below
-            memory.all(modelName, ctx.query, options, cb);
+            memory.all(modelName, query, options, cb);
           } else {
             cb(null, []);
           }
@@ -1672,8 +1682,7 @@ DataAccessObject.find = function find(query, options, cb) {
         } else {
           connector.all(self.modelName, {}, geoCallback);
         }
-      });
-
+      }
       // already handled
       return cb.promise;
     }
@@ -1685,18 +1694,14 @@ DataAccessObject.find = function find(query, options, cb) {
       async.each(data, function(item, callback) {
         var d = item;//data[i];
         var Model = self.lookupModel(d);
+        if (options.notify === false) {
+          buildResult(d);
+        } else {
+          withNotify();
+        }
 
-        context = {
-          Model: Model,
-          data: d,
-          isNewInstance: false,
-          hookState: hookState,
-          options: options,
-        };
-
-        Model.notifyObserversOf('loaded', context, function(err) {
-          if (err) return callback(err);
-          d = context.data;
+        function buildResult(data) {
+          d = data;
 
           var ctorOpts = {
             fields: query.fields,
@@ -1742,7 +1747,22 @@ DataAccessObject.find = function find(query, options, cb) {
           } else {
             callback();
           }
-        });
+        }
+
+        function withNotify() {
+          context = {
+            Model: Model,
+            data: d,
+            isNewInstance: false,
+            hookState: hookState,
+            options: options,
+          };
+
+          Model.notifyObserversOf('loaded', context, function(err) {
+            if (err) return callback(err);
+            buildResult(context.data);
+          });
+        }
       },
       function(err) {
         if (err) return cb(err);
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -86,6 +86,45 @@ module.exports = function(dataSource, should, connectorCapabilities) {
           });
       });
 
+      it('should not trigger hooks, if notify is false', function(done) {
+        monitorHookExecution();
+        TestModel.find(
+          { where: { id: '1' }},
+          { notify: false },
+          function(err, list) {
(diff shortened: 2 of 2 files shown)


lib/dao.js
```diff
@@ -1659,7 +1669,7 @@ DataAccessObject.find = function find(query, options, cb) {
             });
 
             // FIXME: apply "includes" and other transforms - see allCb below
-            memory.all(modelName, ctx.query, options, cb);
+            memory.all(modelName, query, options, cb);
```
bajtos: I think the idea here is that a hook can modify `ctx.query` and the changes will be picked up. Is that preserved by this change?

```js
MyModel.observe('access', function(ctx, next) {
  ctx.query = { /* new query */ };
  next();
});
```

lib/dao.js
```diff
@@ -1629,16 +1629,26 @@ DataAccessObject.find = function find(query, options, cb) {
...
+          hookState: hookState,
+          options: options,
+        };
+        self.notifyObserversOf('access', context, function(err, ctx) {
+          if (err) return cb(err);
+          optGeo(ctx.query);
+        });
+      }
 
+      function optGeo(query) {
```
bajtos: What does `opt` stand for? How about `findWithGeo` or `queryGeo`?

lib/dao.js
```diff
@@ -1683,18 +1692,14 @@ DataAccessObject.find = function find(query, options, cb) {
       async.each(data, function(item, callback) {
         var d = item;//data[i];
         var Model = self.lookupModel(d);
+        if (options.notify === false) {
+          opt(d);
```
bajtos: What does `opt` stand for? How about `buildModel` or `buildResult`?

test/persistence-hooks.suite.js
```diff
@@ -1668,7 +1668,6 @@ module.exports = function(dataSource, should, connectorCapabilities) {
             } else {
               triggered.should.eql([
                 'access',
-                'loaded',
```
bajtos: How comes `loaded` is no longer triggered? Is that why we are implementing this fix? In that case I think you can get rid of `if (dataSource.connector.updateOrCreate)` because both branches report the same hooks. Correct?

test/persistence-hooks.suite.js
```diff
@@ -2122,18 +2110,8 @@ module.exports = function(dataSource, should, connectorCapabilities) {
...
-                // loaded hook is triggered twice in non-atomic version:
-                // 1) It gets triggered once by "find()" in this chain:
-                //    "replaceORCreate()->findOne()->find()",
-                //    which is a bug; Please see this ticket:
-                //    loopback-datasource-juggler/issues#836.
-                // 2) It, also, gets triggered in "replaceAttributes()"
-                //    in this chain replaceORCreate()->replaceAttributes()
                 triggered.should.eql([
                   'access',
-                  'loaded',
```
bajtos: Ditto, please join both branches as the list of hooks is (should be?) the same.

test/persistence-hooks.suite.js
```diff
@@ -2453,32 +2431,15 @@ module.exports = function(dataSource, should, connectorCapabilities) {
...
-                //    in this chain replaceORCreate()->replaceAttributes()
-                observedContexts.should.eql([
-                  aTestModelCtx({
-                    data: {
-                      id: existingInstance.id,
-                      name: 'first',
-                    },
-                    isNewInstance: false,
-                    options: { notify: false },
-                  }),
```
bajtos: Can we get rid of `if (dataSource.connector.replaceOrCreate)` and use the same code as in the other branch?

bajtos:
@Amir-61 please add at least some sort of a smoke test to verify that `find` with `{notify:false}` does not trigger any hooks. My point is that we should have a test verifying this feature directly, we should not rely on other code to verify this as a side effect.

Amir-61:
@slnode test please

lib/dao.js
Amir-61: Actually we already have a test covering this; please see [here](https://github.com/strongloop/loopback-datasource-juggler/blob/063983ccc9257db2e500eb273a585e6a9dad6652/test/persistence-hooks.suite.js#L102)

Anyway, I just added a similar test to make sure there is no regression... Please see [here](https://github.com/strongloop/loopback-datasource-juggler/pull/898/files#diff-145ab49855b38984ba190919e6e7397dR104)

Amir-61:
@slnode test please

Amir-61:
>@Amir-61 please add at least some sort of a smoke test to verify that find with {notify:false} does not trigger any hooks. My point is that we should have a test verifying this feature directly, we should not rely on other code to verify this as a side effect.

I have added tests as well... Please review :-)

Also all the `loopback-connector-*` and `loopback-datasource-juggler` pass CI tests; regarding `loopback-connector-mongodb` it passes on all servers, except `amazon-4` where it failed the first time, but CI re-runs the tests and it passed the second time... Please see [here](http://ci.strongloop.com/job/loopback-connector-mongodb/1084/label=amazon-4/console), choose `full log` on top...

Thanks :-)

Amir-61:
@slnode test please

lib/dao.js
bajtos: I am confused, what's the difference between the existing test and the new one?

bajtos: Are you sure your new test is triggering this specific branch of code? I think the trick is to include geo query inside `where`.

Amir-61: Please see: https://github.com/strongloop/loopback-datasource-juggler/pull/898/files#diff-145ab49855b38984ba190919e6e7397dR96

test/persistence-hooks.suite.js
```diff
@@ -78,6 +78,42 @@ module.exports = function(dataSource, should, connectorCapabilities) {
           });
       });
 
+      it('should not trigger hooks, if notify is false', function(done) {
+        monitorHookExecution();
+        TestModel.find(
+          { where: { id: '1' }}, { notify: false },
```
bajtos: Nitpick: it's better to place one arg per line here.

```js
TestModel.find(
  { where: { id: '1' }}, 
  { notify: false },
  function(err, list) {
    // ...
  });
```

test/persistence-hooks.suite.js
```diff
@@ -78,6 +78,42 @@ module.exports = function(dataSource, should, connectorCapabilities) {
...
+            triggered.should.be.empty();
+            done();
+          });
+      });
+
+      it('should not trigger hooks for geo queries, if notify is false',
+      function(done) {
+        monitorHookExecution();
+
+        TestModel.find({ where: { geo: { near: '10,20' }}}, { notify: false },
```
bajtos: ditto

bajtos:
@Amir-61 I left few comments, please also rebase on top of the current master. Sorry for the merge conflicts I introduced.

lib/dao.js
bajtos: Oh, now I understand why this change is needed and why it's correct: the `query` is an argument of `queryGeo` which is set to `ctx.query` by the caller. I think we can consider this thread as resolved.

bajtos:
LGTM, no further review is necessary.

PR #904, opened 2016-04-14
Implement operation hooks for EmbedsOne methods

bajtos:
create() triggers
 - before save
 - after save

udpate() triggers
 - before save
 - after save

destroy() triggers
 - before delete
 - after delete

The implementation here is intentionally left with less features
than the regular DAO methods provide, the goal is to get a partial
(but still useful!) version released soon.

Connect to #497 
See the original implementation proposed by @fabien in #516

/to @raymondfeng @Amir-61 please review

**Things left out**
 - `before save` and `after save` hooks should include `ctx.isNewInstance`
 - Async validations are not supported yet
 - `persist` and `loaded` hooks are not triggered at all
 - Updating embedded instances triggers update of the parent (owning) model. The context provided by `before save` and `after save` hooks on the parent model is sort of arbitrary and may include wrong/extra data. The same probably applies to the scenario when deleting embedded instances triggers update of parent model.
 - `before delete` hook does not provide `ctx.where` property and it's not possible to change the outcome of `destroy()` using this hook. Note that regular DAO does support this.

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -2107,16 +2107,16 @@ EmbedsOne.prototype.create = function(targetModelData, options, cb) {
 
   var inst = this.callScopeMethod('build', targetModelData);
 
-  var updateEmbedded = function() {
+  var updateEmbedded = function(callback) {
     if (modelInstance.isNewRecord()) {
       modelInstance.setAttribute(propertyName, inst);
       modelInstance.save(options, function(err) {
-        cb(err, err ? null : inst);
+        callback(err, err ? null : inst);
       });
     } else {
       modelInstance.updateAttribute(propertyName,
         inst, options, function(err) {
-          cb(err, err ? null : inst);
+          callback(err, err ? null : inst);
         });
     }
   };
@@ -2124,17 +2124,37 @@ EmbedsOne.prototype.create = function(targetModelData, options, cb) {
   if (this.definition.options.persistent) {
     inst.save(options, function(err) { // will validate
       if (err) return cb(err, inst);
-      updateEmbedded();
+      updateEmbedded(cb);
     });
   } else {
-    var err = inst.isValid() ? null : new ValidationError(inst);
-    if (err) {
-      process.nextTick(function() {
-        cb(err);
-      });
-    } else {
-      updateEmbedded();
-    }
+    var context = {
+      Model: modelTo,
+      instance: inst,
+      options: options || {},
+      hookState: {},
+    };
+    modelTo.notifyObserversOf('before save', context, function(err) {
+      if (err) {
+        return process.nextTick(function() {
+          cb(err);
+        });
+      }
+
+      var err = inst.isValid() ? null : new ValidationError(inst);
+      if (err) {
+        process.nextTick(function() {
+          cb(err);
+        });
+      } else {
+        updateEmbedded(function(err, inst) {
+          if (err) return cb(err);
+          context.instance = inst;
+          modelTo.notifyObserversOf('after save', context, function(err) {
+            cb(err, err ? null : inst);
+          });
+        });
+      }
+    });
   }
   return cb.promise;
 };
@@ -2174,6 +2194,7 @@ EmbedsOne.prototype.update = function(targetModelData, options, cb) {
     cb = options;
     options = {};
   }
+
   var modelTo = this.definition.modelTo;
   var modelInstance = this.modelInstance;
   var propertyName = this.definition.keyFrom;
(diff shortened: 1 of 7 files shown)


test/operation-hooks.suite/embeds-one-create.suite.js
```diff
@@ -0,0 +1,186 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+var ValidationError = require('../..').ValidationError;
+
+var contextTestHelpers = require('../helpers/context-test-helpers');
+var ContextRecorder = contextTestHelpers.ContextRecorder;
+var aCtxForModel = contextTestHelpers.aCtxForModel;
+
+var uid = require('../helpers/uid-generator');
+var HookMonitor = require('../helpers/hook-monitor');
```
Amir-61: Nitpick: Shouldn't these variables be alphabetical order? First variables starting with uppercase and the variables with lowercase and alphabetical ordered?

test/operation-hooks.suite/embeds-one-destroy.suite.js
```diff
@@ -0,0 +1,140 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+var ValidationError = require('../..').ValidationError;
+
+var contextTestHelpers = require('../helpers/context-test-helpers');
+var ContextRecorder = contextTestHelpers.ContextRecorder;
+var aCtxForModel = contextTestHelpers.aCtxForModel;
+
+var uid = require('../helpers/uid-generator');
+var HookMonitor = require('../helpers/hook-monitor');
```
Amir-61: ditto

test/operation-hooks.suite/index.js
```diff
@@ -0,0 +1,17 @@
+var fs = require('fs');
+var path = require('path');
+var debug = require('debug')('test');
```
Amir-61: Please order them alphabaticlaly

lib/relation-definition.js
```diff
@@ -2113,34 +2113,54 @@ EmbedsOne.prototype.create = function(targetModelData, options, cb) {
...
+
+      var err = inst.isValid() ? null : new ValidationError(inst);
+      if (err) {
+        process.nextTick(function() {
+          cb(err);
+        });
+      } else {
+        updateEmbedded(function(err, inst) {
+          if (err) return cb(err);
+          context.instance = inst;
```
Amir-61: Shouldn't `isNewInstance` be in the context for `after save` here?

test/operation-hooks.suite/embeds-one-create.suite.js
bajtos: Well, we don't have any hard rule for ordering and grouping variables (module imports). Could you please post a code snippet showing how would you like to see L6-L13 ordered?

lib/relation-definition.js
bajtos: `ctx.isNewInstance` is marked as `// TODO`. Should I copy those TODO comments over, to make this more clear?

test/operation-hooks.suite/index.js
bajtos: will do.

test/operation-hooks.suite/embeds-one-create.suite.js
Amir-61: >we don't have any hard rule for ordering and grouping variables (module imports)
Oh I see... yea it was just a nitpick...I just saw some Strongloop folks prefer to have ASCI order of variables...
Something like this:
```
var ContextRecorder = contextTestHelpers.ContextRecorder;
var HookMonitor = require('../helpers/hook-monitor');
var ValidationError = require('../..').ValidationError;
var aCtxForModel = contextTestHelpers.aCtxForModel;
var contextTestHelpers = require('../helpers/context-test-helpers');
var uid = require('../helpers/uid-generator');
```

lib/relation-definition.js
Amir-61: I was just missing your other comments... Now I understand...
>Should I copy those TODO comments over, to make this more clear?
TBH I don't have any strong idea... Maybe on top of the page explaining that the contexts are missing some info and we gradually add missing pieces depending on what users demand? thoughts?I leave it up to you...

raymondfeng:
LGTM

test/operation-hooks.suite/embeds-one-create.suite.js
Amir-61: So my comment can be ignored if we don't have this rules... I just thought we follow this rule... Thanks Miroslav for clarification :-)

lib/relation-definition.js
bajtos: That's a great idea! I have updated the PR description and added a list of things not implemented.

bajtos:
@Amir-61 I think all comments have been addressed, I'll go ahead and land this without waiting for your final LGTM.

sebolio:
what about find??

PR #905, opened 2016-04-14
Eslint autofix for 2.x

bajtos:
@Amir-61 I run `eslint --fix` on 2.x code base to fix at least some of the differences between `master` and `2.x`. Will this make it easier for you to back-port changes from master to 2.x?

I'll leave it up to you to decide whether to land this pull request or close it unmerged.

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -526,8 +526,7 @@ function applyFilter(filter) {
         regex += '\\.';
       } else if (char === '*') {
         regex += '\\*';
-      }
-      else {
+      } else {
         regex += char;
       }
     }
@@ -568,12 +567,11 @@ function applyFilter(filter) {
       }
 
       if ('between' in example) {
-        return (testInEquality({ gte:example.between[0] }, value) &&
-                 testInEquality({ lte:example.between[1] }, value));
+        return (testInEquality({ gte: example.between[0] }, value) &&
+          testInEquality({ lte: example.between[1] }, value));
       }
 
       if (example.like || example.nlike) {
-
         var like = example.like || example.nlike;
         if (typeof like === 'string') {
           like = toRegExp(like);
@@ -592,8 +590,8 @@ function applyFilter(filter) {
       }
     }
     // not strict equality
-    return (example !== null ? example.toString() : example)
-      == (value != null ? value.toString() : value);
+    return (example !== null ? example.toString() : example) ==
+      (value != null ? value.toString() : value);
   }
 
   /**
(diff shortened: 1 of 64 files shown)


superkhau:
LGTM, but amir for final check

lib/connectors/memory.js
```diff
@@ -567,9 +567,9 @@ function applyFilter(filter) {
         return compare(example.neq, value) !== 0;
       }
 
-      if ('between' in example ) {
-        return ( testInEquality({gte:example.between[0]}, value) &&
-                 testInEquality({lte:example.between[1]}, value) );
+      if ('between' in example) {
+        return (testInEquality({ gte:example.between[0] }, value) &&
+                 testInEquality({ lte:example.between[1] }, value));
```
Amir-61: Just out of curiosity if we break a line to multiple lines, what is the preferred style?

Based on `Eslint --fix` it seems it adds 9 spaces to indent the second line; should it be two spaces?
```
return (testInEquality({ gte:example.between[0] }, value) &&
  testInEquality({ lte:example.between[1] }, value));
```

Just wanted to know the preference in the team; also the proceeding lines do not have any more indentation, I believe, right?

Example:
```
return (testInEquality({ gte:example.between[0] }, value) &&
  testInEquality({ lte:example.between[1] }, value) &&
  testInEquality({ lte:example.between[2] }, value));
```

Amir-61:
@slnode test please

lib/connectors/memory.js
superkhau: @bajtos can field this one. Please update the style-guide too with this answer as I'm also curious.

Amir-61:
LGTM

lib/connectors/memory.js
bajtos: Good catch! Apparently eslint doesn't care whether its 2 spaces, 8 spaces or 9 spaces in this case.

Your example looks exactly as what I would write myself, @Amir-61:
>```js
return (testInEquality({ gte:example.between[0] }, value) &&
  testInEquality({ lte:example.between[1] }, value) &&
  testInEquality({ lte:example.between[2] }, value));
```

bajtos:
I have rebased on top of the current 2.x. While I was addressing Amir's comment, I fixed all eslint issues I found in the memory connector (see the second commit).

lib/connectors/memory.js
bajtos: > Please update the style-guide too with this answer as I'm also curious.

https://github.com/strongloop-internal/loopback-knowledge-base/pull/10

bajtos:
@Amir-61 PTAL one more time

Amir-61:
LGTM

PR #909, opened 2016-04-19
eslint config 2.0 + remove extra empty lines

bajtos:
Upgrade eslint-config-loopback to ^2.0.0.

Remove extra empty lines to make `npm run lint` pass again.

@superkhau @Amir-61 PTAL


bajtos:
On the second thought, I don't think this is worth reviewing, I'll go ahead and land this.

PR #910, opened 2016-04-19
Implementation of patchById

Amir-61:
I just managed to do very initial implementation of `patchById` and opened this PR to have some discussion to see whether we agree about the `context` of different hooks considering the fact that in this method you would not have access to `instance` and `currentInstance`.

Apart from this I need to write a bunch of tests to verify that `patchById` works well and their operations hooks are triggered in the right way.

@bajtos PTAL :-)

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/752
Thanks!


lib/dao.js
```diff
@@ -2421,6 +2421,108 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+  var connector = Model.getDataSource().connector;
+  assert(typeof connector.update === 'function',
+    'update() must be implemented by the connector');
+
+  var hookState = {};
+  this.applyProperties(data);
+
+  var context = {
+    Model: Model,
+    instanceId: id,
```
bajtos: I am proposing to use the same context as we have for `updateAll`:

- Model
- where
- data
- hookState
- options

Note that instance id can be extracted from the `where` clause, see https://github.com/strongloop/loopback/blob/4cd84dcd8af5f4f90ff62f506a6bf49e6fc2036b/lib/persisted-model.js#L1508-L1518.

We can eventually put `getIdFromWhereByModelId` into a public LoopBack API if there is enough demand.

This is the reason why I am reluctant to add `instanceId`: when we have both `where` and `instanceId`, and a hook changes one of these properties only, it's not clear which one takes precedence. I suppose we could make `instanceId` read-only, although the downside is that `Object.defineProperty` incurs performance penalty :(

lib/dao.js
```diff
@@ -2421,6 +2421,108 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+    data: data,
+    hookState: hookState,
+    options: options,
+  };
+
+  Model.notifyObserversOf('before save', context, function(err, ctx) {
+    if (err) return cb(err);
+
+    doUpdate(ctx.instanceId, ctx.data);
```
bajtos: When we add `ctx.where` instead of `ctx.instanceId`, we need to check what is the `where` filter after hooks have run. See how `updateOrCreate` handles this situation - https://github.com/strongloop/loopback-datasource-juggler/blob/7247b6637e098047daf8d73d4d864455d80a9566/lib/dao.js#L493.

I am proposing to work incrementally and implement only the case when `ctx.where` stays pointing to `instanceId`. If `ctx.where` is changed, then `patchById` should fail with an error. Both branches should be covered with unit-tests though.

@Amir-61 Thoughts?

lib/dao.js
```diff
@@ -2421,6 +2421,108 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+        hookState: hookState,
+        options: options,
+      };
+      Model.notifyObserversOf('loaded', ctx, function(err) {
+        if (err) return cb(err);
+        var context = {
+          Model: Model,
+          isNewInstance: false,
+          hookState: hookState,
+          options: options,
```
bajtos: IIRC, `prototype.updateAttributes` return the full updated instance to the callback. Can we do the same in `patchById`? In which case, there should be `ctx.instance` in the context.

lib/dao.js
```diff
@@ -2421,6 +2421,108 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+      instanceId: instanceId,
+      data: data,
+      isNewInstance: false,
+      hookState: hookState,
+      options: options,
+    };
+    Model.notifyObserversOf('persist', ctx, function(err) {
+      if (err) return cb (err);
+      if (connector.update.length === 5) {
+        connector.update(Model.modelName, where, data, options, patchCallback);
```
bajtos: I think we should call `connector.updateAttributes` here, so that we get back full instance data (see the comment above).

`connector.update` could be used in the scenario when a hook changes `ctx.where`, but let's leave that out of scope of this PR.

lib/dao.js
```diff
@@ -2421,6 +2421,108 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+    try {
+      data = removeUndefined(data);
+      data = Model._coerce(data);
+    } catch (err) {
+      return process.nextTick(function() {
+        cb(err);
+      });
+    }
+
+    function patchCallback(err, info) {
```
bajtos: Could you please move `patchCallback` definition after `connector.update` is called? It will make the code easier to read and understand. At the moment, one has to jump a lot between different parts of this method to understand the order in which different blocks are involved.

Here is what I see now

```js
Model.notifyObserversOf('before save', // ...
// ...
Model.notifyObserversOf('loaded', // ...
// ...
Model.notifyObserversOf('after save', // ...
// ...
Model.notifyObserversOf('persist', // ...
```

And this is what I would like to see instead:

```js
Model.notifyObserversOf('before save', // ...
// ...
Model.notifyObserversOf('persist', // ...
// ...
Model.notifyObserversOf('loaded', // ...
// ...
Model.notifyObserversOf('after save', // ...
```

bajtos:
@Amir-61 one more thought to consider: eventually, I think it will be good to refactor `lib/dao.js` into multiple smaller files, one file per each operation/method: `lib/dao/patch-by-id.js`, `lib/dao/create.js`, etc. (@raymondfeng @superkhau is this in line with what you have in mind for juggler refactor?) Now that we are adding a new DAO method, I think it will be better to start following this future convention, to save us work in the future when we actually get to make the changes.

I am proposing the following:

1) Move the body of `patchById` to a new file `lib/dao/patch-by-id.js`.

```js
module.exports = function patchById(...) {
  // ...
};
```

2) In `lib/dao.js`, import the method from the external file and save it on the DataAcessObject.

```js
DataAccessObject.patchById = require('./dao/patch-by-id');
```

3) A similar approach can be applied to unit-tests. At the moment, most DAO methods have tests in `test/manipulation.test.js` and there is also giant `test/persistence-hooks.suite.js`. I have already created infrastructure for one-file-per-feature for operation hooks in #904, I am proposing to put your new tests into `test/operation-hooks.suite/dao-patch-by-id.suite.js`.

4) As for `test/manipulation.test.js`, I'd create `test/dao.suite/index.js` similar to `test/operation-hooks.suite/index.js`, put your new tests into `test/dao.suite/patch-by-id.suite.js`, and then load `test/dao.suite/index.js` from `test/manipulation.test.js` to make sure they are run as part of the suite run by connectors.

Thoughts?

lib/dao.js
Amir-61: >IIRC, prototype.updateAttributes return the full updated instance to the callback.
Unfortunately `prototype.updateAttributes` returns the number of updated instances as well. something similar to : `{ count: 1 }`

bajtos: I see.

@raymondfeng what's your opinion here? In 3.0, I was thinking about remapping `PATCH /api/my-model/:id` from `prototype.updateAttributes` to `patchById` (`PUT /api/my-model/:id` will be mapped to `replaceById`), so that we can avoid the extra database round-trip which fetches the instance before it's partially updated. However, assuming the current connector implementation, we won't be able to return the updated instance in HTTP response.

To give more context, we have already introduced `replaceById` which should be used in most scenarios where `prototype.updateAttributes` was used now and `replaceById` returns full instance data. As I see it, `patchById` should be used only in specific cases where users intentionally want to do a partial update (or perhaps use MongoDB extended operators), in which case it may be ok to not get back the fully updated instance. I don't think `updateAttributes` is able to parse & apply all MongoDB extended operations, which means the return value is inaccurate anyways. And also if NoSQL databases don't provide an atomic way how to make a partial update and get the full updated object back, then I think it's better to model our API with this limitation in mind and don't pretend we can get accurate updated data when that's not always true.

Having wrote that, I think it's ok to let `PATCH /api/mymodel/:id` to return only an `info` object like `{ count: 1 }`.

**LB 2.x**

`PUT /api/mymodel/:id` calls `prototype.updateAttributes` (partial update) and returns full instance data

**LB 3.x as proposed**

`PUT /api/mymodel/:id` calls `replaceById` (full update/replace) and returns full instance data
`PATCH /api/mymodel/:id` calls `patchById` (partial update) and returns `info` only

Alternatively, we could extend connectors to provide API where we can pass `id` and partial `data` and get back full data of the updated instance (this can be either a flag for `updateAttributes` or a new method), but I think that's a lot of extra work and not all databases provide atomic API that we could leverage.

Thoughts?

I am not sure if my comment makes sense. If it does not, then let's discuss next week on Slack.

Amir-61: >Alternatively, we could extend connectors to provide API where we can pass id and partial data and get back full data of the updated instance (this can be either a flag for updateAttributes or a new method), **but I think that's a lot of extra work and not all databases provide atomic API that we could leverage.**

I have the same opinion especially about the bolded part and I agree!

superkhau:
The refactor we talked about was smaller in scope but the ideas you're proposing are in line with the steps after. As you say, we might as well get it started now since you're already working on it. Also, the way to align tests into suites by feature is also what I had in mind. :shipit:

Amir-61:
@slnode test please

Amir-61:
@slnode test please

test/manipulation.test.js
```diff
@@ -727,6 +729,118 @@ describe('manipulation', function() {
...
+//            .then(function(p) {
+//              p.name.should.equal('Jane');
+//              p.age.should.equal(15);
+//              done();
+//            });
+//        })
+//        .catch(done);
+//    });
+//  });
+
```
Amir-61: I'll delete these commented tests please ignore them...

test/persistence-hooks.suite.js
```diff
@@ -1178,6 +1178,184 @@ module.exports = function(dataSource, should, connectorCapabilities) {
...
+//          if (err) return done(err);
+//          ctxRecorder.records.should.eql(aCtxForModel(TestModel, {
+//            isNewInstance: false,
+//          }));
+//          done();
+//        });
+//      });
+//
+//    });
+
```
Amir-61: I'll delete these commented tests please ignore them...

Amir-61:
@slnode test please

lib/dao.js
```diff
@@ -29,6 +29,7 @@ var util = require('util');
 var assert = require('assert');
 var BaseModel = require('./model');
 var debug = require('debug')('loopback:dao');
+DataAccessObject.patchById = require('./patch-by-id');
```
bajtos: Please move the new file under a new folder, i.e. `dao/patch-by-id.js`.

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
+// Copyright IBM Corp. 2015,2016. All Rights Reserved.
+// Node module: loopback-datasource-juggler
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+var ValidationError = require('./validations').ValidationError;
+var assert = require('assert');
+var dao = require('./dao');
```
bajtos: Cyclic dependencies are better to avoid. Please move the helper functions into a standalone file, e.g. `lib/dao/util.js`.

bajtos: I see we already have `lib/utils.js`, so we can either move the helpers there, but since they are DAO-specific, I think `lib/dao/helpers.js` would be better. Just keep the filename consistent with what we already have (plural `helpers.js`).

lib/dao.js
bajtos: > I use connector.updateAttributes and for now I just return only an info object like { count: 1 } to have consistent result from different connectors. Hopefully you are good with that and it makes sense :-)

Makes sense, I agree.

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
...
+  cb = cb || utils.createPromiseCallback();
+  options = options || {};
+
+  assert((typeof data === 'object') && (data !== null),
+    'The data argument must be an object');
+  assert(typeof options === 'object', 'The options argument must be an object');
+  assert(typeof cb === 'function', 'The cb argument must be a function');
+
+  var connector = Model.getDataSource().connector;
+  assert(typeof connector.update === 'function',
```
bajtos: Since we call `connector.updateAttributes` instead of `connector.update`, you should change this check too.

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
...
+    options: options,
+  };
+
+  Model.notifyObserversOf('before save', context, function(err, ctx) {
+    if (err) return cb(err);
+    var isOriginalQuery = dao.isWhereByGivenId(Model, ctx.where, id);
+
+    if (!isOriginalQuery) {
+      var err = new Error('ctx.where:' + JSON.stringify(ctx.where) + ' is not valid; the `' +
+        idPropertyName + '` original value is ' + id);
```
bajtos: Please improve the error message to mention that `patchById` does not allow hooks to change the `where` query to point to a different instance.

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
...
+      process.nextTick(function() {
+        cb(err);
+      });
+      return cb.promise;
+    }
+
+    if (strict) {
+      var instInfo = { '__unknownProperties': [] };
+      dao.applyStrictCheck(Model, strict, data,
+        instInfo, validateAndCallConnector);
```
bajtos: This doesn't make sense to me, because AFAICT you are discarding `__unknownProperties` and thus they will be never reported.

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
...
+        return doUpdate(ctx.where, ctx.data);
+      }
+
+      // Make sure id is not changed in the context of `before save`
+      checkId(ctx.data[idPropertyName], id, idPropertyName, cb);
+
+      dataObj = new Model(ctx.data);
+      dataObj.unsetAttribute(idPropertyName);
+      // validation required
+      dataObj.isValid(function(valid) {
```
bajtos: This will not work in most cases. Consider a model with required "name" and optional "age" properties, and then call `patchById` setting only `age:42`. The request is perfectly valid, but the validation will fail.

Ideally, we could run validators only on the properties affected by the patch operation. However, I think that's nontrivial and should be left out of scope of this pull request.

OTOH, because we are skipping validation, calling `patchById` makes it easy to get the data into inconsistent state. I think it may be better to not expose this method in the public REST API by default. In which case, we can probably left it out of 3.0 completely and wait until we have more time available to implement it fully with validations.

Thoughts?

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
...
+  // data = data.toObject();
+  // data = removeUndefined(data);
+
+  var hookState = {};
+  this.applyProperties(data);
+
+  var context = {
+    Model: Model,
+    where: where,
+    data: data,
```
bajtos: Should we report `isNewInstance:false` too?

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
...
+      } else {
+        connector.updateAttributes(Model.modelName, where[idPropertyName], ctx.data, patchCallback);
+      }
+    });
+
+    function patchCallback(err, info) {
+      if (err) return cb(err);
+      var ctx = {
+        Model: Model,
+        data: data,
```
bajtos: Are other DAO methods reporting `isNewInstance` for `loaded` hook?

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
...
+        var context = {
+          Model: Model,
+          isNewInstance: false,
+          hookState: hookState,
+          options: options,
+        };
+        Model.notifyObserversOf('after save', context, function(err) {
+          if (!err) Model.emit('changed', info);
+          // Just for consistency we return { count: 1} for all connectors
+          cb(err, { count: 1 });
```
bajtos: Hmm, this is not correct. What if the record-to-be-updated was deleted? I think we should report 0 or 1 depending on the result from the database. Unless the operation fails when the record no longer exists. Please add a unit-test to verify this edge case.

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
...
+        });
+      });
+    }
+  }
+  return cb.promise;
+};
+
+var idValidationFailure;
```
bajtos: I find this suspicious, isn't this a module-global singleton shared by all invocations of `checkId`? I don't think that will work.

test/dao.suite/index.js
```diff
@@ -0,0 +1,17 @@
...
+  operations = operations.filter(function(it) {
+    return it !== path.basename(__filename) &&
+      !!require.extensions[path.extname(it).toLowerCase()];
+  });
+  for (var ix in operations) {
+    var name = operations[ix];
+    var fullPath = require.resolve('./' + name);
+    debug('Loading test suite %s (%s)', name, fullPath);
+    require(fullPath).apply(this, arguments);
+  }
```
bajtos: Makes me wonder, can we put this code into a shared helper?

```js
module.exports = function(dataSource, should, connectorCapabilities) {
  helpers.runAllSuitesInDirExceptFile(__dirname, __filename);
};
```

test/dao.suite/patch-by-id.suite.js
```diff
@@ -0,0 +1,188 @@
...
+        var hashed = false;
+        if (!plain) return;
+        var pos = plain.indexOf('-');
+        if (pos !== -1) {
+          var head = plain.substr(0, pos);
+          var tail = plain.substr(pos + 1, plain.length);
+          hashed = head.toUpperCase() === tail;
+        }
+        if (hashed) return;
+        this.$password = plain + '-' + plain.toUpperCase();
```
bajtos: Can we please move the body of this setter function to a shared method/function? Or perhaps extract the whole `StubUser`

test/dao.suite/patch-by-id.suite.js
```diff
@@ -0,0 +1,188 @@
...
+        Person.patchById(personId, { id: personId + 1, name: 'John' },
+        function(err, info) {
+          should.exist(err);
+          done();
+        });
+      });
+
+    it.skip('should allow model instance on patchById', function(done) {
+      // QUESTION-test[1]: Can we pass undefined here and should we remove undefined in the method?
+      // Please see the other comment in patch-by-id.js
```
bajtos: Yeah, it is a know problem that you cannot pass `age:undefined` to clear a value via updateAttributes. I think this is also a limitation of MongoDB. That's one of the reasons for introducing `replaceById`.

bajtos: > Can we pass undefined here and should we remove undefined in the method?

Yes, the test as written below makes sense to me.

lib/patch-by-id.js
bajtos: Hmm, I see you have a test case for this and I assume it passes. That means I do not fully understand what this piece of code is doing. Could you please explain a bit?

test/dao.suite/patch-by-id.suite.js
```diff
@@ -0,0 +1,188 @@
...
+// Node module: loopback-datasource-juggler
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+module.exports = function(dataSource, should, connectorCapabilities) {
+  describe('patchById', function() {
+    var personId, Person, StubUser;
+    before(function setupDatabase(done) {
+      Person = dataSource.define('Person', {
+        name: String,
```
bajtos: Please make `name` required and add a test for `patchById(id, { age: 15 })`. See one of the earlier comments.

test/operation-hooks.suite/dao-patch-by-id.suite.js
```diff
@@ -0,0 +1,227 @@
...
+    });
+
+    beforeEach(function setupDatabase(done) {
+      TestModel = dataSource.createModel('TestModel', {
+        // Set id.generated to false to honor client side values
+        id: { type: String, id: true, generated: false, default: uid.next },
+        name: { type: String, required: true },
+        extra: { type: String, required: false },
+      });
+      done();
```
bajtos: The other test file(s) in `test/operation-hooks.suite` use promise-based API, could you please do the same in this new file for consistency?

test/operation-hooks.suite/dao-patch-by-id.suite.js
```diff
@@ -0,0 +1,227 @@
...
+            'TestModel:after save',
+          ]);
+          done();
+        });
+    });
+
+    it('triggers `before save` hook', function(done) {
+      TestModel.observe('before save', ctxRecorder.recordAndNext());
+
+      TestModel.patchById(instanceId, { name: 'changed' }, function(err, info) {
```
bajtos: AFAICT, all places pass mostly the same arguments to `patchById`. Could you please refactor this into a helper method, see the other files in this directory for inspiration.

```js
TestModel.observe('before save', ctxRecorder.recordAndNext());
return callPatchById().then(function() {
    // ...
  });
});
```

test/operation-hooks.suite/dao-patch-by-id.suite.js
```diff
@@ -0,0 +1,227 @@
...
+        [err].should.eql([expectedError]);
+        done();
+      });
+    });
+
+    it('applies updates from `before save` hook', function(done) {
+      TestModel.observe('before save', function(ctx, next) {
+        ctx.data.extra = 'extra data';
```
bajtos: Isn't this the same value as you set when creating the model instance in a before-each hook above? I think the hook should not initialise the `extra` property.

test/operation-hooks.suite/dao-patch-by-id.suite.js
```diff
@@ -0,0 +1,227 @@
...
+        done();
+      });
+    });
+
+    it('applies updates from `persist` hook', function(done) {
+      TestModel.observe('persist', ctxRecorder.recordAndNext(function(ctx) {
+        ctx.data.extra = 'hook data';
+      }));
+
+      TestModel.settings.updateOnLoad = true;
```
bajtos: This flag was added to preserve backwards compatibility, please don't use it in new features. Besides that, I believe `patchById` does not return full instance, thus this is not relevant at all.

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
...
+
+    function patchCallback(err, info) {
+      if (err) return cb(err);
+      var ctx = {
+        Model: Model,
+        data: data,
+        hookState: hookState,
+        options: options,
+      };
+      Model.notifyObserversOf('loaded', ctx, function(err) {
```
bajtos: This does not make sense to me. The `loaded` hook is called with the data loaded from the database, which is not the case of `patchById`. I think we should not trigger this hook at all.

lib/patch-by-id.js
```diff
@@ -0,0 +1,187 @@
...
+        if (err) return cb(err);
+        var context = {
+          Model: Model,
+          isNewInstance: false,
```
bajtos: Please add `where` and `data`.

lib/patch-by-id.js
Amir-61: Yea I noticed that restriction too and unfortunately I don't think if there is any efficient alternative either... The only alternative I can think of is calling `findById`, which definitely does not make sense to call another DB query and it makes `patchById` more computational expensive... On the other hand skipping the validation makes `data` inconsistent and as you mentioned I don't think if that's a rational decision either.

Initially the decision for `loopback 3.0` was to map `PUT` to static `replaceById` and `PATCH` to static `patchById` instead of `prototype.updateAttributes`; however now that we cannot have the right validation for static `patchById` as a result of not having access to `instance` I agree with you that we should map `PUT` to static `replaceById` and `PATCH` to `prototype.updateAttributes` and we can say implementation of static `patchById` is out of scope of this milestone.

PR #913, opened 2016-04-26
[WIP] Inherit parent validations

davidcheung:
Currently models defined in js files are not inherited to the submodel
this is to fix https://github.com/strongloop/loopback/issues/1485

### DIFF
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -211,8 +211,8 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
 
   util.inherits(ModelClass, ModelBaseClass);
 
-  //inherit validations
-  if (ModelBaseClass.validations) {
+  // inherit validations
+  if (ModelBaseClass.validations && !settings.realmRequired) {
     ModelClass.validations = traverse(ModelBaseClass.validations).clone();
   }
 
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -18,7 +18,6 @@ var ModelDefinition = require('./model-definition.js');
 var mergeSettings = require('./utils').mergeSettings;
 var MixinProvider = require('./mixins');
 var debug = require('debug')('loopback:dao:modelbuilder');
-var traverse = require('traverse');
 
 // Set up types
 require('./types')(ModelBuilder);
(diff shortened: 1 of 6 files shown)


lib/model-builder.js
```diff
@@ -209,6 +211,11 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
 
   util.inherits(ModelClass, ModelBaseClass);
 
+  // inherit validations
+  if (ModelBaseClass.validations && !settings.realmRequired) {
+    ModelClass.validations = traverse(ModelBaseClass.validations).clone();
```
bajtos: If `validations` is an object, then I am proposing to use prototypal inheritance here:

```js
ModelClass.validations = Object.create(ModelBaseClass.validations);
```

If it's an array, then I think we can simply clone the array only?

```js
ModelClass.validations = ModelBaseClass.validations.slice();
```

Thoughts?

raymondfeng: Really we use inheritance here to achieve mixin. For example,

```
var SubModel = class extends BaseModel {
  // Additional static properties
  // Additional static methods
  // Additional prototype properties
  // Additional prototype methods
}
```
With the code above, SubModel will inherit both static and prototype properties/methods from the BaseModel. But for our purpose, we want to clone the static properties so that the SubModel won't share them with BaseModel.

davidcheung: It's an object like this 
```js
{
  "name": [{
    "validation": "presence",
    "options": {}
  }]
}
```

i tried `ModelClass.validations = Object.create(ModelBaseClass.validations);` it creates a new clone, but 
because in [`isValid()`](https://github.com/strongloop/loopback-datasource-juggler/blob/89bf94245b32751422d7ff6d8432bd9c73c75b06/lib/validations.js#L450)
it does ` var attrs = Object.keys(validations || {});` so it wouldnt get the prototype keys it seems

davidcheung:
Placeholder info for user realm questions

validations does not have this logic to `normalizeCredentials()`, https://github.com/strongloop/loopback/blob/88d41f84fd8eac997a31388ccdbd3ce026ab0dfe/common/models/user.js#L126-L155?

PR #917, opened 2016-04-28
RELEASE-NOTES: describe 30283291

bajtos:
Add an entry for a breaking change made in 30283291 that was not described previously.

/to @Amir-61 please review


Amir-61:
LGTM :shipit:

PR #918, opened 2016-04-29
[SEMVER-MAJOR] DAO.create: don't return the instance

bajtos:
Simplify DataAccessObject.create() and stop returning the instance/array of instances. Users should always use callback (or returned promise) to get the instance(s) created.

See the discussion in #418 around https://github.com/strongloop/loopback/issues/418#issuecomment-74691763 for more details.

/to @Amir-61 @partap please review


Amir-61:
LGTM :shipit: Thanks!

Issue #914, opened 2016-04-26
updateOrCreate does not return error in case of "error"


harshu108:
When my connector returns error in callback to updateOrCreate, the dao.js doesn't mitigate it back.

  ```
     function done(err, data, info) {
          var context = {
            Model: Model,
            data: data,
            isNewInstance: info && info.isNewInstance,
            hookState: ctx.hookState,
            options: options,
          };
          Model.notifyObserversOf('loaded', context, function(err) {
            if (err) return cb(err);
```

The call returns to `function done()` and it straight away notifies observers. The line that checks for error is missing.

Amir-61:
Thanks @harshu108,

Thats an easy fix could you please contribute a patch and please ping me in the PR you send so I'll make sure it gets landed ASAP.

Also please add test case reproducing the issue in our test suite.

Thanks

Amir-61:
@harshu108,

On the second thought since it is an obvious bug, we may not need any test case for that. I just sent [a PR](https://github.com/strongloop/loopback-datasource-juggler/issues/914) to fix this.

Thanks for reporting this issue :-)

Amir-61:
Connect to https://github.com/strongloop/loopback-datasource-juggler/issues/914

Amir-61:
Landed https://github.com/strongloop/loopback-datasource-juggler/pull/954. 

Closing the ticket now!

Thanks for reporting the error!

PR #919, opened 2016-05-01
Updated dao.js updateOrCreate() for issue #914

harshu108:
Whenever updateOrCreate() in connector returns an error, the method in DAO should mitigate it back to the calling source. Added the error check that was missing.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -573,7 +573,7 @@ DataAccessObject.upsert = function(data, options, cb) {
             hookState: ctx.hookState,
             options: options,
           };
-          if(err){
+          if (err) {
             return cb(err);
           }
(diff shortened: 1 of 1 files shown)


slnode:

lib/dao.js
```diff
@@ -573,6 +573,9 @@ DataAccessObject.upsert = function(data, options, cb) {
             hookState: ctx.hookState,
             options: options,
           };
+          if (err) {
+            return cb(err);
+          }
```
superkhau: Move this up to line 570 (first line in the body of the done declaration0

superkhau:
@harshu108 One minor comment, otherwise LGTM.

superkhau:
Can you squash the commits into one also?

PR #928, opened 2016-05-10
Backport/eslint

superkhau:
@bajtos Review please. Backporting with ESLint fixes to make Travis pass.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -67,9 +67,9 @@ function convertSubsetOfPropertiesByType(inst, data) {
   for (var key in data) {
     // Convert the properties by type
     typedData[key] = inst[key];
-    if (typeof typedData[key] === 'object'
-            && typedData[key] !== null
-            && typeof typedData[key].toObject === 'function') {
+    if (typeof typedData[key] === 'object' &&
+        typedData[key] !== null &&
+        typeof typedData[key].toObject === 'function') {
       typedData[key] = typedData[key].toObject();
     }
   }
@@ -907,7 +907,6 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
             }
             if (!err) Model.emit('changed', obj);
           });
-
         } else {
           if (cb.promise) {
             cb(err, [obj, created]);
@@ -924,7 +923,7 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
       where: query.where,
       data: data,
       isNewInstance: true,
-      currentInstance : currentInstance,
+      currentInstance: currentInstance,
       hookState: hookState,
       options: options,
     };
@@ -1748,8 +1747,7 @@ DataAccessObject.find = function find(query, options, cb) {
 
         cb(err, results);
       });
-    }
-    else {
+    } else {
       cb(err, data || []);
     }
   };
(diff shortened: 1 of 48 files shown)


lib/dao.js
```diff
@@ -67,9 +67,8 @@ function convertSubsetOfPropertiesByType(inst, data) {
   for (var key in data) {
     // Convert the properties by type
     typedData[key] = inst[key];
-    if (typeof typedData[key] === 'object'
-            && typedData[key] !== null
-            && typeof typedData[key].toObject === 'function') {
+    if (typeof typedData[key] === 'object' && typedData[key] !== null &&
+        typeof typedData[key].toObject === 'function') {
```
bajtos: One expression per line please. We don't spell it explicitly in our style guide yet, but the examples are already following that rule.

https://github.com/strongloop-internal/loopback-knowledge-base/blob/master/style-guide.md#indentation-of-multi-line-expressions-in-if

```js
if (typeof typedData[key] === 'object' &&
    typedData[key] !== null &&
    typeof typedData[key].toObject === 'function') {
```

lib/datasource.js
```diff
@@ -1905,10 +1899,13 @@ DataSource.prototype.copyModel = function copyModel(Master) {
   hiddenProperty(Slave, 'relations', Master.relations);
 
   if (!(className in dataSource.modelBuilder.models)) {
-
     // store class in model pool
     dataSource.modelBuilder.models[className] = Slave;
-    dataSource.modelBuilder.definitions[className] = new ModelDefinition(dataSource.modelBuilder, md.name, md.properties, md.settings);
+    dataSource.modelBuilder.definitions[className] = new ModelDefinition(
+      dataSource.modelBuilder,
+      md.name, md.properties,
```
bajtos: Should this be split into two lines for consistency?

```js
new ModelDefinition(
  dataSource.modelBuilder,
  md.name,
  md.properties,
  md.settings);
```

lib/geo.js
```diff
@@ -138,11 +138,18 @@ function GeoPoint(data) {
     };
   }
 
-  assert(Array.isArray(data) || typeof data === 'object' || typeof data === 'string', 'must provide valid geo-coordinates array [lat, lng] or object or a "lat, lng" string');
+  assert(
+    Array.isArray(data) ||
+    typeof data === 'object' ||
+    typeof data === 'string',
```
bajtos: I think these last two lines should be indented by one extra level, to make it clear they are part of the first argument, as opposed to the next line.

Notice that you have already applied that rule on lines 145 and 146 below.

```js
assert(
  Array.isArray(data) ||
    typeof data === 'object' ||
    typeof data === 'string',
  'must provide valid geo-coordinates array [lat, lng] or object or a ' +
    '"lat, lng" string');
```

lib/geo.js
```diff
@@ -138,11 +138,18 @@ function GeoPoint(data) {
...
+    'must provide valid geo-coordinates array [lat, lng] or object or a ' +
+      '"lat, lng" string'
+  );
```
bajtos: We don't have an explicit rule for this, but I think we usually put the closing brackets on the preceding line.

bajtos: You are already doing it in other places in this patch, see e.g. https://github.com/strongloop/loopback-datasource-juggler/pull/928/files#diff-b367a815d81f15531e0d0553ea8aaa5aR152

test/memory.test.js
```diff
@@ -281,21 +280,27 @@ describe('Memory connector', function() {
...
       });
     });
 
-    it('should successfully extract 2 users using implied and & and', function(done) {
-      User.find({ where: { name: 'John Lennon', and: [{ role:'lead' }, { vip:true }] }}, function(err, users) {
-        should(users.length).be.equal(1);
-        should(users[0].name).be.equal('John Lennon');
-        done();
+    it('should successfully extract 2 users using implied and & and',
+      function(done) {
```
bajtos: To prevent whitespace changes when fixing lines too long, we usually don't indent the callback:

```js
it('should successfully extract 2 users using implied and & and',
function(done) {
  // indentation stays the same here
});
```

bajtos:
Reviewed. I think the most important thing is to keep the style in `2.x` as close to the `master` branch to make back-porting easy. If any code in `master` branch contradicts what I proposed in my comments above, then please apply the style from `master` instead.

superkhau:
@bajtos I updated in a second commit for you to review the changes applied from your feedback. Will squash after the LGTM. Can you review again please?

Issue #824, opened 2016-01-19
Remove legacy model hooks


bajtos:
Remove legacy model hooks triggered via `inst.trigger`. Users should have migrated to new Operation hooks quite some time ago.

https://github.com/strongloop/loopback-datasource-juggler/pull/473

> **Model hooks**
> beforeValidate
> afterValidate
> beforeCreate
> afterCreate
> beforeSave
> afterSave
> beforeUpdate
> afterUpdate
> beforeDestroy
> afterDestroy

https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/hooks.test.js

davidcheung:
- [ ] Remove use of [inst.trigger](https://github.com/strongloop/loopback-datasource-juggler/blob/5dca27278a37b1e2e0fa572c3ae20681cd500680/lib/dao.js#L2011-L2012)
- [ ] test would probably break because of this
- [ ] update documentation to relfect change 
- [ ] update `3.0-RELEASE-NOTES.md` to document breaking change

0candy:
After discussion with Miroslav, we decided not to include this change for 3.0 as this change is more than just removing deprecated code. There is a feature for `inst.triggerParent` in embedsMany that needs more consideration and possibly a rewrite for embedded relations to continue work. It is for updating the parent model if any changes are made to the relation. Also, embedsOne does not have this feature yet and it should be considered. This part needs a redesign and rewrite to consider the different cases and to make it all consistent.

Here's a proposed rewrite for `inst.triggerParent` which should also be added to other embeddedRelations.
https://github.com/strongloop/loopback-datasource-juggler/commit/d0cc791bc65e5ee7d2447698539da1d1ca0591ee

PR #934, opened 2016-05-13
Remove legacy model hooks and events

0candy:
Connect to https://github.com/strongloop/loopback-datasource-juggler/issues/824


3.0-RELEASE-NOTES.md
```diff
@@ -49,3 +49,31 @@ without any callback, a promise is returned.
 
 See [pull request 918](https://github.com/strongloop/loopback-datasource-juggler/pull/918)
 for more details.
+
+
+##Remove deprecated model hooks and model events
```
bajtos: nitpick: we usually put a space after `##`, see other places in this file

3.0-RELEASE-NOTES.md
```diff
@@ -49,3 +49,31 @@ without any callback, a promise is returned.
...
+beforeValidate
+afterValidate
+beforeCreate
+afterCreate
+beforeSave
+afterSave
+beforeUpdate
+afterUpdate
+beforeDestroy
+afterDestroy
```
bajtos: Could you please add a note that `afterInitialize` is staying around, because it's not a typical (async) model hook, but a just sync method.

lib/dao.js
```diff
@@ -324,82 +324,72 @@ DataAccessObject.create = function(data, options, cb) {
...
-        Model.notifyObserversOf('persist', context, function(err) {
-          if (err) return cb(err);
 
-          if (connector.create.length === 4) {
-            connector.create(modelName, obj.constructor._forDB(context.data), options, createCallback);
-          } else {
-            connector.create(modelName, obj.constructor._forDB(context.data), createCallback);
-          }
+        Model.notifyObserversOf('after save', context, function(err) {
+          if (err) return cb(err);
```
bajtos: I think this line is not necessary, because the only other line in this function is the next one calling `cb(err, obj)`.

lib/dao.js
```diff
@@ -2214,67 +2194,56 @@ DataAccessObject.prototype.save = function(options, cb) {
...
-            data: data,
-            where: byIdQuery(Model, getIdValue(Model, inst)).where,
-            currentInstance: inst,
+            instance: inst,
+            isNewInstance: result && result.isNewInstance,
             hookState: hookState,
             options: options,
           };
+          Model.notifyObserversOf('after save', context, function(err) {
+            if (err) return cb(err, inst);
```
bajtos: ditto, I think this should be enough:

```
Model.notifyObserversOf('after save', context, function(err) {
  cb(err, obj);
});
```

lib/hooks.js
```diff
@@ -22,55 +20,19 @@ function Hookable() {
  * List of hooks available
  */
 Hookable.afterInitialize = null;
```
bajtos: Can we take one step further in this cleanup, remove entire `lib/hooks.js` and then re-implement triggering `afterInitialize` directly in `lib/model.js`?

lib/relation-definition.js
```diff
@@ -2458,20 +2458,6 @@ EmbedsMany.prototype.prepareEmbeddedInstance = function(inst) {
         process.nextTick(callback);
       }
     };
-    var originalTrigger = inst.trigger;
-    inst.trigger = function(actionName, work, data, callback) {
-      if (typeof work === 'function') {
-        var originalWork = work;
-        work = function(next) {
-          originalWork.call(this, function(done) {
-            inst.triggerParent(actionName, function(err, inst) {
```
bajtos: Ha, I was not aware of this feature. Could you please find more details about what was this code performing from user's point of view and then describe it in release notes?

My guess is that this code is ensuring that when you call something like `Categories.products.create(...)`, the hooks are triggered both on Category and Product models. That's just a wild guess. `git blame` and related discussions in pull requests may help you to shed more light on this piece of code.

test/manipulation.test.js
```diff
@@ -176,18 +176,6 @@ describe('manipulation', function() {
         .catch(done);
     });
 
-    it('should work when called without callback', function(done) {
-      Person.afterCreate = function(next) {
```
bajtos: If we want to be correct, then this should have to be rewritten using `after save` hook. However, since this test was written, the API was slightly changed and a callback-less invocation returns a promise, which is already tested elsewhere, and thus it should be safe to remove this test (and other tests similar to this one).

test/manipulation.test.js
```diff
@@ -215,18 +203,6 @@ describe('manipulation', function() {
...
-    it('should work when called with no data and callback', function(done) {
-      Person.afterCreate = function(next) {
-        this.should.be.an.instanceOf(Person);
-        should.not.exist(this.name);
-        should.exist(this.id);
-        Person.afterCreate = null;
-        next();
-        setTimeout(done, 30);
-      };
-      Person.create();
```
bajtos: I am not sure that we have a unit-test for this invocation, could you please check?

test/relations.test.js
```diff
@@ -2527,42 +2527,6 @@ describe('relations', function() {
         done();
       });
     });
-
-    it('should allow to create belongsTo model in beforeCreate hook', function(done) {
```
bajtos: Uff, this is a weird test. Is it related to https://github.com/strongloop/loopback-datasource-juggler/pull/934/files#r63366803? Could you please check whether it is possible to reproduce the same behaviour using operation hooks?

test/relations.test.js
```diff
@@ -2527,42 +2527,6 @@ describe('relations', function() {
...
-      Fear.create(function(err, fear) {
-        should.not.exists(err);
-        should.exists(fear);
-        fear.mindId.should.be.equal(mind.id);
-        should.exists(fear.mind());
-        done();
-      });
-    });
-
-    it('should allow to create belongsTo model in beforeCreate hook with promises', function(done) {
```
bajtos: Yeah, I think it was not necessary to create a promise-based version of the test above, it's good this is going away.

lib/relation-definition.js
0candy: This change was added so that when saving an embedded model, it also correctly updates the parent model.

https://github.com/strongloop/loopback-datasource-juggler/commit/96a276a12b4a75a85b1092ccbd4ad0b31c5fd056

test/manipulation.test.js
0candy: It is tested in a few other files.

https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/basic-querying.test.js
https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/common_test.js

lib/model.js
```diff
@@ -340,7 +344,29 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
       }
     }
   }
-  this.trigger('initialize');
+  this.__trigger('initialize');
+};
+
+ModelBaseClass.prototype.__trigger = function trigger(actionName, work, data, callback) {
```
bajtos: Since `__trigger` is used only for `initialize` and always called as `trigger('initialize')`, we can remove the last three arguments and simplify the code below a lot.

It makes me wonder why cannot we simply change L324 to directly call `this.afterInitialize`? I think it has the same effect as the current solution.

```diff
- this.trigger('initialize');
+ if (this.afterInitialize)
+   this.afterInitialize();
+ else if (this.postInitialize) {
+  this.postInitialize();
```

0candy: I wanted to confirm with you whether I could just remove the trigger function and call afterinitialize directly, but I wasn't sure if we'll be adding anything in the future. I think we can remove that, do you agree?

bajtos: Yes please, let's remove the whole hook/trigger infrastructure and call `afterInitialize` (and/or `postInitialize`) directly.

bajtos:
Closing as abandoned. Since this is a breaking change, we won't be able to land it until 4.0 anyways.

Issue #805, opened 2015-12-23
(node) warning: possible EventEmitter memory leak detected. 11 connected listeners added. Use emitter.setMaxListeners() to increase limit.


OwenBrotherwood:
```
(node) warning: possible EventEmitter memory leak detected. 11 connected listeners added. Use emitter.setMaxListeners() to increase limit.
Trace
    at DataSource.addListener (events.js:239:17)
    at DataSource.once (events.js:265:8)
    at DataSource.ready (/data/loopback-angular-admin/loopback-angular-admin/node_modules/loopback-datasource-juggler/lib/datasource.js:2038:8)

```

https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/datasource.js
contains a

```
var EventEmitter = require('events').EventEmitter;
```

I failed to increase maxlisteners, and avoid warning, in the code.
- https://nodejs.org/docs/latest/api/events.html#events_emitter_setmaxlisteners_n

@raymondfeng   What do you need to help me close the issue?
The issue _may_ be in loopback-datasource-juggler.  

Current problem for me is to find the quickest easiest patch. 
maxlisteners(10) was made for a reason maxlisteners(0) is probably not good. 
https://nodejs.org/docs/latest/api/events.html#events_emitter_getmaxlisteners

---

x-ref loopback issues:
open https://github.com/strongloop/loopback/issues/1186
closed https://github.com/strongloop/loopback/issues/404

downstream issue: https://github.com/beeman/loopback-angular-admin/issues/150

sochka:
I have the same problem. The code which triggers the warning:
```
module.exports = (app, cb) => {
  app.dataSources.db.autoupdate(err => {
    if (err) throw  err;
    cb();
  });
};
```
How do I fix that? process.setMaxListeners(0) doesn't help.

superkhau:
Would one of you guys like to submit a patch? I think it's as simple as using setMaxListenerss as recommended, but somewhere in Juggler's code. I do not have time to look into this ATM, but I could do a code review/merge if you guys send a PR.

superkhau:
Also, which connectors are you guys using? Is the same error happening in multiple connectors or just in-memory?

sochka:
@superkhau for me this happens with PostgreSQL connector but not with memory connector.

sochka:
[...]

PR #939, opened 2016-05-20
[RFR][MINOR] Use Env Variable to configure Max Listeners at runtime

chrisamoore:
@bajtos @raymondfeng please review, simple configuration option to solve, (keeps your hard coded limits, but if I want to push the limits in test scenarios or architecture I should be able to ): 
- https://github.com/strongloop/loopback/issues/404
- https://github.com/strongloop/generator-loopback/issues/89
- https://github.com/strongloop/loopback-datasource-juggler/issues/805

### DIFF
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -185,7 +185,7 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
     var events = new EventEmitter();
     // The model can have more than 10 listeners for lazy relationship setup
     // See https://github.com/strongloop/loopback/issues/404
-    events.setMaxListeners(32);
+    events.setMaxListeners(Number(process.env.MAX_LISTENERS) || 32);
     for (var f in EventEmitter.prototype) {
       if (typeof EventEmitter.prototype[f] === 'function') {
         if (f !== 'on') {


slnode:

chrisamoore:
👍 Bumpy !

bajtos:
@slnode ok to test

bajtos:
@chrisamoore thank you for the pull request. I am not comfortable reading ENV variables so deep inside loopback code. A much cleaner approach is to expose API allowing LoopBack apps to set the MaxListener limit, and then optionally set this limit from ENV in loopback-boot.

Depending on your use cases, I can imagine different options for such API:

1) Simplest to implement: a model setting. Down side: users have to change this value for each model, and cannot change it for models they don't control (e.g. loopback-workspace models).

Implementation:

```js
// in lib/model-builder.js
events.setMaxListeners(ModelClass.settings.maxEventListeners || 32);
```

Usage:

```
// common/models/my-model.json
{
  "name": "MyModel",
  "base": "PersistedModel",
  "maxEventListeners": 64,
  "properties": {
      // ...
  }
}
```

2) A setting on `modelBuilder` instance, exposed via loopback `app.registry` object. 

Implementation:
 
```js
// in juggler's lib/model-builder.js
events.setMaxListeners(this.maxModelEventListeners || 32);

// in loopback-boot
if (process.env.LB_MAX_LISTENERS) {
  app.registry.modelBuilder.maxModelEventListeners = +process.env.LB_MAX_LISTENERS;
}
```

@chrisamoore @raymondfeng thoughts?

chrisamoore:
The issue with the static nature of the model config is that it is always at that limit, where as with the ENV flag I can toggle it very simply  for tests (which is bombing, and on an older project it kept bombing out due to the monolithic nature of the application) For my particular use case this is only for the nature of the tests we are writing which are exhaustive in nature.  cc @bajtos

BerkeleyTrue:
I'd also like to see this in model-config instead of as a env variable.

Maybe a happy medium would be model-config with a env variable global override?

chrisamoore:
@bajtos @BerkeleyTrue Shall I amend the PR to reflect the comment of @BerkeleyTrue above ?

chrisamoore:
Also @bajtos the build looks a little funky in it's failures

bajtos:
Let me summarize my understanding.

In @chrisamoore's use case, one wants to set a single flag in a single place to change the `maxListeners` value in all models, in a way that makes it easy to change this value from the tests.

@BerkeleyTrue would like to change this value on per-model basis.

> Maybe a happy medium would be model-config with a env variable global override?

Providing a setting at the model level with an option to override the value globally is fine with me. However, I am still opposed to control this directly via ENV variables here in juggler. As I wrote in https://github.com/strongloop/loopback-datasource-juggler/pull/939#issuecomment-220985433, the global setting should be kept in `ModelBuilder`.

BTW I created a new issue to keep track of the requirement to configure `modelBuilder` settings via JSON files, see https://github.com/strongloop/loopback/issues/2370.

---

For this pull request, I am proposing the following:

1) Implement my suggestion described in https://github.com/strongloop/loopback-datasource-juggler/pull/939#issuecomment-220985433 with one small change: store the flag in `modelBuilder.options.maxModelEventListeners` instead of `modelBuilder.maxModelEventListeners` to prevent possible clash between setting names and instance methods in the future.

This will allow @chrisamoore to handle `MAX_LISTENERS` ENV variable in his code, e.g. as follows:

```
// server/server.js
var app = loopback();
if (process.env.MAX_LISTENERS) {
  app.registry.modelBuilder.settings.maxModelEventListeners = +process.env.MAX_LISTENERS;
}
boot(app, __dirname);
```

Once https://github.com/strongloop/loopback/issues/2370 is landed, one can configure max listeners e.g. this way:

```js
// server/config.js
{
  "restApiRoot": "/api",
  // ...
  "modelBuilder": { 
    "maxModelEventListeners": "${MAX_LISTENERS}"
  }
}
```

or alternatively, assuming tests are run with `NODE_ENV=test`:

```js
// server/config.test.js
{
  "modelBuilder": {
    "maxModelEventListeners": 256
  }
}
```

(Note that I am assuming the tests are loading the app by importing whole `server/server.js`.)

2) To support @BerkeleyTrue's use case, modelBuilder can take into account model-level settings too. I am fine to keep that out of scope of this pull request though.

chrisamoore:
@bajtos would this need to include a PR to loopback as well to affect the `server.js` as well ?

bajtos:
> would this need to include a PR to loopback as well to affect the server.js as well ? 

I don't think so. For short-term, users of this new feature should edit their `server/server.js` manually in their projects. For long-term, I'd like to implement strongloop/loopback#2370.

Both ways, there is no need to change loopback now, this pull request can be landed on its own.

bajtos:
@chrisamoore ping, what's the status of this patch? Are you still keen to get it finished?

bajtos:
I am closing this pull request as abandoned. If there is anybody willing to take over this work, then feel free to send a new pull request. See https://github.com/strongloop/loopback-datasource-juggler/pull/939#issuecomment-221256488 for guidance.

PR #940, opened 2016-05-21
Model definition strict option defaults to true

### LINKED ISSUE #923: Make strict option default to true for anonymous models
When declaring the model definition JSON file, you can directly declare embedded document properites such as

```javascript
...
{
	"properties":{
		"range": {
			"min":"number",
			"max":"number"
		}
	}
}
```
and it works and also accepts only numbers. The problem with this is that it also accepts any other property in the **range** subdoc (strict option is set to false). This is due to **resolveType** in **model-builder.js** which creates an anonymous model for the subdocument defined (range). The problem is that resolveType defines the model with no strict option set, so it defaults to false. See [here](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L711)

I believe that the most common behaviour would be to only keep properties that are defined in the model.... So default behaviour for strict options should be set to true, and not have to declare a separate definition file for such a small schema like **range** just to declare the **strict** option.

mitsos1os:
When defining a new model use true as default value for strict option. Also updated tests

Fix #923

### DIFF
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -709,7 +709,11 @@ ModelBuilder.prototype.resolveType = function(type) {
       return this.resolveType(type.type);
     } else {
       return this.define(this.getSchemaName(null),
-        type, { anonymous: true, idInjection: false });
+        type, {
+          anonymous: true,
+          idInjection: false,
+          strict: this.settings && this.settings.strictEmbeddedModels || false,
+        });
     }
   } else if ('function' === typeof type) {
     return type;
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -44,6 +44,7 @@ function ModelBuilder() {
   // create blank models pool
   this.models = {};
   this.definitions = {};
+  this.settings = {};
   this.mixins = new MixinProvider(this);
   this.defaultModelBaseClass = DefaultModelBaseClass;
 }
@@ -712,7 +713,7 @@ ModelBuilder.prototype.resolveType = function(type) {
         type, {
           anonymous: true,
           idInjection: false,
-          strict: this.settings && this.settings.strictEmbeddedModels || false,
+          strict: this.settings.strictEmbeddedModels || false,
         });
     }
   } else if ('function' === typeof type) {
--- a/test/loopback-dl.test.js
+++ b/test/loopback-dl.test.js
@@ -1880,4 +1880,79 @@ describe('ModelBuilder options.models', function() {
       assert.equal(M2.M1, M1, 'M1 should be injected to M2');
     });
 
+  it('should use false strict mode for embedded models by default', function() {
+    var builder = new ModelBuilder();
+    var M1 = builder.define('testEmbedded', {
+      name: 'string',
+      address: {
+        street: 'string',
+      },
+    });
+    var m1 = new M1({
+      name: 'Jim',
+      address: {
+        street: 'washington st',
+        number: 5512,
+      },
+    });
+    assert.equal(m1.address.number, 5512, 'm1 should contain number property in address');
+  });
+  it('should use the strictEmbeddedModels setting (true) when applied on modelBuilder', function() {
+    var builder = new ModelBuilder();
+    builder.settings.strictEmbeddedModels = true;
+    var M1 = builder.define('testEmbedded', {
+      name: 'string',
+      address: {
+        street: 'string',
+      },
+    });
+    var m1 = new M1({
+      name: 'Jim',
+      address: {
+        street: 'washington st',
+        number: 5512,
+      },
+    });
+    assert.equal(m1.address.number, undefined, 'm1 should not contain number property in address');
+  });
+  it('should use strictEmbeddedModels setting (validate) when applied on modelBuilder', function() {
+    var builder = new ModelBuilder();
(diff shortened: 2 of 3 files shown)


slnode:

bajtos:
@slnode ok to test

bajtos:
@mitsos1os thank you for the pull request!

Changing the `strict` setting in general is too big breaking change. Could you please limit the change to models created by embedded property types, as described in #923? (https://github.com/strongloop/loopback-datasource-juggler/blob/795952c1823936c37b8dc037e043dd9168c87c5e/lib/model-builder.js#L711-L712).

Also note that when you set `strict:true`, LoopBack will silently discard any extra properties. I think this may confuse users, perhaps we should configure `strict:'throw'` instead? Ideally, I think `strict:'validate'` would be best, but that would require more work to ensure nested properties are validated as part of validation of the root model instance.

Quoting from [docs](https://docs.strongloop.com/display/public/LB/Model+definition+JSON+file#ModeldefinitionJSONfile-Top-levelproperties):

> - `true`: Only properties defined in the model are accepted. Use if you want to ensure the model accepts only predefined properties.
> - `false`: The model is an open model and accepts all properties, including ones not predefined in the model. This mode is useful to store free-form JSON data to a schema-less database such as MongoDB.
> - "validate": The unknown properties will be reported as validation errors.
> - "throw": Throws an exception if properties not defined for the model are used in an operation.
> - undefined: Defaults to false unless the data source is backed by a relational database such as Oracle or MySQL.

mitsos1os:
@bajtos  ΟΚ, this is a really small change... so, what would you like it to be? true or validate?
If you want my opinion I am more biased towards strict true, influenced by the spec of Mongoose embedded schemas that I have earlier used a lot. That helped me get rid of any "garbage data" that I didn't define in my model.

bajtos:
>  ΟΚ, this is a really small change... so, what would you like it to be? true or validate?
If you want my opinion I am more biased towards strict true, influenced by the spec of Mongoose embedded schemas that I have earlier used a lot. That helped me get rid of any "garbage data" that I didn't define in my model.

I don't have a strong opinion. Let' use `strict:true` then.

I have another idea for you to consider. Even when we make this change in embedded property types only, it's still a breaking change that won't get into 2.x series and will have to wait for 3.0 release in few months. However, if we add a feature flag to control this behaviour, then we can land this in 2.x too.

I think the easiest option is to put the feature flag into modelBuilder's settings, let's call it "strictEmbeddedModels" for example. This flag could be set from `server/server.js`:

```js
var app = loopback();
app.registry.modelBuilder.settings.strictEmbeddedModels = true;
boot(app, __dirname);
```

(Setting this from code is not consistent with the way how we configure everything else, but let's leave that out of scope of this pull request and discuss it here: https://github.com/strongloop/loopback/issues/2370.)

Then in [model-builder](https://github.com/strongloop/loopback-datasource-juggler/blob/795952c1823936c37b8dc037e043dd9168c87c5e/lib/model-builder.js#L711-L712), you can read the flag to decide what strict mode to use:

```js
 return this.define(this.getSchemaName(null),
   type, { anonymous: true, idInjection: false, strict: this.settings.strictEmbeddedModels });
```

The benefit of this approach is that this change is completely backwards compatible and can be back-ported to 2.x (`2.x` branch) without affecting existing applications.

In 3.0 (the current `master` branch), we can change the default value of this flag to `true`. This should be done in a new pull request.

@mitsos1os what do you think?

mitsos1os:
@bajtos I agree. I also find the configuration of modelBuilder also a good option. I am kinda lost though on what should be done. Where you talking about separate pull requests on different branches? The modelBuilder setting in 2.x branch and directly strict true to 3.x???

bajtos:
> @mitsos1os I am kinda lost though on what should be done. 

Ah, sorry for not being clear enough. I am proposing the following approach: in this pull request, implement the feature flag in a way that's fully backwards compatible.

Basically make the following change:

> Then in [model-builder](https://github.com/strongloop/loopback-datasource-juggler/blob/795952c1823936c37b8dc037e043dd9168c87c5e/lib/model-builder.js#L711-L712), you can read the flag to decide what strict mode to use:

> ```js
 return this.define(this.getSchemaName(null),
   type, { anonymous: true, idInjection: false, strict: this.settings.strictEmbeddedModels });
```
 
plus add unit-tests etc.

Once this is landed on `master`, I will back-port the change to `2.x` branch myself. 

By adding the feature flag to both `3.x` (`master`) and `2.x` versions, we will make it easy for users to upgrade. They can start using the new feature flag in LB 2.x, and when they upgrade to LB 3.x, their app will keep working in the same way. Another benefit of this feature flag is that it allows users to pick different behavior too, e.g. `strict:'throw'` instead of `strict:true` if it makes more sense for them.

Once this is done, we can discuss whether it's still worth to change the default behavior in LB 3.0. On the second thought, I think it may be better to keep the current default behaviour, as it's closer to what we do for regular top-level models (they are not strict by default).

mitsos1os:
@bajtos I created the change we were talking about. You can take a look! Thanks

bajtos:
I cleaned up the patch a bit and landed it via #955. Thank you for the contribution!

I will back-port it to 2.x shortly.

mitsos1os:
@bajtos You're welcome!

Issue #945, opened 2016-05-25
EmbedsOne save does not call parent save to update parent model with changes


mitsos1os:
see https://github.com/strongloop/loopback/issues/2376 ...

0candy:
The .trigger function is deprecated and will be removed in LB 3.0. 
Please use operation hooks for replacement.

https://docs.strongloop.com/display/APIC/Operation+hooks

mitsos1os:
OK so how will nestRemoting work? Since I do not see any other reference of `prepareEmbeddedInstance` which is how the parent model gets notified for the child changes right??? 
It overrides the .trigger to first call the action on the embedded's model parent

0candy:
nestRemoting is done here:
https://github.com/strongloop/loopback/blob/6964914bab64496c789c434fec39d2914231ee23/lib/model.js#L685

The `prepareEmbeddedInstance` was added so that a separate `save` call was not needed when updating a relation.
https://github.com/strongloop/loopback-datasource-juggler/commit/96a276a12b4a75a85b1092ccbd4ad0b31c5fd056

I think you're right, I don't think the EmbedOne has the parent save. You will have to make a separate call to parent save.

Could you please fork the loopback-sandbox with a sample app so I can better understand what you're trying to do? Thanks!
https://github.com/strongloop/loopback-sandbox

mitsos1os:
@0candy I am sorry but I made a mistake. nestRemoting has nothing to do with it, I was trying it like that and I was too tired, and thought the issue was related.... nestRemoting works correctly at exposing inner model relationships... The problem is as you mentioned that embedsOne does not call the parent save when you call save on the embedded model. In contrast, embedsMany does this correctly through overriding the .trigger I mentioned in prepareEmbeddedInstance... If this is going to be deprecated, then somewhere else for both embedding relations, the save of parent should be also called... 

I have created a repository showing the difference between these two relationships work here:
https://github.com/mitsos1os/loopback-sandbox/tree/embedsone-parent-save

I have created an override that also implements the prepareEmbeddedInstance in embedsOne, but if this is to be deprecated, it is also nogo....

0candy:
Thank you for creating the repository. We will definitely look for a replacement if we remove the deprecated code. I will label this as a feature to add it to embedsOne.

mitsos1os:
[...]

PR #949, opened 2016-05-31
2.x - Save on embedded model (from embedsOne relation) persist to parent

mitsos1os:
Implemented override of trigger for embedsOne relationship embedded document, to allow direct save of changes on embedded model to be persisted on parent document.

ex:
```
Person.embedsOne(Address);
Person.findById(someId)
  .then(function(p){
    var address = p.addressItem();
    address.street = 'new street'
    return address.save(); // This will now persist changes on parent p documentt
  })
```

**Note:** Commit Linter tests fail for some reason, but when I click on details nothing shows up.... on the repository running npm test or npm run lint, completes succesfully...

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -2066,7 +2066,7 @@ EmbedsOne.prototype.related = function(condOrRefresh, options, cb) {
     return;
   }
 
-  var embeddedInstance = modelInstance[propertyName];
+  var embeddedInstance = this.embeddedValue();
 
   if (embeddedInstance) {
     embeddedInstance.__persisted = true;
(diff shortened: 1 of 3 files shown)


slnode:

bajtos:
Hello @mitsos1os, thank you for the pull request!

As @0candy mentioned in https://github.com/strongloop/loopback-datasource-juggler/issues/945, we are dropping model hooks in 3.0. I would prefer to implement your feature using a mechanism that will work in 3.0 too. I think we may be able to do it using new Operation Hooks.

I am proposing to wait until @0candy figures out the way how to implement "trigger parent" for embedsMany without model hooks, and then rework this pull request accordingly.

Thoughts?

mitsos1os:
@bajtos Thank you for your reply. This could be implemented by referencing a hidden parent property in embedded models and overriding **save** or what other function is used to persist changes to also call the updateAttributes on parent with embedded property name... Very abstractly!!!! :-)

As for the pull request I was thinking about implementing it right now at 2.x branch (that's why I made the pull request on this branch) in order to be consistent with the expected behavior which is not currently happening, and when trigger hooks are deprecated also change the code on 2.x branch.

test/relations.test.js
```diff
@@ -3775,6 +3775,30 @@ describe('relations', function() {
...
+        .then(function() {
+          return Person.findById(personId);
+        })
+        .then(function(person) {
+          person.passportItem().toObject().should.eql({ name: 'Jim' });
+          // restore original hasPk
+          Passport.definition.hasPK = originalHasPK;
+          done();
+        })
+        .catch(done);
```
bajtos: You need to restore the PK when the test fails too.

lib/relation-definition.js
```diff
@@ -2081,9 +2081,50 @@ EmbedsOne.prototype.related = function(condOrRefresh, options, cb) {
...
+      var pk = this.definition.keyTo;
+      inst.__persisted = !!inst[pk];
+    } else {
+      inst.__persisted = true;
+    }
+    inst.triggerParent = function(actionName, callback) {
+      if (actionName === 'save' || actionName === 'destroy') {
+        var embeddedValue = self.embeddedValue();
+        modelInstance.updateAttribute(propertyName,
+          embeddedValue, function(err, modelInst) {
```
bajtos: When "destroy" was called, I think we need to unset the embedded value (or at least set it to `null`).

Please add a unit-test to verify this scenario.

bajtos:
> As for the pull request I was thinking about implementing it right now at 2.x branch (that's why I made the pull request on this branch) in order to be consistent with the expected behavior which is not currently happening, and when trigger hooks are deprecated also change the code on 2.x branch.

Makes sense. Could you please address my two comments above and also fix the commit message to follow the [50/72 rule](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)?

This is the message reported by Commit Linter:

```
**  Linting commit logs
**
**  1 problems found:
**    6426204 - added ability to directly save on embedded model of: First line should be 50 characters or less (saw 94)
**
```

mitsos1os:
@bajtos I fixed the things and changed commit message as you asked. Thanks

bajtos:
Landed via bfbdd15, thank you for the contribution!

BTW we decided to defer removal of model hooks, they will stay around in 3.x. I have forward-ported your patch to `master` branch in ea2266e.

PR #952, opened 2016-05-31
Avoid changing PK of instance in operation hook when forceId is set

Amir-61:
* Avoid changing PK for `instance` in `before save` operation hook for replaceById when forceId is set

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/833

*Note: It needs to be back-ported to 2.x.*

To: @bajtos

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2671,6 +2671,13 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
   Model.notifyObserversOf('before save', context, function(err, ctx) {
     if (err) return cb(err);
 
+    if (ctx.instance[pkName] !== id && !Model._warned.cannotOverwritePKInBeforeSaveHook) {
+      Model._warned.cannotOverwritePKInBeforeSaveHook = true;
+      console.warn('WARNING: id property cannot be changed from ' +
+        id + ' to ' + inst[pkName] + ' for model:' + Model.modelName +
+        ' in \'before save\' operation hook');
+    }
+
     data = inst.toObject(false);
 
     if (strict) {
@@ -2722,7 +2729,15 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
           Model.notifyObserversOf('loaded', ctx, function(err) {
             if (err) return cb(err);
 
+            if (ctx.data[pkName] !== id && !Model._warned.cannotOverwritePKInLoadedHook) {
+              Model._warned.cannotOverwritePKInLoadedHook = true;
+              console.warn('WARNING: id property cannot be changed from ' +
+                id + ' to ' + ctx.data[pkName] + ' for model:' + Model.modelName +
+                ' in \'loaded\' operation hook');
+            }
+
             inst.__persisted = true;
+            ctx.data[pkName] = id;
             inst.setAttributes(ctx.data);
 
             var context = {
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -984,6 +984,73 @@ describe('manipulation', function() {
         });
       });
 
+      it('should ignore PK if it is set for `instance`' +
+      'in `before save` operation hook', function(done) {
+        Post.findById(postInstance.id, function(err, p) {
+          if (err) return done(err);
+          changePostIdInHook('before save');
+          p.replaceAttributes({ title: 'b' }, function(err, data) {
+            data.id.should.eql(postInstance.id);
+            if (err) return done(err);
+            Post.find(function(err, p) {
+              if (err) return done(err);
+              p[0].id.should.eql(postInstance.id);
+              done();
+            });
+          });
+        });
+      });
+
+      it('should set cannotOverwritePKInBeforeSaveHook flag, if `instance` in' +
+      '`before save` operation hook is set, so we report a warning just once',
+      function(done) {
+        Post.findById(postInstance.id, function(err, p) {
+          if (err) return done(err);
+          changePostIdInHook('before save');
+          p.replaceAttributes({ title: 'b' }, function(err, data) {
+            if (err) return done(err);
+            Post._warned.cannotOverwritePKInBeforeSaveHook.should.equal(true);
+            data.id.should.equal(postInstance.id);
+            done();
+          });
+        });
+      });
+
+      it('should ignore PK if it is set for `data`' +
+      'in `loaded` operation hook', function(done) {
+        Post.findById(postInstance.id, function(err, p) {
+          if (err) return done(err);
+          changePostIdInHook('loaded');
+          p.replaceAttributes({ title: 'b' }, function(err, data) {
+            data.id.should.eql(postInstance.id);
+            if (err) return done(err);
+            // clear observers to make sure `loaded`
+            // hook does not affect `find()` method
+            Post.clearObservers('loaded');
+            Post.find(function(err, p) {
+              if (err) return done(err);
+              p[0].id.should.eql(postInstance.id);
+              done();
+            });
+          });
+        });
+      });
+
+      it('should set cannotOverwritePKInLoadedHook flag, if `instance` in' +
+      '`before save` operation hook is set, so we report a warning just once',
+      function(done) {
+        Post.findById(postInstance.id, function(err, p) {
+          if (err) return done(err);
+          changePostIdInHook('loaded');
+          p.replaceAttributes({ title: 'b' }, function(err, data) {
+            if (err) return done(err);
+            Post._warned.cannotOverwritePKInLoadedHook.should.equal(true);
+            data.id.should.equal(postInstance.id);
+            done();
+          });
+        });
+      });
+
       it('works without options(promise variant)', function(done) {
         Post.findById(postInstance.id)
       .then(function(p) {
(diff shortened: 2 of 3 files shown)


lib/dao.js
```diff
@@ -2687,6 +2687,14 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
 
   Model.notifyObserversOf('before save', context, function(err, ctx) {
     if (err) return cb(err);
+    if (Model.settings.forceId && ctx.instance[pkName] !== id) {
+      var err = new Error ('id property cannot be changed from ' +
+        id + ' to ' + inst[pkName] + ' in \'before save\' operation ' +
+        'hook when forceId is set to true');
```
bajtos: Is it a good idea to allow hooks to change the id even when `forceId` is not set? Can the connectors handle this?

Regardless of the decision, could you please add a unit-test to demonstrate & verify what happens when a hook modifies the id property?

lib/dao.js
```diff
@@ -2687,6 +2687,14 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
...
+      err.statusCode = 403;
+      process.nextTick(function() { cb(err); });
+      return cb.promise;
```
bajtos: This is nested inside an inner function, the return value is ignored and most likely you are already in a next tick.

I think you can simply

```js
return cb(err);
```

bajtos: BTW if this code was not in a next tick, then L2689 above would need to defer callback to next tick too.

test/optional-validation.test.js
```diff
@@ -133,6 +133,24 @@ describe('optional-validation', function() {
...
+          ModelWithForceId.observe('before save', function(ctx, next) {
+            ctx.instance.id = 50;
+            next();
+          });
+        }
+        ModelWithForceId.create({ name: 'foo' }, function(err, created) {
+          if (err) return done(err);
+          changeIdInBeforeSaveHook();
+          created.replaceAttributes({ name: 'bar' }, function(err, data) {
+            should.exist(err);
```
bajtos: Please verify that the you received the expected error, and not some other error instead.

For example:

```js
err.message.should.match(/id property cannot be changed.*forceId/);
```

bajtos:
@Amir-61 reviewed, see my comments above

lib/dao.js
Amir-61: @bajtos Very good question! I don't know why I didn't see this from a better perspective; in DBs you can never change PK irrespective of whether you set `forceId` or not. also it merits consideration this has been handled under the hood in my implemented methods in connectors; please see how we handle this situation in [loopback-connector for all relational DB](https://github.com/strongloop/loopback-connector/blob/958c6619e3353c4604adc15161fa3a39c8767dfd/lib/sql.js#L617), [memory-connector](https://github.com/strongloop/loopback-datasource-juggler/blob/8f347fbacb3020a86c186ef6e3ec8e0f0a9705db/lib/connectors/memory.js#L745).

I'll add tests to confirm this. So now I believe we have two options:

* Ignore `id` which is passed in `before save` hook, but dont throw an error, just give a warning with an appropriate message saying `the ID, X, which has been passed in before save hook has been ignored. You cannot change ID in replaceById method`

Or
* Throw an error and do not call `connector.replaceById()` anymore

Which one do you think is more appropriate in this case?

bajtos: I don't have a strong opinion either. Here are factors to consider:

 - Backwards compatibility. How likely is it to have an existing app out there that was changing the `id` property in "before save" hook, and which would thus stop working if we start reporting an error?
 - Verbosity. If we print the warning for each hook invocation, the process output will get polluted very quickly.

Having wrote that, I think the ideal solution is to print a warning (that way we preserve backwards compatibility), but make sure to print it only once for each model (so that we don't pollute stdout/stderr).

Something along these lines (please rewrite it in a cleaner style):

```js
if (ctx.instance[pkName] !== id && !(Model._warned || {}).cannotOverwritePrimaryKeyInBeforeSaveHook) {
  Model._warned = Model._warned || {};
  Model._warned.cannotOverwritePrimaryKeyInBeforeSaveHook = true;
  console.warn(...);
}
```

Thoughts?

Amir-61:
@slnode test please

Amir-61:
Hey @rmg,

I keep getting failure for `Amazon-4` server for `loopback-datasource-juggler` tests, although all other servers pass tests; also, when I look into the [logs](http://ci.strongloop.com/job/loopback-datasource-juggler/2084/label=amazon-4/console), I cannot see any info related to my code causing failure. Please see [this](http://ci.strongloop.com/job/loopback-datasource-juggler/2084/label=amazon-4/console). It just says `No failure causes found`.

Any though?

lib/dao.js
```diff
@@ -2742,6 +2748,7 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
 
             inst.__persisted = true;
             inst.setAttributes(ctx.data);
+            inst[pkName] = id;
```
Amir-61: ##### Just to explain and make the review process easier:

You may wonder why I'm not using data which is returned from the connector to make sure data has the `id` which has not been changed, please note that relational databases like Mysql returns something like the following: 

```
OkPacket {
  fieldCount: 0,
  affectedRows: 1,
  insertId: 0,
  serverStatus: 2,
  warningCount: 0,
  message: '(Rows matched: 1  Changed: 1  Warnings: 0',
  protocol41: true,
  changedRows: 1 }
```

So no data is returned for `UPDATE` query.

So thats why I need to set `inset[pkName]` to `id` which is one of the arguments of `replaceById`

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test plese

Amir-61:
@slnode test please

Amir-61:
@slnode test please

lib/dao.js
```diff
@@ -2689,6 +2689,12 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
   Model.notifyObserversOf('before save', context, function(err, ctx) {
     if (err) return cb(err);
 
+    if (ctx.instance[pkName] !== id && !(Model._warned || {}).cannotOverwritePKInBeforeSaveHook) {
+      Model._warned = { cannotOverwritePKInBeforeSaveHook: true };
```
bajtos: When we add more warning in the future, this line will clear all other warnings that may have been already reported/set.

I am proposing to initialize `Model._warned = {}`in the code building Model ctor, so that we can rely on it being initialized everywhere else. See https://github.com/strongloop/loopback-datasource-juggler/blob/b039b51610b5ad14b5abdc5bb9e798baa7c70588/lib/model-builder.js#L222-L233. I would use `hiddenProperty` to define `_warned`, so that we are consistent with other ctor-level properties.

lib/dao.js
```diff
@@ -2689,6 +2689,12 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
...
+      console.warn('WARNING: id property cannot be changed from ' +
+        id + ' to ' + inst[pkName] + ' in \'before save\' operation hook');
```
bajtos: Please include model name (`Model.modelName`) in the message. Otherwise it's difficult to figure out where is the source code triggering this warning without running the debugger.

lib/dao.js
bajtos: Can we please reset the PK column in the data, before it's set in the model instance?

```js
ctx.data[pkName] = id;
inst.setAttributes(ctx.data);
```

Should we detect the case when a "loaded" hook is trying to change PK field and report a warning too?

test/manipulation.test.js
```diff
@@ -984,6 +984,50 @@ describe('manipulation', function() {
         });
       });
 
+      it('should ignore PK if it is set for `instance`' +
+      'in `before save` operation hook', function(done) {
+        function changeIdInBeforeSaveHook() {
```
bajtos: This method is repeated twice (once in each test), is it worth sharing it? If not, then I am proposing to inline it where it's called.

bajtos:
Looks mostly good.

Amir-61:
@bajtos PTAL :-)

Now I also add a warning if the user tries to change `id` for `loaded` hook.

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

test/manipulation.test.js
```diff
@@ -967,6 +967,13 @@ describe('manipulation', function() {
         });
       });
 
+      function changePostIdInBeforeSaveHook(operationHook, ctxField) {
```
bajtos: Please put helper functions at the bottom of the test file, see https://github.com/strongloop-internal/loopback-knowledge-base/blob/master/style-guide.md#layout-of-test-files

test/manipulation.test.js
```diff
@@ -967,6 +967,13 @@ describe('manipulation', function() {
...
+        Post.observe(operationHook, function(ctx, next) {
+          ctx[ctxField].id = 99;
```
bajtos: Can we infer `cxField` ourselves? The hooks always provide only one of `ctx.data` or `ctx.instance` IIRC.

```js
(ctx.data || ctx.instance).id = 99;
next();
```

bajtos:
@Amir-61 two more comments, the patch LGTM otherwise. No further review is needed.

PR #954, opened 2016-06-01
Retun err for UPSERT if the connector returns err

Amir-61:
Return error if the connector returns an error for updateOrCreate

Connect to https://github.com/strongloop/loopback-datasource-juggler/issues/914

Note: It needs to be back-ported to 2.x.

\To: @bajtos


bajtos:
:+1: :shipit:

PR #957, opened 2016-06-01
Retun err for UPSERT if the connector returns err

Amir-61:
Return error if the connector returns an error for updateOrCreate

Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/954


Amir-61:
@slnode test please

Amir-61:
@slnode test please

PR #950, opened 2016-05-31
Add test's description

Amir-61:
@bajtos PTAL

PR #958, opened 2016-06-01
Add test's description

Amir-61:
Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/950


Amir-61:
@slnode test please

PR #953, opened 2016-05-31
Fix error message

Amir-61:
* Fix error message when PK is changed in `replaceById` and test coverage

*Note: It needs to be back-ported to 2.x.*

To: @bajtos

PR #959, opened 2016-06-01
Fix error message

Amir-61:
* Fix error message when PK is changed in `replaceById`

Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/953


Amir-61:
@slnode test please

PR #962, opened 2016-06-03
fix

bajtos:
@Amir-61 @raymondfeng PTAL

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -1559,6 +1559,8 @@ RelationDefinition.hasAndBelongsToMany = function hasAndBelongsToMany(modelFrom,
   var options = { as: params.as, through: params.through };
   options.properties = params.properties;
   options.scope = params.scope;
+  // Forward relation options like "disableInclude"
+  options.options = params.options;
 
   if (params.polymorphic) {
     var polymorphic = polymorphicParams(params.polymorphic);
--- a/test/include.test.js
+++ b/test/include.test.js
@@ -617,6 +617,29 @@ describe('include', function() {
         });
       });
   });
+
+  it('should support disableInclude for hasAndBelongsToMany', function() {
+    var Patient = db.define('Patient', { name: String });
+    var Doctor = db.define('Doctor', { name: String });
+    var DoctorPatient = db.define('DoctorPatient');
+    Doctor.hasAndBelongsToMany('patients', {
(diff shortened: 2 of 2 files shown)


Amir-61:
LGTM. Please just lets wait to see the tests pass; at least connectors(`Mysql`, `MongodB`,...) and `loopback-datasource-juggler`. 👍

Amir-61:
@slnode test please

Amir-61:
Hey @rmg,

It seems we have some problems passing juggler CI tests during the day time, while it works much better when Jenkins is not that much busy. For example, I see [loopback-connector-mongodb](http://ci.strongloop.com/job/loopback-connector-mongodb/) passes on half of servers while it fails on [a few servers](http://ci.strongloop.com/job/loopback-connector-mongodb/label=ubuntu-0.10/1290/console) with error message : `[sl-ci-run INFO  198.57 ]: npm install reports FAILURE`, which I don't think if they have anything with the code. The similar story applies for other connectors: [loopback-connector-postgres](http://ci.strongloop.com/job/loopback-connector-postgresql/),... Should I ignore these failures, or is there any quick fix for these failures to make sure it passes on all servers (a tleast connectors)?

CC: @raymondfeng 

Thanks a lot!

Amir-61:
@slnode test please

raymondfeng:
LGTM

jannyHou:
I am not sure what's included in [`params.options`](https://github.com/strongloop/loopback-datasource-juggler/pull/962/files#diff-20531c2f86840f4a537a4db9f742b0e2R1563) but i am wondering is it proper to pass all options to the [hasMany](https://github.com/strongloop/loopback-datasource-juggler/blob/246f07e3bfe55ef4e6dcf43c4f4e9b413aeebf5e/lib/relation-definition.js#L1579) then?

If `disableInclude` is the only option we need to care about, is it safer to just `options.options.disableInclude = params.options.disableInclude` ?

Amir-61:
@rmg, @raymondfeng Ok the test defined in this PR was failing on `loopback-connector-sqlite3`; now the test is fixed and it passes for all connectors; except we get random `[sl-ci-run INFO  14.14 ]: npm install reports FAILURE]` on some random servers. which I believe it is not related to the code and we cannot do anything about them.

I'm landing the PR now...

PR #964, opened 2016-06-03
Fix

Amir-61:
Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/962


Amir-61:
Merging as the other patch in [2.x](https://github.com/strongloop/loopback-datasource-juggler/pull/962) already got LGTM from Raymond.

PR #965, opened 2016-06-06
[SEMVER-MAJOR] Remove model events

0candy:
Connect to https://github.com/strongloop-internal/scrum-loopback/issues/767


3.0-RELEASE-NOTES.md
```diff
@@ -61,4 +61,16 @@ silently ignored. This has been fixed in 3.0, therefore an undefined mixin
...
+for more details.
+
+##Remove deprecated model hooks and model events
+
+The following deprecated model events are now removed.
+
+Model events
+*changed
+*deletedAll
+*deleted
```
bajtos: The text are formatted in italics, I think you need to put a space after the star character.

bajtos: Also please add instructions explaining how to upgrade code depending on these removed events (i.e. use Operation Hooks).

test/events.js
```diff
@@ -25,52 +25,6 @@ describe('events', function() {
       });
     };
   });
-
-  describe('changed', function() {
```
bajtos: AFAICT, there is no test remaining in this file after this change. Can we remove the entire file?

bajtos:
@0candy I think you missed one more instance - [lib/dao.js#L585-L587](https://github.com/strongloop/loopback-datasource-juggler/blob/remove_events/lib/dao.js#L585-L587):

```
$ grep Model.emit -rn lib
lib/dao.js:586:                Model.emit('changed', inst);
lib/dao.js:2588:  Model.emit('set', inst);
```

I am not sure what's the purpose of the `set` event and I don't see any alternative how to receive that notification, so let's keep `set` around.

bajtos:
LGTM. Please squash the commits before landing.

PR #968, opened 2016-06-08
Give warning if PK is changed in hooks

Amir-61:
* Give warning if PK is changed in `before save` and `loaded`
 operation hooks for replaceById

Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/952


Amir-61:
@slnode test please

Amir-61:
@slnode test please

Issue #867, opened 2016-03-08
options.allowExtendedOperators doesn't work with updateAttributes


Valko54:
Hi,

After many tests, I can't use options.allowExtentedOperations with model.updateAttributes function.

Could you add the option.allowExtentedOperations in dao.js file (line: 2810) ?
`var allowExtendedOperators = connector.settings && connector.settings.allowExtendedOperators;`

I use "strict" option and my data variable is cleaned by this line: 
`if (strict && !allowExtendedOperators) {
      applyStrictCheck(self.constructor, strict, data, inst, validateAndSave);
    }`

The behavior is consistent but, is it correct ?

Thank you.

Amir-61:
Hey @Valko54,
>Could you add the option.allowExtentedOperations in dao.js file (line: 2810)>
Please keep in mind context of the file keeps changing and line numbers are not valid; the better way is to add the link to the line of code you are referring to; however for this question I suggest you send a PR and discuss the changes in your patch... if they make sense we land the PR, or we can discuss further there...

Thanks... Please let me know if you agree and we can close this ticket...

Amir-61:
Closing this ticket now.  As suggested please send a PR and discuss the changes in your patch...

Thanks!

Issue #970, opened 2016-06-09
Wrong setting of allowExtendedOperators in MongoDB connector


mitsos1os:
As mentioned in the doc, in order to use updated operators such as $set in mongoDb models, you have to set the option as mentioned in https://docs.strongloop.com/display/APIC/MongoDB+connector#MongoDBconnector-UsingMongoDBoperatorsinupdateoperations:

> To set the option in the model definition file, set the property settings.mongodb.allowExtendedOperators to true.  For example:
> 
> ```
> {
>   "name": "MyModel",
>   "settings": {
>     "mongodb": {
>       "allowExtendedOperators": true
>     }
>   ...
> ```
> 
> However this is not working. After digging with source code of mongodb connector I found that this is not correct, because in [line 457](https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L457) mongodb  tries to find the option in **modelClass.settings.mongodb** By having the structure mentioned earlier, the option goes into **modelClass.settings.settings.mongodb** so it is not found.

I do not know if this has to change in mongodb connector, or just the doc to show the JSON configuration document properly configured as:

```
  "name": "MyModel",
   "mongodb": {
     "allowExtendedOperators": true
   }
  ...
```

Or it could be set in datasources config for mongo ex.

```
{
  "mongoDs": {
    "host": "localhost",
    "port": 27017,    
    "database": "happyDb",
    "password": "",
    "name": "mongoDs",
    "user": "",
    "connector": "mongodb",
    "allowExtendedOperators": true
  }, 
  "memoryDs": {
    "name": "memoryDs",
    "connector": "memory"
  }
}

```

PR #979, opened 2016-06-16
Refactor logic of options.allowExtendedOperators

mrbatista:
related issue #867

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -181,6 +181,31 @@ DataAccessObject.getConnector = function() {
   return this.getDataSource().connector;
 };
 
+/**
+ * Verify if allowExtendedOperators is enabled
+ * @options {Object} [options] Optional options to use.
+ * @property {Boolean} allowExtendedOperators.
+ * @returns {Boolean} Returns `true` if allowExtendedOperators is enabled, else `false`.
+ */
+DataAccessObject._allowExtendedOperators = function(options) {
+  options = options || {};
+
+  var Model = this;
+  var dsSettings = this.getDataSource().settings;
+  var allowExtendedOperators = dsSettings.allowExtendedOperators;
+  // options settings enable allowExtendedOperators per request (for example if
+  // enable allowExtendedOperators only server side);
+  // model settings enable allowExtendedOperators only for specific model.
+  // dataSource settings enable allowExtendedOperators globally (all models);
+  // options -> model -> dataSource (connector)
+  if (options.hasOwnProperty('allowExtendedOperators')) {
+    allowExtendedOperators = options.allowExtendedOperators === true;
+  } else if (Model.settings && Model.settings.hasOwnProperty('allowExtendedOperators')) {
+    allowExtendedOperators = Model.settings.allowExtendedOperators === true;
+  }
+  return allowExtendedOperators;
+};
+
 // Empty callback function
 function noCallback(err, result) {
   // NOOP
@@ -720,9 +745,9 @@ DataAccessObject.upsertWithWhere = function(where, data, options, cb) {
         function callConnector() {
           try {
             ctx.where = removeUndefined(ctx.where);
-            ctx.where = Model._coerce(ctx.where);
+            ctx.where = Model._coerce(ctx.where, options);
             update = removeUndefined(update);
-            update = Model._coerce(update);
+            update = Model._coerce(update, options);
           } catch (err) {
             return process.nextTick(function() {
               cb(err);
@@ -1110,7 +1135,7 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
     query.limit = 1;
 
     try {
-      this._normalize(query);
+      this._normalize(query, options);
     } catch (err) {
       process.nextTick(function() {
         cb(err);
@@ -1413,10 +1438,12 @@ var operators = {
 /*
  * Normalize the filter object and throw errors if invalid values are detected
  * @param {Object} filter The query filter object
+ * @options {Object} [options] Optional options to use.
+ * @property {Boolean} allowExtendedOperators.
  * @returns {Object} The normalized filter object
  * @private
  */
-DataAccessObject._normalize = function(filter) {
+DataAccessObject._normalize = function(filter, options) {
   if (!filter) {
     return undefined;
   }
@@ -1497,7 +1524,7 @@ DataAccessObject._normalize = function(filter) {
   var handleUndefined =  this._getSetting('normalizeUndefinedInQuery');
   // alter configuration of how removeUndefined handles undefined values
   filter = removeUndefined(filter, handleUndefined);
-  this._coerce(filter.where);
+  this._coerce(filter.where, options);
   return filter;
 };
 
@@ -1555,14 +1582,18 @@ function coerceArray(val) {
 /*
  * Coerce values based the property types
  * @param {Object} where The where clause
+ * @options {Object} [options] Optional options to use.
+ * @property {Boolean} allowExtendedOperators.
  * @returns {Object} The coerced where clause
  * @private
  */
-DataAccessObject._coerce = function(where) {
+DataAccessObject._coerce = function(where, options) {
   var self = this;
   if (!where) {
     return where;
   }
+
+  options = options || {};
 
   var err;
   if (typeof where !== 'object' || Array.isArray(where)) {
@@ -1585,7 +1616,7 @@ DataAccessObject._coerce = function(where) {
       }
 
       for (var k = 0; k < clauses.length; k++) {
-        self._coerce(clauses[k]);
+        self._coerce(clauses[k], options);
       }
 
       continue;
@@ -1703,8 +1734,7 @@ DataAccessObject._coerce = function(where) {
       }
     } else {
       if (val != null) {
-        const dsSettings = this.getDataSource().settings;
-        const allowExtendedOperators = dsSettings.allowExtendedOperators;
+        const allowExtendedOperators = self._allowExtendedOperators(options);
         if (operator === null && val instanceof RegExp) {
           // Normalize {name: /A/} to {name: {regexp: /A/}}
           operator = 'regexp';
@@ -1831,7 +1861,7 @@ DataAccessObject.find = function find(query, options, cb) {
     'all() must be implemented by the connector');
 
   try {
-    this._normalize(query);
+    this._normalize(query, options);
   } catch (err) {
     process.nextTick(function() {
       cb(err);
@@ -2189,7 +2219,7 @@ DataAccessObject.destroyAll = function destroyAll(where, options, cb) {
       try {
         // Support an optional where object
         where = removeUndefined(where);
-        where = Model._coerce(where);
+        where = Model._coerce(where, options);
       } catch (err) {
         return process.nextTick(function() {
           cb(err);
@@ -2345,7 +2375,7 @@ DataAccessObject.count = function(where, options, cb) {
 
   try {
     where = removeUndefined(where);
-    where = this._coerce(where);
+    where = this._coerce(where, options);
   } catch (err) {
     process.nextTick(function() {
       cb(err);
@@ -2616,9 +2646,9 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
   function doUpdate(where, data) {
     try {
       where = removeUndefined(where);
-      where = Model._coerce(where);
+      where = Model._coerce(where, options);
       data = removeUndefined(data);
-      data = Model._coerce(data);
+      data = Model._coerce(data, options);
     } catch (err) {
       return process.nextTick(function() {
         cb(err);
@@ -3083,9 +3113,7 @@ function(data, options, cb) {
   if (isPKMissing(Model, cb))
     return cb.promise;
 
-  var allowExtendedOperators = connector.settings &&
-    connector.settings.allowExtendedOperators;
-
+  var allowExtendedOperators = Model._allowExtendedOperators(options);
   var strict = this.__strict;
   var model = Model.modelName;
   var hookState = {};
--- a/test/allow-extended-operators.test.js
+++ b/test/allow-extended-operators.test.js
@@ -8,32 +8,17 @@
 const DataSource = require('..').DataSource;
 const should = require('should');
 
-describe('Model.settings.allowExtendedOperators', () => {
-  context('DAO.find()', () => {
-    it('converts extended operators to string value by default', () => {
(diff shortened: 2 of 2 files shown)


slnode:

Amir-61:
@mrbatista Thanks for contribution; please add tests which fail without your changes and passes with the changes.

slnode:

slnode:

slnode:

bajtos:
Closing as abandoned.

mrbatista:
@bajtos  which is the correct place to add unit test for verifying this path?  The issue is related to mongoDB connector and not to juggler.

slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

bajtos:
@mrbatista Hi, it's great to see you back :)

>  which is the correct place to add unit test for verifying this path? The issue is related to mongoDB connector and not to juggler.

It turns out we didn't have any tests for allowExtendedOperators yet. I have started a new test suite in #1180, see [test/allow-extended-operators.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/29c60c7b423e05d1ff14832a28da4a2c1d808206/test/allow-extended-operators.test.js).

You will need to rebase your pull request on top of the latest master. In case you are not familiar with git history manipulations, here is the process (assuming `strongloop` is the name of the git remote which points to `strongloop/loopback-datasource-juggler` on GitHub):

```
git fetch strongloop
git rebase strongloop/master
git push -f
```

lib/dao.js
```diff
@@ -2824,6 +2824,14 @@ function(data, options, cb) {
 
   var allowExtendedOperators = connector.settings &&
     connector.settings.allowExtendedOperators;
+  if (options.hasOwnProperty('allowExtendedOperators')) {
+    allowExtendedOperators = options.allowExtendedOperators === true;
```
bajtos: I think it may be better to allow the connector to prevent `options.allowExtendedOperators` to enable this feature. 

```js
if (allowExtendedOperators !== false) {
  if (options.hasOwnProperty('allowExtendedOperators')) {
    allowExtendedOperators = options.allowExtendedOperators === true;
  } 
  // ...
}
```


@fabien thoughts?

lib/dao.js
```diff
@@ -2824,6 +2824,14 @@ function(data, options, cb) {
...
+  } else if (allowExtendedOperators !== false && Model.settings.mongodb &&
+    Model.settings.mongodb.hasOwnProperty('allowExtendedOperators')) {
```
bajtos: Accessing a connector-specific configuration in generic juggler code is violating the current design. Is there any particular reason why you would like to scope `allowExtendedOperators` to MongoDB only, as opposed to having it as a top-level `Model.settings.allowExtendedOperators`?

lib/dao.js
```diff
@@ -2824,6 +2824,14 @@ function(data, options, cb) {
...
+    allowExtendedOperators = Model.settings.mongodb.allowExtendedOperators === true;
+  } else if (allowExtendedOperators === true) {
+    allowExtendedOperators = true;
```
```js
if (allowExtendedOperators === true) {
  allowExtendedOperators = true;
}
```

This is a no-operation, right? Please remove.

mrbatista: Right!

lib/dao.js
bajtos: Could you please add few comments to this code to explain the different behaviour of "options" vs. "model config" in respect to "connector" settings? I am worried that future readers may be puzzled why each branch uses a slightly different approach.

bajtos:
@mrbatista Thank you for the updated code. I still need you to add unit tests, see my earlier https://github.com/strongloop/loopback-datasource-juggler/pull/979#issuecomment-265961806

mrbatista:
@bajtos I need to define some rules for the correct implementation. I presume that the option `allowExtendedOperators` has been implemented only for mongoDB connector and then extended to the other ones. Is it right?
1. Define the hierarchy of which configurations prevail over the others.
2. Define a global configuration of settings the options `allowExtendedOperators` over all connector.

My proposal:
1. See the implementation of mongoDB connector (https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L459-L467) the hierarchy is:
`options -> model settings (with different approch in connector and juggler) -> connector`
2. Move options `allowExtendedOperators` at settings level, like in juggler, and modify implementation on mongoDB connector

mrbatista:
@bajtos Added some example tests to verify this patch. I am waiting for your feedback before complete before proceeding.
see other related issues: 
1. https://github.com/strongloop/loopback-connector-mongodb/issues/157

2. https://github.com/strongloop/loopback-datasource-juggler/issues/970

P.s In your initial test allowExtendedOperators is set on connector settings and not model settings. For this I have modified the tests.

mrbatista:
@bajtos PTAL

mrbatista:
@bajtos I will squash my commits and rebase on master (if necessary) after your review.
Correct me if I am wrong, I think that the logic should be extended to the find method. Am I right?

bajtos:
Hi @mrbatista, sorry for the delay. If my understanding is correct, then you are done and this patch is ready for a detailed review, is that correct?

I took a quick look at the changes proposed in `lib/dao.js`, they look reasonable to me.

> Correct me if I am wrong, I think that the logic should be extended to the find method.

Yes, that would be great! I think it would be best to extract the piece of code resolving `allowExtendedOperators` into a shared method that can be reused from both places.

test/allow-extended-operators.test.js
```diff
@@ -59,4 +46,207 @@ describe('Model.settings.allowExtendedOperators', () => {
...
+    context('DAO.updateAttributes()', () => {
+      it('`options.allowExtendedOperators` override connector settings - disable strict check', () => {
+        const TestModel = createTestModel({allowExtendedOperators: false}, {strict: true});
+        return TestModel.create({value: 'test'}).then(function(instance) {
+          var updateData = {$set: {value: 'changed'}};
+          var ctxData;
+          TestModel.observe('persist', function(ctx, next) {
+            ctxData = ctx.data;
+            next();
+          });
```
bajtos: This code is repeated a lot in the new tests. Can we extract it into a shared method?

For example, `createTestModel` can define `persist` observer that can save `ctx.data` in `TestModel.lastPersistedData` property.

```js
// in the helper method
var TestModel = ds.createModel('TestModel', {value: String}, modelSettings);
TestModel.observe('persist', function(ctx, next) {
   ctx.Model.lastPersistedData = ctx.data;
   next();
});
return TestModel;

// in the test
should(TestModel.lastPersistedData).eql(updateData);
```

mrbatista:
@bajtos PTAL

test/allow-extended-operators.test.js
mrbatista: 👍

test/allow-extended-operators.test.js
```diff
@@ -51,12 +36,193 @@ describe('Model.settings.allowExtendedOperators', () => {
...
+              should(TestModel.lastPersistedData).eql(updateData);
+            });
+        });
+      });
+
+      it('`options.allowExtendedOperators` override connector settings - enable strict check', () => {
+        const TestModel = createTestModel({allowExtendedOperators: true}, {strict: true});
+        return TestModel.create({value: 'test'}).then(function(inst) {
+          return inst.updateAttributes({$set: {value: 'changed'}}, {allowExtendedOperators: false})
+            .catch(function(err) {
```
bajtos: You need to handle the case when `updateAttributes` did not throw any error.

I also think it's safe to assume `err` will be defined when a promise was rejected. To be safe, your assertion can be rewritten in a way that supports undefined `err` too.

```js
return inst.updateAttributes(...).then(
  () => throw new Error('updateAttributes() should have failed.'),
  (err) => should(err).have.property('name', 'ValidationError'));
```

test/allow-extended-operators.test.js
```diff
@@ -51,12 +36,193 @@ describe('Model.settings.allowExtendedOperators', () => {
...
+          });
+        });
+      });
+
+      it('`Model.settings.allowExtendedOperators` override connector settings - ' +
+        'enable strict check', () => {
+        const TestModel = createTestModel({allowExtendedOperators: true},
+          {strict: true, allowExtendedOperators: false});
+        return TestModel.create({value: 'test'}).then(function(inst) {
+          return inst.updateAttributes({$set: {value: 'changed'}}).catch(function(err) {
```
bajtos: Ditto, you need to convert the case when `updateAttributes` passed into a promise rejection.

Alternatively, perhaps don't set the strict mode and let `updateAttributes` to set the value, similarly to what you do with `find` [here](https://github.com/strongloop/loopback-datasource-juggler/pull/979/files#diff-2efdf289a9ad4ebbfcfa83d9a4c3152aR61)?

test/allow-extended-operators.test.js
```diff
@@ -51,12 +36,193 @@ describe('Model.settings.allowExtendedOperators', () => {
...
+              should(TestModel.lastPersistedData).eql(updateData);
+            });
+        });
+      });
+
+      it('`options.allowExtendedOperators` override Model settings - enabled strict check', () => {
+        const TestModel = createTestModel({}, {strict: true, allowExtendedOperators: true});
+        return TestModel.create({value: 'test'}).then(function(inst) {
+          return inst.updateAttributes({$set: {value: 'changed'}}, {allowExtendedOperators: false})
+            .catch(function(err) {
```
bajtos: Ditto.

test/allow-extended-operators.test.js
```diff
@@ -51,12 +36,193 @@ describe('Model.settings.allowExtendedOperators', () => {
...
+          });
+        });
+      });
+
+      it('`connector.settings.allowExtendedOperators` honor Model settings - ' +
+        'enable strict check', () => {
+        const TestModel = createTestModel({allowExtendedOperators: true},
+          {strict: true, allowExtendedOperators: false});
+        return TestModel.create({value: 'test'}).then(function(inst) {
+          return inst.updateAttributes({$set: {value: 'changed'}}).catch(function(err) {
```
bajtos: Ditto

test/allow-extended-operators.test.js
```diff
@@ -51,12 +36,193 @@ describe('Model.settings.allowExtendedOperators', () => {
...
+              should(TestModel.lastPersistedData).eql(updateData);
+            });
+        });
+      });
+
+      it('`Model.settings.allowExtendedOperators` honor options settings - enable strict check', () => {
+        const TestModel = createTestModel({}, {strict: true, allowExtendedOperators: true});
+        return TestModel.create({value: 'test'}).then(function(inst) {
+          return inst.updateAttributes({$set: {value: 'changed'}}, {allowExtendedOperators: false})
+            .catch(function(err) {
```
bajtos: And here too

test/allow-extended-operators.test.js
```diff
@@ -51,12 +36,193 @@ describe('Model.settings.allowExtendedOperators', () => {
...
+              should(TestModel.lastPersistedData).eql(updateData);
+            });
+        });
+      });
+
+      it('`connector.settings.allowExtendedOperators` honor options settings - enable strict check', () => {
+        const TestModel = createTestModel({allowExtendedOperators: true}, {strict: true});
+        return TestModel.create({value: 'test'}).then(function(inst) {
+          return inst.updateAttributes({$set: {value: 'changed'}}, {allowExtendedOperators: false})
+            .catch(function(err) {
```
bajtos: ditto

test/allow-extended-operators.test.js
mrbatista: Right!

mrbatista:
@bajtos please review.

test/allow-extended-operators.test.js
```diff
@@ -47,16 +32,212 @@ describe('Model.settings.allowExtendedOperators', () => {
...
+      });
+
+      it('`options.allowExtendedOperators` override connector settings - enable strict check', () => {
+        const TestModel = createTestModel({allowExtendedOperators: true}, {strict: true});
+        return TestModel.create({value: 'test'}).then((inst) => {
+          return inst.updateAttributes(setCustomData(), {allowExtendedOperators: false})
+            .then(() => {
+              throw new Error('updateAttributes() should have failed.');
+            })
+            .catch(function(err) {
```
bajtos: This will catch the error produced two lines above and then fail on assertion checking that the error was `ValidationError`, hiding the fact that `updateAttributes` actually succeeded.

It's important to pass the assertion function as the second argument of `then()` instead. If you are using regular functions like here, then please name them (we typically use `onSuccess` and `onError`).

mrbatista: Fixed!

test/allow-extended-operators.test.js
```diff
@@ -47,16 +32,198 @@ describe('Model.settings.allowExtendedOperators', () => {
     return new DataSource(connectorSettings);
   }
 
+  function setCustomData() {
+    return {$set: {value: 'changed'}};
+  }
+
+  function onSuccess() {
```
bajtos: Sorry if my comment earlier comment was not clear. This is what I meant:

```js
           return inst.updateAttributes(setCustomData(), {allowExtendedOperators: false})
            .then(function onSuccess() {
              throw new Error('updateAttributes() should have failed.');
            })
            .catch(function onError(err) {
              // ...
            });
```

Extracting a shared function is a great idea though! I find the name `onSuccess` too generic, I would like to see a name that reveals more of the intent, for example `updateShouldHaveFailed`.

mrbatista: Updated.

mrbatista:
@bajtos PTAL.
Only two comments:
1. I need add `options`parameter for `_normalize` and `_coerce` methods to propagate `allowExtendedOperators` by `options`.
2. As soon as this this branch is landed/merged, I should have a look at the mongoDB connector to handle the same logic for enabling extended operators: options -> model -> data source. See https://github.com/strongloop/loopback-connector-mongodb/blob/master/lib/mongodb.js#L465-L473

bajtos:
LGTM. Please squash the commits to a single one and provide a descriptive commit message following the [50/72 rule](http://loopback.io/doc/en/contrib/git-commit-messages.html).

mrbatista:
@slnode test

mrbatista:
@bajtos PTAL

bajtos:
Landed 🎉 
Thank you @mrbatista for the contribution and your patience with our slow review process :)

Issue #849, opened 2016-02-07
Id property in insertion through REST should be ignored


sochka:
Hi,

With REST put and post requests ID property should be ignored otherwise this can result in the following security issue:

- attacker creates an instance of some model though rest with maximum possible ID value for the database (e.g. 2^31 for 32bit ints).
- It's impossible to add any data anymore since autoincrement would overflow.
- the system is broken

I believe this is a serious problem.

loay:
There are few solutions for that: 

1) set ``` "forceId": true ``` 
2) as suggested on a previous issue here a solution that works: 

You can add an attribute  "_id"
```
"_id":{
      "type":"string",
      "id":true,
      "generated":true
    }
```
update the modelfile: `yourmodel.js`
```
module.exports = function(yourModelNameHere) {
  yourModelNameHere.validatesAbsenceOf("_id");
};
```
Documents to be updated: https://github.com/strongloop/loopback/issues/2068

loay:
Hi @sochka 
the above mentioned comment is a temp solution. We are getting it fixed soon.

bajtos:
In my understanding, this issue is specific to databases that use numeric auto-incrementing ids. I think it makes sense to set `forceId` to true in such case.

There are situation where we want the client to provide the id value, for example in offline sync scenario the client is generating UUID string as the id. Therefore we cannot automatically deny clients to provide their own `id` value.

jannyHou:
I think we can add some check here to make default `forcdId: true` for id auto-incrementing datasources
https://github.com/strongloop/loopback-datasource-juggler/blob/1e8fe6a0d7b5c84a3f67ad4d6a05ddaa8d1229ba/lib/datasource.js#L516

bajtos:
> I think we can add some check here to make default `forcdId: true` for id auto-incrementing datasources
https://github.com/strongloop/loopback-datasource-juggler/blob/1e8fe6a0d7b5c84a3f67ad4d6a05ddaa8d1229ba/lib/datasource.js#L516


@jannyHou your proposal looks reasonable to me. @Amir-61 @raymondfeng @superkhau you are more familiar with this part of the codebase than I am, could you PTAL too?

superkhau:
I'm fine with @jannyHou's proposal for auto configuring `forceId` based on data source type.

Amir-61:
@bajtos @raymondfeng 

[...]

PR #982, opened 2016-06-17
forceId=true with auto-increment db

jannyHou:
Connect to strongloop/loopback-datasource-juggler#849

Fix this scenario:

> With REST put and post requests ID property should be ignored otherwise this can result in the following security issue:
attacker creates an instance of some model though rest with maximum possible ID value for the database (e.g. 2^31 for 32bit ints).
It's impossible to add any data anymore since autoincrement would overflow.
the system is broken
I believe this is a serious problem.

For NoSQL database, which doesn't has auto-increment Id, the default `forceId` is undefined.
While for id auto-increment database, the default `forceId` is true, only `settings.forceId` can override it.

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -515,7 +515,11 @@ DataSource.prototype.setupDataAccess = function(modelClass, settings) {
       idProp.type = idType;
       modelClass.definition.rawProperties[idName].type = idType;
       modelClass.definition.properties[idName].type = idType;
-      if (settings.forceId) {
+      var forceId = settings.forceId;
+      if (idProp.generated && forceId !== false) {
+        forceId = true;
+      }
+      if (forceId) {
         modelClass.validatesAbsenceOf(idName, { if: 'isNewRecord' });
       }
     }
--- a/test/crud-with-options.test.js
+++ b/test/crud-with-options.test.js
@@ -13,6 +13,7 @@ describe('crud-with-options', function() {
   before(function(done) {
     db = getSchema();
     User = db.define('User', {
+      id: { type: Number, id: true },
       seq: { type: Number, index: true },
       name: { type: String, index: true, sort: true },
       email: { type: String, index: true },
--- a/test/default-scope.test.js
+++ b/test/default-scope.test.js
@@ -94,6 +94,7 @@ describe('default scope', function() {
 
     Person = db.define('Person', { name: String }, {
       scope: { include: 'things' },
+      forceId: false,
     });
 
     // inst is only valid for instance methods
--- a/test/loopback-dl.test.js
+++ b/test/loopback-dl.test.js
@@ -766,7 +766,7 @@ describe('Models attached to a dataSource', function() {
       title: { type: String, length: 255, index: true },
       content: { type: String },
       comments: [String],
-    });
+    }, { forceId: false });
   });
 
   beforeEach(function(done) {
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -61,6 +61,21 @@ describe('manipulation', function() {
       Person.destroyAll(done);
     });
 
+    describe('forceId', function() {
+      before(function(done) {
+        TestForceId = db.define('TestForceId');
+        db.automigrate('TestForceId', done);
+      });
+
+      it('it defaults to forceId:true for generated id property', function(done) {
+        TestForceId.create({ id: 1 }, function(err, t) {
+          should.exist(err);
+          err.message.should.match(/can\'t be set/);
+          done();
+        });
+      });
+    });
+
     it('should create instance', function(done) {
       Person.create({ name: 'Anatoliy' }, function(err, p) {
         p.name.should.equal('Anatoliy');
@@ -308,7 +323,7 @@ describe('manipulation', function() {
     it('should refuse to create object with duplicate id', function(done) {
       // NOTE(bajtos) We cannot reuse Person model here,
       // `settings.forceId` aborts the CREATE request at the validation step.
-      var Product = db.define('ProductTest', { name: String });
+      var Product = db.define('ProductTest', { name: String }, { forceId: false });
       db.automigrate('ProductTest', function(err) {
         if (err) return done(err);
 
@@ -825,7 +840,7 @@ describe('manipulation', function() {
           title: { type: String, length: 255, index: true },
           content: { type: String },
           comments: [String],
-        });
+        }, { forceId: false });
         ds.automigrate('Post', done);
       });
 
@@ -1046,6 +1061,8 @@ describe('manipulation', function() {
         ds.automigrate('Post', done);
       });
       beforeEach(function(done) {
+        // TODO(bajtos) add API to lib/observer - remove observers for all hooks
+        Post._observers = {};
         Post.destroyAll(function() {
           Post.create({ title: 'a', content: 'AAA' }, function(err, p) {
             if (err) return done(err);
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -4509,10 +4509,13 @@ describe('relations', function() {
 
     it('should setup related items', function(done) {
       Job.create({ name: 'Job 1' }, function(err, p) {
+        if (err) return done(err);
         job1 = p;
         Job.create({ name: 'Job 2' }, function(err, p) {
+          if (err) return done(err);
           job2 = p;
           Job.create({ name: 'Job 3' }, function(err, p) {
+            if (err) return done(err);
             job3 = p;
             done();
           });
@@ -4522,11 +4525,13 @@ describe('relations', function() {
 
     it('should associate items on scope', function(done) {
       Category.create({ name: 'Category A' }, function(err, cat) {
+        if (err) return done(err);
         var link = cat.items.build();
         link.job(job1);
         var link = cat.items.build();
         link.job(job2);
         cat.save(function(err, cat) {
+          if (err) return done(err);
           var job = cat.items.at(0);
           job.should.be.instanceof(Link);
           job.should.not.have.property('jobId');
@@ -4542,6 +4547,7 @@ describe('relations', function() {
 
     it('should include related items on scope', function(done) {
       Category.findOne(function(err, cat) {
+        if (err) return done(err);
         cat.links.should.have.length(2);
 
         // denormalized properties:
@@ -4556,6 +4562,7 @@ describe('relations', function() {
         should.not.exist(cat.items.at(1).job());
 
         cat.items(function(err, items) {
+          if (err) return done(err);
           cat.items.at(0).job().should.be.instanceof(Job);
           cat.items.at(1).job().should.be.instanceof(Job);
           cat.items.at(1).job().name.should.equal('Job 2');
@@ -4566,8 +4573,10 @@ describe('relations', function() {
 
     it('should remove embedded items by id', function(done) {
       Category.findOne(function(err, cat) {
+        if (err) return done(err);
         cat.links.should.have.length(2);
         cat.items.destroy(job1.id, function(err) {
+          if (err) return done(err);
           should.not.exist(err);
           cat.links.should.have.length(1);
           done();
@@ -4577,6 +4586,7 @@ describe('relations', function() {
 
     it('should find items on scope', function(done) {
       Category.findOne(function(err, cat) {
+        if (err) return done(err);
         cat.links.should.have.length(1);
         cat.items.at(0).id.should.eql(job2.id);
         cat.items.at(0).name.should.equal(job2.name);
@@ -4585,6 +4595,7 @@ describe('relations', function() {
         should.not.exist(cat.items.at(0).job());
 
         cat.items(function(err, items) {
+          if (err) return done(err);
           cat.items.at(0).job().should.be.instanceof(Job);
           cat.items.at(0).job().name.should.equal('Job 2');
           done();
@@ -4594,8 +4605,10 @@ describe('relations', function() {
 
     it('should add related items to scope', function(done) {
       Category.findOne(function(err, cat) {
+        if (err) return done(err);
         cat.links.should.have.length(1);
         cat.items.add(job3, function(err, link) {
+          if (err) return done(err);
           link.should.be.instanceof(Link);
           link.id.should.eql(job3.id);
           link.name.should.equal('Job 3');
@@ -4608,6 +4621,7 @@ describe('relations', function() {
 
     it('should find items on scope', function(done) {
       Category.findOne(function(err, cat) {
+        if (err) return done(err);
         cat.links.should.have.length(2);
 
         cat.items.at(0).should.be.instanceof(Link);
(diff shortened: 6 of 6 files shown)


lib/datasource.js
```diff
@@ -513,6 +513,9 @@ DataSource.prototype.setupDataAccess = function(modelClass, settings) {
       idProp.type = idType;
       modelClass.definition.rawProperties[idName].type = idType;
       modelClass.definition.properties[idName].type = idType;
+      if (['oracle', 'mysql'].indexOf(this.connector.name) !== -1) {
```
Amir-61: Is there any reason that you are setting `forceId` just for `oracle` and `mysql`?

jannyHou: @Amir-61 Thank you lol, it's not finished, I haven't checked all auto-increment datasources. I created this pr to see how many repos fail due to my change to estimate the scale i need to take care of.

jannyHou:
@bajtos Please see description above and review it. Thanks.

`loopback` is still failing due to same test issue as I refactored in this pr. I will create a pr in loopback to fix them as well.

lib/datasource.js
```diff
@@ -513,7 +513,12 @@ DataSource.prototype.setupDataAccess = function(modelClass, settings) {
       idProp.type = idType;
       modelClass.definition.rawProperties[idName].type = idType;
       modelClass.definition.properties[idName].type = idType;
-      if (settings.forceId) {
+      var forceId = settings.forceId;
+      if (['cloudant', 'mongodb'].indexOf(this.connector.name) === -1 &&
```
bajtos: Please don't hard-code connector names in juggler. It inverts dependency direction and is difficult to scale.

You should check if the `id` column is auto-incremented instead, that's connector agnostic. I think we should set `forceId` whenever the `id` field is generated by the database in whatever way, not only when it's an incrementing sequence.

In my understanding, `idProp.generated` is exactly the flag we are looking for. Note that this flag is set for non-SQL connectors (like MongoDB) too. I think it makes sense, because for example our memory connector uses auto-incremented numeric ids even though it's considered as NoSQL connector.

Alternatively, if we are not comfortable setting `forceId` for all generated id properties, then we can limit it only to generated id properties of type Number. That will cover SQL & memory, but exclude MongoDB.

bajtos:
I am concerned about the amount of changes in unit-tests. If you had to make those test changes to make the tests pass after your change in implementation, then it would suggests this is a major breaking change that can break many existing applications.

Also please add few new narrowly-focused unit-tests to verify that `forceId` flag is set/is not set depending on the attributes of the `id` model property.

jannyHou:
test cases affected in loopback:
```
3 failing

  1) relations - integration embedsMany "before all" hook: createTodoList:
     ValidationError: The `todoList` instance is not valid. Details: `todoItems` contains invalid item: `1` (`id` can't be set) (value: [ { content: 'Todo 1', id...} ]); `todoItems` contains invalid item: `2` (`id` can't be set) (value: [ { content: 'Todo 1', id...} ]).
      at node_modules/loopback-datasource-juggler/lib/dao.js:2195:19
      at ModelConstructor.<anonymous> (node_modules/loopback-datasource-juggler/lib/validations.js:483:11)
      at ModelConstructor.next (node_modules/loopback-datasource-juggler/lib/hooks.js:80:12)
      at ModelConstructor.<anonymous> (node_modules/loopback-datasource-juggler/lib/validations.js:480:23)
      at ModelConstructor.trigger (node_modules/loopback-datasource-juggler/lib/hooks.js:70:12)
      at ModelConstructor.Validatable.isValid (node_modules/loopback-datasource-juggler/lib/validations.js:446:8)
      at node_modules/loopback-datasource-juggler/lib/dao.js:2191:10
      at doNotify (node_modules/loopback-datasource-juggler/lib/observer.js:98:49)
      at doNotify (node_modules/loopback-datasource-juggler/lib/observer.js:98:49)
      at doNotify (node_modules/loopback-datasource-juggler/lib/observer.js:98:49)
      at doNotify (node_modules/loopback-datasource-juggler/lib/observer.js:98:49)
      at Function.ObserverMixin._notifyBaseObservers (node_modules/loopback-datasource-juggler/lib/observer.js:121:5)
      at Function.ObserverMixin.notifyObserversOf (node_modules/loopback-datasource-juggler/lib/observer.js:96:8)
      at Function.ObserverMixin._notifyBaseObservers (node_modules/loopback-datasource-juggler/lib/observer.js:119:15)
      at Function.ObserverMixin.notifyObserversOf (node_modules/loopback-datasource-juggler/lib/observer.js:96:8)
      at Function.ObserverMixin._notifyBaseObservers (node_modules/loopback-datasource-juggler/lib/observer.js:119:15)

  2) users - integration base-user should create post for a given user:
     Error: expected 200 "OK", got 422 "Unprocessable Entity"
      at Test.assert (node_modules/supertest/lib/test.js:205:15)
      at Server.assert (node_modules/supertest/lib/test.js:132:12)
      at emitCloseNT (net.js:1525:8)
      at node_modules/continuation-local-storage/node_modules/async-listener/glue.js:188:31
      at _combinedTickCallback (internal/process/next_tick.js:71:11)
      at process._tickDomainCallback [as _tickCallback] (internal/process/next_tick.js:122:9)

  3) users - integration base-user should prevent access tokens from being included:
     Uncaught AssertionError: expected [] to have a property 'length' of 1, but got 0
      at Test.<anonymous> (test/user.integration.js:92:36)
      at Test.assert (node_modules/supertest/lib/test.js:221:6)
      at Server.assert (node_modules/supertest/lib/test.js:132:12)
      at emitCloseNT (net.js:1525:8)
      at node_modules/continuation-local-storage/node_modules/async-listener/glue.js:188:31
      at _combinedTickCallback (internal/process/next_tick.js:71:11)
      at process._tickDomainCallback [as _tickCallback] (internal/process/next_tick.js:122:9)
```

Will refactor them after this pr merged.

jannyHou:
@bajtos Since this feature is relational-db specific, do you prefer to add the test cases in each connector's repo or add them here in juggler?

Amir-61:
>@bajtos Since this feature is relational-db specific, do you prefer to add the test cases in each connector's repo or add them here in juggler?

Whenever it is a general test which is applicable for a bunch of connectors we should add tests in `loopback-datasource-juggler` which is run by connectors, as opposed to have duplicate tests in a bunch of connectors. That's my personal opinion.

bajtos:
> Whenever it is a general test which is applicable for a bunch of connectors we should add tests in loopback-datasource-juggler which is run by connectors, as opposed to have duplicate tests in a bunch of connectors. That's my personal opinion.

I agree too, it's better to keep shared tests inside juggler.

bajtos:
@jannyHou is it possible to fix the tests failing in LoopBack before this PR is landed? I am thinking that if the tests are failing because they are setting id in a model with strictId:true, then we need to fix them to not set the id (or change strictId to false), which can be done even before this change is landed.

Thoughts?

jannyHou:
@bajtos @Amir-61 Thanks! Sounds good 👍

superkhau:
>I agree too, it's better to keep shared tests inside juggler.

I've discussed this with @raymondfeng and we're going to keep contractual tests in juggler after the modularization stories -- I'm going to rename them to something more obvious than what it is now though ;))

test/basic-querying.test.js
```diff
@@ -26,6 +26,17 @@ describe('basic-querying', function() {
 
   });
 
+  describe('forceId', function() {
+    it('returns error when default forceId is true', function(done) {
```
bajtos: This test name does not make sense to me. How about "it defaults to forceId:true for generated id property"?

Also "basic-querying.test.js" looks like a wrong place for a test that creates new model instances. I think the "create" block in manipulation.test.js would be a better place - see https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/manipulation.test.js#L58

test/basic-querying.test.js
```diff
@@ -26,6 +26,17 @@ describe('basic-querying', function() {
...
+      User.create({ name: 'a', id: 1 })
+        .then()
```
bajtos: This test will pass when no error is thrown. Remember the golden rule - never trust a unit test you haven't seen fail.

To verify that a method returns rejected promise, one has to do something like this:

```js
User.create({...})
  .then(
    function() { throw new Error('User.create should have failed.'); },
    function(err) { /* assert on err */ });
```

Also if you decide to use promises, then don't mix it with callbacks. You can return a promise from your test function and Mocha will handle it well.

```js
it('...', function() {
  return User.create({...})
    .then(
      function() { throw new Error('User.create should have failed.'); },
      function(err) { /* assert on err */ });
});
```

bajtos:
@jannyHou could you please double-check whether all `{forceId:false}` additions are necessary with the current implementation?

jannyHou:
@bajtos 
> could you please double-check whether all {forceId:false} additions are necessary with the current implementation?

For the question above and the same implementation in [`loopback`](https://github.com/strongloop/loopback/pull/2501#issuecomment-233623576), let's keep discussion here:

I set `forceId: false` because after my change, the default `forceId` for all sql datasources are `true`, if I don't explicit change it to `false`,  those test cases creating a new instance with Id value will throw error `cannot set id`

bajtos:
> I set `forceId: false` because after my change, the default `forceId` for all sql datasources are `true`, if I don't explicit change it to `false`,  those test cases creating a new instance with Id value will throw error `cannot set id` 

There are two things confusing me:

1) You say the default `forceId` for all sql datasources are `true`, but that's not entirely true - the `forceId:true` is set only for auto-generated id properties, right?

2) Those test cases creating a new instance with Id value will throw error `cannot set id` - only when the model depends on auto-generated id column, right? It makes me wonder how can a test setting explicit id value work with auto-generated column and tests in all connectors, when SQL connectors use numeric ids and mongodb uses a string id (IIRC)?

What I am trying to avoid, is the situation where we change all test models to use `forceId:false`, even for models where it's not necessary.

jannyHou:
@bajtos Sorry my bad lol, it should be `auto-generated id` instead of `sql ds`. And from my understanding when `forceId = false`, a datasource using string-type Id like mongodb, will use the given Id instead of auto-generated one if Id provided. 

---
For example: https://github.com/strongloop/loopback-datasource-juggler/blob/60fa7cc5fbf19a12d3e43b6c9df40d763979d92e/test/default-scope.test.js#L828

```js
before(function(done) {
      Person.create({ id: 1, name: 'Person A' }, function(err, person) {
        person.things.create({ name: 'Thing A' }, done);
      });
    });

    it('should find a scoped instance with included relation - things', function(done) {
      Person.findById(1, function(err, person) {
        should.not.exist(err);
        should.exist(person);
        var things = person.things();
        should.exist(things);
        things.should.be.an.instanceOf(Array);
        things.should.have.length(1);
        done();
      });
    });
``` 

This existed test case could pass all connectors, including mongodb, couchdb, etc..
Correct me if I am wrong lol, and I am setting up a simple app using mongodb to test my assumption.

jannyHou:
@slnode test please

jannyHou:
@bajtos I applied your feedback:
1. move test case to manipulation.test.js
2. rename test case and use callback signature instead of promise to keep consistent with mocha callback.

PTAL, thank you :)


And I also write a promise version, please see the following code:

```js
it('it defaults to forceId:true for generated id property', function() {
  return TestForceId.create({ id: 1 })
    .then(function(t) { 
       throw new Error('TestForceId.create should have failed');
    })
    .catch(function(err) {
       should.exist(err);
       err.message.should.match(/can\'t be set/);
    });
})
```

Is it the right way of combining mocha and promise? 
I remember that we discussed in loopback channel before, as we'd better use promise style in the new test cases. So did we reach any agreement of it or any concrete plan for it?
I remember we also mentioned supporting both callback and promise version of test cases.
And I will include `plan for promisify` in homework 2 to bump up a better implementation plan of feature `Pomisify APIs`.

bajtos:
@jannyHou 
> And I also write a promise version, please see the following code:
> Is it the right way of combining mocha and promise? 

Yes, your code looks good to me.

>  remember that we discussed in loopback channel before, as we'd better use promise style in the new test cases. So did we reach any agreement of it or any concrete plan for it?

I think we didn't reach any conclusion. I personally would like to start using promise-based API in tests that are not testing that API, but merely invoking it during setup phase.

As for `forceId:false` in tests, I reviewed the tests and tweaked few of them so that `forceId` is no longer necessary, see 1b8db43.

I also added e38cd5e to add missing error handlers, it was difficult to troubleshoot test failures without that! If you prefer to send this commit in a new PR, then let me know, I can do that too.

superkhau:
>I think we didn't reach any conclusion. I personally would like to start using promise-based API in tests that are not testing that API, but merely invoking it during setup phase.

:+1:

jannyHou:
@bajtos Thank you for helping me fix up test cases. Some dependents fail but not due to this code. I think it's good to land now. PTAL, thanks!

loopback-connector-oracle: amazon6 and ubuntu6 fails due to 
> Error: Cannot find module 'oracledb'
Triggered a [build on master](http://ci.strongloop.com/job/loopback-connector-oracle/) to see if those two platforms have config problem.

loopback-workspace: ubuntu-0.10 and amazon-0.10 fails due to timeout error :(

loopback-connector-mssql: there is a test case failing,  Loay is working on fixing it now: https://github.com/strongloop/loopback-connector-mssql/pull/92

bajtos:
Sure, let's get this landed! Please squash the commits first, keep the last one (1b7b0bb) as a standalone commit.

jannyHou:
loopback passed lol, merging it.

superkhau:
@jannyHou Delete branch?

bajtos:
@superkhau @jannyHou I don't remember, did we consider this a breaking change that's landed in 3.0 only? If it's so, then we need to add an entry to 3.0-RELEASE-NOTES. Some users are already getting confused, see e.g. https://github.com/strongloop/loopback/issues/2642

jannyHou:
@bajtos Yes it's a breaking change and I didn't back port it to 2.x. I will work on updating the docs with Rand and adding it to 3.0-RELEASE-NOTES asap

luncht1me:
This change breaks UPSERT methods...
When you Upsert with an ID, you want to PATCH/PUT that existing instance to update it in the database.
And if it doesn't have an ID, then you create a new one.

Upsert is just completely broken because of this id enforcement not accounting for the use-case and intended behavior of Upsert.
Why not have Upsert check if there's an ID, if there is make sure the document exists, error if it doesn't saying id can't be set. But if the document exists, update the damn thing. Having to double the amount of code for CRUD management in our projects when there `used` to be a method that handled it seems a step backwards to me.

bajtos:
> Why not have Upsert check if there's an ID, if there is make sure the document exists, error if it doesn't saying id can't be set. But if the document exists, update the damn thing. 

@luncht1me thank you for the comment. I think the issue you are observing is very similar to #1235 which was fixed by #1210 and https://github.com/strongloop/loopback-connector/issues/73.

Would you like to contribute the fix yourself, using the two linked pull requests for inspiration?

cc @loay @jannyHou @strongloop/sq-lb-apex

bajtos:
Let's move this discussion to https://github.com/strongloop/loopback-datasource-juggler/issues/1185 please.

PR #986, opened 2016-06-24
Fix #984 Give request options(ex: transactions) to validators

### LINKED ISSUE #984: No options passed in validation causes big concurency problems with postgres
Validation problems because no options given:
=====================================

Facts:
-------
When I want to update an object which needs a validation, the validation is called in the datasource-juggler but the options of my request are not given to the validation function.

The validation function makes a query without my eventual current transaction which should be in the options.

Big concurency problems:
----------------------------------
The connection pool of  postgres-connector has a limited amount of connection. (10 by default)
If  I have a remoteMethod starting a transaction in order to update a model, and if 10 users call this endpoint at the exact same time, then my 10 connection in my connection pool will be used for the 10 different transactions.
It's not a problem most of the time.

But, if my model needs a validation (ex: ValidateUniquenessOf email on a user), then before saving my model, the validation function will make a select query out of my transaction context. My transaction will wait for the result in order to continue.

As my 10 connections are already taken by the transactions, postgre-connector will queue my request until a connection is released. which will never be the case because my transactions are waiting for this validation request.

We are in a **deadlock** situation.

**This make loopback not suitable for production with postgres**

Proposition:
----------------

The datasource-juggler should take an `option` parameter which would potentially be passed to the connector if the connector is suitable with options.
This mean we should modify all the validation functions in `validation.js` (`isValid`, `validationFailed`,`ModelConstructor.validateUniqueness`, ...)

Conclusion:
----------------
This is a big problem which causes serious issues. I will start a PR. Let me know if my solution seems good to you.

robinbiondi:
This PR fixes issue #984.
When calling isValid for validation from dao, I pass the options parameter and propagate it in validators. Validators then, use it in order to do there potential requests. Doing this way, we do the validation request in the same context than the original request => potential same transaction.

This PR is critical as until now, loopback is not suitable for production with postgres due to transaction concurrency issues. (My prod is down 😄) See issue #984

Duplicate of PR #985 with an appropriate target

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -320,7 +320,7 @@ DataAccessObject.create = function(data, options, cb) {
       } else {
         cb(new ValidationError(obj), obj);
       }
-    }, data);
+    }, data, options);
   });
 
   function create() {
@@ -543,7 +543,7 @@ DataAccessObject.upsert = function(data, options, cb) {
               }
             }
             callConnector();
-          }, update);
+          }, update, options);
         }
 
         function callConnector() {
@@ -732,7 +732,7 @@ DataAccessObject.replaceOrCreate = function replaceOrCreate(data, options, cb) {
           inst.isValid(function(valid) {
             if (!valid) return cb(new ValidationError(inst), inst);
             callConnector();
-          }, update);
+          }, update, options);
 
           function callConnector() {
             update = removeUndefined(update);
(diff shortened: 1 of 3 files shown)


slnode:

slnode:

slnode:

slnode:

PR #995, opened 2016-07-11
changes to dao.js and memory.js to add a new method on Upsert based on non-primary key

mountain1234585:
Hi Raymond,
Kindly review the basic changes done to loopback-datasource-juggler lib/dao.js and lib/connectors/memory.js to include a new method called "upsertWithWhere" based on non-primary key.
1)This method leverages the existing methods, find, create, findById and updateAttributes.
2)I have not added notifyObserversOf like in other methods , since this method calls them internally.
3)Once you review the code changes, we will add the unit tests as well.
4) I have tested the operation manually by creating an application, memory db datasource and persisted model in case of create and update scenarios.
Thanks,
Sonali Samantaray


slnode:

slnode:

slnode:

slnode:

lib/connectors/memory.js
```diff
@@ -792,6 +792,105 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
   });
 };
 
+Memory.prototype.upsertWithWhere = function(model,data,options,callback){
```
raymondfeng: The method signature is not correct. The `where` argument is missing.

lib/connectors/memory.js
```diff
@@ -792,6 +792,105 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
...
+
+
+    var response = null;
+    if (Object.keys(data).length === 0)
+      response = false;
+    else
+      response = true;
+
+    if (!data.hasOwnProperty(element)) {
+      data[element] = where[element];
```
raymondfeng: For `upsertWithWhere`, we should not merge things from `where` into `data`. The `where` argument should be the matching criteria while the `data` argument should be the properties to be inserted or updated.

lib/connectors/memory.js
```diff
@@ -792,6 +792,105 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
...
+  try {
+
+    var response;
+
+    var element = Object.keys(where)[0];
+
+    var filter = {
+      "where": where
+    };
+
```
raymondfeng: There are too many blank lines.

lib/connectors/memory.js
```diff
@@ -792,6 +792,105 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
...
+      data[element] = where[element];
+    }
+
+    if (response == true) {
+
+      self.all(filter, function (err, models) {
+
+        if (err) {
+
+          callback(err);
```
raymondfeng: Please wrap `callback` into `process.nextTick()` so that callback only happens after the function is finished.

lib/connectors/memory.js
```diff
@@ -792,6 +792,105 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
...
+        } else {
+
+          var idName = idName(Model);
+
+          if (models.length > 1) {
+
+            callback("Error message:" + "Multiple records exists.Upsert can't be performed.");
```
raymondfeng: Why don't we update multiple matching records?

lib/dao.js
```diff
@@ -68,8 +67,8 @@ function convertSubsetOfPropertiesByType(inst, data) {
     // Convert the properties by type
     typedData[key] = inst[key];
     if (typeof typedData[key] === 'object' &&
-            typedData[key] !== null &&
-            typeof typedData[key].toObject === 'function') {
```
raymondfeng: There are a lot of style changes mixed with functional changes. Please either keep the existing style or commit the style changes in a separate one.

lib/dao.js
```diff
@@ -2783,192 +2836,308 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
...
-        'cannot be updated from ' + inst[idName] + ' to ' + data[idName]);
-      err.statusCode = 400;
-      process.nextTick(function() {
-        cb(err);
-      });
-      return cb.promise;
-    }
-  }
+  var element = Object.keys(where)[0];
+  console.log(element);
```
raymondfeng: Please remove console.log or turn it into a debug statement.

lib/dao.js
```diff
@@ -2783,192 +2836,308 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
...
-};
+    });
 
+  }
+
+
+
+else
+{
+cb("data field can't be empty");
```
raymondfeng: Bad style

slnode:

slnode:

slnode:

lib/connectors/memory.js
Amir-61: Please run our linter; I see lots of lint issues

lib/connectors/memory.js
```diff
@@ -792,6 +792,105 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
...
+          } else if (models.length == 0 && data.hasOwnProperty(idName)) {
+
+            self.create(data, function (err, results) {
+
+              if (err)
+                callback(err);
```
Amir-61: We always use:

```
if (err) callback(err);
```

lib/connectors/memory.js
```diff
@@ -792,6 +792,105 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
...
+
+
+            var idName = idName(Model);
+
+            var idVal = models[0][idName];
+
+            self.findById(idVal, function (err, modelInst) {
+              if (err) {
+
+                callback(err);
```
Amir-61: ditto

lib/connectors/memory.js
```diff
@@ -792,6 +792,105 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
...
+              }
+              else {
+
+                modelInst.updateAttributes(data, function (err, info) {
+
+                  if (err) {
+
+                    callback(err);
```
Amir-61: ditto; please fix in other places as well

lib/connectors/memory.js
```diff
@@ -792,6 +792,105 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
...
+      });
+    } else
+      callback("Error Message: Data/payload can't be empty");
+  } catch (ex) {
+
+    
+    callback(ex);
+  }
+
+
```
Amir-61: I see too many blank lines; please fix them

lib/dao.js
```diff
@@ -83,7 +82,8 @@ function convertSubsetOfPropertiesByType(inst, data) {
 function applyStrictCheck(model, strict, data, inst, cb) {
   var props = model.definition.properties;
   var keys = Object.keys(data);
-  var result = {}, key;
+  var result = {},
+    key;
```
Amir-61: Please avoid changing un-related code in this PR

lib/dao.js
```diff
@@ -106,11 +106,14 @@ function setIdValue(m, data, value) {
 
 function byIdQuery(m, id) {
   var pk = idName(m);
-  var query = { where: {}};
+  var query = {
+    where: {}
+  };
```
Amir-61: Please dont change un-related code in this PR

Amir-61:
@mountain1234585 
Did you sign [CLA](https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler)? Please add tests and run npm test locally and fix issues as well as lint issues. You can type `@slnode test please` to run tests on CI to make sure you are passing tests.

mountain1234585:
Hello Raymond,

Thanks for your detailed review in a short span of time.

Following on the comments you had on the logic :

1) why we don't update multiple instances rather return error if multiple instances are found.
	This was discussed as part of the requirement we had and we plan to restrict this only for single instance update. I had a look at how upsert with foreign key is performed in some of the popularly used endpoints and they follow the similar approach.  

2) why the key value pair mentioned in the "where" is merged into data
        In case of create, we want the field to be updated when a new record is created. And similar is seen in upsert operation with Salesforce as well.

3) method signature is wrong in memory.js
	I have now removed the method in memory.js since the entire logic is handled in dao.js
4) too many empty lines in dao.js
	I have removed them now.


Regards,
Sonali Samantaray

mountain1234585:
@Amir-61
I have fixed the elint errors now.

mountain1234585:
I think I need to take the latest code, since I can see some commits that has happened in between.

mountain1234585:
There are too many conflicts in this PR, while trying to rebase and sync. I am doing these changes and have submitted a PR as part of PR https://github.com/strongloop/loopback-datasource-juggler/pull/1001

PR #996, opened 2016-07-11
Ensure stable order of items in DAO.find()

### LINKED ISSUE #800: Model.find returns items with incorrect order

bajtos:
When post-processing result of find operation, use "async.map" instead of "async.each + array.push" to ensure the order of items is preserved.

This work is based on #864, but preserves parallel post-processing.

Fix #800 

@Amir-61 please review
cc @MichaelKohler

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1678,26 +1678,22 @@ DataAccessObject.find = function find(query, options, cb) {
   }
 
   var allCb = function(err, data) {
-    var results = [];
     if (!err && Array.isArray(data)) {
-      async.each(data, function(item, callback) {
-        var d = item;//data[i];
-        var Model = self.lookupModel(d);
+      async.map(data, function(item, next) {
+        var Model = self.lookupModel(item);
         if (options.notify === false) {
-          buildResult(d);
+          buildResult(item, next);
         } else {
-          withNotify();
+          withNotify(item, next);
         }
 
-        function buildResult(data) {
-          d = data;
-
+        function buildResult(data, callback) {
           var ctorOpts = {
             fields: query.fields,
             applySetters: false,
             persisted: true,
           };
-          var obj = new Model(d, ctorOpts);
+          var obj = new Model(data, ctorOpts);
 
           if (query && query.include) {
             if (query.collect) {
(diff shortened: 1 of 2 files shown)


bajtos:
I have updated the patch to filter out `undefined` entries. While I am not happy that this part of my change is not covered by any unit tests, I decided to leave it at that, because there is no unit tests and no documentation for `query.collect` that I could find, and I don't have bandwidth to research it right now.

@Amir-61 please review.

Amir-61:
@slnode test please

Amir-61:
@jannyHou Do you any thought why `loopback-connector-mongodb` fails for lazyconnect? Here is [the log](http://ci.strongloop.com/job/loopback-connector-mongodb/1364/label=amazon-5/console)

jannyHou:
@Amir-61 I think the test fails due to a change of error msg. 
I didn't see the new error msg "failed to connect to server [127.0.0.1:4] on first connect" before. Let me check is there any related refactor for it recently.

superkhau:
LGTM

Amir-61:
Thanks @bajtos For taking care of this with optimal solution 👍

The code changes LGTM!

1) I see some failures in `loopback-connector-mongodb` for `lazyConnect`, which I don't think it is related to this PR; probably it is due to changes on error message on `loopback-error-handler` CC: @jannyHou
2) Please rebase your code on top of master before landing.

Thanks again 🙇

jannyHou:
@Amir-61 It's caused by the official mongodb driver :(
I created a pr to fix it: https://github.com/strongloop/loopback-connector-mongodb/pull/263

bajtos:
Thank you for the review and taking care of the cause of mongodb CI failures.

I have rebased the patch on top of the current master and going to land it.

terehov:
Awesome! I really need this feature asap. 
Do you have a schedule or something to bring it into the npm package?

PR #999, opened 2016-07-13
Ensure stable order of items in DAO.find() [2.x]

bajtos:
When post-processing result of find operation, use "async.map" instead of "async.each + array.push" to ensure the order of items is preserved.

This is a back-port of #996.

The code has diverged between 2.x and master, the merge leave too many conflicts, so I had to re-do the change again.

/to @Amir-61 please review
/cc @jannyHou @MichaelKohler

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1680,12 +1680,10 @@ DataAccessObject.find = function find(query, options, cb) {
   }
 
   var allCb = function(err, data) {
-    var results = [];
     if (!err && Array.isArray(data)) {
-      async.each(data, function(item, callback) {
-        var d = item;//data[i];
-        var Model = self.lookupModel(d);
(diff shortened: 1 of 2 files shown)


Amir-61:
@bajtos LGTM :shipit:

jannyHou:
@bajtos LGTM 👍 
And one question: why do we setup the hook with a delay?

bajtos:
> @jannyHou why do we setup the hook with a delay?

To reproduce the problem reported in #800:
> Because async.each run in parallel so results.push(obj) is not guaranteed to be run in correct order. That why I get back a list of items in incorrect order.

We are simulating the situation when an iteration of async.each/async.map takes numbers of tick to complete, a different number for each iteration. This typically happens when an async operation is made from the iteration, for example another database request is sent.

PR #1009, opened 2016-07-22
Declare `definition`

Amir-61:
Due to undefined `definition` the following line was broken; this is a fix to declare `definition` which is used in the following error message:

```
cb(new Error('BelongsTo relation ' + definition.name + ' is empty'));
```

/to: @jannyHou
/CC: @bajtos

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -1366,9 +1366,10 @@ BelongsTo.prototype.destroy = function(options, cb) {
     cb = options;
     options = {};
   }
-  var modelTo = this.definition.modelTo;
+
+  var definition = this.definition; 
   var modelInstance = this.modelInstance;
-  var fk = this.definition.keyFrom;
+  var fk = definition.keyFrom;
 
   cb = cb || utils.createPromiseCallback();
 


lib/relation-definition.js
```diff
@@ -1366,9 +1366,10 @@ BelongsTo.prototype.destroy = function(options, cb) {
     cb = options;
     options = {};
   }
-  var modelTo = this.definition.modelTo;
```
Amir-61: Please note `modelTo` was unused, so I deleted it.

jannyHou:
@Amir-61 LGTM.

I am also confused with the usage of `modelTo`, seems like an unused var forgot to delete in pr https://github.com/strongloop/loopback-datasource-juggler/commit/af99a8d34492725a61ef1944c2de3f21bd353f34

Amir-61:
@slnode test please

Amir-61:
Thanks @jannyHou 
I need to back-port it to `2.x` as well.

Amir-61:
Backporting:
Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/1013

bajtos:
Nice work :+1:

PR #1016, opened 2016-07-24
Fix test case typo

supasate:
Fix a test case that coerces date types but it is incorrectly described to boolean types.

Signed-off-by: Supasate Choochaisri <supasate.c@gmail.com>

Issue #123, opened 2014-05-26
Access denied trying to query belongsTo


pbrain19:
![accessdeny](https://cloud.githubusercontent.com/assets/2614807/3078428/b951a384-e471-11e3-83c6-b4e45df5bc6a.png)

PR #1018, opened 2016-07-24
Fix test typo[2.x]

Amir-61:
Backporting
Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/1016

CC: @supasate

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -1622,7 +1622,7 @@ describe('manipulation', function() {
       p1.should.have.property('married', undefined);
     });
 
-    it('should coerce boolean types properly', function() {
+    it('should coerce date types properly', function() {
       var p1 = new Person({ name: 'John', dob: '2/1/2015' });
       p1.dob.should.eql(new Date('2/1/2015'));
 


superkhau:
@Amir-61 Please use the full link in the commit body so that GitHub can convert it to a link automatically.

```
Backport loopback-datasource-juggler#1016
```

to

```
Backport https://github.com/strongloop/loopback-datasource-juggler/pull/1016
```

Amir-61:
@superkhau  

I tried the full URL, but it is not gonna work; please note the commit linter fails; the commit description can be up to 72 char but `Backport https://github.com/strongloop/loopback-datasource-juggler/pull/1016` would be 76 chars.

superkhau:
>I tried the full URL, but it is not gonna work; please note the commit linter fails; the commit description can be up to 72 char but Backport https://github.com/strongloop/loopback-datasource-juggler/pull/1016 would be 76 chars.

Interesting. @bajtos Do you have a solution to this? Should we relax the linter rules for URLs? It's nice to click on the commit message and have a link directly to the backport (and having the full link available in the command line when digging through history). I am also OK with @Amir-61's convention is also ok since it does convey the minimal amount of info to manually look for the PR in the GitHub UI, but the UX is not ideal.

bajtos:
I am fine with using `#1016` as Node.js project does. GitHub is smart and will turn `#1016` into a clickable link #1016 when viewing the commit message on GitHub.com ([example](https://github.com/nodejs/node/commit/46633934fed7dfce289298cf16ea0f4161b77bb7)).

Note that it's important to use on of the following formats: `#123` or `user/repo#123`. `repo#123` is not recognized by GitHub AFAIK.

superkhau:
>Note that it's important to use on of the following formats: #123 or user/repo#123. repo#123 is not recognized by GitHub AFAIK.

@Amir-61 Can you add this to the PR for backporting? Or submit a new on with updates because I think you merged in the other one already.

Amir-61:
@superkhau Sent a PR: https://github.com/strongloop-internal/loopback-knowledge-base/pull/48 :-)

PR #1023, opened 2016-07-29
KeyValue access object + memory connector

superkhau:
KeyValue memory connector RI.

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,3 +22,5 @@ Object.defineProperty(exports, 'test', {
 });
 
 exports.Transaction = require('loopback-connector').Transaction;
+
+exports.KeyValueAccessObject = require('./lib/kvao');
--- a/lib/connectors/kv-memory.js
+++ b/lib/connectors/kv-memory.js
@@ -1,0 +1,143 @@
+'use strict';
+
+var assert = require('assert');
+var Connector = require('loopback-connector').Connector;
+var debug = require('debug')('loopback:connector:kv-memory');
+var util = require('util');
+
+exports.initialize = function initializeDataSource(dataSource, cb) {
+  var settings = dataSource.settings;
+  dataSource.connector = new KeyValueMemoryConnector(settings, dataSource);
+  if (cb) process.nextTick(cb);
+};
+
+function KeyValueMemoryConnector(settings, dataSource) {
+  Connector.call(this, 'kv-memory', settings);
+
+  debug('Connector settings', settings);
+
+  this.dataSource = dataSource;
+  this.DataAccessObject = dataSource.juggler.KeyValueAccessObject;
+
+  this._store = Object.create(null);
+
+  this._setupRegularCleanup();
+};
+util.inherits(KeyValueMemoryConnector, Connector);
+
+KeyValueMemoryConnector.prototype._setupRegularCleanup = function() {
+  // Scan the database for expired keys at a regular interval
+  // in order to release memory. Note that GET operation checks
+  // key expiration too, the scheduled cleanup is merely a performance
+  // optimization.
+  var self = this;
+  this._cleanupTimer = setInterval(
+    function() { self._removeExpiredItems(); },
+    1000);
+  this._cleanupTimer.unref();
+};
+
+KeyValueMemoryConnector._removeExpiredItems = function() {
+  debug('Running scheduled cleanup of expired items.');
+  for (var modelName in this._store) {
+    var modelStore = this._store[modelName];
+    for (var key in modelStore) {
+      if (modelStore[key].isExpired()) {
+        debug('Removing expired key', key);
+        delete modelStore[key];
+      }
+    }
+  }
+};
+
+KeyValueMemoryConnector.prototype._getStoreForModel = function(modelName) {
+  if (!(modelName in this._store)) {
+    this._store[modelName] = Object.create(null);
+  }
+  return this._store[modelName];
+};
+
+KeyValueMemoryConnector.prototype.get =
+function(modelName, key, options, callback) {
+  var store = this._getStoreForModel(modelName);
+  var item = store[key];
+
+  if (item && item.isExpired()) {
+    debug('Removing expired key', key);
+    delete store[key];
+    item = undefined;
+  }
+
+  var value = item ? item.value : null;
+
+  debug('GET %j %j -> %s', modelName, key, value);
+
+  if (/^buffer:/.test(value)) {
+    value = new Buffer(value.slice(7), 'base64');
+  } else if (/^date:/.test(value)) {
+    value = new Date(value.slice(5));
+  } else if (value != null) {
+    value = JSON.parse(value);
+  }
+
+  process.nextTick(function() {
+    callback(null, value);
+  });
+};
+
+KeyValueMemoryConnector.prototype.set =
+function(modelName, key, value, options, callback) {
+  var store = this._getStoreForModel(modelName);
+  var value;
+  if (Buffer.isBuffer(value)) {
+    value = 'buffer:' + value.toString('base64');
+  } else if (value instanceof Date) {
+    value = 'date:' + value.toISOString();
+  } else {
+    value = JSON.stringify(value);
+  }
+
+  debug('SET %j %j %s %j', modelName, key, value, options);
+  store[key] = new StoreItem(value, options && options.ttl);
+
+  process.nextTick(callback);
+};
+
+KeyValueMemoryConnector.prototype.expire =
+function(modelName, key, ttl, options, callback) {
+  var store = this._getStoreForModel(modelName);
+
+  if (!(key in store)) {
+    return process.nextTick(function() {
+      callback(new Error('Cannot expire unknown key ' + key));
+    });
+  }
+
+  debug('EXPIRE %j %j %s', modelName, key, ttl || '(never)');
+  store[key].setTtl(ttl);
+  process.nextTick(callback);
+};
+
+KeyValueMemoryConnector.prototype.disconnect = function(callback) {
+  if (this._cleanupTimer)
+    clearInterval(this._cleanupTimer);
+  this._cleanupTimer = null;
+  process.nextTick(callback);
+};
+
+function StoreItem(value, ttl) {
+  this.value = value;
+  this.setTtl(ttl);
+}
+
+StoreItem.prototype.isExpired = function() {
+  return this.expires && this.expires <= Date.now();
+};
+
+StoreItem.prototype.setTtl = function(ttl) {
+  if (ttl) {
+    this.expires = Date.now() + ttl;
+  } else {
+    this.expires = undefined;
+  }
+};
--- a/lib/kvao/index.js
+++ b/lib/kvao/index.js
@@ -1,0 +1,15 @@
+'use strict';
+
+function KeyValueAccessObject() {
+};
+
+module.exports = KeyValueAccessObject;
+
+KeyValueAccessObject.get = require('./get');
+KeyValueAccessObject.set = require('./set');
+KeyValueAccessObject.expire = require('./expire');
+
+KeyValueAccessObject.getConnector = function() {
+  return this.getDataSource().connector;
+};
+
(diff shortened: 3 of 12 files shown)


bajtos:
@superkhau I added basic implementation of `get` and `set` methods together with tests.

@raymondfeng @ritch please review if this implementation of KeyValue API & connector is going in the right direction.

index.js
```diff
@@ -22,3 +22,5 @@ Object.defineProperty(exports, 'test', {
 });
 
 exports.Transaction = require('loopback-connector').Transaction;
```
raymondfeng: Is `Transaction` needed? If not, we should remove it.

lib/kvao/index.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+function KeyValueAccessObject() {
+};
+
+module.exports = KeyValueAccessObject;
+
+KeyValueAccessObject.get = require('./get');
+KeyValueAccessObject.set = require('./set');
+
```
raymondfeng: Why do we want to split `get` and `set` into separate files?

lib/connectors/kv-memory.js
```diff
@@ -0,0 +1,62 @@
...
+util.inherits(KeyValueMemoryConnector, Connector);
+
+KeyValueMemoryConnector.prototype._getStoreForModel = function(modelName) {
+  if (!(modelName in this._store)) {
+    this._store[modelName] = Object.create(null);
+  }
+  return this._store[modelName];
+};
+
+KeyValueMemoryConnector.prototype.get = function(modelName, key, callback) {
```
raymondfeng: Maybe we should have an optional `options` before `callback`?

lib/connectors/kv-memory.js
```diff
@@ -0,0 +1,62 @@
...
+  } else {
+    value = JSON.parse(value);
+  }
+
+  process.nextTick(function() {
+    callback(null, value);
+  });
+};
+
+KeyValueMemoryConnector.prototype.set = function(modelName, key, value, callback) {
```
raymondfeng: Maybe we should have an optional options before callback?

lib/kvao/index.js
bajtos: To avoid the situation we have in `lib/dao.js`, where the file contains 3000 lines of code and still grows.

By splitting the methods into separate files, the project is also easier navigate around, at least in my experience/opinion.

lib/connectors/kv-memory.js
bajtos: That's a good idea, will do it here and in `set` too.

bajtos: Except I'll make it required to keep the implementation of connectors simpler. KVAO can handle the optionality of that argument.

index.js
bajtos: I have no idea, whether `Transaction` is needed or not. IMO, this is unrelated to the key-value implementation and therefore we should not change it in this pull request.

Removing `exports.Transaction` is also a breaking change.

It makes sense to remove it in 3.0, since users should access `Transaction` directly from `loopback-connector`.

@raymondfeng would you mind opening a pull request for this change yourself? Just don't forget to update 3.0-RELEASE-NOTES.md

bajtos:
@raymondfeng @ritch @superkhau I pushed few more commits implemeting `set(key, value, ttl)` and `expire(key, ttl)`. PTAL

index.js
raymondfeng: Oh, never mind. I was confused. The export is in juggler. We need to keep Transaction as it.

lib/connectors/kv-memory.js
```diff
@@ -0,0 +1,139 @@
+'use strict';
```
raymondfeng: I wonder why we don't use https://github.com/isaacs/node-lru-cache as the base for implementing in-memory KV store.

bajtos: The most important reason is that AFAICT, node-lru does not support EXPIRE operation.

Besides that, LRU cache keeps a fixed amount of items and removes the least-recently-used items when the size limit is hit. IMO, that is very undesirable behaviour for a key-value store implementation.

I guess we could use LRU with `max: Infinity`, but that's not great, as we will be bending the library to our needs that are different from what the library is designed for. In my experience, that usually leads to headaches later in the future, when we hit the limits of how the library can be bent.

> https://github.com/isaacs/node-lru-cache#options
> `max` The maximum size of the cache, checked by applying the length function to all values in the cache. Not setting this is kind of silly, since that's the whole purpose of this lib, but it defaults to Infinity.

bajtos:
At the moment, this pull request contains all that's needed for the first cut of this new feature IMO.

Models attached to a kv-memory datasource will get the following *static* methods:

```js
Color.set(key, value);
Color.set(key, value, ttl);
Color.set(key, value, { ttl: ttl });

Color.get(key);

Color.expire(key, ttl);
```

The proposal in strongloop/loopback#2525 describes prototype methods too. In my view, they are a syntactic sugar that's out of scope of this initial pull request, and should be implemented in the next iteration (e.g. right after this patch is landed).

Another missing item is `Color.ttl(key)`, I am not sure how important that method is? If we think it should be part of the initial release, then we can implement it in the next iteration too.

@raymondfeng @ritch @superkhau please review

superkhau:
LGTM, nice work.

ritch:
> Another missing item is Color.ttl(key), I am not sure how important that method is? If we think it should be part of the initial release, then we can implement it in the next iteration too.

It is important.

superkhau:
@bajtos Added TTL, review please.

lib/connectors/kv-memory.js
```diff
@@ -141,3 +159,10 @@ StoreItem.prototype.setTtl = function(ttl) {
     this.expires = undefined;
   }
 };
+
+StoreItem.prototype.getTtl = function() {
+  // Same behaviour as Redis, see http://redis.io/commands/ttl
```
bajtos: IMO, the Redis API is not great from a Node.js/JavaScript perspective.

- For items that don't expire, I expect `ttl()` to return undefined.
- For items that are already expired but were not deleted yet, the same error about non-existent key should be returned.

lib/connectors/kv-memory.js
```diff
@@ -118,6 +118,24 @@ function(modelName, key, ttl, options, callback) {
   process.nextTick(callback);
 };
 
+KeyValueMemoryConnector.prototype.ttl =
+function(modelName, key, options, callback) {
+  var store = this._getStoreForModel(modelName);
+
+  // key is unknown
+  if (!(key in store)) {
```
bajtos: You should check if the item is expired and remove it, the same way as `get` works.

Please extract L63-L69 into a shared method that can be used by both `get` and `ttl`.

lib/connectors/kv-memory.js
```diff
@@ -118,6 +118,24 @@ function(modelName, key, ttl, options, callback) {
...
+    return process.nextTick(function() {
+      callback(new Error('Cannot get TTL for unknown key ' + key));
+    });
+  }
+
+  debug('TTL %j %j', modelName, key);
```
bajtos: Please include the ttl result in the debug message:

```js
debug('TTL %j %j -> %s', modelName, key, ttl);
```

lib/kvao/ttl.js
```diff
@@ -0,0 +1,30 @@
...
+var assert = require('assert');
+var utils = require('../utils');
+
+/**
+ * Get remaining expiration (TTL) for a given key.
+ *
+ * @param {String} key
+ * @param {Object} options
+ * @callback cb
+ * @param {Error} error
```
bajtos: The callback takes the second argument. Please document it and explain what value is returned when no TTL was set.

test/kvao/ttl.suite.js
```diff
@@ -0,0 +1,104 @@
...
+    var CacheItem;
+    beforeEach(function unpackContext() {
+      CacheItem = helpers.givenCacheItem(dataSourceFactory);
+    });
+
+    context('existing key with expire before expiration time', function() {
+      it('returns ttl - Callback API', function(done) {
+        CacheItem.set('a-key', 'a-value', 10, function(err) {
+          if (err) return done(err);
+          setTimeout(function() {
```
bajtos: I think the timeout is not needed here. It should not matter whether you are checking TTL 9 or 10 seconds before it expires. Unless I am missing something?

test/kvao/ttl.suite.js
```diff
@@ -0,0 +1,104 @@
...
+            CacheItem.ttl('a-key', function(err, ttl) {
+              if (err) return done(err);
+              ttl.should.be.a.Number().lessThan(10).and.greaterThanOrEqual(0);
```
bajtos: Chai provides `within(0, 10)`, is there something similar in `should` too?

test/kvao/ttl.suite.js
```diff
@@ -0,0 +1,104 @@
...
+            return CacheItem.ttl('a-key');
+          })
+          .then(function(ttl) {
+            ttl.should.be.a.Number().lessThan(10).and.greaterThanOrEqual(0);
+          });
+      });
+    });
+
+    context('existing key with expire after expiration time', function(done) {
+      it('returns -2 - Callback API', function(done) {
```
bajtos: No need to test both callback & promise flavour in all tests. Test it once, to ensure both flavours works, and then keep using Promise API only in other tests.

test/kvao/ttl.suite.js
```diff
@@ -0,0 +1,104 @@
...
+              ttl.should.equal(-2);
+              done();
+            });
+          }, 11);
+        });
+      });
+
+      it('returns -2 - Promise API', function() {
+        return CacheItem.set('a-key', 'a-value', 10)
+          .delay(11)
```
bajtos: Please use a longer delay to prevent situation when slow machines and/or slow connections and/or other timing issues may cause the key to expire little bit later. I.e. use `20` here.

test/kvao/ttl.suite.js
```diff
@@ -0,0 +1,104 @@
...
+            return CacheItem.ttl('a-key');
+          })
+          .then(function(ttl) {
+            ttl.should.equal(-2);
+          });
+      });
+    });
+
+    context('existing key without expire', function() {
+      it('returns -1 - Callback API', function(done) {
```
bajtos: No need to test both callback & promise flavour in all tests. Test it once, to ensure both flavours works, and then keep using Promise API only in other tests.

test/kvao/ttl.suite.js
```diff
@@ -0,0 +1,104 @@
...
+            return CacheItem.ttl('a-key');
+          })
+          .then(function(ttl) {
+            ttl.should.equal(-1);
+          });
+      });
+    });
+
+    context('non-existing key', function() {
+      it('returns an error - Callback API', function(done) {
```
bajtos: ditto - No need to test both callback & promise flavour in all tests. Test it once, to ensure both flavours works, and then keep using Promise API only in other tests.

bajtos:
> > Another missing item is Color.ttl(key), I am not sure how important that method is? If we think it should be part of the initial release, then we can implement it in the next iteration too.

> It is important and should be included in the first version.

FWIW, `ttl` method is not available in all key-value stores. For example, memcached does not support it according to [SO](http://stackoverflow.com/questions/5329431/memcache-get-key-expiry-times) and [client API docs](https://www.npmjs.com/package/memcached#api).

I am fine with including `ttl` in the first release, but please move it to a new standalone pull request, so that we can land this one. @superkhau 

/cc @ritch

test/kvao/ttl.suite.js
superkhau: I think it might not be needed if within works as expected. I needed it cause I used `lessThan(10)` and the timeout guarantee some time has passed (ie. doesn't return 10 again). I should've used `lessThanOrEqual(10)`, but your suggestion of `within(0, 10)` is even better.

superkhau:
@bajtos TTL code moved to https://github.com/strongloop/loopback-datasource-juggler/pull/1030, also removed TTL code from this PR.

bajtos:
> @superkhau  TTL code moved to #1030, also removed TTL code from this PR.

Thank you! I'll clean up the commit history and get this landed to both `master` and `2.x`.

bajtos:
@slnode test please

PR #1022, opened 2016-07-26
Support for globalization

Amir-61:
Add support for globalization

to:/ @0candy @superkhau PTAL

PR #1024, opened 2016-08-03
Use g.f instead of utils.format

Amir-61:
Use `g.f(...)` instead of `g.f(utils.format(...))`

Comment from Tesuo is [here](https://github.com/strongloop/loopback-datasource-juggler/pull/1022/files#r73207761)

/to: @jannyHou


jannyHou:
@Amir-61 LGTM 👍

PR #1030, opened 2016-08-06
Ttl

superkhau:
@bajtos PTAL

I removed the last commit on the other branch so you can merge. We can continue reviews here. I branched off your branch so I could get the latest commits too so you'll see 7 commits here, rebase/squash after your merge over there.

### DIFF
--- a/lib/connectors/kv-memory.js
+++ b/lib/connectors/kv-memory.js
@@ -64,19 +64,23 @@ KeyValueMemoryConnector.prototype._getStoreForModel = function(modelName) {
   return this._store[modelName];
 };
 
-KeyValueMemoryConnector.prototype.get =
-function(modelName, key, options, callback) {
+KeyValueMemoryConnector.prototype._removeIfExpired = function(modelName, key) {
   var store = this._getStoreForModel(modelName);
   var item = store[key];
-
   if (item && item.isExpired()) {
     debug('Removing expired key', key);
     delete store[key];
     item = undefined;
   }
+};
 
+KeyValueMemoryConnector.prototype.get =
+function(modelName, key, options, callback) {
+  this._removeIfExpired(modelName, key);
+
+  var store = this._getStoreForModel(modelName);
+  var item = store[key];
   var value = item ? item.value : null;
-
   debug('GET %j %j -> %s', modelName, key, value);
 
   if (/^buffer:/.test(value)) {
@@ -127,6 +131,29 @@ function(modelName, key, ttl, options, callback) {
   process.nextTick(callback);
 };
 
+KeyValueMemoryConnector.prototype.ttl =
+function(modelName, key, options, callback) {
+  this._removeIfExpired(modelName, key);
+
+  var store = this._getStoreForModel(modelName);
+
+  // key is unknown
+  if (!(key in store)) {
+    return process.nextTick(function() {
+      var err = new Error('Cannot get TTL for unknown key ' + key);
+      err.statusCode = 404;
+      callback(err);
+    });
+  }
+
+  var ttl = store[key].getTtl();
+  debug('TTL %j %j -> %s', modelName, key, ttl);
+
+  process.nextTick(function() {
+    callback(null, ttl);
+  });
+};
+
 KeyValueMemoryConnector.prototype.disconnect = function(callback) {
   if (this._cleanupTimer)
     clearInterval(this._cleanupTimer);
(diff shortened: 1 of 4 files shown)


superkhau:
@ritch PTAL too

lib/connectors/kv-memory.js
```diff
@@ -118,6 +122,28 @@ function(modelName, key, ttl, options, callback) {
   process.nextTick(callback);
 };
 
+KeyValueMemoryConnector.prototype.ttl =
+function(modelName, key, options, callback) {
+  this._removeIfExpired(modelName, key);
+
+  var store = this._getStoreForModel(modelName);
+  var error = new Error('Cannot get TTL for unknown key ' + key);
```
bajtos: I find this strange, why are you creating `error` even on the happy path when it's not used?

lib/connectors/kv-memory.js
```diff
@@ -118,6 +122,28 @@ function(modelName, key, ttl, options, callback) {
...
+
+  // key is unknown
+  if (!(key in store))
+    return process.nextTick(function() { callback(error); });
+
+  var ttl = store[key].getTtl();
+  debug('TTL %j %j -> %s', modelName, key, ttl);
+  // item is expired, but not deleted yet
+  if (ttl < 0)
```
bajtos: This should never happen, because you have already called `_removeIfExpired`, right?

I guess this can happen if we run `_removeIfExpired` just a fraction of millisecond before the key expires, and this `if (ttl<0)` right after the expiration, but I think the probability of that is so low, that it's not worth bothering.

bajtos:
@superkhau looks mostly good, I left two minor comments to address.

superkhau:
Connect to https://github.com/strongloop/loopback/issues/2525

lib/connectors/kv-memory.js
superkhau: Yeah, its probably an over-optimization. ;) Removing.

lib/connectors/kv-memory.js
superkhau: It was needed for if (ttl > 0), but gonna remove that now so I can move this into 133 again.

superkhau:
@bajtos Updated, gonna merge after CI passes.

lib/connectors/kv-memory.js
```diff
@@ -120,6 +124,27 @@ function(modelName, key, ttl, options, callback) {
...
+KeyValueMemoryConnector.prototype.ttl =
+function(modelName, key, options, callback) {
+  this._removeIfExpired(modelName, key);
+
+  var store = this._getStoreForModel(modelName);
+
+  // key is unknown
+  if (!(key in store)) {
+    return process.nextTick(function() {
+      callback(new Error('Cannot get TTL for unknown key: ' + key));
```
bajtos: Please set `err.statusCode = 404` so that the REST API returns `404 Not Found` when this error happens.

test/kvao/ttl.suite.js
```diff
@@ -0,0 +1,59 @@
...
+  describe('ttl', function() {
+    var CacheItem;
+    beforeEach(function unpackContext() {
+      CacheItem = helpers.givenCacheItem(dataSourceFactory);
+    });
+
+    it('returns an error when key does not exist', function() {
+      return CacheItem.ttl('key-does-not-exist').then(
+        function() { throw new Error('ttl() should have failed'); },
+        function(err) { err.message.should.match(/key-does-not-exist/); });
```
bajtos: Please check that `err.statusCode` is `404`.

bajtos:
@superkhau two more comments, feel free to address then in a new PR.

test/kvao/ttl.suite.js
```diff
@@ -0,0 +1,65 @@
...
+        return CacheItem.set('key-does-not-exist', 'a-value', 10)
+          .delay(20)
+          .then(function() {
+            return CacheItem.ttl('key-does-not-exist');
+          })
+          .then(
+            function() { throw new Error('ttl() should have failed'); },
+            function(err) {
+              err.message.should.match(/key-does-not-exist/);
+              err.should.have.property('statusCode', 404);
```
superkhau: @bajtos Added a 404 check here too in addition to the ones you mentioned.

superkhau:
@bajtos Applied your new comments, also added one more 404 check in the test suite. Waiting on CI again before merging.

bajtos:
Cool, the patch LGTM now.

superkhau:
All unrelated changes, merging.

PR #1033, opened 2016-08-08
KeyValueAccessObject: return 404 when expiring unknown key

bajtos:

@superkhau please review
cc @ritch


superkhau:
LGTM

PR #1028, opened 2016-08-05
Recursive search for near filter

etsuo:
Fix for #931

Merge this first, then the PR in [loopback-connector-mongodb](https://github.com/strongloop/loopback-connector-mongodb/pull/273)

PR #993, opened 2016-07-06
"Near" fix for Where clauses that contain and/or

cmoore4:
The near filter on geo searches currently only searches a first level
object, eg { "where": { "location": {"near": "30,-90"}}}.  For clauses
that contain an "and" or an "or" clause, the geo filter does not pick up
the operator, and you tend to get a SQL error back.  For instance, Postgres
forms this abomination:  WHERE ("location"point($1,$2)) AND ("name" ~\* $3).

This fix allows for queries with a format such as:
"where":{"and":[{"location":{"near":"29,-90"}},{"name":{"regexp":"/sean/i"}}]}

Additionally adds support for specifying multiple "near" filters.  If a geopoint is
within maxDistance of any of the near geopoints, it will be included in
the return result set.

NB:  This changes none of the actual distance searching functionality; that code has only been touched if necessary for linting.

This is a cleaned up re-submit of PR #876

PR #1036, opened 2016-08-08
Fix geo where clause not working with multiple filter's

### LINKED ISSUE #931: Geo where clause is not working with multiple filter's
In the example below, when I add rental car name to search criteria, it is expected to give 'Avis' rental cars with in 5 miles of given geo location. However, find function is returning all 'Avis' rental car locations with in database. If I just pass geo location, the search is working perfectly and only returning locations with in 5 mile radius. 

I tried 'memory' as well as 'mongo', both have same issue

```
RentalLocation.find({
  // find locations near the provided GeoPoint
  where: {and:[{name:'Avis Rent A Car'}, {geo: {near: here, maxDistance: max}}]},
  // paging
  skip: limit * page,
  limit: limit
}, fn);
```

etsuo:
Merge this first, then the PR in [loopback-connector-mongodb](https://github.com/strongloop/loopback-connector-mongodb/pull/273)

See #931

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,3 @@
 coverage
+node_modules
+support
--- a/lib/geo.js
+++ b/lib/geo.js
@@ -10,6 +10,8 @@ var assert = require('assert');
  */
 
 exports.nearFilter = function nearFilter(where) {
+  if (!where) return false;
+
   var keyPath = [];
   var result = false;
 
@@ -18,19 +20,18 @@ exports.nearFilter = function nearFilter(where) {
   function searchForNear(node) {
     if (node && typeof node === 'object') {
       Object.keys(node).forEach(function(key) {
-        var ex = node[key];
+        var nodeProperty = node[key];
         keyPath.push(key);
 
-        if (ex && ex.near) {
-          result = {
-            near: ex.near,
-            maxDistance: ex.maxDistance,
-            unit: ex.unit,
-            key: keyPath.slice(0),
-          };
-        } else {
-          searchForNear(ex);
+        if (!nodeProperty || !nodeProperty.near) {
+          return searchForNear(nodeProperty);
         }
+        result = {
+          near: nodeProperty.near,
+          maxDistance: nodeProperty.maxDistance,
+          unit: nodeProperty.unit,
+          key: keyPath.slice(0),
+        };
       });
     }
     keyPath.pop();
--- a/lib/geo.js
+++ b/lib/geo.js
@@ -10,23 +10,31 @@ var assert = require('assert');
  */
 
 exports.nearFilter = function nearFilter(where) {
+  var keyPath = [];
   var result = false;
 
-  if (where && typeof where === 'object') {
-    Object.keys(where).forEach(function(key) {
-      var ex = where[key];
+  searchForNear(where);
 
-      if (ex && ex.near) {
-        result = {
-          near: ex.near,
-          maxDistance: ex.maxDistance,
-          unit: ex.unit,
-          key: key,
-        };
-      }
-    });
+  function searchForNear(node) {
+    if (node && typeof node === 'object') {
+      Object.keys(node).forEach(function(key) {
+        var ex = node[key];
+        keyPath.push(key);
+
+        if (ex && ex.near) {
+          result = {
+            near: ex.near,
+            maxDistance: ex.maxDistance,
+            unit: ex.unit,
+            key: keyPath.slice(0),
+          };
+        } else {
+          searchForNear(ex);
+        }
+      });
+    }
+    keyPath.pop();
   }
-
   return result;
 };
 
@@ -278,4 +286,3 @@ function geoDistance(x1, y1, x2, y2, options) {
 
   return 2 * Math.asin(f) * EARTH_RADIUS[type];
 }
-
--- a/test/geo.test.js
+++ b/test/geo.test.js
@@ -6,9 +6,11 @@
 /*global describe,it*/
 /*jshint expr:true */
 
-require('should');
+var should = require('should');
 
 var GeoPoint = require('../lib/geo').GeoPoint;
+var nearFilter = require('../lib/geo').nearFilter;
+
 var DELTA = 0.0000001;
 
 describe('GeoPoint', function() {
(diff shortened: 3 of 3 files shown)


slnode:

slnode:

slnode:

slnode:

etsuo:
@jannyHou closed out #1028 since that wasn't against master. Updated the nearFilter. Let me know if it needs further refinement.

@0candy @Amir-61  - new PR opened ;)

.eslintignore
```diff
@@ -1 +1,3 @@
 coverage
+node_modules
+support
```
0candy: node_module should probably be ignored, but I think support should still be eslinted.

etsuo: @0candy unfortunately, `support/` fails with a huge number of lint problems on the `master` currently... suggestions?

lib/geo.js
```diff
@@ -10,23 +10,31 @@ var assert = require('assert');
...
+      Object.keys(node).forEach(function(key) {
+        var ex = node[key];
+        keyPath.push(key);
+
+        if (ex && ex.near) {
+          result = {
+            near: ex.near,
+            maxDistance: ex.maxDistance,
+            unit: ex.unit,
+            key: keyPath.slice(0),
```
Amir-61: I'm not very familiar with `key` I could not find any documentation about this; what is supposed to return? @0candy @superkhau @jannyHou

lib/geo.js
```diff
@@ -10,23 +10,31 @@ var assert = require('assert');
...
-          key: key,
-        };
-      }
-    });
+  searchForNear(where);
+
+  function searchForNear(node) {
+    if (node && typeof node === 'object') {
+      Object.keys(node).forEach(function(key) {
+        var ex = node[key];
```
Amir-61: We were using `ex` previously as well; not sure what it stands for though...

lib/geo.js
```diff
@@ -10,23 +10,31 @@ var assert = require('assert');
...
+
+        if (ex && ex.near) {
+          result = {
+            near: ex.near,
+            maxDistance: ex.maxDistance,
+            unit: ex.unit,
+            key: keyPath.slice(0),
+          };
+        } else {
+          searchForNear(ex);
```
Amir-61: Always go for short circuit; return first like this:

if (!ex || !ex.near) {
  searchForNear(ex);
}
result = {
  near: ex.near,
  maxDistance: ex.maxDistance,
  unit: ex.unit,
  key: keyPath.slice(0),
}

lib/geo.js
etsuo: it normally returns a string that indicates the property on the where clause that the `buildNearFilter` of the connector uses to mutate `near` (see MongoDB connector PR): https://github.com/strongloop/loopback-connector-mongodb/pull/273

The current implementation of nearFilter assumes the `where` filter is flat... so it doesn't support geo searches with `and` logic. In order to support `and`... the `nearFilter` needs to do a deep search to find the nested `near` property... thus the array so that `buildNearFilter` can find the nested `near` property.

lib/geo.js
```diff
@@ -278,4 +286,3 @@ function geoDistance(x1, y1, x2, y2, options) {
 
   return 2 * Math.asin(f) * EARTH_RADIUS[type];
 }
-
```
Amir-61: Nitpick: Please don not change unrelated code

etsuo: sorry - I didn't do that directly... atom must of... and I didn't catch it when reviewing the diff for the commit.

.eslintignore
Amir-61: @etsuo Thats interesting; do you see a folder like `nodeunit` under `support`?

Amir-61: Better question would be do you see anything other than `describe-operation-hooks.js` under `support`?

lib/geo.js
etsuo: dunno - I was going to change the name to something more meaningful but I didn't want to trigger a possible reason to get the PR rejected... I'm starting to exceed the amount of company time I can spend on this PR since we have a major client project that's behind schedule :/ - let me know if you have a better name, and I'll update the PR.

Amir-61:
@superkhau PTAL too

.eslintignore
etsuo: @Amir-61 here's what I have:
```
drwxr-xr-x   4 jpoveda  staff   136 Aug  8 13:52 .
drwxr-xr-x  26 jpoveda  staff   884 Aug  8 14:09 ..
-rw-r--r--   1 jpoveda  staff  6241 Aug  8 13:52 describe-operation-hooks.js
drwxr-xr-x  22 jpoveda  staff   748 Aug  8 10:56 nodeunit
```

lib/geo.js
Amir-61: No problem :-)

Amir-61:
@slnode test please

Amir-61:
Please make sure you have signed https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler

.eslintignore
0candy: @Amir-61 I only see that one file under `support`, is there suppose to be other stuff there?

etsuo:
@Amir-61 done... third time lol - is it forgetting me?

.eslintignore
Amir-61: Interesting enough I get that weird `nodeunit` folder generated as well; TBH, I do not know how it gets generated and that fails lots of eslint rules. @superkhau do you have any idea? I keep deleting that folder.

lib/geo.js
jannyHou: @etsuo It LGTM 👍 
But I have same question as @Amir-61 mentioned, what does `key` mean and why do we keep track of the keys in query?

@bajtos @raymondfeng @ritch @superkhau Could you help confirm? Thanks!

And the other thought is I am not sure is this the best way to handle recursive query, another way to handle it is checking operator `and` and analyze the subfilters recursively, like what we do in [`setScopeValuesFromWhere`](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/utils.js#L50)
But anyway your implementation also works fine to me 👍

etsuo: @jannyHou - I like that. I'd be happy to refactor it post merge if you want to assign an issue to me. It would also need to account for `or` right? I can do it in this PR, but ideally, if there are no other issues with it, it's be better for me to put that off to later since I have a project deadline kicking me in the face  :)

`key` is used by the connector's implementation of `buildNearFilter` to find the `near` property in the `where` filter - as far as I can tell, only the MongoDB connector uses this, so it shouldn't be a more widely breaking change.

.eslintignore
superkhau: No idea, try commenting out some test files and narrow it down.

Amir-61: We need to sort this out; anyway this should not be solved as this PR; lets move it out of the scope of this PR for now; @bajtos Any idea about this? I can take care of this after hearing your inputs...

lib/geo.js
etsuo: `ex` is renamed in my next commit.

lib/geo.js
etsuo: @Amir-61 implemented in my next commit.

etsuo:
Alright... suggested updates committed... now the PR linter is reporting that a committer should rebase: `INFO: external contribution, a committer should probably rebase for them`. Thx.

.eslintignore
bajtos: AFAICT, this code is based on a commit from "Apr 5, 2016". Let's start with rebasing it on top of the current master, that will hopefully fix these issues.

test/geo.test.js
```diff
@@ -142,3 +144,188 @@ describe('GeoPoint', function() {
...
+      },
+    });
+
+    should.exist(result);
+    result.near.lat.should.equal(0);
+    result.near.lng.should.equal(0);
+    should.not.exist(result.maxDistance);
+    should.not.exist(result.unit);
+    result.key.length.should.be.exactly(1);
+    result.key[0].should.be.exactly('location');
```
bajtos: The assertion failures would produce rather unhelpful messages. Can we simplify to something like the following?

```js
should(result).eql({
  near: {lat: 0, lng: 0},
  key: ['location']
});
```

bajtos: The same comment applies to other tests below too.

bajtos:
Hello, what's the status of this patch? Are you @etsuo still keen to get it finished and landed?

etsuo:
hi @bajtos, unfortunately, I'm buried at the moment and can't focus on this patch. My apologies.

F3L1X79:
Isn't this PR correcting the same thing? https://github.com/strongloop/loopback-datasource-juggler/pull/993

ebarault:
hi, could you check if these PR fix your issue ?
strongloop/loopback-datasource-juggler#1216
strongloop/loopback-connector-mongodb#322

PR #1039, opened 2016-08-09
kv-memory: fix crash in regular cleanup

bajtos:
Fix bug in "_setupRegularCleanup()" where the interval callback was trying to access an object that has been garbage-collected in the meantime.

Symptoms:

```
/Users/bajtos/src/loopback/loopback/node_modules/loopback-datasource-juggler/lib/connectors/kv-memory.js:35
    function() { self._removeExpiredItems(); },
                      ^

TypeError: self._removeExpiredItems is not a function
    at Timeout.<anonymous> (/Users/bajtos/src/loopback/loopback/node_modules/loopback-datasource-juggler/lib/connectors/kv-memory.js:35:23)
    at Timeout._repeat (/Users/bajtos/src/loopback/loopback/node_modules/continuation-local-storage/node_modules/async-listener/glue.js:188:31)
    at Timeout.wrapper [as _onTimeout] (timers.js:417:11)
    at Timer.unrefdHandle (timers.js:454:14)
```

@superkhau please review

### DIFF
--- a/lib/connectors/kv-memory.js
+++ b/lib/connectors/kv-memory.js
@@ -31,8 +31,15 @@ KeyValueMemoryConnector.prototype._setupRegularCleanup = function() {
   // key expiration too, the scheduled cleanup is merely a performance
   // optimization.
   var self = this;
-  this._cleanupTimer = setInterval(
-    function() { self._removeExpiredItems(); },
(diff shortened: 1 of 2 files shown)


superkhau:
LGTM

PR #1043, opened 2016-08-11
prevent duplicate generated foreignKeys

alexkolson:
The belongsTo relation definition should not set a foreignKey if the model has an opposite hasMany relation because hasMany will have already set a foreignKey property.

This should prevent the behavior seen in strongloop/loopback#2587.

This is a rough draft and I would very much appreciate feedback.

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -1247,7 +1247,7 @@ RelationDefinition.belongsTo = function(modelFrom, modelTo, params) {
     pkName = params.primaryKey || modelTo.dataSource.idName(modelTo.modelName) || 'id';
     relationName = params.as || i8n.camelize(modelTo.modelName, true);
     fk = params.foreignKey || relationName + 'Id';
-    
+ 
     // Check for any matching matching mirror relations and don't inject a foreign key property
     // into the model if any are found.
     // Why? Because The mirrored relation defintion will have already ensured a foreign key exists so if BelongsTo
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -1247,8 +1247,17 @@ RelationDefinition.belongsTo = function(modelFrom, modelTo, params) {
     pkName = params.primaryKey || modelTo.dataSource.idName(modelTo.modelName) || 'id';
     relationName = params.as || i8n.camelize(modelTo.modelName, true);
     fk = params.foreignKey || relationName + 'Id';
+    
+    // Check for any matching matching mirror relations and don't inject a foreign key property
+    // into the model if any are found.
+    // Why? Because The mirrored relation defintion will have already ensured a foreign key exists so if BelongsTo
+    // injects a foreignkey it will be a duplicate, representing the same relation.
+    var matchingHasManyRelations = Object.keys(modelTo.relations)
+      .filter(function isMirroredRelation(rk) { return modelTo.relations[rk].modelTo.modelName === modelFrom.modelName });
 
-    modelFrom.dataSource.defineForeignKey(modelFrom.modelName, fk, modelTo.modelName, pkName);
+    if (!matchingHasManyRelations.length) {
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

gunjpan:
@Amir-61 / @jannyHou : Could you review this PR whenever you have time and help @alexkolson get this PR landed. Thank you :)

raymondfeng:
@alexkolson Thank you for the PR. I have two concerns:

1. To ensure `hasOne` and `belongsTo` relations refer to both sides of the same connection between two models, you probably should make sure same FK is used. In your case, the naming convention based approach won't work as it derives two FKs. You should use `foreignKey` property on the relation definition to make it explicit.

2. If you want to intuitively avoid duplicate FKs, the PR is missing one extra step. It should modify the relation definition so that the `foreignKey` should point to the detected one after checking `hasOne` or `hasMany` from `belongsTo` side. The PR should also handle from `hasOne`/`hasMany` to `belongsTo` side too.

alexkolson:
@gunjpan @raymondfeng  First off, thank you for even taking the time to give this PR a look. I imagine it's a lot of (sometimes thankless) work to review community contributions and I appreciate your time.

Second, responding to the concerns:

>  To ensure hasOne and belongsTo relations refer to both sides of the same connection between two models, you probably should make sure same FK is used. In your case, the naming convention based approach won't work as it derives two FKs. You should use foreignKey property on the relation definition to make it explicit

Agreed 100%. The best approach is to explicitly define a foreign key for each relation in the model definition json file. Perhaps all that is needed is some documentation stating very clearly that this behavior can happen if you do not explicitly define the same foreign key for both sides of the relationship. Although I am inclined to think that adding a best-effort attempt to avoid creating duplicate foreign keys would be helpful in addition to the documentation.

> If you want to intuitively avoid duplicate FKs, the PR is missing one extra step. It should modify the relation definition so that the foreignKey should point to the detected one after checking hasOne or hasMany from belongsTo side. The PR should also handle from hasOne/hasMany to belongsTo side too.

Understood. If it is decided that the current approach is the correct one I will add these missing steps.

A harsher approach that would be a breaking change might be to throw an error in the boot phase if conflicting foreign keys are detected and/or modifying the way hasMany/hasOne resolves the foreign key name, so that instead of using the relationName + id it uses modelTo.modelName + id. Just some additional proposals as to how this might be accomplished.

Thanks again for your time and I look forward to your further feedback.

alexkolson:
Shall I add the support that @raymondfeng [mentioned in point 2](https://github.com/strongloop/loopback-datasource-juggler/pull/1043#issuecomment-239923256) or is this something that should be implemented in a different way?

gunjpan:
@alexkolson : It would be great if you can implement it and get Raymond's feedback on what you have there after it.

alexkolson:
@gunjpan Roger that! Will do.

ssh24:
@alexkolson Any updates on this PR? I would like to proceed with the PR. Please rebase your branch. Just as a side note, there will be merge conflicts on [test/relations.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/relations.test.js) so please take a look at it.

alexkolson:
@ssh24 Ah! I assumed it was just rejected! I will rebase my branch and see where things are at.

alexkolson:
@ssh24 Just as a note, I will first be able to take a look next Monday.

ssh24:
@alexkolson No worries. Take a look it next week once you have time.

alexkolson:
@ssh24 Looks like I won't be able to get this for quite some time as other things have come up. It's probably best if this PR is taken over by someone else or if it is closed entirely.

slnode:

slnode:

slnode:

slnode:

ssh24:
@alexkolson Closing this PR in that case. If anyone is willing to pick it up, they can do so by continuing this or creating a new one.

PR #1044, opened 2016-08-12
test/kvao: setup global.Promise on Node v0.10 [2.x only]

bajtos:
When running the tests on Node v0.10, setup `global.Promise` with Bluebird implementation. That way we don't have to run this setup in the test suite of each kv connector.

@superkhau PTAL

Related: https://github.com/strongloop/loopback-connector-kv-redis/pull/5


superkhau:
LGTM

PR #1047, opened 2016-08-13
Rename get test suite to match other test suites

superkhau:
@bajtos PTAL


bajtos:
The test suite tests both `get` and `set`, it's not possible to test one without the other (at least I am not aware of any way).

I am closing this patch as rejected, let me know if you disagree.

superkhau:
> it's not possible to test one without the other (at least I am not aware of any way).

I do disagree. I'm not arguing testing one with the other or anything related to that. The point I wanted to bring up is you have 3 files:

```
expire.suite.js
get-test.suite.js
ttl.suite.js
```

I'm proposing we should name `get-test.suite.js` to get-set.suite.js` to make things consistent OR rename the files to:

```
expire-test.suite.js
get-set-test.suite.js
ttl-test.suite.js
```

superkhau:
I'm update the PR to reflect one choice, LMK if you want to append the `-test` instead.

bajtos:
Oh, must have been kind of blind. Now I see what you mean, and totally agree with the new name `get-set.suite.js`.

:shipit:

bajtos:
@superkhau I landed the change on master, could you please back-port it to 2.x on your own?

superkhau:
Backport at #1054

PR #1049, opened 2016-08-16
kvao: add iterateKeys() and keys()

bajtos:
Add a core implementation of KVAO.iterateKeys() which returns an
AsyncIterator, inspired by
 - https://github.com/tc39/proposal-async-iteration
 - https://www.npmjs.com/package/async-iterators

This way we can safely iterate even large sets of data.

Also add KVAO.keys(), a sugar API converting the result of iterateKeys()
into a single array.

@superkhau @ritch please review

~~The patch is describing `filter.match` option that's not implemented. I'll either add it to this patch or submit a new one, depending on the speed of code reviews.~~

Connect to strongloop/loopback#2592

### DIFF
--- a/lib/kvao/iterate-keys.js
+++ b/lib/kvao/iterate-keys.js
@@ -1,0 +1,35 @@
+'use strict';
+
+var assert = require('assert');
+var utils = require('../utils');
+
+/**
+ * Asynchronously iterate all keys.
+ *
+ * @param {Object} filter An optional filter object with the following
+ * properties:
+ *  - `match` - glob string to use to filter returned keys, e.g. 'userid.*'
+ * @param {Object} options
+ *
+ * @returns {AsyncIterator} An object implementing "next(cb) -> Promise"
+ * function that can be used to iterate all keys.
+ *
+ * @header KVAO.iterateKeys(filter)
+ */
+module.exports = function keyValueIterateKeys(filter, options) {
+  filter = filter || {};
+  options = options || {};
+
+  assert(typeof filter === 'object', 'filter must be an object');
+  assert(typeof options === 'object', 'options must be an object');
+
+  var iter = this.getConnector().iterateKeys(this.modelName, filter, options);
+  // promisify the returned iterator
+  return {
+    next: function(callback) {
+      callback = callback || utils.createPromiseCallback();
+      iter.next(callback);
+      return callback.promise;
+    },
+  };
+};
--- a/lib/kvao/iterate-keys.js
+++ b/lib/kvao/iterate-keys.js
@@ -9,6 +9,10 @@ var utils = require('../utils');
  * @param {Object} filter An optional filter object with the following
  * properties:
  *  - `match` - glob string to use to filter returned keys, e.g. 'userid.*'
+ *    All connectors are required to support `*` and `?`.
+ *    They may also support additional special characters that are specific
+ *    to the backing store.
+ *
  * @param {Object} options
  *
  * @returns {AsyncIterator} An object implementing "next(cb) -> Promise"
--- a/lib/kvao/keys.js
+++ b/lib/kvao/keys.js
@@ -1,0 +1,57 @@
+'use strict';
+
+var assert = require('assert');
+var utils = require('../utils');
+
+/**
+ * Get all keys.
+ *
+ * **NOTE**
+ * Building an in-memory array of all keys may be expensive.
+ * Consider using `iterateKeys` instead.
+ *
+ * @param {Object} filter An optional filter object with the following
+ * properties:
+ *  - `match` - glob string to use to filter returned keys, e.g. 'userid.*'
+ * @param {Object} options
+ * @callback callback
+ * @param {Error=} err
+ * @param {[String]} keys The list of keys.
+ *
+ * @promise
+ *
+ * @header KVAO.keys(filter, callback)
+ */
+module.exports = function keyValueKeys(filter, options, callback) {
+  if (callback === undefined) {
+    if (typeof options === 'function') {
+      callback = options;
+      options = undefined;
+    } else if (options === undefined && typeof filter === 'function') {
+      callback = filter;
+      filter = undefined;
+    }
+  }
+
+  filter = filter || {};
+  options = options || {};
+
+  assert(typeof filter === 'object', 'filter must be an object');
+  assert(typeof options === 'object', 'options must be an object');
+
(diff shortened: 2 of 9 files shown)


bajtos:
Here is the reason for adding `iterateKeys` as the core method:

http://redis.io/commands/KEYS
> Warning: consider KEYS as a command that should only be used in production environments with extreme care. It may ruin performance when it is executed against large databases. This command is intended for debugging and special operations, such as changing your keyspace layout. Don't use KEYS in your regular application code. If you're looking for a way to find keys in a subset of your keyspace, consider using [SCAN](http://redis.io/commands/scan) or sets.

The concept of AsyncIterator is already used by [loopback-sd-iterator-mixin](https://www.npmjs.com/package/loopback-ds-iterator-mixin):
> ```js
var iterator = this.Item.iterate(query)
iterator.next()
  .then(function (item) {
    expect(item.name).to.equal('Item1');
    return iterator.next();
  })
  .then(function (item) {
    expect(item.name).to.equal('Item2');
    return iterator.next();
  })
```

File: lib/kvao/iterate-keys.js [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+ * @param {Object} options
+ *
+ * @returns {AsyncIterator} An object implementing "next(cb) -> Promise"
+ * function that can be used to iterate all keys.
+ *
+ * @header KVAO.iterateKeys(filter)
+ */
+module.exports = function keyValueIterateKeys(filter, options) {
+  if (!filter) filter = {};
+  if (!options) options = {};
```
superkhau: The following is subjective. Yours works too, but maybe this is better?

```
filter = filter || {};
options = options || {};

File: lib/kvao/keys.js [hunk 2]
```diff
@@ -0,0 +1,57 @@
...
+      callback = options;
+      options = undefined;
+    } else if (options === undefined && typeof filter === 'function') {
+      callback = filter;
+      filter = undefined;
+    }
+  }
+
+  if (!filter) filter = {};
+  if (!options) options = {};
```
superkhau: ditto

test/kvao/iterate-keys.suite.js
```diff
@@ -0,0 +1,45 @@
...
+
+    it('returns AsyncIterator supporting Promises', function() {
+      var iterator;
+      return helpers.givenKeys(CacheItem, ['key'])
+        .then(function() {
+          iterator = CacheItem.iterateKeys();
+          return iterator.next();
+        })
+        .then(function(key) {
+          should(key).equal('key');
```
superkhau: Does `key.should.equal('key');` not work in this case?

test/kvao/iterate-keys.suite.js
```diff
@@ -0,0 +1,45 @@
...
+          return iterator.next();
+        })
+        .then(function(key) {
+          should(key).be.undefined;
```
superkhau: This too, `should.not.exist(key)`?

superkhau:
Minor comments that don't really need addressing (optional), LGTM.

test/kvao/iterate-keys.suite.js
bajtos: when `key` happens to be null, then you will get an ugly assertion error.

test/kvao/iterate-keys.suite.js
bajtos: The value has to be `undefined`. I'll add a comment and fix the assertion. Apparently `be.undefined` does not exist and nobody noticed because of [Beware of asserting on property access](https://github.com/moll/js-must#beware-of-libraries-that-assert-on-property-access).

bajtos:
Addresses most of comments and added implementation of `filter.match`.

@superkhau @ritch PTAL again.

test/kvao/iterate-keys.suite.js
superkhau: Weird lol, good to know.

test/kvao/keys.suite.js
```diff
@@ -0,0 +1,105 @@
...
+          'hello',
+          'hxllo',
+          'hllo',
+          'heeello',
+          'foo',
+          'bar',
+        ]);
+      });
+
+      it('supports "h?llo"', function() {
```
superkhau: `supports ? operator (ie. "h?llo")`

test/kvao/keys.suite.js
```diff
@@ -0,0 +1,105 @@
...
+        return CacheItem.sortedKeys({ match: 'h?llo' }).then(function(keys) {
+          should(keys).eql(['hallo', 'hello', 'hxllo']);
+        });
+      });
+
+      it('supports "h*llo"', function() {
```
superkhau: supports * operator (ie. "h*llo")

superkhau:
Two minor comments, LGTM.

lib/kvao/iterate-keys.js
```diff
@@ -0,0 +1,39 @@
...
+
+var assert = require('assert');
+var utils = require('../utils');
+
+/**
+ * Asynchronously iterate all keys.
+ *
+ * @param {Object} filter An optional filter object with the following
+ * properties:
+ *  - `match` - glob string to use to filter returned keys, e.g. 'userid.*'
```
ritch: This looks like a new doc syntax to me... I think we usually do:

```
@param {Object} filter An optional filter object with the following
@param {String} filter.match Glob string to use to filter returned keys, e.g. 'userid.*'
```

lib/kvao/iterate-keys.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+var assert = require('assert');
+var utils = require('../utils');
+
+/**
+ * Asynchronously iterate all keys.
```
ritch: We should add an example. `AsyncIterator` isn't a well known thing.

bajtos: > We should add an example. `AsyncIterator` isn't a well known thing.

Makes sense. @superkhau could you please take care of this as part of your work on API docs?

lib/kvao/iterate-keys.js
bajtos: > This looks like a new doc syntax to me... I think we usually do:
> ```
@param {Object} filter An optional filter object with the following
@param {String} filter.match Glob string to use to filter returned keys, e.g. 'userid.*'
```

@superkhau could you please take care of this as part of your work on API docs?

PR #1061, opened 2016-08-18
Notify find 2.x

Amir-61:
Notify find 2.x

We already implemented `notify` for `find` on master branch; however the implementation of `find` is very different from the one on `2.x`, hence a simple cherry-pick was not going to work; this patch implements `notify` for `2.x`

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1633,15 +1633,25 @@ DataAccessObject.find = function find(query, options, cb) {
       // using all documents
       // TODO [fabien] use default scope here?
 
-      var context = {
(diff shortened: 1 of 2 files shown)


Amir-61:
@slnode test please

bajtos:
FWIW, here is the original pull request against the master branch: https://github.com/strongloop/loopback-datasource-juggler/pull/898

bajtos:
The patch LGTM, :shipit:

Amir-61:
Implement `notify` for find method [2.x]

PR #1063, opened 2016-08-24
test/memory: remove dummy findOrCreate impl

bajtos:
Let the operation-hook tests use the real implementation, now that we have it in place.

@Amir-61 please review


bajtos:
findOrCreate implementation: https://github.com/strongloop/loopback-datasource-juggler/blob/5b464842779d8bedd49dcd766c69645bd800dc64/lib/connectors/memory.js#L285

Amir-61:
LGTM

Amir-61:
Sorry Miroslav I introduced some merge conflicts; please fix them before landing.

PR #1068, opened 2016-08-26
Disable `strict` for a few files

Amir-61:
*Disable `strict` for a few files to disappear warnings

/to: @bajtos, @superkhau


bajtos:
LGTM.

PR #1069, opened 2016-08-29
[SEMVER-MAJOR] Remove updateOnLoad

Amir-61:
- Change default of Model.settings.updateOnLoad to true

Connect to https://github.com/strongloop/loopback/issues/2417

/to: @bajtos

### DIFF
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -102,3 +102,8 @@ bulk `updateOrCreate`, you could use `async.each` or `Promise.all` to
 repeatedly call `updateOfCreate` for each instance.
 
 See [related code change](https://github.com/strongloop/loopback-datasource-juggler/pull/889) here.
+
++## Remove `Model.settings.updateOnLoad`
+ +
+ +See [related code change](https://github.com/strongloop/loopback-datasource-juggler/pull/1069)
+ +for more details.
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -3666,7 +3666,7 @@ describe('relations', function() {
       p.isNewRecord().should.be.true;
       p.passportItem.create({name: 'Fredric'}, function(err, passport) {
         should.not.exist(err);
-        p.passport.should.equal(passport);
+        p.passport.toObject().should.eql({name: 'Fredric'});
         p.isNewRecord().should.be.false;
         done();
       });
@@ -3790,7 +3790,7 @@ describe('relations', function() {
           return Person.findById(personId);
         })
         .then(function(person) {
-          person.passportItem().toObject().should.eql({name: 'Jim'});
+          person.passportItem().toObject().should.eql({name: 'Mitsos'});
           // restore original hasPk
           Passport.definition.hasPK = originalHasPK;
           done();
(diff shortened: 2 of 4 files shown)


Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

test/relations.test.js
```diff
@@ -3790,7 +3790,7 @@ describe('relations', function() {
           return Person.findById(personId);
         })
         .then(function(person) {
-          person.passportItem().toObject().should.eql({name: 'Jim'});
+          person.passportItem().toObject().should.eql({name: 'Mitsos'});
```
Amir-61: I'm very familiar with relations; Is this test valid? I believe name should be `Jim`; any clue/direction about this?

3.0-RELEASE-NOTES.md
```diff
@@ -102,3 +102,8 @@ bulk `updateOrCreate`, you could use `async.each` or `Promise.all` to
 repeatedly call `updateOfCreate` for each instance.
 
 See [related code change](https://github.com/strongloop/loopback-datasource-juggler/pull/889) here.
+
++## Change default of `Model.settings.updateOnLoad` to true
```
bajtos: This is not correct description. AFAICT from the code change, the option is removed entirely.

test/relations.test.js
bajtos: Yes, I think the name should be 'Jim'. 

My guess is that now with `updateOnLoad` is gone, `passport.save` updates the parent instance`person` in a different way, perhaps it does not update all properties as need. If I am correct, then this is a latent bug in the implementation of embedded relations that you have just discovered :)

You can also use `git blame` (or the github UI) to see who wrote this test and ask that person directly.

Amir-61: >You can also use git blame (or the github UI) to see who wrote this test and ask that person directly.

Hi @bajtos,

It seems this is written by @mitsos1os (community contribution) and got reviewed and landed by you; please see: https://github.com/strongloop/loopback-datasource-juggler/pull/949
and https://github.com/strongloop/loopback-datasource-juggler/commit/bfbdd15f1df00a362df13232a2240a463b7c2263

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@slnode test please

test/relations.test.js
bajtos: I see, thank you for looking up the original patch. My conclusion is that the name should indeed be "Jim".

 Honestly, I don't see any obvious reason explaining why you are seeing the original value "Mitsos" now.

mitsos1os: Hi, I just wanted to note for the test. The correct operation would be indeed to see `Jim` which was implemented in my pull request, since you alter the created embedded object that is returned by the create of passport Item and then save it.

This functionality was implemented by adding `EmbedsOne.prototype.prepareEmbeddedInstance` function that calls the parents update when embedded model save requested in a similar way that EmbedsMany relations work.
[See](https://github.com/strongloop/loopback-datasource-juggler/pull/949/files#diff-20531c2f86840f4a537a4db9f742b0e2R2084)

bajtos:
This isn't going to happen at least until v4.0, let's close this pull request for now.

PR #1073, opened 2016-09-02
Fix failures of upsertWithWhere

Amir-61:
* Fix failures of upsertWithWhere

Forwardport of #1052

/cc: @bajtos @superkhau


Amir-61:
@superkhau @bajtos On the second thought I don't think if you need to review it.

PR #1074, opened 2016-09-02
Remove expired item before executing expire

superkhau:
The expire feature is falsely returning 204 instead of 404 because it is
not removing expired items before execution.

Backport of #1072

PR #1072, opened 2016-09-02
Remove expired items before executing expire

superkhau:
The expire feature is falsely returning 204 instead of 404 because it is not
removing expired items before execution.

PR #1075, opened 2016-09-02
Fix test case for expire

superkhau:
Forwardport of #1074. #1072 was missing Promise.resolve before using delay(20)
and causing Travis to fail on 2.x. This PR makes the test code consistent on
both master and 2.x.

@bajtos Merging without review as this is a simple fix for #1072 which has already been reviewed.


superkhau:
@davidcheung Can you follow up with the angular/grunt failures? CI is having issues with installing bower, etc but I'll leave it to you since you're the FA owner.

PR #1077, opened 2016-09-03
Refactor TTL test suite

superkhau:
Simplify TTL tests.

PR #1078, opened 2016-09-04
Refactor TTL tests for KV memory connector

superkhau:
Reduce the complixity in the TTL test suite for the kv-memory connector to
get CI to pass. This change makes 3.x tests equivalent to 2.x tests (which are already merged and passing on CI).

- Tests messages were reworded and reorganized to make the scope of each test clearer
- Some of the code refactoring was done to make CI pass (the original tests were failing due to some expiry issue)

Forwardport of #1077


superkhau:
Unrelated failures, merging.

cc @bajtos

PR #1080, opened 2016-09-05
Fix #1079 - Polymorphic hasMany inverse relation in include filter

### LINKED ISSUE #1079: Polymorphic hasMany inverse relation doesn't work on include filter
Using an include filter where the included relation is a polymorphic hasMany with invert set to true does not correctly take into account the invert option when figuring out which model to query, causing no records for the relation to return.

Given these model definitions:

``` js
Picture = db.define('Picture', { name: String });
Author = db.define('Author', { name: String });
PictureLink = db.define('PictureLink', {});

Author.hasMany(Picture, { through: PictureLink, polymorphic: 'imageable', invert: true });
Picture.hasMany(Author, { through: PictureLink, polymorphic: 'imageable' });
```

When running the below code snippet an empty array will always return for pictures, even if it actually contains valid entries:

``` js
Author.findOne({ include: 'pictures' }...); // this will not work, pictures will always be empty
```

However when querying the relation directly like this, things work fine:

``` js
// Given an author instance...
author.pictures(); // this will work, pictures will include all pictures
```

Here is the failing test case I wrote which shows this behaviour. Please correct me if I have misunderstood how the inverse relations should work:

``` js
  describe('polymorphic hasMany', function() {
    before(function(done) {
      Picture = db.define('Picture', { name: String });
      Author = db.define('Author', { name: String });
      PictureLink = db.define('PictureLink', {});

      Author.hasMany(Picture, { through: PictureLink, polymorphic: 'imageable', invert: true });
      Picture.hasMany(Author, { through: PictureLink, polymorphic: 'imageable' });

      db.automigrate(['Picture', 'Author', 'PictureLink'], done);
    });

    it('should properly query through an inverted relationship', function(done) {
      Author.create({ name: 'Steve' }, function(err, author) {
        author.pictures.create({ name: 'Steve pic 1' }, function(err, pic) {
          Author.findOne({ include: 'pictures' }, function(err, author) {
            if (err) { return done(err); }
            assert.equal(author.pictures().length, 1);
            assert.equal(author.pictures()[0].name, 'Steve pic 1');
            done();
          });
        });
      });
    });
  });
```

This issue might be related to #927, but I'm not sure.

alexpls:
This is a proposed fix to #1079. For more details of the issue please check out the issue.

Note I've targeted this fix to the `2.x-latest` branch. I'm not fully familiar with the v3 source code, but I believe this fix should be applied to the v3 branch as well. Happy to raise another PR for this if you agree.

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -325,10 +325,14 @@ Inclusion.include = function(objects, include, options, cb) {
         inq: uniq(sourceIds),
       };
       if (polymorphic) {
+        var throughModel = polymorphic.invert ?
+          relation.modelTo :
+          relation.modelFrom;
+
         //handle polymorphic hasMany (reverse) in which case we need to filter
         //by discriminator to filter other types
         throughFilter.where[polymorphic.discriminator] =
-          relation.modelFrom.definition.name;
+            throughModel.definition.name;
       }
       /**
        * 1st DB Call of 2 step process. Get through model objects first
@@ -927,4 +931,3 @@ Inclusion.include = function(objects, include, options, cb) {
     }
   }
 };
-
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -5503,4 +5503,32 @@ describe('relations', function() {
       }).should.throw('Invalid relation name: trigger');
     });
   });
+
+  describe('polymorphic hasMany', function() {
+    before(function(done) {
+      Picture = db.define('Picture', {name: String});
+      Author = db.define('Author', {name: String});
+      PictureLink = db.define('PictureLink', {});
+
+      Author.hasMany(Picture, {through: PictureLink, polymorphic: 'imageable', invert: true});
+      Picture.hasMany(Author, {through: PictureLink, polymorphic: 'imageable'});
+
+      db.automigrate(['Picture', 'Author', 'PictureLink'], done);
+    });
+
+    it('should properly query through an inverted relationship', function(done) {
+      Author.create({name: 'Steve'}, function(err, author) {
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

jannyHou:
@slnode test please

Amir-61:
@slnode test please

alexpls:
@slnode test please

devpascoe:
Really need this 😆❤️

jannyHou:
@slnode test please

jannyHou:
@alexpls The change LGTM. while the test case should be in file `test/relation.test.js`:https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/relations.test.js could you move it to the right place? Thanks.

jannyHou:
@slnode test please

alexpls:
@jannyHou Thanks! I've updated the PR with these changes:

- Moved the test case to `test/relation.test.js`
- Rebased to the tip of the `2.x-latest` branch
- Resolved some eslint warnings

All these changes have been squashed into one commit so it keeps the git history clean.

Let me know if anything else needs to be changed!

jannyHou:
@alexpls Cheers the pr landed but sorry I made a mistake...the pr should be checked into 2.x instead of 2.x-latest....could you backport it to 2.x? Or if you don't mind I can do the backport for you.

alexpls:
@jannyHou I'm happy for you to do the backport - thanks!

PR #1084, opened 2016-09-07
[SEMVER-MAJOR] Strict mode cleanup

davidcheung:
connect to strongloop/loopback/issues/1299

### DIFF
--- a/3.0-RELEASE-NOTES.md
+++ b/3.0-RELEASE-NOTES.md
@@ -102,3 +102,55 @@ bulk `updateOrCreate`, you could use `async.each` or `Promise.all` to
 repeatedly call `updateOfCreate` for each instance.
 
 See [related code change](https://github.com/strongloop/loopback-datasource-juggler/pull/889) here.
+
+## Removing strict: validate and strict: throw
+Given a user model definition as follow:
+```js
+ds.define('User', {
+  name: String, required: false,
+  age: Number, required: false
+})
+
+var johndoe = new User({ name: 'John doe', age: 15, gender: 'm'});
+```
+- #### strict: false
+  behavior will remain the same and unknown properties will be saved along with other properties
+  ```js
+  //johndoe would be created as
+  {
+    id: 1,
+    name: 'John doe',
+    age: 15,
+    geneder: 'm'
+  }
+  ```
+
+- #### strict: `'validate'` and strict: `'throw'`
+  are removed in favor of `strict: true`,  and because non-empty string
+  resolves as true, they will share the same behavior as `strict: true`
+
+- #### strict: `true`'s behavior change:
+
+    **Previously:** (2.x)
+
+    `strict:true` would silently removing unknown properties, and User instance will be created with unknown properties discarded
+
+    ```js
+    johndoe.isValid(); //true
+    //johndoe would be created as
+    {
+      id: 1,
+      name: 'John doe',
+      age: 15
+    }
+    ```
+
+    **New Behavior:** (3.x)
+
+    ```js
+    johndoe.isValid(); //false
+    //johndoe would NOT be created and will result in the following error
+    //ValidationError: The User instance is not valid. Details: gender is not defined in the model (value: 'm').
+    ```
+
+See [related code change](https://github.com/strongloop/loopback-datasource-juggler/pull/1084) here.
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -93,10 +93,7 @@ function applyStrictCheck(model, strict, data, inst, cb) {
     key = keys[i];
     if (props[key]) {
       result[key] = data[key];
-    } else if (strict === 'throw') {
-      cb(new Error(g.f('Unknown property: %s', key)));
-      return;
-    } else if (strict === 'validate') {
+    } else if (strict) {
       inst.__unknownProperties.push(key);
     }
   }
--- a/lib/model.js
+++ b/lib/model.js
@@ -94,6 +94,10 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
 
   if (strict === undefined) {
     strict = ctor.definition.settings.strict;
+  } else if (strict === 'throw') {
+    g.warn('Warning: Model %s, {{strict mode: `throw`}} has been removed, ' +
+      'please use {{`strict: true`}} instead, which returns' +
+      '{{`Validation Error`}} for unknown properties,', ctor.modelName);
   }
 
   var persistUndefinedAsNull = ctor.definition.settings.persistUndefinedAsNull;
@@ -141,7 +145,7 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
       },
     });
 
-    if (strict === 'validate') {
+    if (strict) {
       Object.defineProperty(this, '__unknownProperties', {
         writable: true,
         enumerable: false,
@@ -155,7 +159,7 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
     this.__dataSource = options.dataSource;
     this.__strict = strict;
     this.__persisted = false;
-    if (strict === 'validate') {
+    if (strict) {
       this.__unknownProperties = [];
     }
   }
@@ -243,9 +247,7 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
             this.constructor.modelName, p
           ));
         }
-      } else if (strict === 'throw') {
-        throw new Error(g.f('Unknown property: %s', p));
-      } else if (strict === 'validate') {
+      } else {
         this.__unknownProperties.push(p);
       }
     }
--- a/test/datatype.test.js
+++ b/test/datatype.test.js
@@ -192,6 +192,7 @@ describe('datatypes', function() {
       TestModel = db.define(
         'TestModel',
         {
+          name: {type: String, required: false},
           desc: {type: String, required: false},
           stars: {type: Number, required: false},
         },
@@ -220,12 +221,12 @@ describe('datatypes', function() {
 
     it('should convert property value undefined to null', function(done) {
       var EXPECTED = {desc: null, extra: null};
+      var data = {desc: undefined, extra: undefined};
       if (isStrict) {
         // SQL-based connectors don't support dynamic properties
         delete EXPECTED.extra;
+        delete data.extra;
       }
-
-      var data = {desc: undefined, extra: undefined};
       TestModel.create(data, function(err, created) {
         if (err) return done(err);
 
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -544,41 +544,83 @@ describe('manipulation', function() {
         });
     });
 
-    it('should ignore unknown attributes when strict: true', function(done) {
-      // Using {foo: 'bar'} only causes dependent test failures due to the
-      // stripping of object properties when in strict mode (ie. {foo: 'bar'}
-      // changes to '{}' and breaks other tests
-      person.updateAttributes({name: 'John', foo: 'bar'},
-        function(err, p) {
-          if (err) return done(err);
-          should.not.exist(p.foo);
-          Person.findById(p.id, function(e, p) {
-            if (e) return done(e);
-            should.not.exist(p.foo);
-            done();
-          });
+    it('should discard undefined values before strict validation',
+      function(done) {
+        Person.definition.settings.strict = true;
+        Person.findById(person.id, function(err, p) {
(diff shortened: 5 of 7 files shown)


test/datatype.test.js
```diff
@@ -206,7 +206,7 @@ describe('datatypes', function() {
 
     it('should set missing optional properties to null', function(done) {
       var EXPECTED = {desc: null, stars: null};
-      TestModel.create({name: 'a-test-name'}, function(err, created) {
+      TestModel.create({}, function(err, created) {
```
davidcheung: i believe the name has nothing to do with this test case, can someone confirm @Amir-61 @bajtos

and under strict mode it will return validationError, or perhaps we can define it on top https://github.com/strongloop/loopback-datasource-juggler/blob/60d059928d2ef919135efa7bf5425e10a5c83be0/test/datatype.test.js#L192-L200

bajtos: This test checks what happens when you have `persistUndefinedAsNull` configured and create a model instance that does not have values for all properties. 

Your conclusion seems reasonable to me. I guess my only concern is whether it's ok to pass an empty object to `create`. Perhaps we should add the `name` property to the model definition instead?

test/datatype.test.js
```diff
@@ -220,12 +220,12 @@ describe('datatypes', function() {
 
     it('should convert property value undefined to null', function(done) {
       var EXPECTED = {desc: null, extra: null};
+      var data = {desc: undefined, extra: undefined};
       if (isStrict) {
         // SQL-based connectors don't support dynamic properties
         delete EXPECTED.extra;
+        delete data.extra;
```
bajtos: Why is it necessary to delete `data.extra`?

3.0-RELEASE-NOTES.md
```diff
@@ -102,3 +102,37 @@ bulk `updateOrCreate`, you could use `async.each` or `Promise.all` to
 repeatedly call `updateOfCreate` for each instance.
 
 See [related code change](https://github.com/strongloop/loopback-datasource-juggler/pull/889) here.
+
+## Removing strict: validate and strict: throw
+
+- `strict: 'validate'` and `strict: 'throw'` are removed in favor of `strict: true`.
+
+- `strict: true`'s behavior change:
```
bajtos: Please add an item explaining how `strict:false` is behaving now (i.e. no change).

lib/dao.js
```diff
@@ -93,10 +93,11 @@ function applyStrictCheck(model, strict, data, inst, cb) {
     key = keys[i];
     if (props[key]) {
       result[key] = data[key];
-    } else if (strict === 'throw') {
-      cb(new Error(g.f('Unknown property: %s', key)));
-      return;
-    } else if (strict === 'validate') {
+    } else if (strict) {
+      if (strict === 'throw') {
+        g.warn('{{`strict: throw`}} has been removed, use {{`strict: true`}} instead ' +
```
bajtos: This will print a warning for each call of a DAO operation, which can flood the console output and also won't be triggered until an operation is invoked.

I am proposing to check `settings.strict` at the time the model is set up instead. That way the warning will be always printed only once for each model and immediately at startup time.

Also make sure to include model name in the error message, otherwise it's difficult to find which of the application models is having the issue.

bajtos:
I did a quick partial review, didn't reviewed the added/removed test cases.

test/datatype.test.js
davidcheung: agreed, having empty object might cause confusion.
>Perhaps we should add the name property to the model definition instead?

test/datatype.test.js
davidcheung: `mysql` fails with the following otherwise, seems it does not discard undefined properties which leads to
```yml 
1) mysql imported features datatypes model option persistUndefinedAsNull should convert property value undefined to null:
     ValidationError: The `TestModel` instance is not valid. Details: `extra` is not defined in the model (value: undefined); `haha` is not defined in the model (value: undefined).
      at /Users/davidcheung/node-web/loopback-datasource-juggler/lib/dao.js:327:12
      at ModelConstructor.<anonymous> (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/validations.js:495:11)
      at ModelConstructor.next (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/hooks.js:83:12)
      at ModelConstructor.<anonymous> (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/validations.js:492:23)
      at ModelConstructor.trigger (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/hooks.js:73:12)
      at ModelConstructor.Validatable.isValid (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/validations.js:458:8)
      at /Users/davidcheung/node-web/loopback-datasource-juggler/lib/dao.js:323:9
      at doNotify (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/observer.js:99:49)
      at doNotify (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/observer.js:99:49)
      at Function.ObserverMixin._notifyBaseObservers (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/observer.js:122:5)
      at Function.ObserverMixin.notifyObserversOf (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/observer.js:97:8)
      at Function.ObserverMixin._notifyBaseObservers (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/observer.js:120:15)
      at Function.ObserverMixin.notifyObserversOf (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/observer.js:97:8)
      at Function.DataAccessObject.create (/Users/davidcheung/node-web/loopback-datasource-juggler/lib/dao.js:307:9)
      at Context.<anonymous> (/Users/davidcheung/node-web/loopback-datasource-juggler/test/datatype.test.js:229:17)
```

lib/dao.js
davidcheung: agreed, that makes alot more sense
>I am proposing to check settings.strict at the time the model is set up instead. That way the warning will be always printed only once for each model and immediately at startup time.

test/datatype.test.js
davidcheung: I'm not sure what to do with this test case, so the option of `{ persistUndefinedAsNull: true }` disables the discarding of undefined property, while `strict mode` will return validation error. 

in this case deleting `data.extra` is needed ( and i think is correct behavior? )
while the test will still check for the defined property (`desc`) which fulfills the original intent of the test

davidcheung:
@bajtos PTAL

test/datatype.test.js
```diff
@@ -192,6 +192,7 @@ describe('datatypes', function() {
       TestModel = db.define(
         'TestModel',
         {
+          name: {type: String, require: false},
```
bajtos: type, `require` instead of `required`?

test/datatype.test.js
bajtos: I'm surprised the problem was not discovered earlier. Is it really related to the changes made in this pull request? If it isn't, then could you perhaps isolate it in a standalone commit (if not send it as a new PR)?

test/manipulation.test.js
```diff
@@ -544,31 +544,72 @@ describe('manipulation', function() {
         });
     });
 
-    it('should ignore unknown attributes when strict: true', function(done) {
-      // Using {foo: 'bar'} only causes dependent test failures due to the
-      // stripping of object properties when in strict mode (ie. {foo: 'bar'}
-      // changes to '{}' and breaks other tests
```
bajtos: Please preserve this comment.

test/manipulation.test.js
```diff
@@ -544,31 +544,72 @@ describe('manipulation', function() {
...
+              });
+            });
         });
-    });
+      });
 
-    it('should throw error on unknown attributes when strict: throw', function(done) {
+    // `strict: true` used to silently remove unknown properties,
+    // now return validationError upon unknown properties,
+    // same as previous validate behavior
```
bajtos: This comment does not make sense to me when it's before `it should allow unknown attributes when strict: false`

test/datatype.test.js
davidcheung: it is related to this change, because sql database uses `strict:true` which silently remove before, but now returns `err`

test/manipulation.test.js
davidcheung: ure right, going to remove it
going to leave the test below with (which i think is alittle easier to understand)
```js
// `strict: true` used to silently remove unknown properties,
// now return validationError upon unknown properties
it('should return error on unknown attributes when strict: true',
```

test/manipulation.test.js
davidcheung: moved it under `it('should return error on unknown attributes when strict: true',`

test/datatype.test.js
bajtos: Ah, of course! Makes a lot of sense.

test/manipulation.test.js
```diff
@@ -544,31 +544,72 @@ describe('manipulation', function() {
...
+        Person.findById(person.id, function(err, p) {
+          p.updateAttributes({name: 'John', foo: 'bar'},
+            function(err, p) {
+              should.exist(err);
+              err.name.should.equal('ValidationError');
+              err.message.should.containEql('`foo` is not defined in the model');
+              should.not.exist(p.foo);
+              Person.findById(p.id, function(e, p) {
+                if (e) return done(e);
+                should.not.exist(p.foo);
```
bajtos: `p.should.not.have.property('foo')`?

test/manipulation.test.js
```diff
@@ -544,31 +544,72 @@ describe('manipulation', function() {
...
+        // stripping of object properties when in strict mode (ie. {foo: 'bar'}
+        // changes to '{}' and breaks other tests
+        Person.definition.settings.strict = true;
+        Person.findById(person.id, function(err, p) {
+          p.updateAttributes({name: 'John', foo: 'bar'},
+            function(err, p) {
+              should.exist(err);
+              err.name.should.equal('ValidationError');
+              err.message.should.containEql('`foo` is not defined in the model');
+              should.not.exist(p.foo);
```
bajtos: `p.should.not.have.property('foo')`?

test/manipulation.test.js
```diff
@@ -544,31 +544,72 @@ describe('manipulation', function() {
...
-    it('should throw error on unknown attributes when strict: throw', function(done) {
+    it('should allow unknown attributes when strict: false',
+      function(done) {
+        Person.definition.settings.strict = false;
+        Person.findById(person.id, function(err, p) {
+          p.updateAttributes({name: 'John', foo: 'bar'},
+            function(err, p) {
+              if (err) return done(err);
+              should.not.exist(err);
+              should.exist(p.foo);
```
bajtos: `p.should.have.property('foo')`?

test/manipulation.test.js
```diff
@@ -544,31 +544,72 @@ describe('manipulation', function() {
...
 
-    it('should throw error on unknown attributes when strict: throw', function(done) {
+    it('should allow unknown attributes when strict: false',
+      function(done) {
+        Person.definition.settings.strict = false;
+        Person.findById(person.id, function(err, p) {
+          p.updateAttributes({name: 'John', foo: 'bar'},
+            function(err, p) {
+              if (err) return done(err);
+              should.not.exist(err);
```
bajtos: this is redundant, `if (err)` has already handled this case.

test/manipulation.test.js
```diff
@@ -544,31 +544,72 @@ describe('manipulation', function() {
...
+        Person.definition.settings.strict = true;
+        Person.findById(person.id, function(err, p) {
+          p.updateAttributes({name: 'John', unknownVar: undefined},
+            function(err, p) {
+              // if uknownVar was defined, it would return validationError
+              if (err) return done(err);
+              Person.findById(p.id, function(e, p) {
+                if (e) return done(e);
+                p.name.should.equal('John');
+                should.not.exist(p.unknownVar);
```
bajtos: `p.should.not.have.property('unknownVar')`?

lib/model.js
```diff
@@ -95,6 +95,11 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
   if (strict === undefined) {
     strict = ctor.definition.settings.strict;
   }
+  if (strict === 'throw') {
```
bajtos: Perhaps join with the line above?

```js
} else if (string === 'throw') {
```

bajtos:
@slnode test please

davidcheung:
@slnode  test please

PR #1085, opened 2016-09-07
Set `removeAll` as an alias for `destroyAll`

Amir-61:
* Set `removeAll` as an alias for `destroyAll`
* It needs to be backported if the changes are right

/to: @superkhau 
/cc: @bajtos @raymondfeng

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2027,7 +2027,7 @@ DataAccessObject.findOne = function findOne(query, options, cb) {
  * @param {Object) [options] Options
  * @param {Function} [cb] Callback called with (err, info)
  */
-DataAccessObject.remove =
+DataAccessObject.removeAll =
 DataAccessObject.deleteAll =
 DataAccessObject.destroyAll = function destroyAll(where, options, cb) {
   var connectionPromise = stillConnecting(this.getDataSource(), this, arguments);
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2194,7 +2194,7 @@ DataAccessObject.deleteById = function deleteById(id, options, cb) {
 
   var Model = this;
 
-  this.remove(byIdQuery(this, id).where, options, function(err, info) {
+  this.removeAll(byIdQuery(this, id).where, options, function(err, info) {
     if (err) return cb(err);
     var deleted = info && info.count > 0;
     if (Model.settings.strictDelete && !deleted) {
--- a/test/default-scope.test.js
+++ b/test/default-scope.test.js
@@ -540,7 +540,7 @@ describe('default scope', function() {
     });
 
     it('should apply default scope - custom where', function(done) {
-      Widget.remove({name: 'Widget A'}, function(err) {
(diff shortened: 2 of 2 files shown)


superkhau:
@Amir-61 This change makes no sense to me. Why are you removing the original remove when one is instance based and the other is static. I am proposing we reject your changes unless you can explain why it is necessary

Amir-61:
Hey @superkhau,

>@Amir-61 This change makes no sense to me. 

You may be right.

>Why are you removing the original remove when one is instance based and the other is static. I am proposing we reject your changes unless you can explain why it is necessary.

My intention of the changes: why did we have `remove` as an alias of `deleteAll` and `destroyAll`? Doesn't it make sense all use the same name convention (`destroy/delete/remove`+`All`)?

>It needs to be backported if the changes are right

On the second thought this is a breaking change; hence, we may not need to backport it (if the changes are accepted)

superkhau:
>On the second thought this is a breaking change; hence, we may not need to backport it (if the changes are accepted)

That is my number one concern. I'm proposing you add a deprecation flag to the remove alias and dump it in juggler@4 to give anyone who is actually using the static remove enough time to change over to one of the other aliases. IMO though, we should just add `removeAll` like you propose *without* taking out remove for the best of both worlds (non-breaking and consistency). @bajtos @raymondfeng Thoughts?

>My intention of the changes: why did we have remove as an alias of deleteAll and destroyAll? Doesn't it make sense all use the same name convention (destroy/delete/remove+All)?

I actually believe we should have less aliases, but those other ones are already there. IMO, keeping just `delete` would be fine and probably cause less confusion/explaining.

Amir-61:
>'m proposing you add a deprecation flag to the remove alias and dump it in juggler@4 to give anyone who is actually using the static remove enough time to change over to one of the other aliases.

Either your approach or we can deprecate it in `loopback-datasource-juggler@2.x` and this patch can be landed for our master branch which is `loopback-datasource-juggler@3.x`.

>I actually believe we should have less aliases

Thats why we are doing architectural redesign, which I'm spiking in this sprint.

superkhau:
>Either your approach or we can deprecate it in loopback-datasource-juggler@2.x and this patch can be landed for our master branch which is loopback-datasource-juggler@3.x.

I'm fine with this solution too. I'm just not sure when 3x is going out lol. Either way as long as we show a deprecation message somewhere, I think you'll be good. ;)

>Thats why we are doing architectural redesign, which I'm spiking in this sprint.

Cool, jot that one in your notes. ;)

bajtos:
> we should just add removeAll like you propose without taking out remove for the best of both worlds (non-breaking and consistency). @bajtos @raymondfeng Thoughts?

+1 for only adding the new alias without taking out any existing aliases.

> I'm just not sure when 3x is going out lol.

End of September. This means we should finish all breaking changes by the end of this sprint (2016-09-12).

Amir-61:
Please lets get back to this PR after architectural design task of DAO.

Thanks!

PR #1086, opened 2016-09-08
Add missing "done" arg in test/kvao/ttl.suite

bajtos:
Fix regression introduced by #1078 that caused loopback-connector-kv-redis tests to fail.

cc @superkhau


superkhau:
TY for handling the regression. :+1:

PR #1087, opened 2016-09-08
Fix BSON Object ID errors for CI

superkhau:
@0candy PTAL

### DIFF
--- a/test/include.test.js
+++ b/test/include.test.js
@@ -269,9 +269,10 @@ describe('include', function() {
           if (err) return done(err);
 
           var passport = user.passports()[0];
-          passport.id.should.equal(1);
-          passport.ownerId.should.equal(1);
-          passport.number.should.equal('1');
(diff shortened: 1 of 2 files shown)


test/include.test.js
```diff
@@ -478,18 +487,18 @@ describe('include', function() {
...
+
+                          patients.should.have.length(1);
+                          var p = patients[0];
+                          p.name.should.equal('a');
+                          p.addressId.should.eql(patient.addressId);
+                          p.address().id.should.eql(address.id);
+                          p.address().name.should.equal('z');
+
+                          done();
+                        });
```
0candy: Does the linter fail if we revert the extra indent?

superkhau: Yes it does

superkhau: We can address this as part of homework test refactor. This super nested test needs cleanup anyways.

0candy:
LGTM

PR #1091, opened 2016-09-16
Add ilike and nilike operators

### LINKED ISSUE #633: Pass through ilike and nilike parameters so they can be handled by connectors

duffn:
Closes #633

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1400,6 +1400,8 @@ var operators = {
   neq: '!=',
   like: 'LIKE',
   nlike: 'NOT LIKE',
+  ilike: 'ILIKE',
+  nilike: 'NOT ILIKE',
   regexp: 'REGEXP',
 };
 
@@ -1626,6 +1628,8 @@ DataAccessObject._coerce = function(where) {
               break;
             case 'like':
             case 'nlike':
+            case 'ilike':
+            case 'nilike':
               if (!(typeof val === 'string' || val instanceof RegExp)) {
                 err = new Error(g.f('The %s property has invalid clause %j', p, where[p]));
                 err.statusCode = 400;
@@ -1658,7 +1662,8 @@ DataAccessObject._coerce = function(where) {
           operator = 'regexp';
         } else if (operator === 'regexp' && val instanceof RegExp) {
           // Do not coerce regex literals/objects
-        } else if (!((operator === 'like' || operator === 'nlike') && val instanceof RegExp)) {
+        } else if (!((operator === 'like' || operator === 'nlike' ||
+            operator === 'ilike' || operator === 'nilike') && val instanceof RegExp)) {
           val = DataType(val);
         }
       }
@@ -1711,6 +1716,8 @@ DataAccessObject._coerce = function(where) {
  * - neq: !=
  * - like: LIKE
  * - nlike: NOT LIKE
+ * - ilike: ILIKE
+ * - nilike: NOT ILIKE
  * - regexp: REGEXP
  *
  * You can also use `and` and `or` operations.  See [Querying models](http://docs.strongloop.com/display/DOC/Querying+models) for more information.
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -469,6 +469,62 @@ describe('basic-querying', function() {
       });
     });
 
+    var itWhenIlikeSupported = connectorCapabilities.ilike ? it : it.skip.bind(it);
+
+    itWhenIlikeSupported('should support "like" that is satisfied', function(done) {
+      User.find({where: {name: {like: 'John'}}}, function(err, users) {
+        if (err) return done(err);
+        users.length.should.equal(1);
+        users[0].name.should.equal('John Lennon');
+        done();
+      });
+    });
+
+    itWhenIlikeSupported('should support "like" that is not satisfied', function(done) {
+      User.find({where: {name: {like: 'Bob'}}}, function(err, users) {
+        if (err) return done(err);
+        users.length.should.equal(0);
+        done();
+      });
+    });
+
+    var itWhenNilikeSupported = connectorCapabilities.nilike ? it : it.skip.bind(it);
+
+    itWhenNilikeSupported('should support "nlike" that is satisfied', function(done) {
+      User.find({where: {name: {nlike: 'John'}}}, function(err, users) {
+        if (err) return done(err);
+        users.length.should.equal(5);
+        users[0].name.should.equal('Paul McCartney');
+        done();
+      });
+    });
+
+    itWhenIlikeSupported('should support "ilike" that is satisfied', function(done) {
+      User.find({where: {name: {ilike: 'john'}}}, function(err, users) {
+        if (err) return done(err);
+        users.length.should.equal(1);
+        users[0].name.should.equal('John Lennon');
+        done();
+      });
+    });
+
+    itWhenIlikeSupported('should support "ilike" that is not satisfied', function(done) {
+      User.find({where: {name: {ilike: 'bob'}}}, function(err, users) {
+        if (err) return done(err);
+        users.length.should.equal(0);
+        done();
+      });
+    });
+
+    itWhenNilikeSupported('should support "nilike" that is satisfied', function(done) {
+      User.find({where: {name: {nilike: 'john'}}}, function(err, users) {
+        if (err) return done(err);
+        users.length.should.equal(5);
+        users[0].name.should.equal('Paul McCartney');
+        done();
+      });
+    });
+
     it('should only include fields as specified', function(done) {
       var remaining = 0;
 
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -616,8 +616,8 @@ function applyFilter(filter) {
         testInEquality({lte: example.between[1]}, value));
       }
 
-      if (example.like || example.nlike) {
-        var like = example.like || example.nlike;
+      if (example.like || example.nlike || example.ilike || example.nilike) {
+        var like = example.like || example.nlike || example.ilike || example.nilike;
         if (typeof like === 'string') {
           like = toRegExp(like);
         }
@@ -627,6 +627,14 @@ function applyFilter(filter) {
 
         if (example.nlike) {
           return !new RegExp(like).test(value);
+        }
+
+        if (example.ilike) {
+          return !!new RegExp(like, 'i').test(value);
+        }
+
+        if (example.nilike) {
+          return !new RegExp(like, 'i').test(value);
         }
       }
 
--- a/test/init.js
+++ b/test/init.js
@@ -32,3 +32,7 @@ if (!('getModelBuilder' in global)) {
     return new ModelBuilder();
   };
 }
+
+if (!('connectorCapabilities' in global)) {
+  global.connectorCapabilities = {};
+}


slnode:

slnode:

slnode:

slnode:

duffn:
Originally talked about here: https://github.com/strongloop/loopback-datasource-juggler/issues/633#issuecomment-247487564

The original issue was closed because `regexp` was added and it was thought that we didn't need `ilike` and `nilike` because `regexp` could handle all situations. However, we should have `ilike` and `nilike` operators as some databases may not support case-insensitive `regexp`.

superkhau:
Can you add some tests to verify your changes and prevent regressions in the future?

duffn:
I added tests for `ilike` and `nilike` which also required that I added those two operators to the memory connector. 

I also added `like` and `nlike` tests that were missing.

lib/dao.js
```diff
@@ -1661,7 +1665,8 @@ DataAccessObject._coerce = function(where) {
           operator = 'regexp';
         } else if (operator === 'regexp' && val instanceof RegExp) {
           // Do not coerce regex literals/objects
-        } else if (!((operator === 'like' || operator === 'nlike') && val instanceof RegExp)) {
+        } else if (!((operator === 'like' || operator === 'nlike' ||
+          operator === 'ilike' || operator === 'nilike') && val instanceof RegExp)) {
```
superkhau: Please fix indentation. See our style guide at https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md#indentation-of-multi-line-expressions-in-if

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
       });
     });
 
+    it('should support "like" that is satisfied', function(done) {
+      User.find({where: {name: {'like': 'John'}}}, function(err, users) {
```
superkhau: remove single quotes around `like`

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        should.not.exist(err);
```
superkhau: `if (err) return done(err)` for better stack traces.

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        users.should.have.property('length', 1);
```
superkhau: users.length.should.equal(1);

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        users[0].name.should.equal('John Lennon');
+        done();
+      });
+    });
+
+    it('should support "like" that is not satisfied', function(done) {
+      User.find({where: {name: {'like': 'Bob'}}}, function(err, users) {
```
superkhau: Remove single quotes around `like`

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        should.not.exist(err);
```
superkhau: if (err) return done(err);

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        users.should.have.property('length', 0);
```
superkhau: users.length.should.equal(1);

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        done();
+      });
+    });
+
+    it('should support "nlike" that is satisfied', function(done) {
+      User.find({where: {name: {'nlike': 'John'}}}, function(err, users) {
+        should.not.exist(err);
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        users.should.have.property('length', 5);
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+    it('should support "like" that is not satisfied', function(done) {
+      User.find({where: {name: {'like': 'Bob'}}}, function(err, users) {
+        should.not.exist(err);
+        users.should.have.property('length', 0);
+        done();
+      });
+    });
+
+    it('should support "nlike" that is satisfied', function(done) {
+      User.find({where: {name: {'nlike': 'John'}}}, function(err, users) {
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        should.not.exist(err);
+        users.should.have.property('length', 5);
+        users[0].name.should.equal('Paul McCartney');
+        done();
+      });
+    });
+
+    it('should support "ilike" that is satisfied', function(done) {
+      User.find({where: {name: {'ilike': 'john'}}}, function(err, users) {
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        should.not.exist(err);
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        users.should.have.property('length', 1);
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        users[0].name.should.equal('John Lennon');
+        done();
+      });
+    });
+
+    it('should support "ilike" that is not satisfied', function(done) {
+      User.find({where: {name: {'ilike': 'bob'}}}, function(err, users) {
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        should.not.exist(err);
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        users.should.have.property('length', 0);
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        done();
+      });
+    });
+
+    it('should support "nilike" that is satisfied', function(done) {
+      User.find({where: {name: {'nilike': 'john'}}}, function(err, users) {
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        should.not.exist(err);
```
superkhau: ditto

test/basic-querying.test.js
```diff
@@ -469,6 +469,58 @@ describe('basic-querying', function() {
...
+        users.should.have.property('length', 5);
```
superkhau: ditto

lib/dao.js
duffn: I've corrected per the style guide, but the linter does not give any errors or warnings on any of the files.

test/basic-querying.test.js
duffn: I've removed all single quotes.

test/basic-querying.test.js
duffn: I've updated all per the suggestion. This does not match the rest of the tests in this file - I can update those in another PR.

test/basic-querying.test.js
duffn: I've updated all per the suggestion. This does not match the rest of the tests in this file - I can update those in another PR.

lib/dao.js
superkhau: The linter seems to fail to recognize indentation sometimes. See these lines:

>When not feasible, then indent the second and next lines by two levels.
>One level of indentation makes it difficult to tell the difference between the condition and the branch body.

Read the good/bad examples: https://github.com/strongloop/loopback-contributor-docs/blob/master/style-guide.md

test/basic-querying.test.js
superkhau: Thanks, this is fine. All those other parts of the file need to change to this style anyways (if you would like to submit a PR for that, I can help approve that one too ;)).

superkhau: If you decide to go and refactor that file, you can rename the set of tests in this PR too -- the wording is pretty bad -- `it('supports the "like" operator', function(...)` or something along those lines would be better.

superkhau:
test please

superkhau:
@duffn The unfortunate part about this PR is that adding in those operations causes all the downstream connectors to fail (because they have no implementation). Do you have time to submit PRs for those connectors too? This feature should really go in together with the connector updates.

duffn:
Oh my. That's a lot, but I'll consider it. If I do this, is there anything special I should do to coordinate?

superkhau:
>Oh my. That's a lot, but I'll consider it. If I do this, is there anything special I should do to coordinate? 

The ugly thing is that once you create those PRs, we'll need to merge in this PR first before the other PRs pass. I'm reluctant to merge this in until we have PRs for those other connectors too. @bajtos @raymondfeng We really need to do something about this, what should we do in this case? This PR is LGTM, but downstream connectors will fail if I merge.

superkhau:
Related issue for postgres connector: https://github.com/strongloop/loopback-connector-postgresql/pull/93#issuecomment-248625934

bajtos:
> The ugly thing is that once you create those PRs, we'll need to merge in this PR first before the other PRs pass. I'm reluctant to merge this in until we have PRs for those other connectors too. @bajtos @raymondfeng We really need to do something about this, what should we do in this case? This PR is LGTM, but downstream connectors will fail if I merge.

IMO, we need a way for connectors to signal what features they support. The shared test suite in juggler should skip tests for features that are not supported.

duffn:
So regarding the regex discussion here https://github.com/strongloop/loopback-connector-postgresql/pull/93#issuecomment-248779374, that's fine for PostgreSQL and likely many databases, but I'm writing a connector for Redshift and it does not support case-insensitive regex, however, it does support `ilike`, hence this PR.

Amir-61:
@duffn @superkhau 

>IMO, we need a way for connectors to signal what features they support. The shared test suite in juggler should skip tests for features that are not supported.

I believe you can do similar approach to what @bajtos and I did in [PR#866](https://github.com/strongloop/loopback-datasource-juggler/pull/866).


CC: @bajtos

superkhau:
@Amir-61 I agree with the land and add unsupported test case approach for now. We'll need to revisit how connectors broadcast what they support as stated by @bajtos at https://github.com/strongloop/loopback-datasource-juggler/pull/1091#issuecomment-248838125.

superkhau:
>IMO, we need a way for connectors to signal what features they support. The shared test suite in juggler should skip tests for features that are not supported.

This is one key point for the DAO architecture discussions that @Amir-61 is working on.

superkhau:
@bajtos @raymondfeng Any final comments before landing?

superkhau:
@duffn Can you help with submitting PRs for the unsupported comment in the downstream connectors (before landing this so at least we have warning messages to cover those areas).

bajtos:
@superkhau I am opposed to landing this patch until we figure out how to keep downstream CI builds of our connectors green. We have invested a lot of time in the recent weeks to get our CI builds green, landing this patch will undo most of that effort.

> I believe you can do similar approach to what @bajtos and I did in [PR#866](https://github.com/strongloop/loopback-datasource-juggler/pull/866).

I am afraid this approach cannot be used easily, because `test/basic-querying.test.js` does not provide any hook for injecting configuration :(

I see the following options how to move forward:

1) Quick hack: attach the information about connector capabilities to the global `getSchema` method. This will make it easy to use `it` or `it.skip` in the tests:

```js
    var itWhenIlikeSupported = getSchema.connectorCapabilities.ilike ? it : it.skip.bind(it);
  
    itWhenIlikeSupported('should support "ilike" that is satisfied', function(done) {
      User.find({where: {name: {ilike: 'john'}}}, function(err, users) {
        if (err) return done(err);
        users.length.should.equal(1);
        users[0].name.should.equal('John Lennon');
        done();
      });
    });
```

Alternatively, require the connectors to define a global object `connectorCapabilities`, similarly how a global `getSchema` is expected.

2) Refactor the shared test suite to export a function that can accept `connectorCapabilities` and `getSchema` as arguments, to avoid global variables. This can be a follow-up story for 1)

3) Alternatively, store the metadata in the connector/datasource instance. Define a new property `dataSource.capabilities` or `dataSource.feature` that can hold information about additional capabilities supported by the connector. We already have [`dataSource.getType()`](https://github.com/strongloop/loopback-datasource-juggler/blob/24f000d9781cc93db5ccdd782bdc9b8507688f9e/lib/datasource.js#L708-L715) that's backed by connector's [prototype.getType()](https://github.com/strongloop/loopback-datasource-juggler/blob/24f000d9781cc93db5ccdd782bdc9b8507688f9e/lib/connectors/memory.js#L51-L53). The downside is that we won't be able to write `it.skip` tests, because the datasource is created only after all tests have been already declared.

I am personally voting for 1).

@superkhau @raymondfeng @Amir-61 thoughts?

superkhau:
>I am opposed to landing this patch until we figure out how to keep downstream CI builds of our connectors green. We have invested a lot of time in the recent weeks to get our CI builds green, landing this patch will undo most of that effort.

I agree we do not land this unless CI stays green. I'm also gonna vote for solution 1 as that looks like the smoothest path to landing.

@duffn Thoughts? Would you like to help submit the other PRs based on this approach?

duffn:
Yes, I will try to find the time to submit the additional PRs.

duffn:
Is this acceptable?
https://gist.github.com/duffn/0d9880b04c9cd617a8a36a0fb434ad88

bajtos:
> Is this acceptable?
https://gist.github.com/duffn/0d9880b04c9cd617a8a36a0fb434ad88

Looks good to me.

Amir-61:
Sorry for joining late on the conversation.

>>I believe you can do similar approach to what @bajtos and I did in PR#866.

>I am afraid this approach cannot be used easily, because test/basic-querying.test.js does not provide any hook for injecting configuration :(

Oh you are absolutely right!

To solve the problem I totally agree with your first approach which is explained https://github.com/strongloop/loopback-datasource-juggler/pull/1091#issuecomment-249199382

duffn:
connectorCapabilities PR submitted to downstream connector.
- [ ] loopback-connector-db2
- [ ] loopback-connector-mysql
- [ ] loopback-connector-cloudant
- [ ] loopback-connector-dashdb
- [ ] loopback-connector-db2z
- [ ] loopback-connector-informix
- [ ] loopback-connector-kv-redis
- [ ] loopback-connector-mongodb
- [ ] loopback-connector-mssql
- [ ] loopback-connector-oracle
- [ ] loopback-connector-postgresql
- [ ] loopback-connector-remote
- [ ] loopback-connector-rest
- [ ] loopback-connector-sqlite3
- [ ] loopback-ibmdb

connectorCapabilities PR accepted to downstream connector.
- [ ] loopback-connector-db2
- [ ] loopback-connector-mysql
- [ ] loopback-connector-cloudant
- [ ] loopback-connector-dashdb
- [ ] loopback-connector-db2z
- [ ] loopback-connector-informix
- [ ] loopback-connector-kv-redis
- [ ] loopback-connector-mongodb
- [ ] loopback-connector-mssql
- [ ] loopback-connector-oracle
- [ ] loopback-connector-postgresql
- [ ] loopback-connector-remote
- [ ] loopback-connector-rest
- [ ] loopback-connector-sqlite3
- [ ] loopback-ibmdb

bajtos:
@slnode ok to test

bajtos:
As far as I am concerned, the code change in this pull request looks good enough to me. @superkhau could you please work with @duffn to tie up any remaining loose ends? Should we mention the two new test flags somewhere in the documentation for building connectors?

superkhau:
>Should we mention the two new test flags somewhere in the documentation for building connectors?

Yes we should. :+1: 

cc @crandmck

superkhau:
@duffn Can you rebase and squash before we land this? Let's try to land those other related PRs first though.

duffn:
Alright, I think I got all of the PRs submitted.

superkhau:
@duffn I managed to land and backport all the connector capability PRs. Can you rebase this PR and lets see how CI fares.

duffn:
Fingers crossed.

superkhau:
@duffn LGTM, gonna land this as the failures are known failures for those downstream connectors. Thank you for the huge contribution. ;)

superkhau:
@duffn Can you submit another PR to loopback to document the two new flags? See https://github.com/strongloop/loopback.io/

@crandmck Can you guide @duffn to where he can add those notes?

crandmck:
@duffn 

Please do update the docs; you can open a PR for docs just as for the code.

Docs for ilike and nilike operators: http://loopback.io/doc/en/lb2/Where-filter.html#operators - source in https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/lb2/Where-filter.md 

Please at least add it to the table in that section.  If there's more to say about usage, you can also add a subsection as there is for AND/OR and regexp.

Ideally, you would also add an example in http://loopback.io/doc/en/lb2/Where-filter.html#examples

> Should we mention the two new test flags somewhere in the documentation for building connectors?

If you want to address this as well, we'd probably add a new section to http://loopback.io/doc/en/lb2/Building-a-connector.html - source in https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/lb2/Building-a-connector.md.

Don't worry to much about wording and formatting.  I can always clean it up afterward.

Thanks!!

PR #1093, opened 2016-09-17
Add docs for KVAO

superkhau:
@bajtos PTAL, docs matching LoopBack's KV model.

cc @crandmck 


bajtos:
Please back-port to 2.x after landing.

PR #1108, opened 2016-09-23
Fix the bug when near filter is used

Amir-61:
Fix the bug when `near` filter is used

* If near filter was used, users would get:

```
Error: Callback was already called
```

Connect to https://github.com/strongloop/loopback-connector-mongodb/issues/285

/to: @bajtos

### DIFF
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -86,6 +86,19 @@ module.exports = function(dataSource, should, connectorCapabilities) {
           });
       });
 
+      it('triggers correct hooks when near filter is used', function(done) {
+        monitorHookExecution();
+        var query = { where:
+          { location: { near: '10,20', maxDistance: '10', unit: 'meters' }},
+        };
+
+        TestModel.find(query, function(err, list) {
+          if (err) return done(err);
+          hookMonitor.names.should.eql(['access']);
+          done();
+        });
+      });
+
       it('should not trigger hooks, if notify is false', function(done) {
         monitorHookExecution();
         TestModel.find(
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1791,174 +1791,174 @@ DataAccessObject.find = function find(query, options, cb) {
   var near = query && geo.nearFilter(query.where);
   var supportsGeo = !!connector.buildNearFilter;
 
-  if (near) {
-    if (supportsGeo) {
-      // convert it
-      connector.buildNearFilter(query, near);
-    } else if (query.where) {
-      // do in memory query
-      // using all documents
-      // TODO [fabien] use default scope here?
+  if (query.where && near && !supportsGeo) {
+    // do in memory query
+    // using all documents
+    // TODO [fabien] use default scope here?
 
-      if (options.notify === false) {
-        queryGeo(query);
-      } else {
-        withNotifyGeo();
-      }
-
-      function withNotifyGeo() {
-        var context = {
-          Model: self,
-          query: query,
-          hookState: hookState,
-          options: options,
-        };
-        self.notifyObserversOf('access', context, function(err, ctx) {
-          if (err) return cb(err);
-          queryGeo(ctx.query);
-        });
-      }
-      function queryGeo(query) {
-        function geoCallback(err, data) {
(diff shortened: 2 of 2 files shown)


Amir-61:
@slnode test please

bajtos:
@Amir-61 please start by writing a test reproducing the issue.

TBH, the original code is a long spaghetti that I don't have capacity to grok right now. If all existing tests are passing and the new tests covering the bug is passing too, then I am ok to land this.

OTOH, if you can invest little bit more time, then it would be great to clean this code up first, before fixing the bug. For example, there are two places where `this.notifyObserversOf('access')` is called . Many of the `if/else` branches are too long and should be extracted into named functions too. And so on. The idea to is to clean up the code in such way that the commit fixing the bug will become a very small change.

Amir-61:
@slnode test please

Amir-61:
@bajtos 
I fixed more issues and added a test to verify the fixes; could you please review it?

>TBH, the original code is a long spaghetti

I agree, unfortunately that's the case for most of the code in DAO :(

>If all existing tests are passing and the new tests covering the bug is passing too, then I am ok to land this.

AFAICT all tests for juggler and connectors including `loopback-connector-mongodb` and others are passing now; also, I added test which was failing with the previous code base and passes with the changes here.

> there are two places where this.notifyObserversOf('access') is called 

That was the actual bug and I added test to verify the fix.

>Many of the if/else branches are too long and should be extracted into named functions too

I have named functions now `withNotifyGeo`, `queryGeo` and `allCb`, please let me know if you have more suggestions...

>The idea to is to clean up the code in such way that the commit fixing the bug will become a very small change.

The changes are not too much, but since I moved them around the GH changes are not easy to follow... sorry about that :(

test/persistence-hooks.suite.js
```diff
@@ -86,6 +86,18 @@ module.exports = function(dataSource, should, connectorCapabilities) {
...
       });
 
+      it('should trigger access hook once, if near filter is used', function(done) {
+        monitorHookExecution();
+        TestModel.find(
+          { where:
+            { location: { near: '10,20', maxDistance: '10', unit: 'meters' }},
+          }, function(err, list) {
+          if (err) return done(err);
+          hookMonitor.names.should.eql(['access']);
```
Amir-61: Please also note I do not think we have ever had `loaded` hook if `near` filter was used. We may add it in another PR.

test/persistence-hooks.suite.js
```diff
@@ -86,6 +86,18 @@ module.exports = function(dataSource, should, connectorCapabilities) {
           });
       });
 
+      it('should trigger access hook once, if near filter is used', function(done) {
```
bajtos: How about the following name?

```js
it('triggers correct hooks when near filter is used', ...)
```

This way the test can be easily extended later when we fix the implementation to trigger the "loaded" hook too.

test/persistence-hooks.suite.js
bajtos: Is there any GH issue to keep track of this bug? Are you planning to fix this as part of  strongloop/loopback-connector-mongodb#285? Either way, please create a checklist item or a new GH issue so that we don't forget about this problem.

Amir-61: Create an issue and I assigned it to myself.

Amir-61:
@rptally @bachirelkhoury @jainisenough @geodeveloper @jwebcat @Zvika88
Could one of you please verify that strongloop/loopback-connector-mongodb#285 is fixed by this patch juggler@2.x.

Thanks a lot!

Amir-61:
@bajtos Could you please also sign off https://github.com/strongloop/loopback-datasource-juggler/pull/1125

Thanks a lot 🙏

vidhema:
> @bajtos Could you please also sign off #1125
> 
> To-do:
> 
> * [x]  release `loopback-datasource-juggler@2.53.0`
> 
> Thanks a lot

not working at all

bajtos:
@vidhema please open a new GitHub issue and provide us with enough information to allow us to reproduce the issue locally, see https://loopback.io/doc/en/contrib/Reporting-issues.html#loopback-3x-bugs

Issue #1106, opened 2016-09-22
validateNumericality does nullCheck and therefore limits utility


C3realGuy:
Because validateNumericality checks if an attribute is set, and if not, fails, it limits the way you can use this validator.
For example you have an attribute which is optional, but, if it's set, it should be a number. You can't use validatesNumericalityOf because it flags err() even if the attribute is not set.
I think it would be better to stick to the concept, one validator checks one thing and you apply multiple validators to do what you need.
If you want that the attribute is set and numerical, just use validatesPresenceOf + validatesNumericalityOf.

Amir-61:
@C3realGuy 

I see your point... Would you like to send a PR and we discuss there?

C3realGuy:
I will give it a try.
https://github.com/strongloop/loopback-datasource-juggler/pull/1109

C3realGuy:
@Amir-61 
i figured out, this is actually not a bug. This is well expected behaviour. See https://github.com/strongloop/loopback-datasource-juggler/issues/541
But still i discovered some fishy things:

`  describe('numericality', function() {
    it('dummmy should test', function() {
      var t = true;
      t.should.be.false; // does not fail?!
    });
...`

Amir-61:
Thanks for clarification @C3realGuy. I'll work with our doc expert to document `allowNull`, `allowBlank`, `if` and`unless` which are explained in https://github.com/strongloop/loopback-datasource-juggler/issues/541  to avoid more confusion.

PR #1109, opened 2016-09-23
validateNumericality should skip undefined

C3realGuy:
is defined.

Connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1106

Thinks to note:
- This fix needs to be applied to many other validations too. Only did validateNumericality for now.
- I don't know how much we care about performance, but checking each attribute for each validator if it's type is undefined is maybe not the best approach. Still, this was already done before with nullCheck() and should be better, but still not perfect.
- In the original check the validator only checked if the attribute is typeof 'number'. But we also need to check that attribute is not NaN. Because typeof NaN => 'number'.

### DIFF
--- a/lib/validations.js
+++ b/lib/validations.js
@@ -287,7 +287,7 @@ function validateLength(attr, conf, err, options) {
 function validateNumericality(attr, conf, err, options) {
   if (nullCheck.call(this, attr, conf, err)) return;
 
-  if (typeof this[attr] !== 'number') {
+  if (typeof this[attr] !== 'number' || isNaN(this[attr])) {
     return err('number');
   }
   if (conf.int && this[attr] !== Math.round(this[attr])) {
@@ -633,16 +633,26 @@ var defaultMessages = {
   uniqueness: 'is not unique',
 };
 
+/**
+ * Checks if attribute is undefined or null. Calls err function with 'blank' or 'null'.
+ * See defaultMessages. You can affect this behaviour with conf.allowBlank and conf.allowNull.
+ * @param {String} attr Property name of attribute
+ * @param {Object} conf conf object for validator
+ * @param {Function} err
+ * @return {Boolean} returns true if attribute is null or blank
+ */
 function nullCheck(attr, conf, err) {
-  if (this[attr] == null) {
-    if (!conf.allowNull) {
-      err('null');
+  // First determine if attribute is defined
+  if (typeof this[attr] === 'undefined') {
+    if (!conf.allowBlank) {
+      err('blank');
     }
     return true;
   } else {
-    if (blank(this[attr])) {
-      if (!conf.allowBlank) {
-        err('blank');
+    // Now check if attribute is null
+    if (this[attr] === null) {
+      if (!conf.allowNull) {
+        err('null');
       }
       return true;
     }
--- a/test/validations.test.js
+++ b/test/validations.test.js
@@ -558,7 +558,65 @@ describe('validations', function() {
   });
 
   describe('numericality', function() {
-    it('should validate numericality');
+    it('passes when given numeric values', function() {
+      User.validatesNumericalityOf('age');
+      var user = new User({age: 10});
+      user.isValid().should.be.true();
+    });
+
+    it('fails when given non-numeric values', function() {
+      User.validatesNumericalityOf('age');
+      var user = new User({age: 'notanumber'});
+      user.isValid().should.be.false();
+      user.errors.should.eql({age: ['is not a number']});
+    });
+
+    it('fails when given undefined values', function() {
+      User.validatesNumericalityOf('age');
+      var u = new User({});
+      u.isValid().should.be.false();
+      u.errors.should.eql({age: ['is blank']});
+    });
+
+    it('skips undefined values when allowBlank option is true', function() {
+      User.validatesNumericalityOf('age', {allowBlank: true});
+      var user = new User({});
+      user.isValid().should.be.true();
+    });
+
+    it('fails when given non-numeric values when allowBlank option is true', function() {
+      User.validatesNumericalityOf('age', {allowBlank: true});
+      var user = new User({age: 'test'});
+      user.isValid().should.be.false();
+      user.errors.should.eql({age: ['is not a number']});
+
+    });
+
+    it('fails when given null values', function() {
+      User.validatesNumericalityOf('age');
+      var user = new User({age: null});
+      user.isValid().should.be.false();
+      user.errors.should.eql({age: ['is null']});
+    });
+
+    it('passes when given null values when allowNull option is true', function() {
+      User.validatesNumericalityOf('age', {allowNull: true});
+      var user = new User({age: null});
+      user.isValid().should.be.true();
+    });
+
+    it('passes when given float values', function() {
+      User.validatesNumericalityOf('age');
+      var user = new User({age: 13.37});
+      user.isValid().should.be.true();
+    });
+
+    it('fails when given non-integer values when int option is true', function() {
+      User.validatesNumericalityOf('age', {int: true});
+      var user = new User({age: 13.37});
+      user.isValid().should.be.false();
+      user.errors.should.eql({age: ['is not an integer']});
+    });
   });
 
   describe('inclusion', function() {


slnode:

slnode:

slnode:

slnode:

Amir-61:
@C3realGuy 

Thanks for the PR. Please write test cases which fail with the previous implementation and passes with your changes.

C3realGuy:
@Amir-61 never wrote test cases, so give me a hint/link and i will give it a try too.

Amir-61:
>@Amir-61 never wrote test cases, so give me a hint/link and i will give it a try too.

I believe looking into tests in `validations.test.js` will give you a good insight:
https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/validations.test.js

C3realGuy:
This is actually not the real problem which needs to be solved.
Things which need to be solved:
1. validateNumericality checks not if attribute isNaN()
2. there were no tests for validatesNumericalityOf()
3. nullCheck needs first to be fixed. See https://github.com/strongloop/loopback-datasource-juggler/issues/1111
4. Because nullCheck uses blank and blank interprets typeof attribute == 'number' + isNaN(attribute) as blank, we will get some trouble.

I will create a new issue.

lib/validations.js
```diff
@@ -626,21 +631,31 @@ var defaultMessages = {
...
+      err('blank');
     }
     return true;
-  } else {
-    if (blank(this[attr])) {
-      if (!conf.allowBlank) {
-        err('blank');
+  }else{
+    // Now check if attribute is null
+    if(this[attr] === 'null') {
```
C3realGuy: Good point, those '' shouldn't be there. Added test for this case and fixed.

C3realGuy:
@Amir-61 any news?

test/validations.test.js
```diff
@@ -558,7 +559,88 @@ describe('validations', function() {
   });
 
   describe('numericality', function() {
-    it('should validate numericality');
+    it('dummy for should framework which should fail', function(){
+      var t = false;
+      t.should.be.true;
```
Amir-61: The right syntax is:
```
t.should.be.true()
```

Amir-61: Please see: http://shouldjs.github.io/#assertion-true

C3realGuy: @Amir-61 now linter should be happy.
Also now using should with the right syntax. But please look how should is used with the other tests in validations.test.js, it's always without the `()`.

```
  describe('commons', function() {

    describe('skipping', function() {

      it('should NOT skip when `if` is fulfilled', function() {
        User.validatesPresenceOf('pendingPeriod', {if: 'createdByAdmin'});
        var user = new User;
        user.createdByAdmin = true;
        user.isValid().should.be.false;
        user.errors.pendingPeriod.should.eql(['can\'t be blank']);
        user.pendingPeriod = 1;
        user.isValid().should.be.true;
      });
```

C3realGuy:
@Amir-61 any news?

lib/validations.js
```diff
@@ -285,9 +282,15 @@ function validateLength(attr, conf, err, options) {
  * Numericality validator
  */
 function validateNumericality(attr, conf, err, options) {
-  if (nullCheck.call(this, attr, conf, err)) return;
+  if (nullCheck.call(this, attr, conf, err)) {
+    // If conf.allowBlank is set, we need to check typeof and isNaN again. Otherwise non numeric attributes can slip through
+    if (conf.allowBlank && typeof this[attr] === 'number' && isNaN(this[attr])) {
```
Amir-61: The only test I see for `alloBlanck` is `should skip undefined attributes if conf.allowBlank is true` which passes. Please add test to cover the the failure for your logic here.

lib/validations.js
```diff
@@ -263,9 +263,6 @@ function validateAbsence(attr, conf, err, options) {
   }
 }
 
-/*!
- * Length validator
- */
```
Amir-61: Please do not delete irrelevant code.

Amir-61:
@slnode test please

Amir-61:
@C3realGuy Please squash your commit into one single commit and rebase it on top of master.

@superkhau I need a second thought on changes in this patch.

lib/validations.js
C3realGuy: Not needed anymore because nullCheck behaves now more precise.
Will be removed in next commit.

superkhau:
@C3realGuy Reviewed, overall the changes look fine. Just change the `u` to `user` in all test cases.

> This fix needs to be applied to many other validations too. Only did validateNumericality for now.

This is fine, feel free to submit PRs to fix the others. :smile: 

>I don't know how much we care about performance, but checking each attribute for each validator if it's type is undefined is maybe not the best approach. Still, this was already done before with nullCheck() and should be better, but still not perfect.

They can set allowBlank to false to speed it up if necessary. We can optimize then because we'll have a better picture when the person files the issue with regards to actual benchmarks.
 
> In the original check the validator only checked if the attribute is typeof 'number'. But we also need to check that attribute is not NaN. Because typeof NaN => 'number'.

:+1: 

@Amir-61 You can take it from here.

C3realGuy:
Applied requested changes. Will squash it if you're fine with it.

Amir-61:
>Applied requested changes. Will squash it if you're fine with it.

Right now the commit message is failing... Please see [the failure](https://ci.sl.strongloop.com/job/strongloop-commit-linter-jobs/job/loopback-datasource-juggler/568/console); also the PR needs to be rebased on top of master see [failure](https://ci.sl.strongloop.com/job/strongloop-pr-linter-jobs/job/loopback-datasource-juggler/568/console); unfortunately since these two first check are failling CI is not running other tests.. could you please solve them and see the CI result after that... Thanks for contribution...

test/validations.test.js
```diff
@@ -558,7 +558,58 @@ describe('validations', function() {
   });
 
   describe('numericality', function() {
-    it('should validate numericality');
+    it('should succes if property is a number', function() {
+      User.validatesNumericalityOf('age');
+      var user = new User({age: 10});
+      user.isValid().should.be.true();
+    });
+    it('fails when given non-numeric values', function() {
```
Amir-61: Please add one blank line between each test to make the code more readable... This applies for all your tests

test/validations.test.js
```diff
@@ -558,7 +558,58 @@ describe('validations', function() {
...
+      var user = new User({age: 13.37});
+      user.isValid().should.be.true();
+    });
+    it('fails when given non-integer values when int option is true', function() {
+      User.validatesNumericalityOf('age', {int: true});
+      var user = new User({age: 13.37});
+      user.isValid().should.be.false();
+      user.errors.should.eql({age: ['is not an integer']});
+    });
+
```
Amir-61: no blank line here

test/validations.test.js
```diff
@@ -558,7 +558,58 @@ describe('validations', function() {
...
+      user.isValid().should.be.false();
+      user.errors.should.eql({age: ['is not a number']});
+    });
+    it('fails when given undefined values', function() {
+      User.validatesNumericalityOf('age');
+      var u = new User({});
+      u.isValid().should.be.false();
+      u.errors.should.eql({age: ['is blank']});
+    });
+    it('skips undefined values when allowBlank option is true', function() {
```
Amir-61: ditto and other places

Amir-61:
FYI:We follow these rules to add commit description: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

C3realGuy:
Commit linter is now happy too, also applied requested changes by @Amir-61 and the one test renaming missed before by @superkhau.

Amir-61:
@slnode test please

C3realGuy:
Can't access cis-jenkins.swg-devops.com. What are those 7 failing checks about?
Besides this I think all of @superkhau s requested changes are already applied.

Amir-61:
@C3realGuy 
>What are those 7 failing checks about?

I believe they are irrelevant to your changes... 

LGTM from me. @bajtos could you please also give the final LGTM as well?

Thanks!

superkhau:
>I believe they are irrelevant to your changes... 

@Amir-61 Do not assume and read the details of CI to determine whether they are relevant or not.

@C3realGuy They fails are not related your changes and are downstream known failures.

Amir-61:
@C3realGuy Thanks for the contribution. Landing now 🎊

test/validations.test.js
```diff
@@ -558,7 +558,65 @@ describe('validations', function() {
...
+      var user = new User({});
+      user.isValid().should.be.true();
+    });
+
+    it('fails when given non-numeric values when allowBlank option is true', function() {
+      User.validatesNumericalityOf('age', {allowBlank: true});
+      var user = new User({age: 'test'});
+      user.isValid().should.be.false();
+      user.errors.should.eql({age: ['is not a number']});
+
```
superkhau: @Amir-61 Should remove this blank line.

test/validations.test.js
```diff
@@ -558,7 +558,65 @@ describe('validations', function() {
...
+      var user = new User({age: 'notanumber'});
+      user.isValid().should.be.false();
+      user.errors.should.eql({age: ['is not a number']});
+    });
+
+    it('fails when given undefined values', function() {
+      User.validatesNumericalityOf('age');
+      var u = new User({});
+      u.isValid().should.be.false();
+      u.errors.should.eql({age: ['is blank']});
```
superkhau: @Amir-61 This user is still u, we should make them all `u` or all `user` for consistency.

superkhau:
@Amir-61 Can you submit patch fixes for the 2 points above in a follow up PR since you landed this already?

Amir-61:
@superkhau 

>@Amir-61 Can you submit patch fixes for the 2 points above in a follow up PR since you landed this already?

https://github.com/strongloop/loopback-datasource-juggler/pull/1147 addresses your concerns. PTAL

PR #1119, opened 2016-09-28
Update translation files - round#2

0candy:
https://github.com/strongloop-internal/scrum-loopback/issues/1097


Amir-61:
>@amir You can’t compare the string order, you have to compare the hash. The order is different

LGTM; however I'm still confused how come whole files are completely changed and the order is not the same; the order did not change in other repos, just this repo; thats why it confuses me... anyway LGTM. :shipit:

PR #1125, opened 2016-10-04
Add test for operation hooks

Amir-61:
* Add test for operation hook when near filter is used

Connect to https://github.com/strongloop/loopback-connector-mongodb/issues/285

### DIFF
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -99,6 +99,18 @@ module.exports = function(dataSource, should, connectorCapabilities) {
           });
       });
 
+      it('triggers correct hooks when near filter is used', function(done) {
+        monitorHookExecution();
+        var query = {
+          where: {location: {near: '10,20', maxDistance: '10', unit: 'meters'}},
+        };
+        TestModel.find(query, function(err, list) {
+          if (err) return done(err);
+          hookMonitor.names.should.eql(['access']);
+          done();
+        });
+      });
+
       it('should not trigger hooks for geo queries, if notify is false',
       function(done) {
         monitorHookExecution();


Amir-61:
Hey @bajtos,

I do not think if https://github.com/strongloop/loopback-connector-mongodb/issues/285 affects LB 3.x. I added test and it passes on loopback-datasource-juggler@3.x. Please note the code change for `DAO.find()` is totally different between juggler@2.x Vs. juggler@3.x.

test/persistence-hooks.suite.js
```diff
@@ -99,6 +99,18 @@ module.exports = function(dataSource, should, connectorCapabilities) {
           });
       });
 
+      it('should trigger access hook once, if near filter is used', function(done) {
+        monitorHookExecution();
+        TestModel.find(
+          {where:
+            {location: {near: '10,20', maxDistance: '10', unit: 'meters'}},
+          }, function(err, list) {
```
bajtos: one arg per line please, see https://github.com/strongloop/loopback-contributor-docs/blob/198d4ebeb7171997474c05d6cc067adb60af3848/style-guide.md#one-arg-per-line

Alternatively, store the long query in a variable.


```js
var query = {
  where: {location: {near: '10,20', maxDistance: '10', unit: 'meters'}},
};

TestModel.find(query, function(err, list) {
  // ...
});
```

Amir-61:
>Interesting. Did you try to reproduce the problem described in strongloop/loopback-connector-mongodb#285 with LB 3.x?

>Perhaps we should start with adding the test to 2.x branch of juggler to see it fail?


Yes I could reproduce the issue with juggler@2.x and added test which failed and passed with changes in https://github.com/strongloop/loopback-datasource-juggler/pull/1108; AFAICT, the bug does not exist in juggler@3.x, it only affected juggler@2.x.

@rptally @bachirelkhoury @jainisenough @geodeveloper @jwebcat @Zvika88
Could one of you please verify that https://github.com/strongloop/loopback-connector-mongodb/issues/285 is fixed by https://github.com/strongloop/loopback-datasource-juggler/pull/1108 for juggler@2.x. 

Thanks alot

Amir-61:
@bajtos PTAL.

jainisenough:
Thanks, @bajtos & @Amir-61

PR #1128, opened 2016-10-05
Fix release notes

Amir-61:
* Fix release notes

/to: @superkhau


Amir-61:
Oh I guess we are removing these release notes now; please see: https://github.com/strongloop/loopback-datasource-juggler/pull/1126

Issue #1132, opened 2016-10-07
upsertWithWhere bug in JSDoc


cojack:
There is:

```
 * * @param {Object} [where]  `where` filter, like
```

should be:

```
 * @param {Object} [where]  `where` filter, like
```

Because of this in generated documentation (http://apidocs.strongloop.com/loopback/#persistedmodel-upsertwithwhere), first argument `where` doesn't apear.

Regards.

superkhau:
Would you like to submit a PR for the fix? See https://github.com/strongloop/loopback/blob/master/lib/persisted-model.js#L121

Amir-61:
Fixed by https://github.com/strongloop/loopback-datasource-juggler/pull/1133

PR #1133, opened 2016-10-11
Fix JSDoc issue

Amir-61:
* Fix JSDoc 

Connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1132

/cc: @superkhau

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -636,7 +636,7 @@ DataAccessObject.upsert = function(data, options, cb) {
  * If there is a single instance retrieved, update the retrieved model.
  * Creates a new model if no model instances were found.
  * Returns an error if multiple instances are found.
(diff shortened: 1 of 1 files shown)


Amir-61:
On the second though I merge it since it is just a trivial JSDoc fix

lib/dao.js
```diff
@@ -636,7 +636,7 @@ DataAccessObject.upsert = function(data, options, cb) {
  * If there is a single instance retrieved, update the retrieved model.
  * Creates a new model if no model instances were found.
  * Returns an error if multiple instances are found.
- * * @param {Object} [where]  `where` filter, like
+ * @param {Object} [where]  `where` filter, like
```
superkhau: Second eyes help sometimes. You should remove the extra space in front of the `]`. ;)

PR #1140, opened 2016-10-18
test/kvao: add connectorCapabilities options

bajtos:
- canExpire
 - canQueryTtl
 - ttlPrecision
 - canIterateKeys
 - canIterateLargeKeySets

These options allow connectors to disable shared tests for features that are not supported/implemented.

Connect to strongloop-internal/scrum-loopback#1128
See also the spike - #1129

@superkhau please review

### DIFF
--- a/test/kvao/ttl.suite.js
+++ b/test/kvao/ttl.suite.js
@@ -1,11 +1,25 @@
 'use strict';
 
+var bdd = require('../helpers/bdd-if');
 var should = require('should');
 var helpers = require('./_helpers');
 var Promise = require('bluebird');
 
 module.exports = function(dataSourceFactory, connectorCapabilities) {
-  describe('ttl', function() {
+  var TTL_PRECISION = connectorCapabilities.ttlPrecision;
+
+  // Use ~1s for stores with precision of 1 ms,
+  // about 3s for stores with precision of 1s.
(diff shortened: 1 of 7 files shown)


test/kvao/ttl.suite.js
```diff
@@ -14,19 +23,19 @@ module.exports = function(dataSourceFactory, connectorCapabilities) {
     it('gets TTL when key with unexpired TTL exists - Promise API',
     function() {
       return Promise.resolve(
-          CacheItem.set('a-key', 'a-value', {ttl: 1000}))
-        .delay(1)
+          CacheItem.set('a-key', 'a-value', {ttl: INITIAL_TTL}))
+        .delay(Math.max(1, Math.floor(TTL_PRECISION / 10)))
```
superkhau: maybe move this up to a var in line 26 to describe the intent? I see you're adding some sort of delay at quick glance but don't see what the formula extracts easily.

```
var minorDelayAfterInitial = Math.max(1, Math.floor(TTL_PRECISION / 10));
...
.delay(minorDelayAfterInitial)
```

or a comment like you have above on line 11 explaining the formula.

superkhau:
One minor comment that should be addressed, otherwise LGTM.

test/kvao/ttl.suite.js
bajtos: Great point! As I was writing down the comment, I realised I have a bug in my formula, I'll fix it along the way.

PR #1142, opened 2016-10-18
DataAccessObject._coerce now continues after encountering a logical op

doublemarked:
I discovered this while working on https://github.com/strongloop/loopback/issues/2824 and felt it needed to be fixed. Please let me know if there's some non-obvious reason why logical operators should halt coercion.

### DIFF
--- a/test/loopback-dl.test.js
+++ b/test/loopback-dl.test.js
@@ -1421,6 +1421,16 @@ describe('DataAccessObject', function() {
     assert.deepEqual(where, {or: [{age: 10}, {vip: true}]});
   });
 
+  it('continues to coerce properties after a logical operator', function() {
+    var clause = {and: [{age: '10'}], vip: 'true'};
+
+    // Key order is predictable but not guaranteed. We prefer false negatives (failure) to false positives.
+    assert(Object.keys(clause)[0] === 'and', 'Unexpected key order.');
+
+    where = model._coerce(clause);
+    assert.deepEqual(where, {and: [{age: 10}], vip: true});
+  });
+
   it('should throw if the where property is not an object', function() {
     try {
       // The where clause has to be an object
@@ -1467,6 +1477,20 @@ describe('DataAccessObject', function() {
     try {
       // The or operator only takes an array of objects
       model._coerce({or: ['x']});
+    } catch (err) {
+      error = err;
+    }
+    assert(error, 'An error should have been thrown');
+  });
+
+  it('throws an error when malformed logical operators follow valid logical clauses', function() {
+    var invalid = {and: [{x: 1}], or: 'bogus'};
+
+    // Key order is predictable but not guaranteed. We prefer false negatives (failure) to false positives.
+    assert(Object.keys(invalid)[0] !== 'or', 'Unexpected key order.');
+
+    try {
+      model._coerce(invalid);
     } catch (err) {
       error = err;
     }
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1559,7 +1559,8 @@ DataAccessObject._coerce = function(where) {
         err.statusCode = 400;
         throw err;
       }
-      return where;
+
+      continue;
     }
     var DataType = props[p] && props[p].type;
     if (!DataType) {


slnode:

slnode:

slnode:

slnode:

Amir-61:
@slnode test please

test/loopback-dl.test.js
```diff
@@ -1473,6 +1483,20 @@ describe('DataAccessObject', function() {
     assert(error, 'An error should have been thrown');
   });
 
+  it('should throw when malformed logical operators follow valid logical clauses', function() {
```
Amir-61: `throws error when malformed logical operators follow by valid logical clauses`

test/loopback-dl.test.js
```diff
@@ -1421,6 +1421,16 @@ describe('DataAccessObject', function() {
     assert.deepEqual(where, {or: [{age: 10}, {vip: true}]});
   });
 
+  it('should continue to coerce properties after a logical operator', function() {
+    var clause = {and: [{age: '10'}], vip: 'true'};
+
+    // Key order is predictable but not guaranteed. We prefer false negatives (falure) to false positives.
```
Amir-61: typo: `falure` -> `failure`

Amir-61:
There is a commit linter issue:
```
**  Linting commit logs
**
**  1 problems found:
**    f42aff0 - DataAccessObject._coerce now continues after encoun: First line should be 50 characters or less (saw 76)
```

We follow [these rules](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) for commit descroiption

Amir-61:
This patch looks almost good to me... @superkhau What's your take on it?

Amir-61:
@doublemarked Also please rebase the changes on top of master and squash commits as one single commit

test/loopback-dl.test.js
```diff
@@ -1473,6 +1483,20 @@ describe('DataAccessObject', function() {
     assert(error, 'An error should have been thrown');
   });
 
+  it('should throw when malformed logical operators follow valid logical clauses', function() {
+    var invalid = {and: [{x: 1}], or: 'bogus'};
+
+    // Key order is predictable but not guaranteed. We prefer false negatives (falure) to false positives.
```
superkhau: falure -> failure

superkhau:
I don't see any obvious issues -- @bajtos Second opinion please.

bajtos:
The fix makes sense to me, thank you @doublemarked for taking the effort to send us a pull request! I'll leave detailed review to @Amir-61 and @superkhau.

doublemarked:
Thanks for your feedback, @bajtos. I'll clean up things according to the other comments and resubmit.

test/loopback-dl.test.js
doublemarked: Hey @Amir-61, I'm happy to make this change, however - this seems inconsistent with the grammar of the rest of the tests which all include the 'should' prefix. What do you advise?

Amir-61:
@slnode test please

test/loopback-dl.test.js
Amir-61: @doublemarked unfortunately in many places tests start with `should` and we should eventually make them consistent. For now as a team we decided to follow [these documented rules](https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/contrib/style-guide.md#test-naming) where it says: 
```
Use imperative mood, do not start test names with should.
```

Lets follow them.

Thanks :-)

superkhau: @doublemarked I would say don't worry about it since you're are following the same consistency within a file.

@Amir-61 I noticed this issue coming a lot in PRs. We should just accept the change if they are consistent within the file and send follow up PRs to remove all the "shoulds" before enforcing it on the community. @bajtos Thoughts?

Amir-61: @superkhau,

I would personal say we should not care about whether test starts with `should` or not, however I have seen you are so careful about that in PRs and have documented [here](https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/contrib/style-guide.md#test-naming), so I followed that. If it's me I would say most of the tests currently start with `should` and we should follow that or at least we should follow what the majority of tests use in that specific file and follow them to have the tests in that file consistent.

I also propose to delete the following like from https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/contrib/style-guide.md#test-naming:

```
Use imperative mood, do not start test names with should.
```

superkhau: @Amir-61 Those guidelines were carefully documented by @bajtos as how we would like to move forward. That said, the files in our current codebase does not agree -- ie. we didn't submit PRs to all 60+ LB projects to fix this when creating this guideline. The problem is users tend to follow the styles already existing in the current codebase as best practice in general when submitting PRs to any project. So IMO, the only way to "ease" in the rule is to relax the requirement of forcing everyone to clean our house before we even do it -- so we don't end up with two styles in one file too (of course some of this has happened already -- we should fix those ones too).

doublemarked:
Those Jenkins links are not loading, making it difficult for me to understand the source of these failures.

Amir-61:
@slnode test please

Amir-61:
```
npm ERR! 404 no such package available : karma
npm ERR! 404 
npm ERR! 404 'karma' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 It was specified as a dependency of 'loopback'
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```
I just see the above ^ failure which is a glitch based on what @rmg said. I have re-built so many times and still we get this error; @rmg could you clear registry. it may help. 🙏

rmg:
@slnode test please

superkhau:
>I personally prefer to clean things incrementally. New code follows the new guidelines. The old code is updated as part of related changes. That way the parts of code that we look at often are cleaned soon. The parts that don't need changes can stay in the old style and it that's fine, because nobody is working with that code anyways.

That's the problem, there ARE people working on that part of the code and automatically follow what's already in the file (it ALREADY happened in this case and many others). 

>The other approach - doing a big one-time clean up is IMO a waste of time.

That's just it, I also prefer the incremental update approach (and that works fine internally because we all know the rules). The problem is everyone else who is working on the code is assuming otherwise (and they don't read through all the pages in our style guide off the bat before submitting PRs). Which leads to the next problem -- we're gonna have to explain to everyone that we do N and you can see why in the style guide anyways. After that comes "but it's already like that in the code" and then we have to explain why some code has the old vs new style, etc. This goes for most PRs I notice that modify (or add) test case messages. I see a few choices:

1. Do what we do now and explain for each incoming PR that modifies/adds tests
2. Land PRs with the updates ourselves that modify the messages (so we don't have to explain)
3. Accept either or and fix it when we decide to clean up that file (if ever)

>My trouble with this approach is that new additions add to the overall mess, which makes the task of doing a big cleanup even more time consuming and thus less likely to happen at all.

I don't have a good answer TBH, but I think repeating ourselves for all PRs is a huge time waster. At the same time I also agree with your other view that going back and changing everything is a huge time waster also. Thoughts?

@bajtos @strongloop/loopback-dev

bajtos:
> I don't have a good answer TBH, but I think repeating ourselves for all PRs is a huge time waster

I agree it costs us time. On the other hand, it's also a good opportunity to educate new contributors, which makes the cost worth it IMO.

Amir-61:
@slnode test please

Amir-61:
@doublemarked Landed. Thanks for the contribution! :-)

superkhau:
>On the other hand, it's also a good opportunity to educate new contributors, which makes the cost worth it IMO.

I agree this is a good chance to introduce devs (new and old) to the style guide/newly revamped contributing code sections on loopback.io. Seems like the costs are justified then, TY.

PR #1143, opened 2016-10-18
Coercing conditions with array-like objects into arrays

doublemarked:
Fixes https://github.com/strongloop/loopback/issues/2824

Please let me know if this is aligned with your goals for `DataAccessObject._coerce`.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1526,6 +1526,27 @@ function NumberType(val) {
   return !isNaN(num) ? num : val;
 }
 
+function coerceArray(val) {
+  if (Array.isArray(val)) {
+    return val;
+  }
+
+  if (!utils.isPlainObject(val)) {
+    throw new Error(g.f('Value is not an {{array}} or {{object}} with sequential numeric indices'));
+  }
+
+  var arrayVal = new Array(Object.keys(val).length);
+  for (var i = 0; i < arrayVal.length; ++i) {
+    if (!val.hasOwnProperty(i)) {
+      throw new Error(g.f('Value is not an {{array}} or {{object}} with sequential numeric indices'));
+    }
+
+    arrayVal[i] = val[i];
+  }
+
+  return arrayVal;
+}
+
 /*
  * Coerce values based the property types
  * @param {Object} where The where clause
@@ -1550,14 +1571,16 @@ DataAccessObject._coerce = function(where) {
     // Handle logical operators
     if (p === 'and' || p === 'or' || p === 'nor') {
       var clauses = where[p];
-      if (Array.isArray(clauses)) {
-        for (var k = 0; k < clauses.length; k++) {
-          self._coerce(clauses[k]);
-        }
-      } else {
-        err = new Error(g.f('The %s operator has invalid clauses %j', p, clauses));
+      try {
+        clauses = coerceArray(clauses);
+      } catch (e) {
+        err = new Error(g.f('The %s operator has invalid clauses %j: %s', p, clauses, e.message));
         err.statusCode = 400;
         throw err;
+      }
+
+      for (var k = 0; k < clauses.length; k++) {
+        self._coerce(clauses[k]);
       }
 
       continue;
@@ -1614,15 +1637,21 @@ DataAccessObject._coerce = function(where) {
           switch (operator) {
             case 'inq':
             case 'nin':
-              if (!Array.isArray(val)) {
-                err = new Error(g.f('The %s property has invalid clause %j', p, where[p]));
+            case 'between':
+              try {
+                val = coerceArray(val);
+              } catch (e) {
+                err = new Error(g.f('The %s property has invalid clause %j: %s', p, where[p], e));
                 err.statusCode = 400;
                 throw err;
               }
-              break;
-            case 'between':
-              if (!Array.isArray(val) || val.length !== 2) {
-                err = new Error(g.f('The %s property has invalid clause %j', p, where[p]));
+
+              if (operator === 'between' && val.length !== 2) {
+                err = new Error(g.f(
+                  'The %s property has invalid clause %j: Expected precisely 2 values, received %d',
+                  p,
+                  where[p],
+                  val.length));
                 err.statusCode = 400;
                 throw err;
               }
@@ -1632,7 +1661,10 @@ DataAccessObject._coerce = function(where) {
             case 'ilike':
             case 'nilike':
               if (!(typeof val === 'string' || val instanceof RegExp)) {
-                err = new Error(g.f('The %s property has invalid clause %j', p, where[p]));
+                err = new Error(g.f(
+                  'The %s property has invalid clause %j: Expected a string or RegExp',
+                  p,
+                  where[p]));
                 err.statusCode = 400;
                 throw err;
               }
@@ -1649,6 +1681,14 @@ DataAccessObject._coerce = function(where) {
         }
       }
     }
+
+    try {
+      // Coerce val into an array if it resembles an array-like object
+      val = coerceArray(val);
+    } catch (e) {
+      // NOOP when not coercable into an array.
+    }
+
     // Coerce the array items
     if (Array.isArray(val)) {
       for (var i = 0; i < val.length; i++) {
--- a/test/loopback-dl.test.js
+++ b/test/loopback-dl.test.js
@@ -1412,6 +1412,53 @@ describe('DataAccessObject', function() {
     assert.deepEqual(where, {and: [{age: 10}], vip: true});
   });
 
+  const COERCIONS = [
+    {
+      in: {scores: {0: '10', 1: '20'}},
+      out: {scores: [10, 20]},
+    },
+    {
+      in: {and: {0: {age: '10'}, 1: {vip: 'true'}}},
+      out: {and: [{age: 10}, {vip: true}]},
+    },
+    {
+      in: {or: {0: {age: '10'}, 1: {vip: 'true'}}},
+      out: {or: [{age: 10}, {vip: true}]},
+    },
+    {
+      in: {id: {inq: {0: 'aaa', 1: 'bbb'}}},
+      out: {id: {inq: ['aaa', 'bbb']}},
+    },
+    {
+      in: {id: {nin: {0: 'aaa', 1: 'bbb'}}},
+      out: {id: {nin: ['aaa', 'bbb']}},
+    },
+    {
+      in: {scores: {between: {0: '0', 1: '42'}}},
+      out: {scores: {between: [0, 42]}},
+    },
+  ];
+
+  COERCIONS.forEach(coercion => {
+    var inStr = JSON.stringify(coercion.in);
+    it('coerces where clause with array-like objects ' + inStr, () => {
+      assert.deepEqual(model._coerce(coercion.in), coercion.out);
+    });
+  });
+
+  const INVALID_CLAUSES = [
+    {scores: {inq: {0: '10', 1: '20', 4: '30'}}},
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

davidcheung:
in the light of this would it be better if we handle this earlier in the cycle where the query object is built, then we can make it configurable whether user wants their `array-like-objects` to be converted to `arrays`

davidcheung:
@slnode ok to test

davidcheung:
thanks @doublemarked! the changes LGTM, can you fix the commit title to be under 50chars? 

@Amir-61 can you please take a look too?  some line seems quite long, but I believe juggler has more lenient linting conventions, can you confirm?

doublemarked:
@davidcheung thanks for your feedback. Yes, happy to fix the conflicts and improve linting compliance. Let's see if @Amir-61 has anything to add.

lib/dao.js
```diff
@@ -1526,6 +1526,27 @@ function NumberType(val) {
...
+function coerceArray(val) {
+  if (Array.isArray(val)) {
+    return val;
+  }
+
+  if (!utils.isPlainObject(val)) {
+    throw new Error(g.f('Value is not an {{array}} or {{array-like object}}'));
+  }
+
+  var arrayVal = new Array(Object.keys(val).length);
```
bajtos: If I understand this correctly, then this is opening a security vulnerability, where the attacker can send an array-like object `{ 4000000000: true }` and we will create a 4GB array containing a single item at a large index. This is reverting the prevention measures implemented by qs.

https://github.com/strongloop/loopback/issues/2824#issuecomment-254535482
> Also in that Issue they reveal that arrayLength exists to prevent params like `key[999999]` auto-vivifying an array of that length, which I'll keep in mind to avoid as well.

I think we need to find a different solution here.

doublemarked: @bajtos Are you sure? I tried to implement this in a way that specifically avoids such an issue.

```
Scallop:Projects heath$ node -v
v6.9.1
Scallop:Projects heath$ node
> keys = Object.keys({ 4000000000: true })
[ '4000000000' ]
> a = new Array(keys.length)
[  ]
```

bajtos: @doublemarked 

```
$ node
> const arg = { 4000000000: true };
undefined
> var result = new Array(Object.keys(arg).length)
undefined
> result[4000000000]=true;
true
> result
[ ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ,
  ... 3999999901 more items ]
```

bajtos: Fortunately, V8 seems smart enough to allocate a sparse array where empty items do not take any (or very little) memory.

However, notice that the maximum array size is limited:

```
> new Array(5000000000)
RangeError: Invalid array length
```

In general, I am not comfortable overriding `qs` decisions about security in LoopBack. I mean if the team working on `qs` believe that automatically converting indexed query parameters into an array is a security vulnerability, then I trust them and don't want to undo their security-hardening measurements in loopback. If you believe they are wrong and it's ok to allow arbitrarily large arrays, then please talk to them and convince them to remove this restriction.

doublemarked: Hey @bajtos :) So, there are two aspects here, the code's behavior and parity with qs, both of which I'll address.

Regarding the code from this PR - I believe that this code in fact specifically avoids the issue you're concerned about, but in the very least, the fact that this is not clear means that something needs to change here. That said, let me decompose the code a bit to explain it.

The code operates on an assumption that a well-formed "array-like object" contains a series of sequential numeric indices starting from zero. It initializes the `arrayVal` array with the length of the keys from the source content (`Object.keys(arg).length`, which is not an arbitrarily large number), and is very careful to never perform an assignment operation like that from your example (`result[4000000000]=true`).

To avoid assignments like that, it iterates from 0 up until the keys length, and hard fails in case of any missing or unexpected keys. In addition to protecting against the arbitrary-sized key, it also rejects any objects that are cluttered with unexpected non-sequential non-numeric keys.  Does this make sense?

Now, regarding qs's approach - yeah generally I agree that it's better to avoid the potential for subtle discrepancies. However, they seem to have decided to entirely disable this array parsing by default in a future release, meaning even single element array-like parameters would be represented as objects. The premise behind this PR was to anticipate that upcoming change, and my implementation aims to provide a more conservative approach to array processing than `qs` is currently performing. 

So it seems like there are several routes to go here regarding qs's arrayLimit,
   1. Do nothing, however this will be a decision to (in the future) disable by default support for array-like parameters (`filter[where][someprop][inq][0]=...`).
   2. Pick more a reasonable default for `arrayLimit` as part of the default Loopback config. Currently it defaults to 20, which is easy to unexpectedly stumble into (potentially disastrously). A substantially larger value (1000?) would create no practical risk and likely avoid the problem in almost all cases.
   3. A patch like the one from this PR, which seeks to provide support for array-like objects in the future.

Thoughts?

bajtos: (face palm) I misunderstood what your code was doing, thank you for detailed explanation! Your proposal looks very reasonable now.

I'll be on vacation in the next few days, I'll try to review this in more details later next week, so that we can move your pull request forward. Sorry for the delay!

lib/dao.js
```diff
@@ -1526,6 +1526,27 @@ function NumberType(val) {
...
+  for (var i = 0; i < arrayVal.length; ++i) {
+    if (!val.hasOwnProperty(i)) {
+      throw new Error(g.f('Value is not an {{array}} or {{array-like object}}'));
```
bajtos: If I understand this code (and the test cases) correctly, when processing my example `result[4000000000]=true`, you will only access `result[0]` and bail out with an error.

I am wondering whether we should iterate over all keys in the input `val`, thus converting `result[4000000000]=true` to a single-item array `[true]`?

```js
var arrayVal = [];
for (var key in val) {
  arrayVal.push(val[key]);
}
return arrayVal;
```

If we want to reject such requests, then I would like to see a better error message reported back to the user. Right now, `coerceArray({ 40000: true })` fails with `Value is not an array or array-like object`, which is misleading to me, because the input (`result[40000]=true`) is an array-like object.

@doublemarked Thoughts?

lib/dao.js
```diff
@@ -1550,15 +1571,18 @@ DataAccessObject._coerce = function(where) {
...
       var clauses = where[p];
-      if (Array.isArray(clauses)) {
-        for (var k = 0; k < clauses.length; k++) {
-          self._coerce(clauses[k]);
-        }
-      } else {
+      try {
+        clauses = coerceArray(clauses);
+      } catch (e) {
         err = new Error(g.f('The %s operator has invalid clauses %j', p, clauses));
```
bajtos: Please log the `e` error (at least via `debug`), so that users have means how to investigate what exactly is wrong with their clauses.

lib/dao.js
```diff
@@ -1613,14 +1637,16 @@ DataAccessObject._coerce = function(where) {
           switch (operator) {
             case 'inq':
             case 'nin':
-              if (!Array.isArray(val)) {
+            case 'between':
+              try {
+                val = coerceArray(val);
+              } catch (e) {
```
bajtos: Ditto - please log `e`. I think it may be best to include it in the error message?

```js
debug('Cannot coerce array clause: ', e);
err = new Error(g.f('The %s property has invalid clause %j: %s', p, where[p], e.message));
```

test/loopback-dl.test.js
```diff
@@ -1421,6 +1421,44 @@ describe('DataAccessObject', function() {
     assert.deepEqual(where, {or: [{age: 10}, {vip: true}]});
   });
 
+  it('should be able to coerce where clauses with array-like objects', function() {
+    where = model._coerce({scores: {0: '10', 1: '20'}});
+    assert.deepEqual(where, {scores: [10, 20]});
+
```
bajtos: One assert per test please. Split this test into multiple smaller tests, each test calling `_coerce` + `assert` only once.

(I do realise the existing code does not follow this best practice.)

test/loopback-dl.test.js
```diff
@@ -1421,6 +1421,44 @@ describe('DataAccessObject', function() {
...
+  });
+
+  it('should throw on malformed array-like objects for array-only operators', function() {
+    var invalid = [
+      {scores: {inq: {0: '10', 1: '20', 4: '30'}}},
+      {scores: {inq: {0: '10', 1: '20', bogus: 'true'}}},
+      {scores: {between: {0: '10', 1: '20', 2: '30'}}},
+    ];
+
+    invalid.forEach(function(where) {
```
bajtos: Ditto, please split this into multiple tests. You can write a parameterised test case, e.g.

```js
context('malformed array-like objects for array-only operators', function() {
  var INVALID_WHERES = [
    {scores: {inq: {0: '10', 1: '20', 4: '30'}}},
    {scores: {inq: {0: '10', 1: '20', bogus: 'true'}}},
    {scores: {between: {0: '10', 1: '20', 2: '30'}}},
  ];

  INVALID_WHERE.forEach(function(where) {
    it('should reject ' + JSON.stringify(where), function() {
      expect(function() { model._coerce(where) })
        .to.throw(/*...*/);
    });
  });
});
```

lib/dao.js
doublemarked: > If I understand this code (and the test cases) correctly, when processing my example result[4000000000]=true, you will only access result[0] and bail out with an error.

Yes, correct!

> I am wondering whether we should iterate over all keys in the input val, thus converting result[4000000000]=true to a single-item array [true]?

My objective when writing this was to maintain a strict understanding of an "array-like object" to limit the opportunity for subtle bugs, as I didn't see any circumstance where that form of input data is produced that is not the result of some mistake. In particular I was concerned that the presence of non-index keys was an indication that the object was not intended to be interpreted as an array.

But honestly, looking at this again, the rather few places that arrays are permitted (`inq`, `nin`,`between`, `and`, `or`) leaves little room for unintended consequences. One ambiguity I see is with `between`, which depends on correct element order, but a key sort would probably match intended results. 

So anyways, in summary - I'd be happy to refactor the patch to loosen the parsing a bit if you feel that would be an improvement.

bajtos: To be honest, I don't have a strong opinion. Considering that it's usually easier to loosen a strict implementation, than to make a loose implementation more strict, I think we should proceed with your current proposal.

bajtos:
@doublemarked what's the status of this pull request? Are you waiting for any information from me? I'll be on vacation from Friday to the end of the year, I would like to resolve any blockers before I leave.

doublemarked:
@bajtos yes, I was waiting on your feedback before cracking the code back open to fix the more minor things. I should be able to do this by EOD tomorrow, so we can try to get it closed out before your leave. Thanks!

bajtos:
Landed, thank you for the contribution! 🙇

Issue #1137, opened 2016-10-13
Backport latest ESLint conventions


superkhau:
Use `eslint: ^2.13.1` and `eslint-config-loopback: ^4.0.0` to make backporting easier.

Amir-61:
Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/1146

PR #1146, opened 2016-10-19
2.x eslint

Amir-61:
* Use eslint: ^2.13.1 and eslint-config-loopback: ^4.0.0 to make backporting easier.

Connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1137

/to: @superkhau

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,8 +6,9 @@
       "ignoreUrls": true,
       "ignorePattern": "^\\s*var\\s.+=\\s*(require\\s*\\()|(/)"
     }],
-    // NOTE(bajtos) we should eventuall remove this override
+    // NOTE we should eventually remove this override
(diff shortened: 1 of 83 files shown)


.eslintrc
```diff
@@ -6,8 +6,9 @@
       "ignoreUrls": true,
       "ignorePattern": "^\\s*var\\s.+=\\s*(require\\s*\\()|(/)"
     }],
-    // NOTE(bajtos) we should eventuall remove this override
+    // NOTE we should eventually remove this override
     // and fix all of those 100+ violations
-    "one-var": "off"
+    "one-var": "off",
```
superkhau: I believe we don't need this rule override anymore, please confirm and send another PR to remove it if linting works without it.

Issue #1135, opened 2016-10-11
There is no `loaded` hook if `near` filter is used for `DAO.find()`


Amir-61:
There is no `loaded` hook if `near` filter is used for `DAO.find()`

Amir-61:
Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/1108

PR #1149, opened 2016-10-24
Add support for `loaded` operation hook for DAO.find() when near is used

Amir-61:
* Add support for `loaded` operation hook for DAO.find() when near is used

Connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1135

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1821,7 +1821,7 @@ DataAccessObject.find = function find(query, options, cb) {
       }
 
       function queryGeo(query) {
-        function geoCallback(err, data) {
+        function geoCallbackWithoutNotify(err, data) {
           var memory = new Memory();
           var modelName = self.modelName;
 
@@ -1847,6 +1847,29 @@ DataAccessObject.find = function find(query, options, cb) {
           }
         }
 
+        function geoCallbackWithNotify(err, data) {
+          if (err) return cb(err);
+
+          async.map(data, function(item, next) {
+            var context = {
+              Model: self,
+              data: item,
+              isNewInstance: false,
+              hookState: hookState,
+              options: options,
+            };
+
+            self.notifyObserversOf('loaded', context, function(err) {
+              if (err) return next(err);
+              next(null, context.data);
+            });
+          }, function(err, results) {
+            if (err) return cb(err);
+            geoCallbackWithoutNotify(null, results);
+          });
+        }
+
+        var geoCallback = options.notify === false ? geoCallbackWithoutNotify : geoCallbackWithNotify;
         if (connector.all.length === 4) {
           connector.all(self.modelName, {}, options, geoCallback);
         } else {
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -6,10 +6,12 @@
 'use strict';
 var ValidationError = require('../').ValidationError;
 
+var async = require('async');
 var contextTestHelpers = require('./helpers/context-test-helpers');
 var ContextRecorder = contextTestHelpers.ContextRecorder;
 var deepCloneToObject = contextTestHelpers.deepCloneToObject;
 var aCtxForModel = contextTestHelpers.aCtxForModel;
+var GeoPoint = require('../lib/geo.js').GeoPoint;
 
 var uid = require('./helpers/uid-generator');
 var getLastGeneratedUid = uid.last;
@@ -24,7 +26,7 @@ module.exports = function(dataSource, should, connectorCapabilities) {
   }
   describe('Persistence hooks', function() {
     var ctxRecorder, hookMonitor, expectedError;
-    var TestModel, existingInstance;
+    var TestModel, existingInstance, GeoModel;
     var migrated = false;
 
     var undefinedValue = undefined;
@@ -41,12 +43,25 @@ module.exports = function(dataSource, should, connectorCapabilities) {
         extra: {type: String, required: false},
       });
 
+      GeoModel = dataSource.createModel('GeoModel', {
+        id: {type: String, id: true},
+        name: {type: String, required: false},
+        location: {type: GeoPoint, required: false},
+      });
+
       uid.reset();
 
       if (migrated) {
-        TestModel.deleteAll(done);
+        async.series([
+          function(cb) {
+            TestModel.deleteAll(cb);
+          },
+          function(cb) {
+            GeoModel.deleteAll(cb);
+          },
+        ], done);
       } else {
-        dataSource.automigrate(TestModel.modelName, function(err) {
+        dataSource.automigrate([TestModel.modelName, 'GeoModel'], function(err) {
           migrated = true;
           done(err);
         });
@@ -64,7 +79,12 @@ module.exports = function(dataSource, should, connectorCapabilities) {
 
           TestModel.create({name: 'second'}, function(err) {
             if (err) return done(err);
-            done();
+            var location1 = new GeoPoint({lat: 10.2, lng: 6.7});
+            GeoModel.create([
+              {name: 'Rome', location: location1},
+            ], function(err) {
+              done(err);
+            });
           });
         });
       });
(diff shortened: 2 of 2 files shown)


Amir-61:
## Please note:

https://github.com/strongloop/loopback-datasource-juggler/pull/996 is implemented in a way that for `DAO.find()` without `near` filter, it triggers `loaded` hook for each item when the result is an array. To clarify please see [here](https://github.com/strongloop/loopback-datasource-juggler/blob/d9a736a4a8e5eae8ca1f86625909861d1a9d121c/lib/dao.js#L1862-L1944) that `withNotify` methods is inside `async.map` for each item, which I believe is right the logic.

## What is the problem:
For a connector like `loopback-connector-mongodb` [the added test](https://github.com/strongloop/loopback-datasource-juggler/pull/1149/files#diff-145ab49855b38984ba190919e6e7397dR109) where we expect `access` and `loaded` hooks to be triggered in my current PR#1149 works, because it does not have `buildNearFilter` and `supportsGeo` is false [here](https://github.com/strongloop/loopback-datasource-juggler/blob/d9a736a4a8e5eae8ca1f86625909861d1a9d121c/lib/dao.js#L1796), so it goes to the other branch where I added support for `loaded` hook.

However, for memeory connector where it has `buildNearFilter` and  `supportsGeo` is True it goes [here](https://github.com/strongloop/loopback-datasource-juggler/blob/d9a736a4a8e5eae8ca1f86625909861d1a9d121c/lib/dao.js#L1862-L1948) where `loaded` hook gets triggered for each item of the array result set based on the logic in https://github.com/strongloop/loopback-datasource-juggler/pull/996. and for added test [here](https://github.com/strongloop/loopback-datasource-juggler/pull/1149/files#diff-145ab49855b38984ba190919e6e7397dR109), since the result set is an empty array, no `loaded` operation hook gets triggered. and test fails for memory connector.

## Attempted Solutions:
#### 1:
Tried to change the test case in a way to populate the TestModel with some properties of type `GeoPoint` and then use `DAO.find()` with near filter not to get an empty array result; however to use the `GeoPoint` type I need to require `loopback`, which does not make sense that much to me:
```
        var location1 = new loopback.GeoPoint({lat: 10.2, lng: 6.7});
        var location2 = new loopback.GeoPoint({lat: 10.4, lng: 5.8});
        var location3 = new loopback.GeoPoint({lat: 10.3, lng: 5.9});
        TestModel.create([
          {location: location1},
          {location: location2},
          {location: location3}
        ]);
```
#### 2: 
I check whether the connector implements `buildNearFilter` or not and set the expected result based on that.


**I decided to go with the second option, I would like to hear feedback on this** 🙇

Amir-61:
@slnode test please

Amir-61:
@slnode test please

Amir-61:
@bajtos PTAL

bajtos:
> however to use the GeoPoint type I need to require loopback, which does not make sense to add loopback as a dependency for just this simple test:

`GeoPoint` is defined here in juggler and exported in juggler's public API, see
https://github.com/strongloop/loopback-datasource-juggler/blob/5091f7141674554e25c49943f74c13202f7320c5/index.js#L13

test/persistence-hooks.suite.js
```diff
@@ -106,7 +106,11 @@ module.exports = function(dataSource, should, connectorCapabilities) {
         };
         TestModel.find(query, function(err, list) {
           if (err) return done(err);
-          hookMonitor.names.should.eql(['access']);
+          if (dataSource.connector.buildNearFilter)
+            hookMonitor.names.should.eql(['access']);
```
bajtos: I am confused. AFAICT, there is no `TestModel` instance created with `location` set. How comes that a connector is returning an non-empty result then? Is this an interference between tests? In that case I think we should either change the model name to a unique value or call `deleteAll` in a before/beforeEach hook.

bajtos:
> where loaded hook gets triggered for each item of the array result set

IMO, this is correct behaviour. Could you please add a unit-test to describe it and prevent regressions in the future?

lib/dao.js
```diff
@@ -1846,10 +1846,40 @@ DataAccessObject.find = function find(query, options, cb) {
...
+              if (err) return next(err);
+              next(null, context.data);
+            });
+          }, function(err, results) {
+            if (err) return cb(err);
+            geoCallbackWithoutNotify(null, results);
+          });
+        }
+
+        if (options.notify === false) {
```
bajtos: I am wondering if it's worth to simplify these four cases as follows:

```js
var geoCallback = options.notify ? geoCallbackWithNotify : geoCallbackWithoutNotify;
if (connector.all.length === 4) {
  connector.all(self.modelName, {}, options, geoCallback);
} else {
  connector.all(self.modelName, {}, geoCallback);
}
```

Thoughts?

test/persistence-hooks.suite.js
```diff
@@ -41,12 +42,19 @@ module.exports = function(dataSource, should, connectorCapabilities) {
...
+        id: {type: String, id: true},
+        name: {type: String, required: false},
+        location: {type: GeoPoint, required: false},
+      });
+
       uid.reset();
 
       if (migrated) {
-        TestModel.deleteAll(done);
+        TestModel.deleteAll();
```
bajtos: You should wait for `TestModel.deleteAll` to complete too. E.g. using `async.series`.

lib/dao.js
Amir-61: @bajtos Thanks for this great suggestion; I love it 👍  however please note the right code would be the following:
```
        var geoCallback = options.notify === false ? geoCallbackWithoutNotify : geoCallbackWithNotify;
        if (connector.all.length === 4) {
          connector.all(self.modelName, {}, options, geoCallback);
        } else {
          connector.all(self.modelName, {}, geoCallback);
        }
```

The reason is if options.notify is `undefined` we want to go for `geoCallbackWithNotify`, so we need to check it's value with `false`.

Thanks for this great suggestion which makes the code cleaner :-)

Amir-61:
@slnode test please

Amir-61:
>I am concerned that DataAccessObject.find is already way too long and we are making it even longer. I am proposing to schedule a task to refactor this method and split it into smaller chunks. Ideally, the work should start immediately after this patch is landed.

Yea I totally agree with you; thats the case for most of the methods in DAO where we have a chain of Operation Hooks; thats even worse for `DAO.find()` where we are dealing with geo queries; I created https://github.com/strongloop/loopback-datasource-juggler/issues/1152 for refactroing `DAO.find()`; however with that being said I'm not sure how much room we have to make it more clear, because there are already many chunks of logics as methods for example `withNotifyGeo`, `queryGeo`, `geoCallbackWithoutNotify`, `geoCallbackWithNotify`, `allCb`, `buildResult`, `withNotify` and ...

Amir-61:
@slnode test please

Amir-61:
@bajtos PTAL

test/persistence-hooks.suite.js
```diff
@@ -22,9 +24,9 @@ module.exports = function(dataSource, should, connectorCapabilities) {
     console.warn('The connector does not support a recently added feature:' +
       ' replaceOrCreateReportsNewInstance');
   }
-  describe('Persistence hooks', function() {
+  describe.only('Persistence hooks', function() {
```
bajtos: Forgotten `.only`, please revert.

test/persistence-hooks.suite.js
```diff
@@ -41,12 +43,27 @@ module.exports = function(dataSource, should, connectorCapabilities) {
...
-        TestModel.deleteAll(done);
+        async.series([
+          function(cb) {
+            TestModel.deleteAll(cb);
+          },
+          function(cb) {
+            GeoModel.deleteAll(cb);
+          },
+        ], function(err) {
+          done(err);
```
bajtos: You can pass `done` directly as `async.series` callback:

```js
async.series([
  // ...
], done);
```

PR #1162, opened 2016-11-10
Update chinese simplified translation file

0candy:


0candy:
@Amir-61 PTAL

superkhau:
Those mssql CI fails look new -- also the cloudant ones I thought were working now @jannyHou.

PR #1165, opened 2016-11-15
Drop support for Node v0.10 and v0.12

bajtos:
Connect to strongloop/loopback#2807

cc @strongloop/loopback-dev


bajtos:
Note: juggler does not work on Node v7 - see https://github.com/strongloop/loopback-datasource-juggler/issues/1166

PR #1169, opened 2016-11-24
Allow nested properties to be coerced

nVitius:
Currently, nested properties (i.e. properties of an embedded model) will
be skipped in the coerce step. For string values, this is not an issue.
For other types, such as dates, this means that comparisons may not
work.

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -352,27 +352,47 @@ Memory.prototype.fromDb = function(model, data) {
   if (!data) return null;
   data = deserialize(data);
   var props = this._models[model].properties;
+
   for (var key in data) {
-    var val = data[key];
-    if (val === undefined || val === null) {
-      continue;
-    }
-    if (props[key]) {
-      switch (props[key].type.name) {
-        case 'Date':
-          val = new Date(val.toString().replace(/GMT.*$/, 'GMT'));
-          break;
-        case 'Boolean':
-          val = Boolean(val);
-          break;
-        case 'Number':
-          val = Number(val);
-          break;
-      }
-    }
-    data[key] = val;
+    data[key] = this._castPropertyValue(key, data[key], props);
   }
+
   return data;
+};
+
+Memory.prototype._castPropertyValue = function(prop, val, props) {
+  if (val === undefined || val === null || !props[prop]) {
+    return val;
+  }
+
+  if (Array.isArray(val)) {
+    var self = this;
+    return val.map(function(val) {
+      return self._castPropertyValue(prop, val, props);
+    });
+  }
+
+  var isArray = Array.isArray(props[prop].type);
+  var propType = isArray ? props[prop].type[0] : props[prop].type;
+
+  switch (propType.name) {
+    case 'Date':
+      val = new Date(val.toString().replace(/GMT.*$/, 'GMT'));
+      break;
+    case 'Boolean':
+      val = Boolean(val);
+      break;
+    case 'Number':
+      val = Number(val);
+      break;
+    case 'ModelConstructor':
+      for (var subProp in val) {
+        val[subProp] = this._castPropertyValue(subProp, val[subProp], propType.definition.properties);
+      }
+      break;
+  }
+
+  return val;
 };
 
 function getValue(obj, path) {
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1532,10 +1532,14 @@ function NumberType(val) {
  * @returns {Object} The coerced where clause
  * @private
  */
-DataAccessObject._coerce = function(where) {
+DataAccessObject._coerce = function(where, props) {
   var self = this;
   if (!where) {
     return where;
+  }
+
+  if (!props) {
+    props = self.definition.properties;
   }
 
   var err;
@@ -1545,7 +1549,6 @@ DataAccessObject._coerce = function(where) {
     throw err;
   }
 
-  var props = self.definition.properties;
   for (var p in where) {
     // Handle logical operators
     if (p === 'and' || p === 'or' || p === 'nor') {
@@ -1562,7 +1565,24 @@ DataAccessObject._coerce = function(where) {
 
       continue;
     }
+
+    if (p.match(/\./)) {
+      var model = p.split('.')[0];
+      var prop = p.split('.').slice(1);
+
+      if (props[model]) {
+        var clause = {};
+        clause[prop] = where[p];
+        where[p] = Array.isArray(props[model].type) ?
+          self._coerce(clause, props[model].type[0].definition.properties)[prop] :
+          self._coerce(clause, props[model].type.definition.properties)[prop];
+
+        continue;
+      }
+    }
+
     var DataType = props[p] && props[p].type;
+
     if (!DataType) {
       continue;
     }
--- a/test/memory.test.js
+++ b/test/memory.test.js
@@ -158,6 +158,7 @@ describe('Memory connector', function() {
         city: String,
         state: String,
         zipCode: String,
+        occupiedSince: Date,
         tags: [
           {
             tag: String,
@@ -502,6 +503,14 @@ describe('Memory connector', function() {
       });
     });
 
+    it('should support nested date property in query', function(done) {
+      User.find({where: {'address.occupiedSince': new Date('2016-01-01')}},
+        function(err, users) {
+          should(users.length).be.equal(1);
+          done();
+        });
+    });
+
     it('should support nested property with regex over arrays in query', function(done) {
       User.find({where: {'friends.name': {regexp: /^Ringo/}}}, function(err, users) {
         should.not.exist(err);
@@ -571,6 +580,7 @@ describe('Memory connector', function() {
             city: 'San Jose',
             state: 'CA',
             zipCode: '95131',
+            occupiedSince: new Date('2016-01-01'),
             tags: [
                 {tag: 'business'},
                 {tag: 'rent'},


slnode:

slnode:

slnode:

slnode:

nVitius:
I ran into this issue when trying to query on a date property that was nested in an embedded model. Not sure how to go about writing tests to cover this use case, so any help there would be appreciated.

nVitius:
Had a failing test because I didn't account for Arrays. Pushed code with change.

nVitius:
In trying to write a test for this, I found an issue where the memory connector wasn't casting nested properties properly. I fixed that and added a test case that covers both commits. I'm not too happy with how the test is, but it was the easiest thing that came to mind. If anyone has a better idea, I'll take it.

chrisamoore:
👍

bajtos:
Hi @nVitius, thank you for the pull request. The problem you are describing makes total sense to me, however I am not sure if your patch is fixing the right place.

IIUC, there are two parts of your patch:

The first part modifies `fromDB` to restore certain types in the data, because JSON is loosing type information. This is the part I am concerned about. I believe none of our connectors are applying this kind of type conversions yet, it's the responsibility of juggler `Model` to convert raw data from the connector to a Model instance.

In your patch, you are casting `Booleans` and `Numbers` among other types. This does not make sense to me, because JSON roundtrip (`JSON.parse(JSON.stringify(value))`) should preserve these two types. If the type returned from the database does not match the property type, then I think it's a bug in juggler that should be fixed elsewhere.

IMO, there are only two types that need to be handled specially when using JSON-based persistence: `Date` and `Buffer`. For these two, I agree it makes sense to modify the format used by the memory connector to preserve the type information.

Note that LoopBack allows free-form models, where clients can set arbitrary properties beyond those specified in model definition. (This is the default.) For these extra properties, there will be no type information available beyond what's stored in JSON.

The second part of your patch adds deep coercion for queries using nested properties. This makes perfect sense to me 👍 

@raymondfeng @ritch do you have any opinion here?

bajtos:
@slnode ok to test

nVitius:
@bajtos Thanks for the response! Regarding the changes made to the `fromDb` function:
The actual casting in the memory connector of data to Number/Bool/Date was already in the code. All my commit (https://github.com/strongloop/loopback-datasource-juggler/pull/1169/commits/4044707384ae4b30adeae51949f1c9e3512b75ff) does, is make it so that this functionality also works on array types and embedded models.  
I actually noticed this issue only after trying to write the test for the fix on the coercion step. I was able to make the search work using the mongo connector, but it wouldn't work in the tests with the memory connector. Using a debugger, I narrowed the issue down to how the properties were being cast in the `fromDb` function of the memory connector.  
I don't think this type of change would have to be implemented in any other connector, as I believe it is unique functionality to the memory connector.

lib/connectors/memory.js
```diff
@@ -351,26 +351,54 @@ Memory.prototype.fromDb = function(model, data) {
...
   data = deserialize(data);
   var props = this._models[model].properties;
+  function cast(prop, val, props) {
+    if (Array.isArray(val)) {
+      return val.map(function(val) {
+        return cast(prop, val, props);
+      });
+    }
+
+    var isArray = Array.isArray(props[prop].type);
```
bajtos: As I mentioned in my comments, `props[prop]` may be undefined when the model has `strict` mode disabled (which is the default).

lib/connectors/memory.js
```diff
@@ -351,26 +351,54 @@ Memory.prototype.fromDb = function(model, data) {
...
+        val = new Date(val.toString().replace(/GMT.*$/, 'GMT'));
+        break;
+      case 'Boolean':
+        val = Boolean(val);
+        break;
+      case 'Number':
+        val = Number(val);
+        break;
+      case 'ModelConstructor':
+        for (var subProp in val) {
```
bajtos: I still don't think this is a good idea. Could you please show me the use case where it's needed to convert plain-data objects into Model instances?

lib/connectors/memory.js
```diff
@@ -351,26 +351,54 @@ Memory.prototype.fromDb = function(model, data) {
...
+
+    return val;
+  }
+
   for (var key in data) {
     var val = data[key];
     if (val === undefined || val === null) {
       continue;
     }
     if (props[key]) {
```
bajtos: Please move this check into `cast` function:

```js
function cast(...) {
  if (!props[key]) return val;
  // cast the value
}
```

bajtos:
> The actual casting in the memory connector of data to Number/Bool/Date was already in the code. 

Oh, I see, are you referring to L360-L370? In that case, I guess it's better to keep these conversions in place for backwards compatibility.

> I don't think this type of change would have to be implemented in any other connector, as I believe it is unique functionality to the memory connector.

Yes, I agree!

lib/connectors/memory.js
```diff
@@ -351,26 +351,54 @@ Memory.prototype.fromDb = function(model, data) {
   if (!data) return null;
   data = deserialize(data);
   var props = this._models[model].properties;
+  function cast(prop, val, props) {
```
bajtos: Please move this function out of `fromDb`, perhaps into `Memory.prototype._castPropertyValue`?

nVitius: I can do this, but keep in mind that the only reason I'm moving this logic into its own function is so that it can call itself for the nested properties. The logic is still unique to the `fromDb` method and won't be used elsewhere.

nVitius:
This is an overview of what the change introduced by the PR in the memory connector actually does.  
For this example, we'll be running the `it('should support nested properties in query')` test in `memory.test.js`. The test is simple and looks like this:  
```
it.only('should support nested property in query', function(done) {
  User.find({where: {'address.city': 'San Jose'}}, function(err, users) {
    should.not.exist(err);
    users.length.should.be.equal(1);
    for (var i = 0; i < users.length; i++) {
      users[i].address.city.should.be.eql('San Jose');
    }
    done();
  });
});
```
and the User model is defined like this:
```
{
  seq: {type: Number, index: true},
  name: {type: String, index: true, sort: true},
  email: {type: String, index: true},
  birthday: {type: Date, index: true},
  role: {type: String, index: true},
  order: {type: Number, index: true, sort: true},
  vip: {type: Boolean},
  address: {
    street: String,
    city: String,
    state: String,
    zipCode: String,
    tags: [
      {
        tag: String,
      },
    ],
  },
  friends: [
    {
      name: String,
    },
  ],
  foo: {
    date: {type: Date},
  },
}
```

Now, the `find` method of the memory connector calls `Memory.prototype.fromDb` with the model definition and all the data in that collection. In `fromDb` we iterate all the entries in the collection and cast the values of each of their properties (w/ the `Model.prototype._casPropertyValue` method).

In the cast step, to the model's property list, as well as the name of the property we're looking at and its current value. In the property list (`props`), each property is an Object that has a `type` property. `type` itself has a `name` which tells us what the `type` is defined as in the model definition. Based off of that, we try to cast the value to what it should be. Now, the actual step that this patch adds is this:  

If the `name` is `ModelConstructor`, it means we're looking at an embedded object.  In that case, we have to iterate over each of its properties and cast those. The embedded object's own property list is stored in `type.definition.properties`. We take the value of each of those, along with the list, and pass it back down to the `_castPropertyValue` function. The following screenshot was taken in the cast step when looking at the `address` property of the User model:
screen shot 2016-12-07 at 10 55 06 am

lib/connectors/memory.js
bajtos: > The logic is still unique to the fromDb method and won't be used elsewhere.

Sure, that's fine. `fromDb` is called from a single method too, and yet it's a regular method.

lib/connectors/memory.js
bajtos: You are right! Sorry, I should have read the code more carefully.

lib/connectors/memory.js
```diff
@@ -352,29 +352,61 @@ Memory.prototype.fromDb = function(model, data) {
...
   data = deserialize(data);
   var props = this._models[model].properties;
+
   for (var key in data) {
     var val = data[key];
-    if (val === undefined || val === null) {
-      continue;
-    }
+
     if (props[key]) {
```
bajtos: As I mentioned in an earlier [comment](https://github.com/strongloop/loopback-datasource-juggler/pull/1169#discussion_r91296999), please move this check into `_castPropertyValue`.

lib/connectors/memory.js
```diff
@@ -352,29 +352,61 @@ Memory.prototype.fromDb = function(model, data) {
...
+  }
+
+  if (Array.isArray(val)) {
+    var self = this;
+    return val.map(function(val) {
+      return self._castPropertyValue(prop, val, props);
+    });
+  }
+
+  var isArray = Array.isArray(props[prop].type);
```
bajtos: `props[prop]` may be undefined, please handle this case too. Start by writing a failing test case, e.g.

```js
User.find({where: {'foo.extra': true}},...)
```

(notice `foo` does not have any `extra` property defined).

bajtos: I also recommend to store `props[prop]` in a new variable, so that you don't have to repeat this snippet again and again in the code below.

```js
var propType = props[prop]; // or even better: var propertyType = ...
if (!propType) return val;
// ...
```

lib/connectors/memory.js
```diff
@@ -352,29 +352,61 @@ Memory.prototype.fromDb = function(model, data) {
...
+  var name = isArray ? props[prop].type[0].name : props[prop].type.name;
```
bajtos: What if the type is an array, but the value is not? Please add test cases for both array type + array value and array type + scalar value.

lib/connectors/memory.js
```diff
@@ -352,29 +352,61 @@ Memory.prototype.fromDb = function(model, data) {
...
+    case 'Number':
+      val = Number(val);
+      break;
+    case 'ModelConstructor':
+      for (var subProp in val) {
+        val[subProp] = isArray ?
+          this._castPropertyValue(
+            subProp,
+            val[subProp],
+            props[prop].type[0].definition.properties
```
bajtos: The logic of choosing between `props[prop].type` vs. `props[prop].type[0]` is duplicated. Please extract the item type only once.

Something like this looks best to me:

```js
var isArray = ...
if (isArray) {
  propType = propType[0];
}

var name = propType.name;
switch (...) {
  // ...
  case 'ModelConstructor':
    for (var subProp in val) {
      this._castPropertyValue(subProp, val[subProp], propType.definition.properties);
    }
}
```

lib/dao.js
```diff
@@ -1562,7 +1565,23 @@ DataAccessObject._coerce = function(where) {
 
       continue;
     }
+
     var DataType = props[p] && props[p].type;
+    if (p.match(/\./)) {
```
bajtos: Please move this code above `DataType`, since `DataType` is not used for nested properties.

lib/connectors/memory.js
```diff
@@ -352,29 +352,61 @@ Memory.prototype.fromDb = function(model, data) {
...
+    case 'Date':
+      val = new Date(val.toString().replace(/GMT.*$/, 'GMT'));
+      break;
+    case 'Boolean':
+      val = Boolean(val);
+      break;
+    case 'Number':
+      val = Number(val);
+      break;
+    case 'ModelConstructor':
```
bajtos: Please add a unit-test verifying this case.

lib/dao.js
```diff
@@ -1562,7 +1565,23 @@ DataAccessObject._coerce = function(where) {
...
+      var model = p.split('.')[0];
+      var prop = p.split('.').slice(1);
+
+      if (props[model]) {
+        var clause = {};
+        clause[prop] = where[p];
+        where[p] = Array.isArray(props[model].type) ?
+          self._coerce(clause, props[model].type[0].definition.properties)[prop] :
+          self._coerce(clause, props[model].type.definition.properties)[prop];
```
bajtos: What if `props[model]` is a scalar value, e.g. `{ type: 'string' }`? Please add a unit test for this case, e.g.

```js
User.find({where: {'vip.prop': true}},...)
```

lib/dao.js
```diff
@@ -1562,7 +1565,23 @@ DataAccessObject._coerce = function(where) {
...
     var DataType = props[p] && props[p].type;
+    if (p.match(/\./)) {
+      var model = p.split('.')[0];
+      var prop = p.split('.').slice(1);
+
+      if (props[model]) {
+        var clause = {};
+        clause[prop] = where[p];
+        where[p] = Array.isArray(props[model].type) ?
+          self._coerce(clause, props[model].type[0].definition.properties)[prop] :
```
bajtos: What if `props[model]` is an array of scalar values, e.g. `{ type: ['string'] }`? Please add a unit test for this case, e.g.

```js
User.find({where: {'tags.prop': true}},...)
```

test/memory.test.js
```diff
@@ -169,6 +169,9 @@ describe('Memory connector', function() {
           name: String,
         },
       ],
+      foo: {
```
bajtos: Please use a more descriptive name than `foo`. Can we perhaps move the `date` field to `address` and call it e.g. `since`, meaning the date since the address is valid (when the person moved in)?

Then in tests, you would use `address.since` instead of `foo.date`.

lib/connectors/memory.js
nVitius: I did move this. The diff is me removing it from `fromDb`. Look at line 370

test/memory.test.js
nVitius: Yeah, I like that idea. I'll add it in along with the other test cases.

lib/connectors/memory.js
bajtos: Line 370 is `if (val === undefined || val === null) {`, I am referring to a different check - `if (props[key]) {`

nVitius: Ah, okay. I misunderstood; my bad.

lib/connectors/memory.js
nVitius: So, the properties from the "where" query aren't taken into account in the `fromDb` mehod. It's only taking the existing collection and actual model definition.

lib/connectors/memory.js
nVitius: The type will only be an array if that is what is defined in the model schema. In that case, the value should also be an array.

lib/dao.js
nVitius: Would this not be the test I added on L508 of memory.test.js?

lib/dao.js
nVitius: done

lib/connectors/memory.js
nVitius: done

lib/connectors/memory.js
nVitius: Any test that queries on a nested property will cover this. More specifically though, the cast will realistically only do anything if the property is a date. For that use-case, take a look at the test on L506 of memory.test.js

lib/connectors/memory.js
nVitius: done

lib/connectors/memory.js
```diff
@@ -352,29 +352,49 @@ Memory.prototype.fromDb = function(model, data) {
...
-    }
-    data[key] = val;
+    data[key] = this._castPropertyValue(key, data[key], props);
   }
+
   return data;
 };
 
+Memory.prototype._castPropertyValue = function(prop, val, props) {
+  if (val === undefined || val === null || !props[prop]) {
```
bajtos: See https://github.com/strongloop/loopback-datasource-juggler/pull/1169#discussion_r91471941

> I also recommend to store `props[prop]` in a new variable, so that you don't have to repeat this snippet again and again in the code below.

```js
var propertyType = props[prop]; 
if (val === undefined || val === null || !propertyType)
  return val;
// ...
```

lib/connectors/memory.js
bajtos: > So, the properties from the "where" query aren't taken into account in the fromDb method. It's only taking the existing collection and actual model definition.

I don't understand what do you mean, could you please explain in more detail?

lib/dao.js
```diff
@@ -1532,20 +1532,23 @@ function NumberType(val) {
  * @returns {Object} The coerced where clause
  * @private
  */
-DataAccessObject._coerce = function(where) {
+DataAccessObject._coerce = function(where, props) {
```
bajtos: Please rename `props` to `propertyDefinitions` and add a jsdoc entry describing this new parameter.

lib/dao.js
bajtos: That test passes with the current implementation. I need a test that is failing now and passes with your changes in place.

The test you are referring to is querying a string value, so there is no need to perform coercion. We need to query a date value instead.

lib/dao.js
bajtos: > Any test that queries on a nested property would cover this. There's already plenty of these.
  
None of them is failing with the current implementation though. Either add a test that fails with the current implementation, or simplify your changes by removing parts that are not needed to make newly added tests to pass.

See https://en.wikipedia.org/wiki/Test-driven_development#Test-driven_development_cycle

test/memory.test.js
```diff
@@ -502,6 +503,14 @@ describe('Memory connector', function() {
       });
     });
 
+    it('should support nested date property in query', function(done) {
+      User.find({where: {'address.occupiedSince': new Date('2016-01-01')}},
+        function(err, users) {
+          should(users.length).be.equal(1);
```
bajtos: When an assert like this fails, it's hard to tell what went wrong, because the error message provides very little details.

A better way is to assert on some property of the returned users, e.g. name.

```js
var actualNames = users.map(function(u) { return u.name });
should(actualNames).be.eql(['John Lennon']);
```

lib/dao.js
nVitius: In that case, the test for querying on nested date that I added covers this and fails with the current implementation.

bajtos:
Hi @nVitius, just a quick check - what's the status of this pull request? Are you waiting for any information from me? I'll be on vacation from Friday until end of year, I'd like to clarify any items that may be blocking further progress before that.

nVitius:
@bajtos I've had a busy couple of weeks and forgot about this PR. I'll finish those tests in the next couple of days. Thanks for following up!

bajtos:
@nVitius ping, are you still keen to finish this pull request? If so, then please start with rebasing the patch on top of the latest `master`.

bajtos:
Closing as abandoned. Feel free to reopen if/when you get time to work on this patch again.

PR #1172, opened 2016-11-28
Tests cleanup

Amir-61:
Changes in `test/manipulation.test.js`, `test/scope.test.js`:

* `create()` does not return the result anymore in LB 3.0 so fixed-up tests accordingly

* use `if (err) return done(err);` instead of `should.not.exist(err);` for better error reorting

* remove un-used variables

/to: @superkhau

**Update 29/11/16**:
 - Review done, just need to update PR and land.

**Update 30/11/16**:
 - Changes made, @Amir-61 will check again and merge.

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -83,9 +83,10 @@ describe('manipulation', function() {
     it('should create instance', function(done) {
       Person.create({name: 'Anatoliy'}, function(err, p) {
         p.name.should.equal('Anatoliy');
-        should.not.exist(err);
+        if (err) return done(err);
         should.exist(p);
         Person.findById(p.id, function(err, person) {
+          if (err) return done(err);
           person.id.should.eql(p.id);
           person.name.should.equal('Anatoliy');
           done();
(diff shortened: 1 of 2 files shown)


test/manipulation.test.js
```diff
@@ -1551,6 +1565,7 @@ describe('manipulation', function() {
     it('should allow delete(id) - fail with error', function(done) {
       Person.settings.strictDelete = true;
       Person.findOne(function(e, u) {
+        if (e) return done(e);
```
superkhau: `if (err) return done(err);` shoudl work here too

test/manipulation.test.js
```diff
@@ -1592,7 +1592,7 @@ describe('manipulation', function() {
         function(done) {
           var now = Date.now();
 
-          var myCustomModel = CustomModel.create(function(err, m) {
+          CustomModel.create(function(err, m) {
```
superkhau: m -> model

superkhau: for all the below too I guess

PR #1175, opened 2016-11-29
Fixed example for creating ValidationError

BostjanPisler:


slnode:

slnode:

slnode:

slnode:

Amir-61:
@slnode test please

lib/validations.js
```diff
@@ -776,7 +776,7 @@ function ErrorCodes(messages) {
  *   }
  *
  *   // This line shows how to create a ValidationError
- *   err = new ValidationError(model);
+ *   var err = new MyModel.ValidationError(model);
```
Amir-61: I believe this should be `var err = new ValidationError(model);` for example see how it is used in other places as well: https://github.com/strongloop/loopback-datasource-juggler/blob/5bf18b07280e22d713e722b8e23f8aee14c6587e/lib/relation-definition.js#L2195

BostjanPisler: Using it like that in a remote method throws a `ValidationError is not defined` error.
Using proposed method works, the ideas is from here: https://github.com/strongloop/loopback/blob/master/lib/persisted-model.js#L438

Amir-61: Thanks for clarification!

PR #1177, opened 2016-11-30
Add mysql to ignore list

jannyHou:
Triggered a CI build of loopback-connector-mysql:
http://ci.strongloop.com/job/loopback-connector-mysql/2154/node=6.x,os=windows/console
http://ci.strongloop.com/job/loopback-connector-mysql/2154/node=4.x,os=windows/console

And issue to fix it is already filed in https://github.com/strongloop-internal/scrum-loopback/issues/1104

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
       "loopback-connector-informix",
       "loopback-connector-mqlight",
       "loopback-connector-mssql",
+      "loopback-connector-mysql",
       "loopback-connector-oracle"
     ]
   }


jannyHou:
@superkhau PTAL. Thanks.

superkhau:
Rejecting as we do not need to add to downstream ignore. MySQL needs to be fixed via that issue.

superkhau:
@jannyHou I'm also gonna add back Oracle and MSSQL FYI, shouldn't have removed those in the first place: https://github.com/strongloop/loopback-datasource-juggler/pull/1178

PR #1178, opened 2016-12-01
Remove valid connectors from downstream ignore list

superkhau:
- Remove Microsoft SQL connector
- Remove Oracle connector

@jannyHou We know these connectors should be fully functional on CI, removing from downstream ignore list. The real solution is to pass those issues you create to fix CI to the relevant squad leads to get it planned/interrupted if necessary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,9 +57,7 @@
     "downstreamIgnoreList": [
       "loopback-connector-db2z",
       "loopback-connector-informix",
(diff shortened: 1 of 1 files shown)


superkhau:
@slnode test please

superkhau:
Landing as something is wrong with CI.

@rmg Commit linter and PR linter is hanging.

PR #1179, opened 2016-12-05
Upgrade eslint & config to latest

bajtos:
- eslint ^3.11.1
 - eslint-config-loopback: ^6.0.0
 - fix linter errors (mostly no-undef)

@superkhau or @Amir-61 please review

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -5,6 +5,7 @@
 
 'use strict';
 
+/* global window:false */
 var g = require('strong-globalize')();
 var util = require('util');
 var Connector = require('loopback-connector').Connector;
@@ -823,7 +824,7 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
     // Calling _createSync to update the collection in a sync way and
     // to guarantee to create it in the same turn of even loop
     return self._createSync(model, data, function(err, id) {
-      if (err) return process.nextTick(function() { cb(err); });
+      if (err) return process.nextTick(function() { callback(err); });
       self.saveToFile(id, function(err, id) {
         self.setIdValue(model, data, id);
         callback(err, self.fromDb(model, data), {isNewInstance: true});
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -6,6 +6,7 @@
 // This test written in mocha+should.js
 'use strict';
 
+/* global getSchema:false, connectorCapabilities:false */
 var should = require('./init.js');
 var async = require('async');
 var db, User;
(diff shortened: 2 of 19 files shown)


bajtos:
The failed downstream builds are unrelated AFAICT - two timeouts and an SQL error.

test/basic-querying.test.js
```diff
@@ -6,6 +6,7 @@
 // This test written in mocha+should.js
 'use strict';
 
+/* global getSchema:false, connectorCapabilities:false */
```
superkhau: We should fix (ie. not use) these globals at some point. ;) That is outside the scope of this PR though.

lib/connectors/memory.js
```diff
@@ -823,7 +824,7 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
     // Calling _createSync to update the collection in a sync way and
     // to guarantee to create it in the same turn of even loop
     return self._createSync(model, data, function(err, id) {
-      if (err) return process.nextTick(function() { cb(err); });
+      if (err) return process.nextTick(function() { callback(err); });
```
superkhau: Why not cb? I still prefer `cb` over `callback` unless we've decided to do this all over the place.

Amir-61: @bajtos @superkhau either one is fine with me; consistency is what I would follow... I did a quick search in `lib/connectors/memory.js` and it seems most of the methods use `callback`, so I would personally go for `callback` in this case; however I agree if I had the option with not lots of `callback`s in the file, I would go for `cb` rather than `callback` but it is a nitpick.

@superkhau this one actually was a bug since there was no `cb` in this method and that's why @bajtos changed it to `callback`.

test/basic-querying.test.js
bajtos: I agree with both! Should we create an issue to keep track of this work? I am not sure if this problem important enough, there is no point in creating issues that nobody will work for months.

bajtos:
Landed, thank you!

test/basic-querying.test.js
superkhau: I believe it is important to keep our house as clean as possible and as soon as possible. We should not delay these minor issues for months IMO, but I'll leave that decision up to @ritch. Issue created at https://github.com/strongloop/loopback-datasource-juggler/issues/1186 and assigned to @siddhipai

lib/connectors/memory.js
superkhau: It's a nitpick anyways, 2 commits (one for bug and one for the eslint changes) would've been better for the reviewer to determine WHY the cb/callback thing was changed since the bugfix is not a requirement of ESLint since this PR was for ESLint changes. I'm not saying make a separate PR, just another commit on top of the current changes.

PR #1180, opened 2016-12-05
Honour allowExtendedOperators in "DAO.find"

bajtos:
Modify the coercion of filter.where to hounour "allowExtendedOperators" and don't coerce property values of type object (extended operators). This allows MongoDB users to query using MongoDBs operators like `{ $exists: true, $ne: null }`.

See https://github.com/strongloop/loopback/issues/646

@raymondfeng @superkhau or @Amir-61 please review

@fabien IIRC, you are using MongoDB quite a lot. Could you please take a look too?

### DIFF
--- a/test/allow-extended-operators.test.js
+++ b/test/allow-extended-operators.test.js
@@ -1,0 +1,62 @@
+// Copyright IBM Corp. 2015,2016. All Rights Reserved.
+// Node module: loopback-datasource-juggler
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+'use strict';
+
+const DataSource = require('..').DataSource;
+const should = require('should');
+
+describe('Model.settings.allowExtendedOperators', () => {
+  context('DAO.find()', () => {
+    it('converts extended operators to string value by default', () => {
+      const TestModel = createTestModel();
+      return TestModel.find(extendedQuery()).then((results) => {
+        should(results[0].value).eql('[object Object]');
+      });
+    });
+
+    it('preserves extended operators wit allowExtendedOperators set', () => {
+      const TestModel = createTestModel({allowExtendedOperators: true});
+      return TestModel.find(extendedQuery()).then((results) => {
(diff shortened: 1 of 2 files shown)


bajtos:
The first commit is from https://github.com/strongloop/loopback-datasource-juggler/pull/1179, I'll rebase it away.

test/allow-extended-operators.test.js
```diff
@@ -0,0 +1,63 @@
...
+        should(results[0].value).eql({$exists: true});
+      });
+    });
+
+    function extendedQuery() {
+      // datasource modifies the query,
+      // we have to build a new object for each test
+      return {where: {value: {$exists: true}}};
+    }
+
```
superkhau: remove empty space

test/allow-extended-operators.test.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  all(model, filter, options, callback) {
+    // return the raw "value" query
+    var instanceFound = {
+      value: filter.where.value,
+    };
+    callback(null, [instanceFound]);
+  }
+}
```
superkhau: Why are these declared outside describe and extendedQuery is inside? Put all inside or put all outside for consistency?

Amir-61: I agree... It seems reasonable to have consistency...

Amir-61:
Nice Happy ES6 coding 🎊

bajtos:
> Nice...Happy ES6 coding 🎊  Could you please let us know what is the plan regarding ES6? Are we going to write tests in ES6 from now on?

@Amir-61 Yes, we can and should use ES6 in `master` branches now that we no longer support Node v0.10/v0.12. See https://github.com/strongloop/loopback.io/pull/187 and http://loopback.io/doc/en/contrib/style-guide.html

Note that when back-porting changes to older versions (typically 1.x and 2.x), we should stick to ES5, to make sure we don't break LoopBack for people still running on pre-v4 versions of Node.

PR #1181, opened 2016-12-05
Fix MySql CI server Failure

loay:
loopback-connector-mysql is red on juggler CI
This PR is the fix for this failure. 
This PR is prerequisite for https://github.com/strongloop/loopback-connector-mysql/pull/211

### DIFF
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -44,7 +44,7 @@ module.exports = function(dataSource, should, connectorCapabilities) {
       });
 
       GeoModel = dataSource.createModel('GeoModel', {
-        id: {type: String, id: true},
+        id: {type: String, id: true, default: uid.next},
         name: {type: String, required: false},
         location: {type: GeoPoint, required: false},
       });


loay:
@jannyHou @Amir-61 @superkhau  PTAL. Thanks.

loay:
@jannyHou 
Correct .. Here is some background:
loopback-connector-mysql server was failing on juggler here and giving this error:
`Persistence hooks "before each" hook: createTestData for "triggers hooks in the correct order":
    Error: done() invoked with non-Error: Error: ER_NO_DEFAULT_FOR_FIELD: Field 'id' doesn't have a default value,Error: ER_NO_DEFAULT_FOR_FIELD: Field 'id' doesn't have a default value`

So the possible solutions could be:
1) L46:  `id: {type: String, id: false}` but I don’t think that’s right since we might need to associate IDs or primary keys in other tests.
2)  Same as in `TestModel` line 41 ..
`  id: {type: String, id: true, generated: false, default: uid.next},`
3) simply `generated: true` so it auto increments or generates
4)  Line85:           `  {id: 1, name: 'Rome', location: location1},` -> not sure if I should do that, because in future tests, if a new instance is created without an ID, it will still fail

so I chose to add the default value, which is autogenerated, .. so solution 2 without generated` parameter.

superkhau:
Please create a follow up issue to verify autoinc/autoinc reset support in underlying driver so we can use instead of the helper uid lib.

PR #1183, opened 2016-12-06
Back-port fixes for linter errors from master

bajtos:
Back-port the relevant parts of #1179

cc @superkhau @Amir-61


bajtos:
I checked all 14 failed builds, none of them seems to be caused by my change :( I am going to ignore them and land this patch.

PR #1197, opened 2016-12-20
Fix HasOne.update to propagate options arg

bajtos:
While working on https://github.com/strongloop/loopback/pull/3023, I discovered that `HasOne.prototype.update` does not propagate the `options` argument when calling `targetModel.updateAttributes`.

This patch fixes the issue.

Connect to strongloop/loopback#1495


bajtos:
@slnode test please

bajtos:
Ignoring unrelated build failures.

PR #1201, opened 2016-12-21
Fix HasOne.update to propagate options arg [2.x]

bajtos:
Back-port #1197
Connect to strongloop/loopback#1495


bajtos:
Ignoring unrelated build failures:

> `npm ERR! 404 no such package available : semver`

PR #1204, opened 2016-12-24
Throw error when model relation name is trigger

schempy:
### Description
Defining a model relation with the name "trigger" causes the model not
able to insert records. No error is thrown when a model relation with
the name "trigger" is defined. Adding a check for the model relation
name "trigger" will now throw an error.

#### Related issues
@bajtos This was originally submitted for a fix in strongloop/loopback#2988. Per your request to apply the patch to this repository. Please review.

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -461,6 +461,8 @@ DataSource.prototype.defineRelations = function(modelClass, relations) {
     Object.keys(relations).forEach(function(rn) {
       var r = relations[rn];
       assert(DataSource.relationTypes.indexOf(r.type) !== -1, 'Invalid relation type: ' + r.type);
+      assert(isValidRelationName(rn), 'Invalid relation name: ' + rn);
+
       var targetModel, polymorphicName;
 
       if (r.polymorphic && r.type !== 'belongsTo' && !r.model) {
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

lib/datasource.js
```diff
@@ -456,11 +456,21 @@ DataSource.prototype.defineRelations = function(modelClass, relations) {
     }
   };
 
+  var hasInvalidRelationName = function(relationName) {
```
bajtos: Since this function does not need anything from `defineRealations` closure, it's better to move it to the root scope of this file.

Also `isValidRelationName` is a better name, as it prevents double-negative in `!hasInvalidRelationName`.

test/relations.test.js
```diff
@@ -5609,4 +5609,26 @@ describe('relations', function() {
       .catch(done);
     });
   });
+
+  describe('relation names', function() {
+    it('should throw error when a relation name is `trigger`', function(done) {
```
bajtos: This test is sync, therefore there is no need to use `done` argument.

lib/datasource.js
```diff
@@ -456,11 +456,21 @@ DataSource.prototype.defineRelations = function(modelClass, relations) {
...
+    var invalidRelationNames = ['trigger'];
+
+    return invalidRelationNames.some(function(element) {
```
bajtos: Since we are doing a simple value-comparison here, `Array.prototype.indexOf` is sufficient.

bajtos:
@slnode ok to test

bajtos:
Landed, thank you for the contribution! 🎉

PR #1206, opened 2016-12-31
Add flush operation to KVAO

superkhau:
Used to delete all keys (and values) associated to the current model.

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/1379

cc @bajtos

### DIFF
--- a/lib/connectors/kv-memory.js
+++ b/lib/connectors/kv-memory.js
@@ -201,6 +201,12 @@ KeyValueMemoryConnector.prototype.disconnect = function(callback) {
   process.nextTick(callback);
 };
 
+KeyValueMemoryConnector.prototype.flush =
+function(modelName, options, callback) {
+  this._store = Object.create(null);
+  callback();
+};
+
 function StoreItem(value, ttl) {
   this.value = value;
   this.setTtl(ttl);
--- a/lib/kvao/flush.js
+++ b/lib/kvao/flush.js
@@ -1,0 +1,30 @@
+'use strict';
+
+var assert = require('assert');
+var utils = require('../utils');
+
+/**
+ * Delete all keys (and values) associated to the current model.
+ *
+ * @options {Object} options Unused ATM, placeholder for future options.
+ * @callback {Function} callback
+ * @param {Error} err Error object.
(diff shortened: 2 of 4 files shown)


superkhau:
@slnode test please

superkhau:
@slnode test please

superkhau:
@slnode test please

superkhau:
Oracle is failing due to executing `./run-tests` on windows -- it is a known failure and unrelated to this PR.

bajtos:
@superkhau I am rather unhappy that addition of a new API was landed without any peer review :(

I personally find the new method name `flush` confusing and ambiguous.

Quoting google result:
> cleanse (something, especially a toilet) by causing large quantities of water to pass through it.
>*"she flushed the loo"*
> - (of a toilet) be cleansed by flushing.
>  *"Cally heard the toilet flush"*
> - remove or dispose of (an object or substance) by flushing.
> *"I flushed the pills down the lavatory"*
> - cause (a liquid) to flow through something.
> *"0.3 ml of saline is gently flushed through the tube"*

Can we use a better name please, for example one of `removeAllKeys`, `deleteAllKeys`, `clearAllKeys`, `clearDatabase`, `deleteAll`?

Considering that `lib/dao.js` provides a similar method under the names `remove`, `deleteAll` and `destroyAll`, I think `deleteAll` may be the best choice.

@ritch thoughts?

lib/connectors/kv-memory.js
```diff
@@ -201,6 +201,12 @@ KeyValueMemoryConnector.prototype.disconnect = function(callback) {
   process.nextTick(callback);
 };
 
+KeyValueMemoryConnector.prototype.flush =
+function(modelName, options, callback) {
+  this._store = Object.create(null);
```
bajtos: I forgot to submit my comments yesterday. Based on our last discussion, this comment is no longer relevant, as `flush` will be deleting only keys of a single model, i.e. the implementation will be different.

superkhau:
Discussed with @bajtos over hangouts, will be changing `flush` to `deleteAll` + various other changes.

PR #1208, opened 2016-12-31
Flush data during test setup

superkhau:
Downstream KV connector tests were returning false positives due to
improper test teardown. This fix ensures test fixtures are returned to a
clean state between each test.

Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/1208

cc @bajtos


superkhau:
Closing in favour of https://github.com/strongloop/loopback-datasource-juggler/pull/1218

Issue #1185, opened 2016-12-06
`updateOrCreate` does not work with default value of forceId: true


Amir-61:
`forceId` is set to true by default by https://github.com/strongloop/loopback-datasource-juggler/pull/982 . `updateOrCreate` does not work with default value of forceId set to true when we want to update an instance and provide the PK of instance in data object.

Failure: `ValidationError: The 'Model' instance is not valid. Details: 'id' can't be set (value: 9)`

/cc: @jannyHou @bajtos

jannyHou:
👍  Thanks.

ssh24:
Closing this. Please follow here: https://github.com/strongloop-internal/scrum-apex/issues/82

bajtos:
I am reopening this issue because of https://github.com/strongloop/loopback-datasource-juggler/pull/982#issuecomment-289599727 by @luncht1me:

>This change breaks UPSERT methods...
> When you Upsert with an ID, you want to PATCH/PUT that existing instance to update it in the database.
> And if it doesn't have an ID, then you create a new one.
>
> Upsert is just completely broken because of this id enforcement not accounting for the use-case and intended behavior of Upsert.
> Why not have Upsert check if there's an ID, if there is make sure the document exists, error if it doesn't saying id can't be set. But if the document exists, update the damn thing. Having to double the amount of code for CRUD management in our projects when there `used` to be a method that handled it seems a step backwards to me.

Here is my reply for completeness:

> I think the issue you are observing is very similar to #1235 which was fixed by #1210 and https://github.com/strongloop/loopback-connector/issues/73.
>
> Would you like to contribute the fix yourself, using the two linked pull requests for inspiration?

cc @loay @jannyHou @strongloop/sq-lb-apex

PR #1210, opened 2017-01-04
Fix forceId validation error

loay:
### Description
Fix Validation error

#### Related issues

connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1235

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -840,6 +840,11 @@ DataAccessObject.replaceOrCreate = function replaceOrCreate(data, options, cb) {
     return this.create(data, options, cb);
   }
 
+  var forceId = this.settings.forceId;
+  if (forceId) {
+    return Model.replaceById(id, data, options, cb);
+  }
+
   var inst;
   if (data instanceof Model) {
     inst = data;
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -12,6 +12,7 @@ var should = require('./init.js');
 
 var db, Person;
 var ValidationError = require('..').ValidationError;
+var bdd = require('./helpers/bdd-if.js');
 
 var UUID_REGEXP = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
 
@@ -1096,6 +1097,52 @@ describe('manipulation', function() {
       });
     });
   }
+
+  var hasReplaceById = !!getSchema().connector.replaceById;
+  bdd.describeIf(hasReplaceById, 'replaceOrCreate when forceId is true', function() {
+    var Post;
+    before(function(done) {
+      var ds = getSchema();
+      Post = ds.define('Post', {
+        title: {type: String, length: 255},
+        content: {type: String},
+      }, {forceId: true});
+      ds.automigrate('Post', done);
+    });
+
+    it('fails when id does not exist in db', function(done) {
+      var post = {id: 123, title: 'a', content: 'AAA'};
+      Post.replaceOrCreate(post, function(err, p) {
+        err.statusCode.should.equal(404);
+        done();
+      });
+    });
+
+    it('works on create if the request does not include an id', function(done) {
+      var post = {title: 'a', content: 'AAA'};
+      Post.replaceOrCreate(post, function(err, p) {
+        if (err) return done(err);
+        p.title.should.equal(post.title);
+        p.content.should.equal(post.content);
+        done();
+      });
+    });
+
+    it('works on update if the request includes an existing id in db', function(done) {
+      Post.create({title: 'a', content: 'AAA'},
+           function(err, post) {
+             if (err) return done(err);
+             post = post.toObject();
+             delete post.content;
+             post.title = 'b';
+             Post.replaceOrCreate(post, function(err, p) {
+               if (err) return done(err);
+               p.id.should.equal(post.id);
+               done();
+             });
+           });
+    });
+  });
 
   if (!getSchema().connector.replaceById) {
     describe.skip('replaceAttributes/replaceById - not implemented', function() {});
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -793,7 +793,9 @@ Memory.prototype.replaceById = function(model, id, data, options, cb) {
   var modelData = cachedModels && this.collection(model)[id];
   if (!modelData) {
     var msg = 'Could not replace. Object with id ' + id + ' does not exist!';
-    return process.nextTick(function() { cb(new Error(msg)); });
+    var error = new Error(msg);
+    error.statusCode = error.status = 404;
+    return process.nextTick(function() { cb(error); });
   }
 
   var newModelData = {};


test/manipulation.test.js
```diff
@@ -1096,6 +1096,221 @@ describe('manipulation', function() {
...
+        Post = ds.define('Post', {
+          title: {type: String, length: 255, index: true},
+          content: {type: String},
+          comments: [String],
+        }, {forceId: true});
+        ds.automigrate('Post', done);
+      });
+
+      it('works without options on create (promise variant)', function(done) {
+        var post = {id: 123, title: 'a', content: 'AAA'};
```
jannyHou: As we discussed, I think this test case should expect throwing error.

loay:
Hi @bajtos 
PTAL .. thanks

test/manipulation.test.js
```diff
@@ -1096,6 +1096,157 @@ describe('manipulation', function() {
...
+      before(function(done) {
+        Post = ds.define('Post', {
+          title: {type: String, length: 255, index: true},
+          content: {type: String},
+          comments: [String],
+        }, {forceId: true});
+        ds.automigrate('Post', done);
+      });
+
+      it('fails without options on create', function(done) {
```
bajtos: replaceOrCreate does not fail on all create calls - only when the data includes `id`. Please update the test name and also add a test for the case where the create operation succeeds.

test/manipulation.test.js
```diff
@@ -1096,6 +1096,157 @@ describe('manipulation', function() {
...
+        var post = {id: 123, title: 'a', content: 'AAA'};
+        Post.replaceOrCreate(post, function(err, p) {
+          should.exist(err);
+          err.should.match(/'Could not replace. Object with id ' + post.id + ' does not exist!'/);
+          done();
+        });
+      });
+
+      it('fails with options on create', function(done) {
```
bajtos: ditto.

> replaceOrCreate does not fail on all create calls - only when the data includes id. Please update the test name and also add a test for the case where the create operation succeeds.

test/manipulation.test.js
```diff
@@ -1096,6 +1096,157 @@ describe('manipulation', function() {
...
+      });
+
+      it('works without options on update (promise variant)', function(done) {
+        var post = {title: 'a', content: 'AAA', comments: ['Comment1']};
+        Post.create(post)
+          .then(function(created) {
+            created = created.toObject();
+            delete created.comments;
+            delete created.content;
+            created.title = 'b';
```
bajtos: Since the intention of this test is to verify how `forceId` is handled, I believe there is no need to duplicate testing of how modified/deleted properties are handled. So modify only a single property and assume other cases are already covered by existing tests.

test/manipulation.test.js
```diff
@@ -1096,6 +1096,157 @@ describe('manipulation', function() {
...
+            return Post.replaceOrCreate(created)
+            .then(function(p) {
+              should.exist(p);
+              p.should.be.instanceOf(Post);
+              p.id.should.equal(created.id);
+              p.should.not.have.property('_id');
+              p.title.should.equal('b');
+              p.should.have.property('content', undefined);
+              p.should.have.property('comments', undefined);
```
bajtos: Ditto, most of this asserts are unrelated to `forceId` settings and should be removed.

test/manipulation.test.js
```diff
@@ -1096,6 +1096,157 @@ describe('manipulation', function() {
...
+                should.not.exist(p.content);
+                should.not.exist(p.comments);
+                done();
+              });
+            });
+          })
+        .catch(done);
+      });
+
+      it('works with options on update (promise variant)', function(done) {
```
bajtos: IMO, there is no need to test all variants of `updateOrCreate`, because `forceId` flag is not affecting them. Pick a single variant and discard the rest. I would personally choose promise API + no options, as it's the simplest one.

test/manipulation.test.js
```diff
@@ -1096,6 +1096,157 @@ describe('manipulation', function() {
...
+                should.not.exist(p.content);
+                should.not.exist(p.comments);
+                done();
+              });
+            });
+          })
+        .catch(done);
+      });
+
+      it('works without options on update (callback variant)', function(done) {
```
bajtos: Ditto, this test is not needed IMO.

test/manipulation.test.js
```diff
@@ -1096,6 +1096,157 @@ describe('manipulation', function() {
...
+                p.title.should.equal('b');
+                should.not.exist(p.content);
+                should.not.exist(p.comments);
+                done();
+              });
+            });
+          });
+      });
+
+      it('works with options on update (callback variant)', function(done) {
```
bajtos: Ditto, this test is not needed IMO.

test/manipulation.test.js
```diff
@@ -1096,6 +1096,157 @@ describe('manipulation', function() {
       });
     });
   }
+  if (!getSchema().connector.replaceById) {
```
bajtos: Use `describeIf` from [test/helpers/bdd-if.js](https://github.com/strongloop/loopback-datasource-juggler/blob/66c54a96467b39609fec07f41b6d8018c5fc3819/test/helpers/bdd-if.js)

```js
var hasReplaceById = !!getSchema().connector.replaceById;
describeIf(hasReplaceById, 'replaceOrCreate when forceId is true', function() {
 // ...
});

**Comment by bajtos** [comment]
@Amir-61 I think you are most familiar with `replaceById`, could you PTAL too?

> The  error `  Error: Could not replace. Object with id 123 does not exist!` only happens in juggler but it doesn't happen in the connectors so the connectors fail the test . It only happens when an id `123` is already provided and not generated. 
Do you think that requires work within the connectors themselves within the replaceByID function as it doesn't give an error message there. thoughts?

@loay What is the test failure you are observing? Is `replaceById` failing with a different error message, or is it passing?

IMO, `replaceById` should always fail for `id` that does not exist in the database. If there is a connector that does not fail, then we have a bug to fix in such connector.

However, the tests in juggler should not rely on a specific error message text. I am proposing that the error reported by `replaceById` on unknown `id` should have `statusCode` set to 404 (Not Found) and then the tests in juggler should assert this fact. The reason for 404 is that `replaceById` is typically mapped on something like `PUT /api/customers/:id` and 404 is the right way how to signal the caller that there is no resource at the URL they provided (because of an unknown id).

File: lib/dao.js [hunk 10]
```diff
@@ -840,6 +840,11 @@ DataAccessObject.replaceOrCreate = function replaceOrCreate(data, options, cb) {
     return this.create(data, options, cb);
   }
 
+  var forceId = this.settings.forceId;
+  if (forceId && id) {
```
Amir-61: I don't think if you need to check `id` here; if `id` is `undefined` or `null` it already calls `this.create(...)` please see L840; in other words I think you need to change your condition to:
```
if(Model.settings.forceId && isOriginalQuery) {
  return self.replaceById(id, data, options, cb);
}

```

lib/dao.js
```diff
@@ -840,6 +840,11 @@ DataAccessObject.replaceOrCreate = function replaceOrCreate(data, options, cb) {
...
+    return this.replaceById(data.id, data, options, cb);
```
Amir-61: why `data.id`; L838 already set `id` to the value of PK; so the right code would be:
```
return self.replaceById(id, data, options, cb);
```

lib/dao.js
loay: Actually you do, as you can have `forceId` set to true and `id` is null or undefined, and it fails some cases .. Anyways I got to review the test cases first before I make the assumption

Amir-61:
Also it is worth mentioning using 404 HTTP status code (Not Found) on unknown id is needed not only for `replace` methods but also for `update` methods for this issue: https://github.com/strongloop/loopback-datasource-juggler/issues/1185

loay:
@bajtos @Amir-61 

So let me clarify a bit the current situation:
when forceId === true, we have 3 situations:

1) request includes id, there is a model instance with such id in the database - SUCCESS (replace)
- That works fine without a problem..  I just need to remove the unnecessary test cases in the next commit. 
2) request includes id, there is no instance with such id - ERROR 404
Which is where the case has a provided id = 123 in the added test cases. 
There are 2 problems here: 
- the error message given is not error 404 but this one: ` Error: Could not replace. Object with id 123 does not exist!`
- if I catch the error as I did in the current test cases, then the test case passes, however the fail in few connectors such as mongodb, mssql,postgresql, cloudant and couple more. The reason it fails is that there is no error message but it creates the instance when it does not exist. 

BTW, the existing code where the forceId is false, this situation does not generate an error, but it creates the instance instead. 

3) request does not include id - SUCCESS (create)
- It works fine

Amir-61:
@loay 

Regarding the second scenario:

>the error message given is : Error: Could not replace. Object with id 123 does not exist!
--> to be fixed and set the statusCode to 404 (Not Found)

This is what you are fixing just for memory connector in `loopback-datasource-juggler`, and that's just one part of the work; this needs to be fixed for dependent connectors as well; in other words, the mentioned error message and 404 HTTP status code are expected when `replaceById` is used for a non-existing id not only for memory connector but also for other connectors; otherwise other connectors fail and you cannot land this PR. The good point about `replaceById` is that for all SQL connectors the implementation is one single method in `loopback-connector`; so fixing that would fix all SQL connectors; [Here](https://github.com/strongloop/loopback-connector/blob/e384aa53ee639c35e01f9f9f9faf03ca2427271e/lib/sql.js#L605-L609) is the implementation.

CC: @bajtos

Amir-61:
@bajtos @loay 

>whether the behaviour you are observing is intentional or not?

No the behaviour is not intentional and it is definitely  bug; if they create an instance for unknown id that would be a bug; however, I'm not sure it can be skipped before landing this PR; @bajtos what do you think?

Amir-61:
> this fails in few connectors such as mongodb, mssql,postgresql, cloudant and couple more. The reason it fails is that there is no error message but it creates the instance when it does not exist.

Actually I just test with `loopback-connector-mysql` and `loopback-connector-mongodb` and replaceById **does not** create instance for non-existing Ids; however DAO for some reason is returning whatever you pass as data arguments (which is a bug, but not a big problem and should be fixed); here is how I tested @loay:
```
      customers.replaceById(999, {firstName: 'John'}, function(err, inst) {
        console.log('>replaced data from DAO:', err, inst);
        if (err) return done(err);
        customers.find(function(err, data){
          console.log('Final Data:', data);
          done(err);
        });
      });
```
Id 999 did not exist previously and thats the result: 
```
>replaced data from DAO: null { firstName: 'John', id: 999 }
Final Data: []
```

I hope that clarifies everything...

test/manipulation.test.js
```diff
@@ -1097,6 +1098,71 @@ describe('manipulation', function() {
...
+        err.statusCode.should.equal(404);
+        done();
+      });
+    });
+
+    it('works on create if the request does not include an ID', function(done) {
+      var post = {title: 'a', content: 'AAA'};
+      Post.replaceOrCreate(post, function(err, p) {
+        if (err) return done(err);
+        p.should.not.have.property('_id');
```
jannyHou: Confused why we check `_id` doesn't exist here, when will it be silently created? I understand this check means we are avoiding add it here, but curious about the reason.

test/manipulation.test.js
```diff
@@ -1097,6 +1098,71 @@ describe('manipulation', function() {
...
+        if (err) return done(err);
+        p.id.should.equal(post.id);
+        p.should.not.have.property('_id');
+        p.title.should.equal(post.title);
+        p.content.should.equal(post.content);
+        done();
+      });
+    });
+
+    it('works without options on update', function(done) {
```
jannyHou: I might miss the context of this test case, it looks same as "works on update if the request includes an exisiting ID in db", could you elaborate more which "option on update" is it checking?

jannyHou:
Hmm...wait, mysql and mssql are failing on new test cases :( 
oracle's failure is not related, a npm install error.

lib/dao.js
```diff
@@ -840,6 +840,19 @@ DataAccessObject.replaceOrCreate = function replaceOrCreate(data, options, cb) {
...
   }
 
+  var forceId = this.settings.forceId;
+  if (forceId && id) {
+    this.findById(id, function(err, instance) {
+      if (!instance) {
+        var msg = g.f('Cannot find unknown {{id}} "%s". ', id);
+        var error = new Error(msg);
+        error.statusCode = error.status = 404;
+        cb(error);
```
jannyHou: Try `return cb(err)`? mysql and mssql doesn't get the statusCode.

test/manipulation.test.js
Amir-61: This test is not needed. Please delete it.

lib/dao.js
Amir-61:
```
return cb(error);
```

test/manipulation.test.js
```diff
@@ -1097,6 +1098,71 @@ describe('manipulation', function() {
...
+            post.title = 'b';
+            Post.replaceOrCreate(post, function(err, p) {
+              if (err) return done(err);
+              p.id.should.equal(post.id);
+              done();
+            });
+          });
+    });
+  });
+
```
Amir-61: I think these added tests ☝️ are breaking the original previous tests probably because they are using the same model name called `Post`; you should either move your tests after the original tests [here](https://github.com/strongloop/loopback-datasource-juggler/blob/b530ec2a6571ef2ab16abffb6298f35c1d975691/test/manipulation.test.js#L1295) or you may consider changing the model name not to have test model conflicts; that's just my guess to fix the broken tests [here](https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-mysql~master/250/console); you need to investigate this further.

test/manipulation.test.js
Amir-61: This is added specifically for `loopback-connector-mongodb`.

loay: You are right. I was cleaning up the old test cases and I missed this one !

test/manipulation.test.js
loay: Same as the above one .. I rewrote the test cases and I forgot this one but you are right . it is the same as the previous test case and we only need one.

lib/dao.js
```diff
@@ -840,6 +840,19 @@ DataAccessObject.replaceOrCreate = function replaceOrCreate(data, options, cb) {
     return this.create(data, options, cb);
   }
 
+  var forceId = this.settings.forceId;
+  if (forceId && id) {
+    this.findById(id, function(err, instance) {
```
loay: removing the id check used to break the previous tests when forcedId was set to false, however it does not need that anymore, so I don't think I need it for now. I will keep an eye on it.

test/manipulation.test.js
```diff
@@ -1097,6 +1098,71 @@ describe('manipulation', function() {
...
+    before(function(done) {
+      Post = ds.define('Post', {
+        title: {type: String, length: 255, index: true},
+        content: {type: String},
+        comments: [String],
+      }, {forceId: true});
+      ds.automigrate('Post', done);
+    });
+
+    it('fails when id does not exist in db', function(done) {
```
loay: typo .. fixed

test/manipulation.test.js
loay: Not sure what do you mean, but it is not broken on my end .. all tests pass on my end. I will commit and see if anything breaks !

loay:
Concerning the optimal way of the fix, this fix in juggler is for the memory or the general connector. 
The `!instance` part was not meant for the other connectors. Some connectors are still failing and I agree that I still need go and fix the method in each connector individually. 
In our current situation, I still need to make the `!instance` part in order to insert the 404 error message as we agreed with Miroslav.

loay:
@bajtos PTAL again.

Now, I don't think I can merge this PR even if it is finalized till I fix the method within the connectors `mysql` and `mssql` and add the same error message there when the id doesn't exist in db, then it should detect a 404.

lib/dao.js
```diff
@@ -840,6 +840,19 @@ DataAccessObject.replaceOrCreate = function replaceOrCreate(data, options, cb) {
...
+  if (forceId) {
+    this.findById(id, function(err, instance) {
+      if (err || !instance) {
+        var msg = g.f('Cannot find unknown {{id}} "%s". ', id);
+        var error = new Error(msg);
+        error.statusCode = error.status = 404;
+        return cb(error);
+      }
+    });
+    return Model.replaceById(id, data, options, cb);
```
jannyHou: I think `return Model.replaceById(id, data, options, cb);` this line should move into the callback function of `findById`?

```
this.findById(id, function(err, instance) {
  if (err) {
    // handle error
  } else {
    return Model.replaceById(id, data, options, cb);
  }
});
```
probably can solve the connector failures.

Amir-61: ☝️ That's right!

loay: Nope, that doesn't work .. I have tried it before and it doesn't work. The only way it works is the way the code is currently written

loay:
@bajtos @Amir-61  @jannyHou 

As for Amir's comments above, 
For 1, I will work the error as we discussed and we will wait for Miroslav's opinion. 
I am submitting another commit now. 

for 2 and 4, I totally agree, and as you guys can see, I have already created issues within every affected connector , waiting for it to get prioritized so we can work on them later. 

for 3, I think @rmg  removed some of the downstreams .. this pr was being tested on around 70 servers previously but it got reduced to 32. I am not sure tho .. Maybe Ryan can confirm.

loay:
@bajtos 
Please ignore the connectors failure, because that will be fixed within loopback-connector and other connectors, but can you PTAL at  current pr as is for juggler?
Thanks

test/manipulation.test.js
```diff
@@ -1097,6 +1098,53 @@ describe('manipulation', function() {
     });
   }
 
+  var hasReplaceById = !!getSchema().connector.replaceById;
+  bdd.describeIf(hasReplaceById, 'replaceOrCreate when forceId is true', function() {
+    var Post, foundId;
+    var ds = getSchema();
```
bajtos: I think we should initialise `ds` inside a `before` hook (i.e. just before we run the tests), not at the time when `describe` is invoked. My main concern is that if `getSchema` is statefull, then the `ds` instance may become invalid by the time the tests are run, because other `before` hooks are calling `getSchema` in the meantime.

Also if you tell mocha to run a subset of tests, e.g. via `mocha -g <pattern>` or by using `it.only/describe.only`, then your current code will always create the datasource, even though "replaceOrCreate" tests are not run at all. When `getSchema()` is moved to a `before` hook, it won't be run if "replaceOrCreate" tests are skipped.

test/manipulation.test.js
```diff
@@ -1097,6 +1098,53 @@ describe('manipulation', function() {
...
+    before(function(done) {
+      Post = ds.define('Post', {
+        title: {type: String, length: 255, index: true},
+        content: {type: String},
+        comments: [String],
```
bajtos: AFAICT, `comments` field is not used in any of the tests, please remove it from the model definition.

bajtos: I think the `content` field is not needed either, and the `title` field dos not need `length` nor `index` config.

Please keep things as simple as possible, to prevent future confusion about whether this particular model configuration is required for the tests (and why?).

test/manipulation.test.js
```diff
@@ -1097,6 +1098,53 @@ describe('manipulation', function() {
...
+
+    it('works on create if the request does not include an id', function(done) {
+      var post = {title: 'a', content: 'AAA'};
+      Post.replaceOrCreate(post, function(err, p) {
+        if (err) return done(err);
+        p.title.should.equal(post.title);
+        p.content.should.equal(post.content);
+        Post.findById(p.id, function(err, p) {
+          if (err) return done(err);
+          foundId = p.id;
```
bajtos: Sharing state between tests is evil. First, if this test fails, the second one will always fail too, which is confusing. Secondly, you cannot run the second test on its own.

test/manipulation.test.js
```diff
@@ -1097,6 +1098,53 @@ describe('manipulation', function() {
...
+          done();
+        });
+      });
+    });
+
+    it('works on update if the request includes an exisiting id in db', function(done) {
+      var post = {id: foundId, title: 'b', content: 'BBBB'};
```
bajtos: Please rework this test to create the model instance you are relying on.

test/manipulation.test.js
```diff
@@ -1097,6 +1098,53 @@ describe('manipulation', function() {
...
+        p.content.should.equal(post.content);
+        Post.findById(p.id, function(err, p) {
+          if (err) return done(err);
+          foundId = p.id;
+          done();
+        });
+      });
+    });
+
+    it('works on update if the request includes an exisiting id in db', function(done) {
```
bajtos: typo, should be `existing`

test/manipulation.test.js
loay: correct. Index is not needed, but I kinda used comments and contents in the new test to replace the one that was using the id from the previous test.

loay: actually kept content and removed comments.

test/manipulation.test.js
loay: fixed

loay:
@slnode test please

Amir-61:
@loay  As explained in https://github.com/strongloop/loopback-datasource-juggler/pull/1210#pullrequestreview-16072818 can you create 3 issues for connectors to return 404 Error message when non-existing id is used for `replaceById`; issues should be under the following repos:
- [ ] `loopback-connector` which covers all relational connectors
- [ ] `loopback-connector-mongodb`
- [ ] `loopback-connector-cloudant`

Please make sure you plan them as soon as possible otherwise this PR cannot be landed.

CC: @0candy @bajtos

loay:
@slnode test please

loay:
@slnode test please

bajtos:
@loay please squash the changes before landing.

bajtos:
@loay AFAICT, the only downstream build triggered by https://github.com/strongloop/loopback-connector/pull/73 was in loopback-datasource-juggler. I think that means all of our connectors are using an older semver-major version of loopback-connector, in which case we need to upgrade them to use the latest loopback-connector version before this patch can be landed.

@Amir-61 thoughts?

loay:
@bajtos 
I can create PRs to update the sql connectors to version 3.0 of `loopback-connector` or I can backport if that is an available solution.

loay:
@slnode test please

loay:
@bajtos everything passes now, except Oracle which is not related but a driver issue that needs to be fixed.

PR #1211, opened 2017-01-05
Rename KVAO.flush to deleteAll

superkhau:
- Add new test to ensure side effects of deleting do not spill over to
  other models
- Fix kv-memory implementation to pass new test

@bajtos PTAL


lib/connectors/kv-memory.js
```diff
@@ -201,9 +201,11 @@ KeyValueMemoryConnector.prototype.disconnect = function(callback) {
   process.nextTick(callback);
 };
 
-KeyValueMemoryConnector.prototype.flush =
+KeyValueMemoryConnector.prototype.deleteAll =
 function(modelName, options, callback) {
-  this._store = Object.create(null);
+  var modelStore = this._store[modelName];
```
bajtos:
```js
var store = this._getStoreForModel(modelName);
```

See e.g. `KeyValueMemoryConnector.prototype.iterateKeys`. Feel free to use `modelStore` as the variable name though.

test/kvao/delete-all.suite.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+const bdd = require('../helpers/bdd-if');
+const helpers = require('./_helpers');
+const should = require('should');
+
+module.exports = function(dataSourceFactory, connectorCapabilities) {
+  var supportsDeleteAllOperation =
+    connectorCapabilities.supportsDeleteAllOperation !== false;
```
bajtos: As I commented earlier in another PR, the test suite should skip new features that are not supported by the connector-under-test yet. That way downstream builds stay green and we can land a new feature before it's implemented in connectors.

```js
var supportsDeleteAllOperation =  !!connectorCapabilities.supportsDeleteAllOperation;
```

I would also consider using a shorter flag name, e.g. `canDeleteAll`, `hasDeleteAll` or `providesDeleteAll`. In fact, can we check the `connector` object to detect whether it implements deleteAll or not?

```js
var supportsDeleteAll = 'deleteAll' in dataSourceFactory().connector;
```

That way there is no need for a flag in `connectorCapabilities`.

test/kvao/delete-all.suite.js
```diff
@@ -0,0 +1,45 @@
...
+
+  bdd.describeIf(supportsDeleteAllOperation, 'deleteAll', function() {
+    let CacheItem;
+    beforeEach(function unpackContext() {
+      CacheItem = helpers.givenCacheItem(dataSourceFactory);
+    });
+
+    it('removes all keys-value pairs associated to the given model', function() {
```
bajtos: `key-value pairs`, not `keys-...`

test/kvao/delete-all.suite.js
```diff
@@ -0,0 +1,45 @@
...
+      return helpers.givenKeys(CacheItem, ['key1', 'key2'])
+        .then(() => CacheItem.deleteAll())
+        .then(() => CacheItem.keys())
+        .then((keys) => {
+          should(keys).eql([]);
+        });
+    });
+
+    it('does not remove keys-value pairs unassociated to the given model',
```
bajtos: `key-value pairs`, not `keys-...`

I also find the difference between test names that say "associated" and "unassociated" too small, it can be easily overlooked. How about

```js
it('does not remove data of other models')
```

test/kvao/delete-all.suite.js
```diff
@@ -0,0 +1,45 @@
...
+    function() {
+      var NonCacheItem = dataSourceFactory().createModel('NonCacheItem', {
+        key: String,
+        value: 'any',
```
bajtos: Do we actually need to specify these properties? I think a KV model does not require properties, since the KV methods are accepting arbitrary javascript entity as a value.

test/kvao/delete-all.suite.js
```diff
@@ -0,0 +1,45 @@
...
+        .then(() => CacheItem.deleteAll())
+        .then(() => CacheItem.keys())
+        .then((keys) => {
+          should(keys).eql([]);
+        });
+    });
+
+    it('does not remove keys-value pairs unassociated to the given model',
+    function() {
+      var NonCacheItem = dataSourceFactory().createModel('NonCacheItem', {
```
bajtos: Perhaps `AnotherModel` would be a better name? This will make the distinction from `CacheItem` much better.

test/kvao/delete-all.suite.js
```diff
@@ -0,0 +1,45 @@
...
+        key: String,
+        value: 'any',
+      });
+      return helpers.givenKeys(CacheItem, ['key1', 'key2'])
+        .then(() => helpers.givenKeys(NonCacheItem, ['key3', 'key4']))
+        .then(() => CacheItem.deleteAll())
+        .then(() => CacheItem.keys())
```
bajtos: No need to verify `CacheItem` keys, this has been already verified by the previous test.

bajtos:
I am glad there was no release since this new API was added. Otherwise we would have to preserve `flush` as a (deprecated) alias for backwards compatibility.

The PR description says that #1211 depends on #1212, but then #1212 says that it depends on #1211 too. ❓ 😱 ❓ If don't think #1212 is needed as long as you implement my proposal from https://github.com/strongloop/loopback-datasource-juggler/pull/1211#discussion_r94783632

Issue #1212, opened 2017-01-05
Use KVAO connector supported deleteAll or fall back to non-atomic deleteAll


superkhau:
KVAO should use connector-defined `deleteAll` implementation or fall back to non-atomic `deleteAll` (ie. loop and call connector-defined `delete`).

# Prerequisites

- [ ] [Renaming `flush` to deleteAll` in Juggler](https://github.com/strongloop/loopback-datasource-juggler/pull/1211)

# Prerequisite of

- [Configurable mapname in KV Extreme Scale connector](https://github.com/strongloop/loopback-connector-kv-extreme-scale/pull/12)

superkhau:
Closing in favour of #1214

PR #1213, opened 2017-01-05
Add kvao delete

superkhau:
Delete a single key-value pair via key.

@bajtos PTAL

# Prerequisite of

- #1212


test/kvao/delete.suite.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const bdd = require('../helpers/bdd-if');
+const helpers = require('./_helpers');
+const should = require('should');
+
+module.exports = function(dataSourceFactory, connectorCapabilities) {
+  describe('delete', function() {
```
bajtos: This test suite should be disabled unless the connector supports the newly added feature, see https://github.com/strongloop/loopback-datasource-juggler/pull/1211#discussion_r94783632

test/kvao/delete.suite.js
```diff
@@ -0,0 +1,44 @@
...
+    function() {
+      return helpers.givenKeys(CacheItem, ['key1', 'key2'])
+        .then(() => CacheItem.delete('key1'))
+        .then(() => CacheItem.keys())
+        .then((keys) => {
+          should(keys).eql(['key2']);
+        });
+    });
+
+    it('does not remove key-value pairs unassociated to the given key',
```
bajtos: I wouldn't bother with this test. What kind of implementation would delete a single key across all models?

superkhau: true -- the point was to ensure deleting a single key does not have side effects that spill over to other models

superkhau:
>Alternatively, it may be better to send a single pull request implementing all three parts - delete, optimized deleteAll and fall-back deleteAll.

Gonna do this instead, make it less confusing to figure out prerequisites, etc.

superkhau:
Closing in favour of #1214

PR #1214, opened 2017-01-06
Refactor/kvao flush to deleteAll

superkhau:
Aggregate pull request for refactoring `kvao.flush` to `kvao.deleteAll`. Implementation consists of 3 parts:

- [x] `delete`
- [x] `deleteAll`
- [x] Fall back `deleteAll`


superkhau:
@rmg I think something is wrong with CI, PTAL.

superkhau:
@slnode test please

test/kvao/_helpers.js
```diff
@@ -2,8 +2,14 @@
 
 var Promise = require('bluebird');
 
-exports.givenCacheItem = function(dataSourceFactory) {
+exports.givenCacheItem = function(dataSourceFactory, excludeFromConnector) {
   var dataSource = dataSourceFactory();
+
+  if (excludeFromConnector) {
+    // `delete connector.deleteAll` doesn't work
+    dataSource.connector[excludeFromConnector] = undefined;
```
bajtos: This is not the way how we force a datasource to use fall-back (unoptimized) methods.

See `memory` connector tests for the correct solution: https://github.com/strongloop/loopback-datasource-juggler/blob/66c54a96467b39609fec07f41b6d8018c5fc3819/test/memory.test.js#L887-L906.

The idea is to run the shared test suite twice for the built-in `kv-memory` connector. Once we run against the "full" connector, the second time we delete optimized functions from the connector instance after the datasource was created.

Th change in this file should be reverted.

test/kvao/delete-all-fallback.suite.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+const bdd = require('../helpers/bdd-if');
+const DataSource = require('../..').DataSource;
+const helpers = require('./_helpers');
+const kvMemory = require('../../lib/connectors/kv-memory');
+const should = require('should');
+
+module.exports = function(dataSourceFactory, connectorCapabilities) {
+  // override injected data source factory
```
bajtos: See my comment above, the decision fallback vs regular deleteAll is made by the connector. There is no point to test how connectors implementing optimized deleteAll are handling the fall-back variant, because the fall-back variant will be never invoked.

This whole file should be deleted and [test/kv-memory.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/66c54a96467b39609fec07f41b6d8018c5fc3819/test/kv-memory.js#L5-L11) should be modified to run the shared test suite twice.

test/kvao/delete-all.suite.js
```diff
@@ -0,0 +1,40 @@
...
+          should(keys).eql([]);
+        });
+    });
+
+    it('does not remove data from other existing models', function() {
+      var AnotherModel = dataSourceFactory().createModel('AnotherModel');
+      return helpers.givenKeys(CacheItem, ['key1', 'key2'])
+        .then(() => helpers.givenKeys(AnotherModel, ['key3', 'key4']))
+        .then(() => CacheItem.deleteAll())
+        .then(() => CacheItem.keys())
```
bajtos: As I commented in your earlier pull request, there is no need to verify that `CacheItem.deleteAll` deletes CacaheItem keys - that has been already verified by the previous test.

rmg:
@slnode test please

package.json
```diff
@@ -31,14 +31,14 @@
...
   "devDependencies": {
-    "async-iterators": "^0.2.2",
     "eslint": "^3.11.1",
     "eslint-config-loopback": "^6.0.0",
     "mocha": "^2.1.0",
     "should": "^8.0.2"
   },
   "dependencies": {
     "async": "~1.0.0",
+    "async-iterators": "^0.2.2",
```
superkhau: I simply moved this from a devDep to a dep. It's actually not needed as a full on dep anymore with your suggested changes for deleteAll, so removing this as it's unused now.

lib/kvao/delete-all.js
```diff
@@ -0,0 +1,62 @@
...
+  callback = callback || utils.createPromiseCallback();
+
+  var connector = this.getConnector();
+  if (typeof connector.deleteAll === 'function') {
+    connector.deleteAll(this.modelName, options, callback);
+  } else if (typeof connector.delete === 'function') {
+    debug('falling back to unoptimized `deleteAll`');
+    iterateAndDelete(this.modelName, connector, options, callback);
+  } else {
+    debug('connector does not support `deleteAll`');
```
bajtos: I think this is not entirely correct. How about `Connector does not support deleting key/values.` (Also notice the error message is a proper sentence, starting with a capital letter and ending with a dot. I don't mind the missing dot, but I believe we are starting error messages with a capital letter in other places.)

lib/kvao/delete-all.js
```diff
@@ -0,0 +1,62 @@
...
+    process.nextTick(function() {
+      var err = new Error('connector does not support `deleteAll`');
```
bajtos: Ditto, `new Error('Connector does not support deleting key/values')`.

(Feel free to use a different wording for "key/values", just stay consistent in the error message and the debug log.)

lib/kvao/delete-all.js
```diff
@@ -0,0 +1,62 @@
...
+      err.statusCode = 501;
+      callback(err);
+    });
+  }
+  return callback.promise;
+};
+
+function iterateAndDelete(modelName, connector, options, callback) {
```
bajtos: Nitpick: I would personally change the order of arguments from `(modelName, connector,...)` to `(connector, modelName, ...)`, Start with the thing that's changing the least (connector remains the same for the lifetime of a datasource), follow with things that change more often.

lib/kvao/delete.js
```diff
@@ -0,0 +1,44 @@
...
+
+  assert(typeof key === 'string' && key, 'key must be a non-empty string');
+
+  callback = callback || utils.createPromiseCallback();
+
+  var connector = this.getConnector();
+  if (typeof connector.delete === 'function') {
+    connector.delete(this.modelName, key, options, callback);
+  } else {
+    debug('connector does not support `delete`');
```
bajtos: I think it may be better to use the same message I proposed for "deleteAll" above, i.e. `Connector does not support deleting key/values`.

lib/kvao/delete.js
```diff
@@ -0,0 +1,44 @@
...
+    process.nextTick(function() {
+      var err = new Error('connector does not support `delete`');
```
bajtos: ditto

superkhau:
@slnode test please

superkhau:
@slnode test please

superkhau:
Created follow up issue for long term Oracle fail at https://github.com/strongloop/loopback-connector-oracle/issues/99. Merging as everything else is green.

PR #1216, opened 2017-01-06
"Near" fix for where clauses

viki53:
### Description

Modifies the `geo.nearFilter` to provide an in-depth search (so that it also searches in sub-queries, such as `$and` arrays)

~Requires~ PR strongloop/loopback-connector-mongodb#322 *helps make it* work *properly* in MongoDB and ~probably~ similar modifications in other connectors *might help too*.

#### Related issues

- Required for PR strongloop/loopback-connector-mongodb#322
- Based on strongloop/loopback-datasource-juggler#993

### Checklist

- [x] Code conforms with the [style  guide](http://loopback.io/doc/en/contrib/style-guide.html)

### DIFF
--- a/lib/geo.js
+++ b/lib/geo.js
@@ -12,82 +12,109 @@ var assert = require('assert');
  */
 
 exports.nearFilter = function nearFilter(where) {
-  var result = false;
+  function nearSearch(clause, parentKeys) {
+    if (typeof clause !== 'object') {
+      return false;
+    }
+    parentKeys = parentKeys || [];
 
-  if (where && typeof where === 'object') {
-    Object.keys(where).forEach(function(key) {
-      var ex = where[key];
-
-      if (ex && ex.near) {
-        result = {
-          near: ex.near,
-          maxDistance: ex.maxDistance,
-          minDistance: ex.minDistance,
-          unit: ex.unit,
-          key: key,
-        };
+    Object.keys(clause).forEach(function(clauseKey) {
+      if (Array.isArray(clause[clauseKey])) {
+        clause[clauseKey].forEach(function(el, index) {
+          var ret = nearSearch(el, parentKeys.concat(clauseKey).concat(index));
+          if (ret) return ret;
+        });
+      } else {
+        if (clause[clauseKey].hasOwnProperty('near')) {
+          var result = clause[clauseKey];
+          nearResults.push({
+            near: result.near,
+            maxDistance: result.maxDistance,
+            minDistance: result.minDistance,
+            unit: result.unit,
+            // If key is at root, define a single string, otherwise append it to the full path array
+            mongoKey: parentKeys.length ? parentKeys.concat(clauseKey) : clauseKey,
+            key: clauseKey,
+          });
+        }
       }
     });
   }
+  var nearResults = [];
+  nearSearch(where);
 
-  return result;
+  return (!nearResults.length ? false : nearResults);
 };
 
 /*!
- * Filter a set of objects using the given `nearFilter`.
+ * Filter a set of results using the given filters returned by `nearFilter()`.
+ * Can support multiple locations, but will include results from all of them.
+ *
+ * WARNING: "or" operator with GeoPoint does not work as expected, eg:
+ * {where: {or: [{location: {near: (29,-90)}},{name:'Sean'}]}}
+ * Will actually work as if you had used "and".  This is because geo filtering
+ * takes place outside of the SQL query, so the result set of "name = Sean" is
+ * returned by the database, and then the location filtering happens in the app
+ * logic.  So the "near" operator is always an "and" of the SQL filters, and "or"
+ * of other GeoPoint filters.
+ *
+ * Additionally, since this step occurs after the SQL result set is returned,
+ * if using GeoPoints with pagination the result set may be smaller than the
+ * page size.  The page size is enforced at the DB level, and then we may
+ * remove results at the Geo-app level.  If we "limit: 25", but 4 of those results
+ * do not have a matching geopoint field, the request will only return 21 results.
+ * This may make it erroneously look like a given page is the end of the result set.
  */
 
-exports.filter = function(arr, filter) {
-  var origin = filter.near;
-  var max = filter.maxDistance > 0 ? filter.maxDistance : false;
-  var min = filter.minDistance > 0 ? filter.minDistance : false;
-  var unit = filter.unit;
-  var key = filter.key;
-
-  // create distance index
+exports.filter = function(rawResults, filters) {
   var distances = {};
-  var result = [];
+  var results = [];
 
-  arr.forEach(function(obj) {
-    var loc = obj[key];
+  filters.forEach(function(filter) {
+    var origin = filter.near;
+    var max = filter.maxDistance > 0 ? filter.maxDistance : false;
+    var min = filter.minDistance > 0 ? filter.minDistance : false;
+    var unit = filter.unit;
+    var key = filter.key;
 
-    // filter out objects without locations
-    if (!loc) return;
+    // create distance index
+    rawResults.forEach(function(result) {
+      var loc = result[key];
 
-    if (!(loc instanceof GeoPoint)) {
-      loc = GeoPoint(loc);
-    }
+      // filter out results without locations
+      if (!loc) return;
 
-    if (typeof loc.lat !== 'number') return;
-    if (typeof loc.lng !== 'number') return;
+      if (!(loc instanceof GeoPoint)) loc = GeoPoint(loc);
 
-    var d = GeoPoint.distanceBetween(origin, loc, {type: unit});
+      if (typeof loc.lat !== 'number') return;
+      if (typeof loc.lng !== 'number') return;
 
-    if (min && d < min) {
-      return;
-    }
-    if (max && d > max) {
-      // dont add
-    } else {
-      distances[obj.id] = d;
-      result.push(obj);
-    }
+      var d = GeoPoint.distanceBetween(origin, loc, {type: unit});
+
+      // filter result if distance is either < minDistance or > maxDistance
+      if ((min && d < min) || (max && d > max)) return;
+
+      distances[result.id] = d;
+      results.push(result);
+    });
+
+    results.sort(function(resA, resB) {
+      var a = resA[key];
+      var b = resB[key];
+
+      if (a && b) {
+        var da = distances[resA.id];
+        var db = distances[resB.id];
+
+        if (db === da) return 0;
+        return da > db ? 1 : -1;
+      } else {
+        return 0;
+      }
+    });
   });
 
-  return result.sort(function(objA, objB) {
-    var a = objA[key];
-    var b = objB[key];
-
-    if (a && b) {
-      var da = distances[objA.id];
-      var db = distances[objB.id];
-
-      if (db === da) return 0;
-      return da > db ? 1 : -1;
-    } else {
-      return 0;
-    }
-  });
+  return results;
 };
 
 exports.GeoPoint = GeoPoint;
@@ -285,4 +312,3 @@ function geoDistance(x1, y1, x2, y2, options) {
 
   return 2 * Math.asin(f) * EARTH_RADIUS[type];
 }
-
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -38,6 +38,7 @@ describe('basic-querying', function() {
           name: String,
         },
       ],
+      addressLoc: {type: 'GeoPoint'},
     });
 
     db.automigrate(done);
@@ -553,6 +554,147 @@ describe('basic-querying', function() {
       });
     });
 
+    describe('geo queries', function() {
+      describe('near filter', function() {
+        it('supports a basic "near" query', function(done) {
+          User.find({
+            where: {
+              addressLoc: {
+                near: '29.9,-90.07',
+              },
+            },
+          }, function(err, users) {
+            if (err) done(err);
+            users.should.have.property('length', 3);
+            users[0].name.should.equal('John Lennon');
+            users[0].addressLoc.should.not.be.null();
+            done();
+          });
+        });
+
+        it('supports "near" inside a coumpound query with "and"', function(done) {
+          User.find({
+            where: {
+              and: [
+                {
+                  addressLoc: {
+                    near: '29.9,-90.07',
+                  },
+                },
+                {
+                  vip: true,
+                },
+              ],
+            },
+          }, function(err, users) {
+            if (err) done(err);
+            users.should.have.property('length', 2);
+            users[0].name.should.equal('John Lennon');
+            users[0].addressLoc.should.not.be.null();
+            users[0].vip.should.be.true();
+            done();
+          });
+        });
+
+        it('supports "near" inside a complex coumpound query with multiple "and"', function(done) {
+          User.find({
+            where: {
+              and: [
+                {
+                  and: [
+                    {
+                      addressLoc: {
+                        near: '29.9,-90.07',
+                      },
+                    },
+                    {
+                      order: 2,
+                    },
+                  ],
+                },
+                {
+                  vip: true,
+                },
+              ],
+            },
+          }, function(err, users) {
+            if (err) done(err);
+            users.should.have.property('length', 1);
+            users[0].name.should.equal('John Lennon');
+            users[0].addressLoc.should.not.be.null();
+            users[0].vip.should.be.true();
+            users[0].order.should.equal(2);
+            done();
+          });
+        });
+
+        it('supports multiple "near" queries with "or"', function(done) {
+          User.find({
+            where: {
+              or: [
+                {
+                  addressLoc: {
+                    near: '29.9,-90.04',
+                    maxDistance: 300,
+                  },
+                },
+                {
+                  addressLoc: {
+                    near: '22.97,-88.03',
+                    maxDistance: 50,
+                  },
+                },
+              ],
+            },
+          }, function(err, users) {
+            if (err) done(err);
+            users.should.have.property('length', 2);
+            users[0].addressLoc.should.not.be.null();
+            users[0].name.should.equal('Paul McCartney');
+            users[1].addressLoc.should.not.equal(null);
+            users[1].name.should.equal('John Lennon');
+            done();
+          });
+        });
+
+        it('supports multiple "near" queries with "or" ' +
+          'inside a coumpound query with "and"', function(done) {
+          User.find({
+            where: {
+              and: [
+                {
+                  or: [
+                    {
+                      addressLoc: {
+                        near: '29.9,-90.04',
+                        maxDistance: 300,
+                      },
+                    },
+                    {
+                      addressLoc: {
+                        near: '22.7,-89.03',
+                        maxDistance: 50,
+                      },
+                    },
+                  ],
+                },
+                {
+                  vip: true,
+                },
+              ],
+            },
+          }, function(err, users) {
+            if (err) done(err);
+            users.should.have.property('length', 1);
+            users[0].addressLoc.should.not.be.null();
+            users[0].name.should.equal('John Lennon');
+            users[0].vip.should.be.true();
+            done();
+          });
+        });
+      });
+    });
+
     it('should only include fields as specified', function(done) {
       var remaining = 0;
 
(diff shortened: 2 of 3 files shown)


slnode:

slnode:

slnode:

slnode:

lib/geo.js
```diff
@@ -9,80 +9,119 @@ var assert = require('assert');
...
-  var unit = filter.unit;
-  var key = filter.key;
-
-  // create distance index
   var distances = {};
   var result = [];
 
-  arr.forEach(function(obj) {
-    var loc = obj[key];
+  filters.forEach(function(filter) {
```
0candy: This is causing an error:
```ReferenceError: filters is not defined
      at Object.exports.filter (lib/geo.js:78:3)```

**Comment by viki53** [comment]
Updated to keep the near key as a simple string if it's a root key (no depth array) to help with retro-compatibility.

Also ran `eslint --fix .`, so there might be more changes than necessary to the code (but should help keeping it coherent all around)

**Comment by jannyHou** [comment]
Test mongodb with juggler PR https://github.com/strongloop/loopback-datasource-juggler/pull/1216 on local, all mongodb tests pass 👍 
While the three new test cases added in juggler fail:
```
 1) mongodb imported features basic-querying find should support nested GeoPoint near queries:
     Uncaught AssertionError: expected Array [] to have property length of 2 (got 0)
      at Assertion.fail (/Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/node_modules/should/lib/assertion.js:92:17)
      at Assertion.Object.defineProperty.value [as property] (/Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/node_modules/should/lib/assertion.js:164:19)
      at /Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/test/basic-querying.test.js:529:27
      at /Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/lib/connectors/memory.js:474:7

  2) mongodb imported features basic-querying find should support very nested GeoPoint near queries:
     Uncaught AssertionError: expected Array [] to have property length of 1 (got 0)
      at Assertion.fail (/Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/node_modules/should/lib/assertion.js:92:17)
      at Assertion.Object.defineProperty.value [as property] (/Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/node_modules/should/lib/assertion.js:164:19)
      at /Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/test/basic-querying.test.js:543:27
      at /Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/lib/connectors/memory.js:474:7

  3) mongodb imported features basic-querying find should support multiple GeoPoint near queries:
     Uncaught AssertionError: expected Array [] to have property length of 2 (got 0)
      at Assertion.fail (/Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/node_modules/should/lib/assertion.js:92:17)
      at Assertion.Object.defineProperty.value [as property] (/Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/node_modules/should/lib/assertion.js:164:19)
      at /Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/test/basic-querying.test.js:560:27
      at /Users/jannyhou/lb/triage/nearArray/loopback-datasource-juggler/lib/connectors/memory.js:474:7
```
@viki53 This pr LGTM and ready to land, but better to land with juggler pr together to make sure CI always pass. Do you also get those 3 juggler test cases fail on your end? Any idea to fix them?

**Comment by ebarault** [comment]
@0candy : can we move on with this PR?

**Comment by 0candy** [comment]
@ebarault I would like to, but the 3 geopoint near test cases are failing.

**Comment by ebarault** [comment]
@0candy : isn't that because they are requiring the changes in strongloop/loopback-connector-mongodb#322 ?

**Comment by 0candy** [comment]
@ebarault I have synced with @jannyHou , she will work on merging the 2 prs together. Thanks!

**Comment by ebarault** [comment]
thank both of you! 👍

**Comment by ebarault** [comment]
@jannyHou : what is the status on this PR please? Can we help you in any way?

**Comment by jannyHou** [comment]
@ebarault This pr depends on another juggler pr with quite many line's change, and I could not get tests pass in juggler. I try my best to make time to review the juggler pr while I am new to `Geopoint` and our team also have more urgent tasks to finish. I will dig into the juggler pr asap and see why those tests fail.

**Comment by ssh24** [comment]
@ebarault Looks like this PR is from your fork. Not sure if you or @viki53 worked on it but could either of you rebase the code? There will be merge conflicts on `test/basic-querying.test.js` file. Please take a look at it.

File: lib/geo.js [hunk 2]
```diff
@@ -9,80 +9,123 @@ var assert = require('assert');
...
+      } else {
+        if (clause[clauseKey].hasOwnProperty('near')) {
+          var result = clause[clauseKey];
+          nearResults.push({
+            near: result.near,
+            maxDistance: result.maxDistance,
+            unit: result.unit,
+            // If key is at root, define a single string, otherwise append it to the full path array
+            // key: parentKeys.length ? parentKeys.concat(clauseKey) : clauseKey,
+            key: clauseKey
```
ebarault: i rolled back the line from the mysql fix (see #993)

**L42** works for mongoDB but fails for the unit-tests (and would fail for mysql) as latter on for non mongoDB connectors the `filter()` method will be called and it expects a string, not an array for [`var loc = obj[key];`](https://github.com/strongloop/loopback-datasource-juggler/pull/1216/files#diff-b367a815d81f15531e0d0553ea8aaa5aR90).

so here's the situation: 
- how can we unit-test in the juggler the specific case where the juggler will call the geoNear operator under the hood (are there integration tests?)
- how to detect upstream in the `nearFilter()` method when to trigger one way (non mongoDB connector) or the other (mongoDB connecter) to build the `key` property ?

ebarault:
@ssh24: could you have a look please so I can finalize this pr? Thanks

ssh24:
@ebarault I do not know much about "Near filter" and this PR in particular. I looked at the code and your comment and have some comments on it.

Since this fix works for MongoDB, for this line [`var loc = obj[key];`](https://github.com/strongloop/loopback-datasource-juggler/pull/1216/files#diff-b367a815d81f15531e0d0553ea8aaa5aR90), we could do a check with a **string** for other non-mongodb connectors. 

For example, the non-mongodb connectors will look for strings where as mongodb would be fine with the array. Something like this: `var loc = obj[key] || str;`

ebarault:
> the issue is upstream, when building the key, used in `var loc = obj[key];`

```js
// mongodb
key: parentKeys.length ? parentKeys.concat(clauseKey) : clauseKey,

// non mongodb
key: clauseKey
```
the juggler must select one or the other or store both and let the connector pick the good one (i could adapt your proposal here to something like: `var loc = obj[key] || mongoKey;`)

👉  can/should the juggler know the downstream connector?

ssh24:
Does non-mongodb connector have `parentKeys`? For example, if we just have 
```
// for all connectors (mongodb and non-mongodb)
key: parentKeys.length ? parentKeys.concat(clauseKey) : clauseKey,
```

Then, non-mongodb connectors would have the string `clauseKey` if parentKeys doesn't exists and for mongo-db it would just work the way we were expecting it to (assuming parentKeys exists).

To answer your question **can/should the juggler know the downstream connector?**, on the tests you could do something like `connector.name` to check for the connector and specify the logic you want for it but I prefer to not do that way.

ebarault:
thanks for following on this @ssh24 

> Does non-mongodb connector have parentKeys? 

yes, that's the main issue, they are part of the query injected at loopback level.
For example in unit tests 
`key: parentKeys.length ? parentKeys.concat(clauseKey) : clauseKey` produces an array that the fallback `filter()` method can not handle. This would be the same if plugging any non-mongodb connector.

> on the tests you could do something like connector.name to check for the connector and specify the logic you want for it but I prefer to not do that way.

Yes, i understand the requirement to keep the juggler independent from the connectors. That being said, implementing a `keyMongoDB` property in the juggler also breaks this rule.

👉  To ease my research for a solution, i'd like to implement an integration test that would use the mongodb connector underneath: are there such integration tests ATM?

ssh24:
@ebarault 
> For example in unit tests
key: parentKeys.length ? parentKeys.concat(clauseKey) : clauseKey produces an array that the fallback filter() method can not handle.

Why won't the `filter()` function not be able to handle this?

I prefer to modify the `filter()` function so it can take care of the in-depth search situation for all connectors. I am willing to discuss ways to do that, but will need some insights on what the `filter()` function requires/does from you.

> i'd like to implement an integration test that would use the mongodb connector underneath: are there such integration tests ATM?

Do you mean integration tests on juggler that would use mongodb connector to verify this? Shouldn't that still fail the case for non-mongodb connectors when they run their own downstream tests?

ebarault:
so let's roll back a bit: just to make sure we're on the same track here:

the logic in geo.js is the following: 
1. loopback fires a request with a **where** object including **near** operator(s)
2. the nearFilter() method is called to adapt the query from the where statement to downstream 
connectors
3. if the connector supports a **near** operator (mongdb case), the query is passed to the db driver

or
    3. if the connector does not support a **near** operator, the connector runs a query without the near operators, get all the results and pass them back to the juggler's filter() method which filters the result according to the original query, including the **near** operators

Agreed ?

Having said that i understand that the reason why the juggler's filter() method is failing with the mongodb nearFilter format is that there's a functional gap between what mongodb near queries do, and what the juggler's geo.filter() method does.
👉  But it's not really realistic i'm afraid to reproduce all the power of the mongodb geo near queries inside the juggler.

ssh24:
Is it the case all non-mongodb connectors do not support **near** operator as of yet? In that case, this becomes interesting and I think the only way we could do it is as you said
```
// mongodb
mongoKey: parentKeys.length ? parentKeys.concat(clauseKey) : clauseKey,

// non mongodb
key: clauseKey
```

ebarault:
i'll do some background check in the connectors to confirm, but **yes** i'm 99% sure.

ssh24:
In that case, I believe we should just add a specific key for `mongodb` connector and leave the other connectors point to another key.

ebarault:
👍  good
regarding integration tests, how to make sure the juggler + mongodb-connector work well together and prevent future regression?

ssh24:
How big is the functional gap between what the juggler **nearFilter** function does versus the expected **near** operators in mongodb? We need to make sure everything what **nearFilter** does is covered in mongodb's **near** operators and any shortcoming should be well documented in the docs so that the users know.

ebarault:
see: https://docs.mongodb.com/manual/reference/operator/query/near/
but the original issue in this PR is that all sort of nested query combinations are not addressed by juggler's geoNear

for example (from the unit tests): 
```js
where: {
          and: [
            {
              or: [
                {
                  addressLoc: {
                    near: '29.9,-90.04',
                    maxDistance: 300,
                  },
                },
                {
                  addressLoc: {
                    near: '22.97,-88.03',
                    maxDistance: 300,
                  },
                },
              ],
            },
            {
              vip: true,
            },
          ],
        },
      }
```

mongoKey won't be an isolate case if latter on we want to cover other geo operators: geoWithin, geoIntersect, etc.

ssh24:
I see. This might be something we want to discuss with Raymond.

@raymondfeng What is your take on this?

ssh24:
@ebarault Wanted to get on with this PR. 

> mongoKey won't be an isolate case if latter on we want to cover other geo operators: geoWithin, geoIntersect, etc.

Is using `mongoKey` the best way to support `geoNear`?

ebarault:
yes, thanks for reactivating this one : i'm really depending on you guys here for the design policy:

- either the juggler stores a different set of keys in a mongo specific object and returns it back to the connector
- or the downstream connector injects some sort of command to the juggler to trigger what is returned

ssh24:
I believe this is what we went with:

> either the juggler stores a different set of keys in a mongo specific object and returns it back to the connector

And then we can have the specific `mongoKey` for mongo-db and `key` for non mongo-db. 

i.e 
```
// mongodb
mongoKey: parentKeys.length ? parentKeys.concat(clauseKey) : clauseKey,

// non mongodb
key: clauseKey
```

ebarault:
i am good with that proposal, i leave a couple more hours for people to comment, and then i'll move forward so we can land it.

0candy:
Need to fix up some linting errors:
```
/home/travis/build/strongloop/loopback-datasource-juggler/lib/geo.js
  43:27  error  Missing trailing comma  comma-dangle
/home/travis/build/strongloop/loopback-datasource-juggler/test/basic-querying.test.js
  594:87  error  A space is required after ','   comma-spacing
  640:10  error  Missing trailing comma          comma-dangle
  649:61  error  Infix operators must be spaced  space-infix-ops
```
and rebase before merging.

ssh24:
@ebarault Any updates on cleaning up the PR and moving forward with it? Just as a side note, there will be merge conflicts when you rebase on [./lib/geo.js](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/geo.js), so please take care of it.

ebarault:
@ssh24 : i'll try to finish it tonight or tomorrow. Have already quite a backlog in loopback.

ebarault:
@ssh24 : i added the `mongoKey` param and updated mongodb connector to use it, see https://github.com/strongloop/loopback-connector-mongodb/pull/322

i also rebased and handled the merge conflict, all tests pass locally, PTAL

test/geo.test.js
```diff
@@ -147,8 +147,8 @@ describe('GeoPoint', function() {
         },
       };
       var filter = nearFilter(where);
-      filter.key.should.equal('location');
-      filter.should.have.properties({
+      filter[0].key.should.equal('location');
+      filter[0].should.have.properties({
         key: 'location',
```
ebarault: updated test from min distance PR: nearFilter now returns an array

test/geo.test.js
```diff
@@ -187,15 +187,15 @@ describe('GeoPoint', function() {
...
+      var filter = [{
         key: 'location',
         near: {
           lat: 30.278562,
           lng: 120.139846,
         },
         unit: 'meters',
         minDistance: 10000,
-      };
+      }];
```
ebarault: updated test from min distance PR: filter now expects an array

lib/geo.js
```diff
@@ -9,85 +9,128 @@ var assert = require('assert');
...
 
-    if (!(loc instanceof GeoPoint)) {
-      loc = GeoPoint(loc);
-    }
+      // filter out objects without locations
+      if (!loc) return;
 
-    if (typeof loc.lat !== 'number') return;
-    if (typeof loc.lng !== 'number') return;
+      if (!(loc instanceof GeoPoint)) {
```
ssh24: `if (!(loc instanceof GeoPoint)) loc = GeoPoint(loc);`

lib/geo.js
```diff
@@ -9,85 +9,128 @@ var assert = require('assert');
...
-    if (max && d > max) {
-      // dont add
-    } else {
-      distances[obj.id] = d;
-      result.push(obj);
-    }
-  });
+      var d = GeoPoint.distanceBetween(origin, loc, {type: unit});
+
+      if (min && d < min) {
```
ssh24: `if (min && d < min) return;`

lib/geo.js
```diff
@@ -9,85 +9,128 @@ var assert = require('assert');
...
+        return;
+      }
+      if (max && d > max) {
```
ssh24: shouldn't this be `if (max && d > max) return;` as well?

test/basic-querying.test.js
```diff
@@ -553,6 +554,132 @@ describe('basic-querying', function() {
...
 
+    it('supports basic a basic query with GeoPoint "near"', function(done) {
+      User.find({
+        where: {
+          addressLoc: {
+            near: '29.9,-90.07',
+          },
+        },
+      }, function(err, users) {
+        if (err) return done(err);
```
ssh24: `if (err) done(err);` is fine

test/basic-querying.test.js
```diff
@@ -553,6 +554,132 @@ describe('basic-querying', function() {
...
+                near: '29.9,-90.07',
+              },
+            },
+            {
+              vip: true,
+            },
+          ],
+        },
+      }, function(err, users) {
+        if (err) return done(err);
```
ssh24: just `if (err) done(err);` is fine.

test/basic-querying.test.js
```diff
@@ -553,6 +554,132 @@ describe('basic-querying', function() {
...
+                },
+              ],
+            },
+            {
+              vip: true,
+            },
+          ],
+        },
+      }, function(err, users) {
+        if (err) return done(err);
```
ssh24: `if (err) done(err);`

test/basic-querying.test.js
```diff
@@ -553,6 +554,132 @@ describe('basic-querying', function() {
...
+            {
+              addressLoc: {
+                near: '22.97,-88.03',
+                maxDistance: 300,
+              },
+            },
+          ],
+        }
+      }, function(err, users) {
+        if (err) return done(err);
```
ssh24: `if (err) done(err);`

test/basic-querying.test.js
```diff
@@ -553,6 +554,132 @@ describe('basic-querying', function() {
...
+                },
+              ],
+            },
+            {
+              vip: true,
+            },
+          ],
+        },
+      }, function(err, users) {
+        if (err) return done(err);
```
ssh24: `if (err) done(err);`

test/basic-querying.test.js
```diff
@@ -553,6 +554,132 @@ describe('basic-querying', function() {
...
+      User.find({
+        where: {
+          addressLoc: {
+            near: '29.9,-90.07',
+          },
+        },
+      }, function(err, users) {
+        if (err) return done(err);
+        users.should.have.property('length', 3);
+        users[0].addressLoc.should.not.equal(null);
```
ssh24: Can we check for something more additional to this only being `null`?

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
@ebarault I had to fix the commit linter for CI so please make sure to pull the latest before committing the changes requested.

ssh24:
@slnode test please

test/basic-querying.test.js
viki53: Still need to return to stop the execution of this function. Otherwise it'll use un-necessary resources.

lib/geo.js
```diff
@@ -9,85 +9,115 @@ var assert = require('assert');
...
-      result.push(obj);
-    }
-  });
+      var d = GeoPoint.distanceBetween(origin, loc, {type: unit});
 
-  return result.sort(function(objA, objB) {
-    var a = objA[key];
-    var b = objB[key];
+      // filter result if distance is either < minDistance or > maxDistance
+      if ((min && d < min) || (max && d > max)) return;
```
ebarault: this is much cleaner now

lib/geo.js
```diff
@@ -9,85 +9,115 @@ var assert = require('assert');
...
 
-exports.filter = function(arr, filter) {
-  var origin = filter.near;
-  var max = filter.maxDistance > 0 ? filter.maxDistance : false;
-  var min = filter.minDistance > 0 ? filter.minDistance : false;
-  var unit = filter.unit;
-  var key = filter.key;
-
-  // create distance index
+exports.filter = function(rawResults, filters) {
```
ebarault: changing the param names to make them more readable

test/basic-querying.test.js
```diff
@@ -553,6 +554,147 @@ describe('basic-querying', function() {
       });
     });
 
+    describe('geo queries', function() {
+      describe('near filter', function() {
```
ebarault: added to describe blocks to better classify the tests

ebarault:
@ssh24 : i fully reviewed the code once more, improved the tests, cleaned the code from the other commit adding min-distance, as per your proposals, rebased and squashed. This should be all good now.

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
@ebarault Thank you for the fix. Code looks good. Waiting for CI to finish and then can be landed.

viki53:
All green lights @ssh24

ebarault:
thanks @ssh24 for following this through ! 👍

PR #1218, opened 2017-01-10
Detect deleteAll support in KVAO tests

superkhau:
Clear CacheItem data between tests if connector supports `deleteAll`. Tests fail with 501 if this check is not used.

PR #1220, opened 2017-01-12
Add missing return in KVAO keys test suite

superkhau:
Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/1218

cc @bajtos


superkhau:
Follow up issue for oracle at strongloop/loopback-connector-oracle#100

PR #1221, opened 2017-01-12
User Provided Default Model Value Callbacks

tatemz:
### Description
In order to use custom defined `defaultFn` I added the following code to a `server/boot/manipulation.js` file to test this feature:

```
module.exports = function(server) {
  server.dataSources.mysql.defineDefaultFn(function MyDefaultName(prop, data) {
    return "John Doe"
  }, ['my-default-name']);
};
```

And my `model.json`:
```
{
  "name": "Human",
  "plural": "humans",
  "base": "PersistedModel",
  "idInjection": false,
  "options": {
  },
  "properties": {
    "name": {
      "type": "String",
      "required": true,
      "defaultFn": "MyDefaultName"
    },
  },
  "validations": [],
  "relations": {},
  "acls": [],
  "methods": {}
}

```

#### Related issues
- #458
- strongloop/loopback#292

### Checklist

- [ ] Add test to `manipulation.test.js` to register and handle custom `defaultFn`
- [ ] Community testing & feedback
- [ ] Review and approve how to define custom callbacks
- [x] Sign Contributor License Agreement (CLA)

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -118,6 +118,8 @@ function DataSource(name, settings, modelBuilder) {
   // Initialize it before calling setup as the connector might register operations
   this._operations = {};
 
+  this.__defaultFns = {};
+
   this.setup(name, settings);
 
   this._setupConnector();
@@ -784,6 +786,32 @@ DataSource.prototype.defineProperty = function(model, prop, params) {
   var resolvedProp = this.getModelDefinition(model).properties[prop];
   if (this.connector && this.connector.defineProperty) {
     this.connector.defineProperty(model, prop, resolvedProp);
+  }
+};
+
+/**
+ * Define a new custom default functions for creating default model values
+ * @param {function()} type Type constructor.
+ * @param {string[]=} aliases Optional list of alternative names for this type.
+ */
+DataSource.prototype.defineDefaultFn = function(fn, names) {
+  names = names || [];
+  names = names.concat([fn.name]);
+  this.__defaultFns = this.__defaultFns || {};
+  for (var n = 0; n < names.length; n++) {
+    this.__defaultFns[names[n].toLowerCase()] = fn;
+  }
+};
+
+/**
+ * Get a defaultFn by its name
+ * @param {string} name The name of the callback being retrieved
+ */
+DataSource.prototype.getDefaultFn = function(name) {
+  if (this.__defaultFns.hasOwnProperty(name) > -1) {
+    return this.__defaultFns[name.toLowerCase()];
+  } else {
+    return false;
   }
 };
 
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

raymondfeng:
@tatemz Thank you for the patch. I like the idea to allow custom functions to be registered and referenced in model definitions.

I'm not sure if `DataSource` is the best holder of such functions though. What about adding the capability to `app`? Please note each model can access the app via `MyModel.app`.

bajtos:
@tatemz cool, thank you for the pull request!

I agree with @raymondfeng that datasource is not the best place where to keep a registry of default value factories.

> Please note each model can access the app via `MyModel.app`.

This is true for LoopBack models, but does not hold for loopback-datasource-juggler. 

In juggler, we are using `modelRegistry` as the object holding various lookup tables: not only models, but also types (see [lib/types.js](https://github.com/strongloop/loopback-datasource-juggler/blob/5a7283386c7dbda1b73ab07690a3e91b08867417/lib/types.js) and [initialization in model-builder](https://github.com/strongloop/loopback-datasource-juggler/blob/5a7283386c7dbda1b73ab07690a3e91b08867417/lib/model-builder.js#L24)).

I think we should use the same design for the registry of `defaultFn` functions:

 - In juggler, `ModelBuilder` keeps this registry, e.g. as `this.defaultFunctions`.
 - In LoopBack, it is the [registry](https://github.com/strongloop/loopback/blob/d2e9e7ee77c0ec64e0aa69c8aa8c286e1db714ce/lib/registry.js#L26) which hold the single `ModelBuilder` instance used across all models and datasource of the app. We should add a new API for registering custom `defaultFn` functions there too.

bajtos:
@tatemz we will also need you to sign our Contributor License Agreement (CLA) - see one of the failed checks. Here is the link for your convenience: https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler

tatemz:
I agree that DataSource is probably not the best spot. I originally tried to apply it within `model.js` but had some roadblocks. I will try again and document what roadblocks those were to get you guys more details on how we can best structure this feature.

Another note is that this feature should probably be structured in such a way to handle custom callbacks that involve returning promises. :-/

bajtos:
> Another note is that this feature should probably be structured in such a way to handle custom callbacks that involve returning promises. :-/

I am afraid that will be difficult if possible at all. Default values are created and assigned when a new model instance is created via JavaScript constructor call (`new MyModel()`), which is always strictly synchronous.

If you need asynchronous computation of default values, then I think you should use a different mechanism, for example you can use a `before save` or `persist` hook to fill in any missing values.

bajtos:
@tatemz ping, what's the status of this pull request?

bajtos:
Closing as abandoned. Feel free to reopen if/when you get time to work on this patch again.

PR #1222, opened 2017-01-13
Add missing return for KVAO delete all

superkhau:
Connect to https://github.com/strongloop/loopback-connector-kv-extreme-scale/pull/12

cc @bajtos


superkhau:
Oracle fail is a known issue: https://github.com/strongloop/loopback-connector-oracle/issues/100
MySQL fail is unrelated.

Landing.

PR #1240, opened 2017-01-27
Ensure replaceById returns 404 when id not found

loay:
### Description

Add a test case to ensure that replaceById will return error 404 when id does not exist in db. 
#### Related issues

- reference:
https://github.com/strongloop/loopback-connector/pull/73#issuecomment-275041949

connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1244

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -1340,6 +1340,27 @@ describe('manipulation', function() {
     });
   }
 
+  hasReplaceById = !!getSchema().connector.replaceById;
+  bdd.describeIf(hasReplaceById, 'replaceById', function() {
+    var Post;
+    before(function(done) {
+      var ds = getSchema();
+      Post = ds.define('Post', {
+        title: {type: String, length: 255},
+        content: {type: String},
+      }, {forceId: true});
+      ds.automigrate('Post', done);
+    });
+
+    it('fails when id does not exist in db using replaceById', function(done) {
+      var post = {id: 123, title: 'a', content: 'AAA'};
+      Post.replaceById(post.id, post, function(err, p) {
+        err.statusCode.should.equal(404);
+        done();
+      });
+    });
+  });
+
   describe('findOrCreate', function() {
     it('should create a record with if new', function(done) {
       Person.findOrCreate({name: 'Zed', gender: 'male'},


test/manipulation.test.js
```diff
@@ -1118,6 +1118,14 @@ describe('manipulation', function() {
       });
     });
 
+    it('fails when id does not exist in db using replaceById', function(done) {
```
bajtos: I am afraid the test for `replaceById` does not belong to `replaceOrCreate` describe block.

It looks like we don't have any explicit tests for `replaceById` yet and test `replaceAttributes` only ([source](https://github.com/strongloop/loopback-datasource-juggler/blob/c145b9359bae974ab2a18e641050109a292abec2/test/manipulation.test.js#L1158)). It's time to start a new `describe` block then! You can add it e.g. after `describe('updateAttributes')` [here](https://github.com/strongloop/loopback-datasource-juggler/blob/c145b9359bae974ab2a18e641050109a292abec2/test/manipulation.test.js#L1348-L1349).

loay:
@bajtos PTAL. Thanks.

test/manipulation.test.js
```diff
@@ -1338,6 +1338,25 @@ describe('manipulation', function() {
         });
       }
     });
+    describe('replaceById', function() {
```
bajtos: I think you need to skip this block when the connector does not implement `replaceById`:

```js
var connectorHasReplaceById = !!getSchema().connector.replaceById;
bdd.describeIf(connectorHasReplaceById, 'replaceById', function() {
  // ...
});
```

See [test/kvao/delete.suite.js](https://github.com/strongloop/loopback-datasource-juggler/blob/736a4ad1856077c23a053a14165a3707799abd77/test/kvao/delete.suite.js) for an example how to setup and use `bdd.describeIf` and [test/manipulation.test.js#L885](https://github.com/strongloop/loopback-datasource-juggler/blob/c9283cffdaf76f79e5611270f9777fd88056271d/test/manipulation.test.js#L885) for an existing example detecting `replaceId` availability.

loay: I thought I did. the describe block is within the else block, so the previous `bdd.describeif` is applicable here. I will re-arrange the code one more time and re-commit

loay:
@bajtos Fixed the code block. PTAL again. Thanks

test/manipulation.test.js
bajtos: Ah, it was my fault then, it's entirely possible I did not spot that `describe` was nested inside the `else` branch.

The new way is much cleaner 👍

PR #1241, opened 2017-01-31
Fix datasource to report connector-loading errors

bajtos:
### Description

Before this change, when resolving full connector path, all errors were
ignored. As a result, when the connector was installed but not
correctly built (e.g. loopback-connector-db2 which uses a native addon),
a very confusing message was reported by LoopBack.

In this commit, I am fixing the code handling `require()` errors
to ignore only MODULE_NOT_FOUND errors.

#### Related issues

- None

This problem resurfaced when troubleshooting client installation, cc @kjdelisle @qpresley.

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -224,7 +224,16 @@ function tryModules(names, loader) {
     try {
       mod = loader(names[m]);
     } catch (e) {
-      /* ignore */
+      var notFound = e.code === 'MODULE_NOT_FOUND' &&
+        e.message && e.message.indexOf(names[m]) > 0;
+
+      if (notFound) {
+        debug('Module %s not found, will try another candidate.', names[m]);
+        continue;
+      }
+
+      debug('Cannot load connector %s: %s', names[m], e.stack || e);
+      throw e;
     }
     if (mod) {
       break;
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "async-iterators": "^0.2.2",
     "eslint": "^3.12.2",
     "eslint-config-loopback": "^8.0.0",
+    "loopback-connector-throwing": "file:./test/fixtures/loopback-connector-throwing",
     "mocha": "^3.2.0",
     "should": "^8.4.0"
   },
--- a/test/datasource.test.js
+++ b/test/datasource.test.js
@@ -25,4 +25,24 @@ describe('DataSource', function() {
       });
     }).should.throw(/loopback-connector-throwing/);
   });
+
+  it('reports helpful error when connector init via short name throws', function() {
+    (function() {
+      // this is what LoopBack does
+      return new DataSource({
+        name: 'dsname',
+        connector: 'throwing',
+      });
+    }).should.throw(/expected test error/);
+  });
+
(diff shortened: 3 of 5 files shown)


bajtos:
The solution is not 100% reliable, as it is swallowing all MODULE_NOT_FOUND errors, regardless of what module was not found. If the connector is installed but cannot be loaded because its dependency is missing, then LoopBack will treat such connector as if it was not installed at all.

Unfortunately, Node.js core does not provide the name of the module being loaded in the error, the only extra property is `code` set to `MODULE_NOT_FOUND`. As a result, we cannot (easily) check which module was not found.

```
$ node -v
v6.9.4
$ node
> try { require('asd'); } catch(err) { console.log(err); }
{ Error: Cannot find module 'asd'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at repl:1:7
    at sigintHandlersWrap (vm.js:22:35)
    at sigintHandlersWrap (vm.js:96:12)
    at ContextifyScript.Script.runInThisContext (vm.js:21:12)
    at REPLServer.defaultEval (repl.js:346:29)
    at bound (domain.js:280:14) code: 'MODULE_NOT_FOUND' }
```

I guess we could check whether the error message contains the string that we are trying to `require()`, but I am not sure if that won't backfire on us, e.g. if the error message reported by Node.js changes in the future.

Thoughts?

kjdelisle:
@bajtos Needs rebasing -- conflict with `package.json`

loay:
LGTM
Needs rebasing as there is a conflict as mentioned by @kjdelisle

lib/datasource.js
```diff
@@ -224,7 +224,10 @@ function tryModules(names, loader) {
     try {
       mod = loader(names[m]);
     } catch (e) {
-      /* ignore */
+      if (e.code !== 'MODULE_NOT_FOUND') {
+        throw e;
```
kjdelisle: Building off your comment about checking the string for the module name:

We could still make that check here and if it fails, return a more generic error.
```
// snip
} catch (e) {
  if (e.code === 'MODULE_NOT_FOUND') {
    var isMissing = _.some(names, function(name) {
      return _.contains(message, name);
    });
    if (!isMissing) {
      // If none of them match, then the message isn't what we expected.
      e.message = util.format("One or more dependencies could not be found.",
      "See stack trace for details.");
    }
  }
  throw e;
}
```

The downside is that this will run into false positives (modules with names that are common English words), and is a much more involved check.

bajtos:
> We could still make that check here and if it fails, return a more generic error.

I prefer to preserve the original error, because it contains important information needed to resolve the problem.

I choose a simpler solution, please take a look at the updated code.

As I was working on this, I found that the test connector was actually not triggering the new path. It's fixed now.

kjdelisle:
@bajtos LTGM, but there's still some failing tests

bajtos:
@kjdelisle oops, I forgot to run `npm test` locally. Apparently my code cleanup in `tryModules` broke some of the tests, so I have reverted it back.

PR #1245, opened 2017-02-03
Add proper statusCode for duplicate

loay:
### Description
Add proper statusCode for http client error of Duplicate

#### Related issues

connect to https://github.com/strongloop/loopback/issues/2512

### Checklist

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -220,8 +220,11 @@ Memory.prototype._createSync = function(model, data, fn) {
     this.collection(model, {});
   }
 
-  if (this.collection(model)[id])
-    return fn(new Error(g.f('Duplicate entry for %s.%s', model, idName)));
+  if (this.collection(model)[id]) {
+    var error = new Error(g.f('Duplicate entry for %s.%s', model, idName));
+    error.statusCode = error.status = 409;
+    return fn(error);
+  }
 
   this.collection(model)[id] = serialize(data);
   fn(null, id);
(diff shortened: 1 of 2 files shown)


loay:
@bajtos @jannyHou PTAL. Thanks.

loay:
No it does not use status either ..  for example, in mysql example:
`'ER_DUP_ENTRY', errno: 1062, sqlState: '23000', index: 0 `
So I am not sure if I should be using the test case modification after all !

bajtos:
+1 for returning 409 instead of 500 when a duplicate id is detected in CREATE operation.

I think we need to fix all connectors to detect the duplicate entry and report a correct error back before we can land the change in the shared test suite.

lib/connectors/memory.js
```diff
@@ -220,8 +220,12 @@ Memory.prototype._createSync = function(model, data, fn) {
     this.collection(model, {});
   }
 
-  if (this.collection(model)[id])
-    return fn(new Error(g.f('Duplicate entry for %s.%s', model, idName)));
+  if (this.collection(model)[id]) {
+    var errMsg = g.f('Duplicate entry for %s.%s', model, idName);
+    var error = new Error(errMsg);
```
bajtos: Nitpick - feel free to ignore: I would personally simplify this lines to a single statement.

```js
var error = new Error(g.f(
  'Duplicate entry for %s.%s', 
  model, idName));
```

bajtos:
Alternatively, you can move the test for status code to memory-connector tests only, so that we can land this fix sooner and defer fixing other connectors for later.

Issue #1253, opened 2017-02-16
DataAccessObject.upsert doesn't pass options to Model.findOne


azatoth:
### Bug or feature request

- [x] Bug
- [ ] Feature request

### Description of feature (or steps to reproduce if bug)

On [lib/dao.js#L638](https://github.com/strongloop/loopback-datasource-juggler/blob/0cf863a483090b8927c76290da218817718b278c/lib/dao.js#L638), `upsert` creates a new `opts` object with a 'notify: false' and optionally with the transaction object.

### Expected result

`Model.find()` should have access to all given options.

### Actual result (if bug)

This results in that other entries like `accessToken` will not be visible in subsequent `findOne`call. The original `options` is only visible in the next `save()` or `updateAttributes()`

### Additional information (Node.js version, LoopBack version, etc)

lb3

bajtos:
@azatoth As I said on Gitter, this is a bug and I am pretty sure there will be more places like this one where are not passing the options through.

I am proposing to use the following approach in places where we need to modify the `options` object before it's passed down:

```js
var opts = Object.create(options);
opts.notify = false;
// add any other overrides as needed
Model.findOne({where: ctx.query.where}, opts, function(err, inst) {
  // ...
});
```

PR #1254, opened 2017-02-16
Fix DAO.upsert to pass options to DAO.find

azatoth:
Adding reuse of give options to subsequent `Model.find()` calls instead of passing a new stripped opts object.

- #1253

Closes: #1253

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -631,10 +631,9 @@ DataAccessObject.upsert = function(data, options, cb) {
         }
       });
     } else {
-      var opts = {notify: false};
-      if (ctx.options && ctx.options.transaction) {
-        opts.transaction = ctx.options.transaction;
-      }
+      var opts = Object.create(options);
+      opts.notify = false;
+
       Model.findOne({where: ctx.query.where}, opts, function(err, inst) {
         if (err) {
           return cb(err);
@@ -798,10 +797,9 @@ DataAccessObject.upsertWithWhere = function(where, data, options, cb) {
         }
       });
     } else {
-      var opts = {notify: false};
-      if (ctx.options && ctx.options.transaction) {
-        opts.transaction = ctx.options.transaction;
-      }
+      var opts = Object.create(options);
+      opts.notify = false;
+
       self.find({where: ctx.query.where}, opts, function(err, instances) {
         if (err) return cb(err);
         var modelsLength = instances.length;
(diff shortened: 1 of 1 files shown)


slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

bajtos:
@azatoth thank you for the pull request. Please add unit-tests to verify these changes - they should fail on current `master` and pass with your fix in place.

azatoth:
@bajtos Sadly I'm unable to write any test for this issue as I do not know how to check what options `Model.findOne()` actually received.

bajtos:
Usually, one would add an operation hook to observe what `options` are passed in the context. However, in the case of `upsert`, the `findOne` method is not triggering any hooks because `options.notify` is set to false.

This raises a question, why do you actually want the options to be passed from `upsert` to `findOne`? How is your application using these options?

Here are few options how to observe what `options` were passed to `findOne` in your tests:
 - override Model's `findOne` method with a custom implementation that stores the options
 - override connector's `all` method with a custom implementation to store the options

A mock-up example to show what I mean, the test should be added at the end of [describe('updateOrCreate')](https://github.com/strongloop/loopback-datasource-juggler/blob/76c73d55d28dfa150b9e004aeae12aafe0ce6475/test/manipulation.test.js#L728):

```js
var hasUpdateOrCreate =  !!getSchema().connector.updateOrCreate;
bdd.itIf(!hasUpdateOrCreate, 'should pass options down to findOne', function() {
  var findOneOptions; 
  var originalFindOne = Post.findOne;
  Post.findOne = function(query, options, cb) {
    if (!cb && typeof options === 'function') {
      cb = options;
      options = undefined;
    }
    findOneOptions = options;
    return originalFindOne.apply(this, arguments);
  };

  return Post.replaceOrCreate({id:1, title: 'test'}, {accessToken: 'a-token'})
    .then(() => {
      expect(findOneOptions).to.eql({accessToken: 'a-token'});
    })
    .finally(() => Post.findOne = originalFindOne);
});
```

bajtos:
@azatoth ping, what's the status of this pull request?

bajtos:
> Sadly I've transitioned to another project that doesn't use loopback, so I will not be able to finish this pull request.

Boo 😢 Thank you for letting us know and have fun in your new adventure!

PR #1255, opened 2017-02-16
Add missing options to scope method

azatoth:
- None

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -128,7 +128,7 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
           queryRelated.where[IdKey] = collectTargetIds(data, IdKey);
         }
 
-        relatedModel.find(queryRelated, cb);
+        relatedModel.find(queryRelated, options, cb);
       } else {
         cb(err, data);
       }


slnode:

slnode:

slnode:

slnode:

bajtos:
@slnode ok to test

ssh24:
@azatoth Could you please add some unit tests to go with your patch to stop future regressions as per @bajtos's comment?

felixruponen:
Hello! 

I would like to have this merged and can write some tests for it, I'm  a bit uncertain of where they should go. Perhaps you can push me in the right direction. Should I make a new test file or add a test case in an existing one, say "include.test.js".

@azatoth @bajtos

bajtos:
@felixruponen 
> I would like to have this merged and can write some tests for it

That's great to hear!

> I'm a bit uncertain of where they should go. Perhaps you can push me in the right direction. Should I make a new test file or add a test case in an existing one, say "include.test.js".

Considering that the change is made in `lib/scope.js`, I think the new tests should be added to [test/scope.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/d095cc116101d3b838c7c742a990b2a169205f34/test/scope.test.js).

I may be wrong since I am not very familiar with this part of juggler, can somebody from @strongloop/sq-lb-apex help please?

ssh24:
>  I think the new tests should be added to [test/scope.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/d095cc116101d3b838c7c742a990b2a169205f34/test/scope.test.js)

@azatoth Ditto!

jannyHou:
@azatoth thank you for the PR, I tend to add test cases in `test/relations.test.js`, since the change happens in function `related` and it's especially for a hasManyThrough relation.

Could you add some test case to explain which **options ** do you want to use? It's quite important to know where the `options` are expected to consume:
It either passes into the fetch function of throughModel, see https://github.com/azatoth/loopback-datasource-juggler/blob/020d07880ba3a196fc8dcfea5403310e4daf81fc/lib/scope.js#L105
or passes into the one of relatedModel as your pr does, but **NOT BOTH** of them.

At first I pass `options` into the throughModel's fetch but now I realize that relatedModel is probably the right place. So I agree with you, and please remove the `options` in Line105 as well :) Thanks.

jannyHou:
@azatoth Thank you so much for the contribution, I think same fix checked in with PR https://github.com/strongloop/loopback-datasource-juggler/pull/1426.

We really appreciate your code while I may need to close this PR as a duplicate, and the fix will in the new release. Thanks for understanding :)

PR #1259, opened 2017-02-16
Included models from include operations do not change defined `strict` model option

### LINKED ISSUE #1252: Saving a model retrieved from find operation that uses include filter for hasMany relation, also saves the related model as property in the database
### Bug or feature request

- [x] Bug
- [ ] Feature request

### Description of feature (or steps to reproduce if bug)
When you have two models **modelA** with `hasMany` relationship to **modelB** and query for **modelA** documents with `include` filter for **modelB** it will return model instances that will have a property with the relation name inside their data and the included document. The problem is that it will have set the `__strict` option of the model class to false, so if you make a change to this document and save it, it will also save the included document as a property in the database.

This is due to this line [here](https://github.com/strongloop/loopback-datasource-juggler/blob/26d25d60fdc90234ef0b2d73fa91ddd7603910d5/lib/include.js#L875) in **include.js** that sets the property `__strict` to `false` no matter what... I don't believe this is excepted behavior or it serves some purpose to set hardcoded the value of strict to false

### Expected result
It should add it as a property but not change strict flag of model. So when save it should follow the original configuration of the model and leave it with a `true` value if originally set like that

### Actual result (if bug)
It changes the option `__strict` to `false` no matter what for `hasMany` relation...

### Additional information (Node.js version, LoopBack version, etc)
**Node version: 7.5.0**
**Loopback version: 2.38.0**

mitsos1os:
### Description
Changed behaviour when including related models in find operations forces to change the flag of strict to false, so any change to the model will also save in the datasource tha included properties...
I believe it should not change the default strict setting as in the original model definition

#### Related issues

- #1252 

### Checklist

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -205,16 +205,17 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
       var relationType = ctor.relations[p].type;
 
       var modelTo;
-      /* Issue #1252
       if (!properties[p]) {
         modelTo = ctor.relations[p].modelTo || ModelBaseClass;
         var multiple = ctor.relations[p].multiple;
         var typeName = multiple ? 'Array' : modelTo.modelName;
         var propType = multiple ? [modelTo] : modelTo;
         properties[p] = {name: typeName, type: propType};
+        /* Issue #1252
         this.setStrict(false);
+        */
       }
-      */
+
       // Relation
       if (relationType === 'belongsTo' && propVal != null) {
         // If the related model is populated
--- a/lib/model.js
+++ b/lib/model.js
@@ -205,6 +205,7 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
       var relationType = ctor.relations[p].type;
 
       var modelTo;
+      /* Issue #1252
       if (!properties[p]) {
         modelTo = ctor.relations[p].modelTo || ModelBaseClass;
         var multiple = ctor.relations[p].multiple;
@@ -213,7 +214,7 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
         properties[p] = {name: typeName, type: propType};
         this.setStrict(false);
       }
-
+      */
       // Relation
       if (relationType === 'belongsTo' && propVal != null) {
         // If the related model is populated
--- a/test/include.test.js
+++ b/test/include.test.js
@@ -60,6 +60,42 @@ describe('include', function() {
     });
   });
 
+  it('should not have changed the __strict flag of the model', function(done) {
+    const originalStrict = User.definition.settings.strict;
+    User.definition.settings.strict = true; // Change to test regression for issue #1252
+    const finish = (err) => {
+      // Restore original user strict property
+      User.definition.settings.strict = originalStrict;
+      done(err);
+    };
+    User.find({include: 'posts'}, function(err, users) {
+      if (err) return finish(err);
+      users.forEach(user => {
+        user.should.have.property('__strict', true); // we changed it
+      });
+      finish();
+    });
+  });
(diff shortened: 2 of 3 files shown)


slnode:

slnode:

slnode:

ssh24:
@slnode ok to test

test/include.test.js
```diff
@@ -60,6 +60,42 @@ describe('include', function() {
...
+      User.definition.settings.strict = originalStrict;
+      done(err);
+    };
+    User.findOne({where: {id: 1}, include: 'posts'}, function(err, user) {
+      if (err) return finish(err);
+      user.save(function(err) { // save the returned user
+        if (err) return finish(err);
+        // should not store in db the posts
+        JSON.parse(User.dataSource.adapter.cache.User[1]).should.not.have.property('posts');
+        finish();
```
ebarault: why not doing an explicit find rather than getting the user from cache?
IMO it would make the test easier to read.

lib/model.js
```diff
@@ -213,7 +214,7 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
         properties[p] = {name: typeName, type: propType};
         this.setStrict(false);
       }
-
+      */
       // Relation
```
ebarault: could you comment why this code should be removed in the context of this PR?

test/include.test.js
mitsos1os: I accessed the Memory storage directly, because using a direct find, it would adhere to the defined model properties and since `posts` is not an actual model property it would not be attached to the model beucase it would be stripped of under the build process of the ModelConstructor. However it would still be in the db storage.
This it the same case with MongoDB. Retrieved models do not have the extra properties, but if you look into the database directly you will see them saved as a whole embedded document

lib/model.js
mitsos1os: I think you are right, this is an other place that just sets the strict false in hardcoded. But it also assigns the relation as a dynamically set property in the model. I could leave this and just comment the `setStrict` statement. I will revisit it

mitsos1os:
@ssh24 I cannot view the errors of the builds... The repository tests are all successful except this one here: https://github.com/mitsos1os/loopback-datasource-juggler/blob/issue-1252/test/loopback-dl.test.js#L1744 Which also fails on the master branch

ssh24:
@slnode test please

test/include.test.js
ebarault: oh yes, i see. And in the case of the memory connector I suppose we have no mean to access the data directly through a raw connector (such as with native mongo driver) ? 
Or shall we consider it IS the only way to access the memory storage directly? (other than persisiting to file and parsing that file)

mitsos1os: Yes, that's the point of memory connector! No file! :smile:

I got this property path by looking into the memory connector and where it saves the data

mitsos1os:
I cannot understand why the tests are failing... The travis ci build is failing on
`npm ERR! 404 'loopback-connector-throwing' is not in the npm registry.`

The Jenkins build do not load a webpage when clicking on the details link...

I would appreciate it if we resolved these issues so that the request gets accepted.

ssh24:
@slnode test please

ssh24:
@slnode ok to test

ssh24:
@mitsos1os I was looking into your changes and it is breaking change to our existing code base. Currently all the databases we have as a downstream to juggler: `dashdb, mongodb, mssql, mysql, oracle, postgresql` are failing because of your change and more importantly failing your test case as well. 

Please take a look into all of those datasources and fix your test case according to that.

Issue #1247, opened 2017-02-06
Server crashes when entering invalid date


joostdebruijn:
### Bug or feature request

- [x] Bug
- [ ] Feature request

### Description of feature (or steps to reproduce if bug)

1. Add a model with a property of the type date;
2. Make a custom function (e.g. with Async) which updates the date property;
3. Call this function with an invalid date as a string.
4. The server crashes.

Remark: It does not happen when you update the model via a non-wrapped function call, because the error is catched by the error handler mechanism of LoopBack.

### Link to sample repo to reproduce issue (if bug)

See: https://github.com/joostdebruijn/loopback-date-issue

1. Clone the repo.
2. Run `npm install`.
3. Start the server with `node .`
4. Create notes:

```
POST /api/Notes
[
  {
    "title": "Test 1",
    "content": "Test",
    "date": "2017-02-06"
  },
  {
    "title": "Test 2",
    "content": "Test",
    "date": "2017-02-06"
  }
]
``` 
5. Update notes in bulk:

```
POST /api/Notes/updateDatesInBulk
[
  {"id": 1, "date": "thiswillfail"}
]
``` 
6. The server will crash.

### Expected result
Receive a validation error in the callback which could be handled.

### Actual result (if bug)
The server crashes because the error is thrown, trace:

```
\node_modules\loopback-datasource-juggler\lib\model-builder.js:531
    throw new Error(g.f('Invalid date: %s', arg));

Error: Invalid date: thiswillfail
    at DateType (node_modules\loopback-datasource-juggler\lib\model-builder.js:531:11)
    at ModelConstructor.set [as startTime] (node_modules\loopback-datasource-juggler\lib\model-builder.js:467:81)
    at ModelConstructor.setAttribute (node_modules\loopback-datasource-juggler\lib\dao.js:2760:14)
    at ModelConstructor.setAttributes (node_modules\loopback-datasource-juggler\lib\dao.js:2793:10)
    at validateAndSave (node_modules\loopback-datasource-juggler\lib\dao.js:3101:12)
    at node_modules\loopback-datasource-juggler\lib\dao.js:3085:7
    at doNotify (node_modules\loopback-datasource-juggler\lib\observer.js:99:49)
    at doNotify (node_modules\loopback-datasource-juggler\lib\observer.js:99:49)
    at doNotify (node_modules\loopback-datasource-juggler\lib\observer.js:99:49)
[...]

PR #1266, opened 2017-03-15
dao: catch sync errors on setAttributes

kjdelisle:
### Description
Wrap inst.setAttributes in try-catch to prevent synchronous errors
from crashing the application.

#### Related issues
#1247 

### Checklist

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -685,6 +685,15 @@ describe('manipulation', function() {
         });
       });
 
+    it('should fail if field validation fails', function(done) {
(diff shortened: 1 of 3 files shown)


test/manipulation.test.js
```diff
@@ -685,6 +685,15 @@ describe('manipulation', function() {
         });
       });
 
+    it('should fail if it fails field validation', function(done) {
```
ssh24: could be named better: `should fail if field validation fails`

kjdelisle: true true, changing

PR #1267, opened 2017-03-16
Update error message

loay:
### Description

Update error message to reflect proper translation syntax.


0candy:
Please do not check in the 2 `intl/zz` files.

PR #1276, opened 2017-03-21
Fixes #1275. `Include` filter transforms fields property into array.

### LINKED ISSUE #1275: Issue with incorrect include on HasManyThrough relation.
### Bug or feature request

- [x] Bug
- [ ] Feature request

### Description of feature (or steps to reproduce if bug)
There is a bug when someone sends a request with incorrect `include` filter containing "fields" equal to a String instead of an Object/Array on a model that has a relation HasManyThrough.

eg: 
```
{
  "include": {
    "relation": "oneExistingRelation",
    "scope": {
      "fields": "dummyStringProperty"
    }
  }
}
```

### Link to sample repo to reproduce issue (if bug)
https://github.com/nikooiko/loopback-sandbox

You just have to `npm start` && send this request:

```
http://0.0.0.0:3000/api/AppUsers?filter=%7B%22include%22%3A%20%7B%22relation%22%3A%20%22roles%22%2C%20%22scope%22%3A%20%7B%22fields%22%3A%20%22dummyField%22%7D%7D%7D
```

### Expected result
Application should probably return the results with the included field containing only the "id" and the "dummyStringProperty" 

### Actual result (if bug)
After a request like this the application simply dies after receiving this error:
TypeError: Cannot read property 'toString' of undefined
   at linkManyToMany: ...

(you can see the actual error at provided sandbox).

### Additional information (Node.js version, LoopBack version, etc)
node v7.7.0

### Related issues

Can be related to #951

nikooiko:
`Include` filter takes into consideration string property 'fields' and transforms it into an array containing this string.

#### Related issues

- #1275

### Checklist

### DIFF
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -835,6 +835,24 @@ describe('relations', function() {
         });
       });
 
+      context('findById with include filter that contains string fields', function() {
+        it('should accept string and convert it to array', function(done) {
+          var includeFilter = {include: {relation: 'patients', scope: {fields: 'name'}}};
+          var physicianId = physician.id;
+          Physician.findById(physicianId, includeFilter, function(err, result) {
+            should.not.exist(err);
+            should.exist(result);
+            result.id.should.eql(physicianId);
+            should.exist(result.patients);
+            result.patients().should.be.an.instanceOf(Array);
+            should.exist(result.patients()[0]);
+            should.exist(result.patients()[0].name);
+            should.not.exist(result.patients()[0].age);
+            done();
+          });
+        });
+      });
+
       function createSampleData(done) {
         Physician.create(function(err, result) {
           result.patients.create({name: 'a', age: '10'}, function(err, p) {
--- a/lib/include.js
+++ b/lib/include.js
@@ -97,12 +97,16 @@ function execTasksWithInterLeave(tasks, callback) {
   // Context Switch BEFORE Heavy Computation
   process.nextTick(function() {
     // Heavy Computation
-    async.parallel(tasks, function(err, info) {
-      // Context Switch AFTER Heavy Computation
-      process.nextTick(function() {
-        callback(err, info);
+    try {
+      async.parallel(tasks, function(err, info) {
+        // Context Switch AFTER Heavy Computation
+        process.nextTick(function() {
+          callback(err, info);
+        });
       });
-    });
+    } catch (err) {
+      callback(err);
+    }
   });
 }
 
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

ssh24:
@nikooiko Could you please add some unit tests to stop future regressions?

nikooiko:
Ì'm going to put some tests tomorrow

test/relations.test.js
```diff
@@ -835,6 +835,23 @@ describe('relations', function() {
         });
       });
 
+      context('findById with filter include that contains scope with string filter fields', function() {
```
ssh24: change this to `context('findById with include filter that contains string fields')`

test/relations.test.js
```diff
@@ -835,6 +835,23 @@ describe('relations', function() {
...
+        it('should be accepted and convert fields to array.', function(done) {
```
ssh24: change this to `it('should accept string and convert it to array')`

test/relations.test.js
```diff
@@ -835,6 +835,23 @@ describe('relations', function() {
...
+          var includeFilter = {include: {relation: 'patients', scope: {fields: 'name'}}};
+          var physicianId = physician.id;
+          Physician.findById(physicianId, includeFilter, function(err, ch) {
```
ssh24: please call it `result` instead of `ch` in the callback and replace all the subsequent `ch` with `result`

test/relations.test.js
```diff
@@ -835,6 +835,23 @@ describe('relations', function() {
...
+            should.not.exist(err);
+            should.exist(ch);
+            ch.id.should.eql(physicianId);
+            should.exist(ch.patients);
```
ssh24: add another assertion to actually check for the array instance, something like `result.patients().should.be.an.instanceOf(Array);`

ssh24:
@slnode ok to test

ssh24:
@nikooiko Thank you for the contribution and the patch! The PR is ready to be landed

While in this case everything worked out fine, just an advice **Please do not use your fork's master to submit PR**. It is not a good practice to use the fork's master to commit changes and submit a PR for it. The best practice is to create a branch (specific for each patch) from your fork's master and submit a PR from that branch specific to the patch you are submitting. That results in a cleaner way to rebase your code and easier for us to land the code as well.

Happy coding!

PR #1277, opened 2017-03-21
Added notify flag for create and upsert

jsheely:
create and patch now honor the option.notify flag. You may not always want to notify operational hooks that something has changed. 

Hopefully I did this pull-request correctly this time.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -398,10 +398,14 @@ DataAccessObject.create = function(data, options, cb) {
                   hookState: hookState,
                   options: options,
                 };
-
-                Model.notifyObserversOf('after save', context, function(err) {
-                  cb(err, obj);
-                });
+                                
+                if(options.notify !== false) {
+                  Model.notifyObserversOf('after save', context, function(err) {
+                    cb(err, obj);
+                  });
+                }else {
+                  cb(null, obj);
+                }
               });
             });
           });
@@ -623,9 +627,13 @@ DataAccessObject.upsert = function(data, options, cb) {
                 options: options,
               };
 
-              Model.notifyObserversOf('after save', context, function(err) {
-                cb(err, obj);
-              });
+              if(options.notify !== false) {
+                Model.notifyObserversOf('after save', context, function(err) {
+                  cb(err, obj);
+                });
+              }else {
+                cb(null, obj);
+              }
             }
           });
         }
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

ssh24:
@jsheely This PR is correct this way! :)

Could you please add some unit tests to stop future regressions?

lib/dao.js
```diff
@@ -398,10 +398,13 @@ DataAccessObject.create = function(data, options, cb) {
                   hookState: hookState,
                   options: options,
                 };
-
-                Model.notifyObserversOf('after save', context, function(err) {
-                  cb(err, obj);
-                });
+                if (options.notify !== false) {
```
ssh24: better way to write it, `if (options.notify)`

lib/dao.js
```diff
@@ -623,9 +626,13 @@ DataAccessObject.upsert = function(data, options, cb) {
                 options: options,
               };
 
-              Model.notifyObserversOf('after save', context, function(err) {
-                cb(err, obj);
-              });
+              if (options.notify !== false) {
```
ssh24: better way to write it, `if (options.notify)`

jsheely: Except the default is to notify. options.notify is null by default.

ssh24: Right. It should be `if (!options.notify)`

jsheely: That is also not valid because of options.notify being null by default. 

Unless you explicitly say you don't want to be notified. You're getting notified.

```
Model.create(model, {notify: false});
```

ssh24:
@jsheely Could you please add some unit tests to go with this patch?

jsheely:
I'll give it a try

ssh24:
@jsheely Any update on adding some unit tests? I want to proceed with this PR as it has been idle for a while.

jsheely:
Trying to figure out what the proper thing to do is for throwing an error on a function that should never get called.

```javascript
it('aborts when `after save` fires when option to notify is false', function (done) {
	TestModel.observe('after save', function () {
	  done(new Error('should not have notified')); // Don't think this is right.
	});

	TestModel.create({ name: 'created' }, { notify: false });
});
```

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
@jsheely Thank you for the contribution! PR is ready to land.

PR #1288, opened 2017-03-27
Add name mapping to Discovery API

kjdelisle:
Add the mapName prototype to allow consistent mapping across
discovery functions, and add hooks into each discovery function
to allow in-flight override of the name mapping.

Additionally, the prototype can also be overridden to replace the
behaviour across all discovery functions.

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -911,6 +911,18 @@ DataSource.prototype.autoupdate = function(models, cb) {
 };
 
 /**
+ * Name mapper function for converting names of certain object types to
+ * the appropriate name. Override this function to provide custom mapping.
+ */
+DataSource.prototype.mapName = function(type, name) {
+  if (type === 'table' || type === 'model') {
+    return fromDBName(name, false);
+  } else {
+    return fromDBName(name, true);
+  }
+};
+
+/**
  * Discover existing database tables.
  * This method returns an array of model objects, including {type, name, onwer}
  *
@@ -992,7 +1004,9 @@ DataSource.prototype.discoverModelProperties = function(modelName, options, cb) 
   cb = cb || utils.createPromiseCallback();
 
   if (this.connector.discoverModelProperties) {
-    this.connector.discoverModelProperties(modelName, options, cb);
+    var nameMapper = options.nameMapper || this.mapName;
+    var name = nameMapper('table', modelName);
+    this.connector.discoverModelProperties(name, options, cb);
   } else if (cb) {
     process.nextTick(cb);
   }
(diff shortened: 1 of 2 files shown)


test/discovery.test.js
```diff
@@ -609,6 +672,27 @@ describe('discoverExportedForeignKeys', function() {
...
+
+  it('should use custom nameMapper function to `discoverExportedForeignKeys`', function(done) {
+    // Custom version of fake -- need to assert that modelName is being overridden!
+    ds.connector.discoverExportedForeignKeys = function(modelName, options, cb) {
+      modelName.should.equal('INVENTORY');
+      process.nextTick(function() {
+        cb(null, exportedForeignKeys);
+      });
+    };
+    var customMapper = function(type, name) {
```
b-admike: you can make this customMapper() function global and call it within your tests.

test/discovery.test.js
```diff
@@ -609,6 +672,27 @@ describe('discoverExportedForeignKeys', function() {
...
+      return name.toUpperCase();
+    };
+    // Table name is intentionally lowercase!
+    ds.discoverExportedForeignKeys('inventory', {
+      nameMapper: customMapper,
+    }, function(err, schemas) {
+      if (err) return done(err);
+      schemas.should.be.eql(exportedForeignKeys);
```
jannyHou: nitpick: `schemas` are used in all discover functions, while not all of them return schemas, probably change it to `result`

lib/datasource.js
```diff
@@ -992,7 +1004,9 @@ DataSource.prototype.discoverModelProperties = function(modelName, options, cb)
   cb = cb || utils.createPromiseCallback();
 
   if (this.connector.discoverModelProperties) {
-    this.connector.discoverModelProperties(modelName, options, cb);
+    var nameMapper = options.nameMapper || this.mapName;
+    var name = nameMapper('table', modelName);
+    this.connector.discoverModelProperties(name, options, cb);
```
jannyHou: function `fromDBName` in `mapName` will change the value of `modelName` here, and same for all discover apis.
e.g. the postgresql fails due to `fromDBName` convert `customer_test` to `CustomerTest` 
cc @loay can confirm with other connectors.

kjdelisle: This is part of the problem. Currently, the tests are failing because their expectation is unreasonable; they mock up a database-specific relationship between the Model and the table in postgres:
```
'name': 'CustomerTest',
        'options': {
          'idInjection': false,
          'postgresql': {
            'schema': 'public',
            'table': 'customer_test',
          },
        },
 // etc
```

As a result, this test is hiding the fact that the connector itself does not map the name of the Model (what the user would expect to see) to the database table.

The goal of this is ultimately to give them a default, let each connector override that default if necessary, and finally, give the user a way to arbitrarily map however they please if all else fails.

kjdelisle:
Going to close this for now.

PR #1289, opened 2017-03-27
review model settings merge/inheritance policy (util/mergeSettings() ) : relations, acls, ...

ebarault:
### Description

ModelClass.extend(className, subclassProperties, subclassSettings) calls utils.mergeSettings(settings, subclassSettings) to merge the settings from subclass into the settings of the baseModel.

implementation of utils.mergeSettings() for objects is rather straight forward and merge recursively nested values with the following logic:

- `if` the **base** model item is **null**, **undefined** or **not an object**
  - **replace** the base model item with the subclass value  
- `else If` the **target** model item is **null**, or **not an object**  
  - **replace** the base model item with the subclass item  
- `else` 
  - **call mergeSettings on the nested object**  

**Issue:** A given child's relation should not be merged with an equally named parent's relation, but should replace it

Consider the following situation encountered in https://github.com/strongloop/loopback/issues/3308 and https://github.com/strongloop/loopback/issues/3307:

Built-in AccessToken model has the following relation with the User model:

```js
"user": {
      "type": "belongsTo",
      "model": "User",
      "foreignKey": "userId"
    }
```

CustomAccessToken, extending AccessToken has the following relation:
```js
"user": {
      "type": "belongsTo",
      "idName": "id",
      "polymorphic": {
        "idType": "string",
        "foreignKey": "userId",
        "discriminator": "principalType"
      }
    }
```

then the result of the merge for this relation in CustomAccessToken is:

```js
"user": {
      "type": "belongsTo",
      "idName": "id",
      "model": "User", // <--------- extraneous
      "foreignKey": "userId", // <--------- extraneous
      "polymorphic": {
        "idType": "string",
        "foreignKey": "userId",
        "discriminator": "principalType"
      }
    }
```

Based on these inconsistent settings, the juggler is not capable of setting the relation which ends undefined.

This PR fixes the utils.merge() function to ensure a child relation replaces an equally named relation from the parent.
 

#### Related issues

- https://github.com/strongloop/loopback/issues/3308   -
- https://github.com/strongloop/loopback/issues/3307

### Checklist


test/util.test.js
```diff
@@ -208,9 +208,62 @@ describe('mergeSettings', function() {
 
     should.deepEqual(dst.acls, expected.acls, 'Merged settings should match the expectation');
   });
+
+  it('should merge relations correctly', function() {
+    var src = {
```
ssh24: Maybe change the name to `child` instead of `src` for clarity

test/util.test.js
```diff
@@ -208,9 +208,62 @@ describe('mergeSettings', function() {
...
+          idName: 'id',
+          polymorphic: {
+            idType: 'string',
+            foreignKey: 'userId',
+            discriminator: 'principalType',
+          },
+        },
+      },
+    };
+    var tgt = {
```
ssh24: Also change the name to `parent` instead of `tgt` for clarity

test/util.test.js
```diff
@@ -208,9 +208,62 @@ describe('mergeSettings', function() {
...
+        },
+        someRelation: {
+          type: 'hasMany',
+          model: 'someModel',
+          foreignKey: 'modelId',
+        },
+      },
+    };
+
+    var dst = mergeSettings(tgt, src);
```
ssh24: `mergeSettings(parent, child)`

bajtos: +1, I would personally also switch the order in which the variables are defined - start with the parent, then define a child. I think it would be helpful to add comments explaining what these settings are mimicking (LoopBack's built-in AccessToken model, a custom model extending the built-in AccessToken model).

ebarault:
Thanks for your reviews. I'll apply the requested changes on tests.

> This implementation will pick relations key at any level of the settings object.

I don't see valid scenarios where it's safe to build a relation based on the merge of 2 objects. Relations can have quite complex configurations. I believe if one expects to alter a relation inherited from the base class, he assumes that he needs to provide the full definition of the relation.

Could we consider this approach as valid as-is and quickly react in case we see related issues?
Or i need to add another param to utils.mergeSettings() so that the method knows this is the first call (with the full settings object)

bajtos:
> I don't see valid scenarios where it's safe to build a relation based on the merge of 2 objects. Relations can have quite complex configurations. I believe if one expects to alter a relation inherited from the base class, he assumes that he needs to provide the full definition of the relation.

My concern is about scenarios where the same key `relations` is used somewhere deep inside settings for something that's not related to LoopBack/Juggler relation settings.

> Could we consider this approach as valid as-is and quickly react in case we see related issues?
Or i need to add another param to utils.mergeSettings() so that the method knows this is the first call (with the full settings object)

That's the question! I'd like to hear the opinion of more people before making the decision.

BTW instead of adding many new parameters to `utils.mergeSettings`, perhaps we can change only the place calling this util method? Something like the following:

```js
    // Merge the settings
    var originalSubclassSettings = subclassSettings;
    subclassSettings = mergeSettings(settings, subclassSettings);
    // relation config of individual relations must be copied, not merged
    for (var key in settings.relations) {
      if (key in originalSubclassSettings) return;
      subclassSettings.relations[key] = settings.relations[key];
    }
```

That reminds me of an important scenario to include in our tests: inheritance of model relations that are not changed by the subclass.

Consider the following loopback code:

```js
app.registry.createModel({
  name: 'Customer', 
  base: 'User',
});
```

Since the built-in `User` model comes with "hasMany accessTokens" relation, the `Customer` model should inherit this relation as-is. 

Thoughts?

ebarault:
> My concern is about scenarios where the same key relations is used somewhere deep inside settings for something that's not related to LoopBack/Juggler relation settings.

yes, i got you. But in that case where it's unrelated to LoopBack, what's a probability that a nested value of a LoopBack-unrelated relation object would conflict with a parent model Loopback-unrelated relation object with the same name, at the exact same nesting level? 
And why the developer would like a merge of these relation objects (which it does right now) rather than a replacement?

I prepared the following code to be ready for further exceptions as i suspect the current merging policy is/can be the origin of other problems.

```js
var isRelations = key === 'relations';
var replace = isRelations; // could be completed like = isRelations || isACLs || etc.
```

That being said, i can modify the code to only modify relations at root level, i took good note of your implementation proposal.

> About further tests

yes the test coverage is rather poor considering the key role of this little function. I can easily add a bunch of other tests, starting if your proposal.

The more we add test, the more we will detect inconsistencies generated by the current merge policy i think.

ssh24:
+1 with @bajtos's proposal to only look and modify for `relations` object in the root level only.

ebarault:
hi, 
On second thoughts, i'm not in favor of modifying the place that calls the function, as i believe relations are not an isolate case, take for example ACLs, they should not be merged either, but replaced by the child definition. So it will result in balancing the merging logic both in modelClass.extend() and in util.mergeSettings(), which is not good.

It was already hard enough to isolate the root cause of this bug, we need to fix in one place.

Given the critical role played by util.mergeSettings() in model inheritance we should rather consolidate this function by: 
- mapping which property should be merged,
- which should not be merged,
- and what it is the default policy for non-loopback-settings-properties. *

\* For the latter, i think it should be 'replace', not 'merge', as the latter in counter-intuitive.


On a different level, i also think the place of this merge function in an util package is not quite right, as it carries business logic. It's only used in the modelClass.extend() function to merge settings and to merge properties.
At some point if there are different merging policies for properties and settings, then maybe there should be 2 different functions.

Waiting for some discussion here before going on.

ssh24:
> Given the critical role played by util.mergeSettings() in model inheritance we should rather consolidate this function by: 
> - mapping which properties should be merged,
> - which should not be merged,
> - and what it is the default policy for non-loopback-settings-properties. *

@ebarault Which properties should be merged rather than replaced? Is it not the case that all properties (relations, acls, etc.) of a `base` model be replaced by the respective properties in the `child` model?

ebarault:
@ssh24 thanks for keeping the ball rolling!
so that's an excellent question, let's inspect the model json spec: according to [Model-definition-JSON-file](http://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#top-level-properties), + my own experience, the top level properties of type object or array are:

**acls, description, options, properties, relations, scopes, _validations_, methods, mixins, hidden (+ alias hiddenProperties), protected (+ alias protectedProperties)**

my analysis: 
- acls: each acl redefined in child should replace the parent's one
- description: (can be an array in case of multiline desc) : should not be merged, imagine the mess...
- [options](http://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#options): see below
- properties: compound properties redefined in child should replace the parent's ones
- relations: each relation redefined in child should replace the parent's one
- [scopes](http://loopback.io/doc/en/lb3/Model-definition-JSON-file.html#scopes): very risky IMO to merge any of inner scopes
- validations: not implemented yet, but these will be complex compound objects, individual validations should not be merged IMO
- methods: not officially supported, but works. again these are very complex compound objects, very risky to merge
- mixins: same think
- hidden/protected : these arrays could me merged i believe, but that means has no mean to unhide/unprotect a hidden/protected property inherited from the parent, this can be intentional -> to be discussed


**about options**: this is clearly a grey area: the doc is far from being up-to-date, there can be a bunch of things in there. I believe not many will be inherited from built-in LoopBack models. So it's mainly about model inheritance from custom base models. Here i'd apply also the replace/dontMerge logic, as for any user provided property inside model settings, as it's far too risky and unintuitive to merge inner arrays/properties.

**So long story short**: yes i'd apply the replace/dontMerge policy nearly everywhere:
- **LoopBack top level properties of type array**: description (hidden and protected to be discussed)
- **inner objects of LoopBack top level properties of type object**
- **any unknown top level objects/arrays** (custom, non LoopBack)

@ssh24, @bajtos : please confirm/discuss so i can finalize this

ssh24:
This looks good to me. 

However, the current `acl` logic for merging looks really funky to me.
> the current algorithm for array merge looks if the exact same ACL exist, and add it otherwise

I agree with you, `principalType + principalId` would be a good way to characterize a unique ACL and then use that key to replace any existing base model ACL's with the child model if it exists.

Would like some comments from @bajtos about this.

bajtos:
Hello, I am afraid I don't have any definite answer on this topic. I'd like @davidcheung and/or @jannyHou to join the discussion, IIRC one of them were looking into model options inheritance in the past - see https://github.com/strongloop/loopback/issues/157.

Here is a list of issues related to model inheritance - there may in relevant information in them. I am afraid I don't have bandwidth to go through them and distill a short summary for our discussion here. Can somebody else do that please?

- #2274
- #2173
- #1927
- #1485
- #324
- #157
- #2270

Before digging into the details of the proposal, we need to keep in mind backwards-compatibility requirements. If we change the way how model inheritance works, then most likely we will need a feature flag allowing users to opt into this new behaviour.

> acls: each acl redefined in child should replace the parent's one, but it's a bit tricky because acls in an array and can omit a number of optional fields, see below

My knowledge of ACLs is limited. The proposal to use principalType + principalId to identify "same" ACL entries looks reasonable to me FWIW.

@raymondfeng @ritch could you please join this discussion?

> description: (can be an array in case of multiline desc) : should not be merged, imagine the mess...

Makes sense.

> properties: compound properties redefined in child should replace the parent's ones
> relations: each relation redefined in child should replace the parent's one
> methods: not officially supported, but works. again these are very complex compound objects, very risky to merge
> scopes: very risky IMO to merge any of inner scopes

This makes sense to me as long as the merge/replace decision is made for each individual property/relation/method. I.e. we must not replace whole `settings.properties` object, but only replace entries in that object.

> mixins: same think

If `Base` model wants to apply `MixinX` with `ConfigA`, and `Child` model wants to apply `MixinX` with `ConfigA`, then perhaps we should apply the mixin twice, once for each configuration?

> validations: not implemented yet, but these will be complex compound objects, individual validations should not be merged IMO. (/!\ it's an array)

I think validations should be additive - validations defined by child models should be added to those inherited from the parent.

> hidden/protected : these arrays could me merged i believe, but that means has no mean to unhide/unprotect a hidden/protected property inherited from the parent, this can be intentional -> to be discussed

I think it's ok to merge these arrays for now.

ebarault:
thanks @bajtos for keeping the ball rolling here: i'll undertake the review of the issues you provided unless someone wants to backup me or help in the quest.

> only replace entries in that object.

sure 👍 

> feature flag

some of these are pure bugs to me, we shall decide on a per settings.property basis
👉  we could introduce the new merge/replace algorithm for ACL as a feature flag to implement it quickly without too much risks

> mixins

hmmm, need to think more about this one

**how to progress:**
- First review the issues 
- To make sure the new code is readable and maintainable, given its importance, i'll map a spec of the settings with policies to apply for each groups of inner properties, something like
```js
{
  description: {replace: true},
  relations: {replace: true},
  acls: {replace: true, keys: {principalType: true, principalId: true}},
  validation: {merge: true},
  // ...
}
```

👇  would should i appoint as reviewer for this one? i'd like someone with sufficient bandwidth/overall knowledge to challenge my proposals please. Thanks 👍

raymondfeng:
@ebarault For ACL merging, we can possibly give each entry different weight so that the ones from the sub model will have precedence over those inherited from the base model. The `ACL.getScore` method can be enhanced to honor such factor.

ebarault:
@raymondfeng : i see. yes it's maybe less error prone than identifying an ACL based on `principalType + principalId`.
Then we'd need to add something like a `rank` property to ACLs on the fly when merging settings from a parent model to a child model, iteratively. This `rank` factor would then contribute in the ACL score computing.

Is this what you have in mind?

raymondfeng:
@ebarault Yes.

ebarault:
@raymondfeng : excellent, let's do this.

superkhau:
@ebarault's proposal looks good to me. I've always hated the non-replace ACL behaviour, ie. I tell it to allow, and I still get deny because it's ranked higher. It really should just do what I tell it to and not force me to learn the ranking algorithm.

jannyHou:
I am afraid my knowledge of model builder is very limited, from a user's perspective, I would expect  
```json
{
  "objectTypeSetting": {
    "objitem": "value" //root level item
  },
  "arrayTypeSetting": [
    {"arritem": "value"} //root level item
  ],
  "nonObjArrSetting1": "str",
  "nonObjArrSetting2": "boolean"
}
```

- **objectTypeSettings**: 
  - root level item exist in both childModel & parentModel, do replace, merging items seems risky
  - root level item exist only in parentModel, keep it.
- **arrayTypeSetting**: 
  - acl: agree with https://github.com/strongloop/loopback-datasource-juggler/pull/1289#issuecomment-291641482
  - others: e.g.validations, not quite sure...if the object contains same properties, do replace, otherwise array.push()
- **nonObjArrSetting**: do replace

For relation inheritance, I personally tend to remove all relations from parentModel and define everything in childModel from scratch, but give users a warning and explicitly tell them to redefine relations.

ebarault:
@jannyHou @ssh24 @bajtos @superkhau @raymondfeng 

could you PTAL? i took into account most (if not all) comments in my last commit. Some details should still be improved/tinkered as well as some tests added (feedbacks welcome regarding test additions) but I'd like your general feedbacks before going further.

**NOTE**: currently one test is failing given the new ACL rank property. I expect comments from you whether we should bring this new behavior with a feature flag or if I should change the failing test to account for this change.

lib/utils.js
```diff
@@ -375,9 +375,29 @@ function parseSettings(urlStr) {
  *
  * @param {Object} target The target settings object
  * @param {Object} src The source settings object
+ * @param {Boolean} replaceTarget Should the src object replace the target value if it exists
  * @returns {Object} The merged settings object
  */
-function mergeSettings(target, src) {
+function mergeSettings(target, src, replaceTarget) { // NOTE: src is subclassSettings
+  const mergePolicy = {
+    description: {replace: true}, // string or array
```
raymondfeng: Would `override` be a better term?

lib/utils.js
```diff
@@ -375,9 +375,29 @@ function parseSettings(urlStr) {
...
+    options: {innerReplace: true}, // object
```
raymondfeng: Maybe `patch: true`?

lib/utils.js
```diff
@@ -392,27 +412,46 @@ function mergeSettings(target, src) {
...
+    let baseModelAclRank;
     if (target != null && typeof target === 'object') {
       // Add properties from target to dst
       Object.keys(target).forEach(function(key) {
+        if (key === 'acls' && Array.isArray(target[key])) {
+          let targetACLs = target[key];
+          targetACLs.forEach(function(acl) {
+            acl.rank = acl.rank || 1;
+            // capture the base model ACL rank
+            baseModelAclRank = baseModelAclRank || acl.rank;
```
raymondfeng: Don't we have to find out the maximum rank from target ACLs? IIUC, you code will get the last positive rank from all entries.

lib/utils.js
```diff
@@ -375,9 +375,29 @@ function parseSettings(urlStr) {
...
+    hidden: {replace: false}, // array
+    protected: {replace: false}, // array
+    indexes: {innerReplace: true}, // object
+    methods: {innerReplace: true}, // object
+    mixins: {innerReplace: true}, // object
+    // validations: {innerReplace: true}, // object // NOTE: not implemented
+    relations: {innerReplace: true},
+    scope: {replace: true}, // object
+    scopes: {innerReplace: true}, // object
+    acls: {replace: false}, // array
```
raymondfeng: ACLs shadow the base ones by rank.

lib/utils.js
```diff
@@ -375,9 +375,29 @@ function parseSettings(urlStr) {
...
+    // any matching non array/object property is replaced
+    // any matching custom top-level object/array is replaced
+    default: {replace: true},
+  };
+
```
raymondfeng: How to describe a `remove`? Is it `replace with undefined`? Sometimes the sub model decides to remove properties/methods/... from the base model. For example, we allow the following syntax today:
```
{
  base: 'User',
  properties: {
     username: null; // or false
  }
}
```

lib/utils.js
ebarault: as new acls are pushed in the array, the last one always has the highest rank so there's no need to check for the max.
but i'll go on with you proposal to make it more solid.

lib/utils.js
ebarault: are you suggesting to use the shadow keywork in the merge policy spec regarding the acls algorithm?
`acls: {shadow: true}`

anyhow this makes sense to make the ranking algo generic

lib/utils.js
ebarault: the current code handling this is
```
if (srcValue == null || typeof srcValue !== 'object') {
  // The source item value is null, undefined or not an object
  dst[key] = srcValue;
}
```
**NOTE**: `src` reffers to subClass settings

it's pretty straight forward do we have extra configuration to add here?

ebarault:
@raymondfeng thanks for your review. I left inlined comments.

Also, i thought of one important thing + one question:
- ACLs can also be persisted by user in db. Relevant rank should be added on the fly to persisted ACLs, maybe an operation hook. We won't be able to support direct management of ACLs at db level outside of the juggler's control.
-  Is there added value in the ACL score calculation to keep classified by strict inheritance rank ?
for example, consider modelA>modelB>modelC, modelA & C have ACLs, modelB has none. Should we prefer
    - **option 1:** modelA.ACL.rank=1, modelC.ACL.rank=2
    - **option 2:** or modelA.ACL.rank=1, modelC.ACL.rank=3

the way i plan to modify the scoring algorithm i don't see the requirement for option 2 but i prefer to leave the door open for your comments

raymondfeng:
@ebarault Thanks for the prompt response.

> ACLs can also be persisted by user in db. Relevant rank should be added on the fly to persisted ACLs, maybe an operation hook. We won't be able to support direct management of ACLs at db level outside of the juggler's control.

I feel ACL entries from DB have higher ranks than static ones from code/json. Maybe we can define an option flag for that. WDYT? 

> Is there added value in the ACL score calculation to keep classified by strict inheritance rank ?
for example, consider modelA>modelB>modelC, modelA & C have ACLs, modelB has none. Should we prefer
option 1: modelA.ACL.rank=1, modelC.ACL.rank=2
option 2: or modelA.ACL.rank=1, modelC.ACL.rank=3

I would prefer option 2 to be consistent. Do we want to expose `rank` to api/json?

> Feature flag or not ?

Probably yes. It's a breaking change, right?

lib/utils.js
raymondfeng: From merging perspective, `{replace: false}` will instruct LoopBack to combine ACL rules from both base and sub model. I'm wondering if we should introduce a way to generalize the idea of `ranking by inheritance hierarchy`. Really, our inheritance is much closer to mixin based composition.

lib/utils.js
raymondfeng: I'm more thinking toward to have a way to express `delete` behaviors against settings from the base model.

bajtos:
> the merge function now comes with an optionated spec matching the (up-to-my-knowledge) model JSON definition spec:
``` js
const mergePolicy = {
    description: {replace: true}, // string or array
    // ...
    scopes: {innerReplace: true}, // object
```

The proposed behaviour seems reasonable to me. I think it's not possible to foresee all possible implications/breaking changes of this configuration, so let's start with what you have and iterate on top of that using user feedback.

> `acls: {replace: false}, // array`

I'll leave this up to you and @raymondfeng, my knowledge of ACL config is limited.

> any top-level property outside the model JSON spec is handled with the default replace policy
```js
    // any matching non array/object property is replaced
    // any matching custom top-level object/array is replaced
    default: {replace: true},
```

My only concern is whether this behaviour is backwards-compatible with the merging algorithm we have right now? It's a common use case to define custom model settings (top-level properties), e.g. in loopback we are adding `User.settings.ttl` and `User.settings.maxTTL`. These two are scalars therefore not affected by this merging algorithm. However, there may be  components/applications that use object/array custom settings and depend on the current  behaviour of our inheritance merging for these custom properties.

> I'm not quite OK with the current location of this spec within an util.js file. It should probably brought back directly on model-builder.js. Any comments/suggestions?

I think that ideally, there should be a generic `merge` implementation accepting two object values (parent, child settings) and a merge-config object, this function should live in `utils` or perhaps in a standalone file. Then the model-specific configuration of the merge algorithm can be defined in model-builder, I think that's the logical place for it.

ebarault:
Thanks @bajtos for taking the time to review and notably bring light on the following architecture concern:
> I think that ideally, there should be a generic merge implementation accepting two object values (parent, child settings) and a merge-config object, this function should live in utils or perhaps in a standalone file. Then the model-specific configuration of the merge algorithm can be defined in model-builder, I think that's the logical place for it.

In the meantime i finally got sick of tuning the `util.mergeSettings()` function to the new requirements (mainly because of the recursion calls that prevent to isolate top-level properties in a clean way), and after starting to pass a merge-config object as you propose, i decided to leave the original `util.mergeSettings()` untouched (which is actually a straight forward deep merge) and bring back the new handling logic right into `ModelClass.extend()` as it's already done for merging of properties. I can already push this for you to have a look, i am in the process of fixing tests that now needs to be performed on `ModelClass.extend()` rather than directly on `util.mergeSettings()`

lib/model-builder.js
```diff
@@ -388,9 +411,58 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
+          });
+          return;
+        }
+
+        // in case no merge policy matched, apply a deep merge
+        mergedSettings[key] = mergeSettings(settings[key], subclassSettings[key]);
+      });
+      // assign final sub class settings
+      subclassSettings = mergedSettings;
+    }
```
ebarault: of course this can be moved back to a separate function or separate file later on in this PR.

👉  the main point at this stage is that it leaves the original mergeSettings function untouched, and relies on it for legitimate deep merge operations

lib/model-builder.js
```diff
@@ -161,6 +162,7 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
       assert(ModelBaseClass, 'Base model is not found: ' + baseClass);
     }
   }
+  ModelBaseClass.rank = ModelBaseClass.rank || 0;
```
ebarault: adding a new property `rank` right on model instance to reflect the model inheritance rank.
This is not finalized yet. I propose to start at rank `0` if base model defaults to model.js, but rank `1` if a different base model is passed with `options.base` or param `parent` (one extra call to `.extend()`)

lib/utils.js
```diff
@@ -423,6 +425,19 @@ function mergeSettings(target, src) {
...
+  if (!Array.isArray(array) || !Number.isFinite(rank))
+    return array;
+
+  array.forEach(function(el) {
+    el.rank = rank;
+  });
+  return array;
+}
+
+/**
```
ebarault: new helper to add `rank` property to all array elements. Needs to be tinkered to check if elements are objects

ebarault:
you can check the interim commit [here](https://github.com/strongloop/loopback-datasource-juggler/pull/1289/commits/200673a42a93b944c8ab421d94afb6e826958767)

**description:** 
- restored original util.mergeSettings() helper (deep merge)
- moved merge policy spec to modelbuilder
- moved settings merging logic to modelbuilder
- created a new test file for modelbuilder class
  . migrated a couple of tests from modeldefinition test file
  . started new modelbuilder inheritance unit tests

test/model-definition.test.js
```diff
@@ -257,43 +257,6 @@ describe('ModelDefinition class', function() {
...
-
-  it('should ignore inherited options.super', function() {
-    var modelBuilder = memory.modelBuilder;
-    var base = modelBuilder.define('base');
-    var child = base.extend('child', {}, {super: 'base'});
-    var grandChild = child.extend('grand-child');
-    assert.equal('child', grandChild.base.modelName);
-    assert(grandChild.prototype instanceof child);
-  });
-
```
ebarault: these tests are not related to modelDefinition class

ebarault:
@raymondfeng 
> I'm thinking toward to have a way to express delete behaviors against settings from the base model.

i added a `__delete` setting in the merge policy specification to control which value allows deleting a setting. does it match your expectations?

> I feel ACL entries from DB have higher ranks than static ones from code/json. Maybe we can define an option flag for that. WDYT?

I agree with your proposal. it will greatly simplify the implementation.

👉  I added several commits, could you PTAL?
**changes**
- restored original util.mergeSettings() helper (deep merge)
- moved merge policy spec to modelbuilder
- moved settings merging logic to modelbuilder
- added a `__rank` property to `ModelClass`
- created a new test file for modelbuilder class
  - migrated a couple of tests from modeldefinition test file
  - added more model merge/inheritance unit tests
- added a model level feature flag to control the new model merge behavior
- added an option to control which value controls the deletion of a setting

lib/model-builder.js
```diff
@@ -17,7 +17,8 @@ var deprecated = require('depd')('loopback-datasource-juggler');
 var DefaultModelBaseClass = require('./model.js');
 var List = require('./list.js');
 var ModelDefinition = require('./model-definition.js');
-var mergeSettings = require('./utils').mergeSettings;
+var deepMerge = require('./utils').deepMerge;
```
ebarault: `mergeSettings` is not a good term, as
- it is also used for properties
- it is now misleading in the context of the new merge policy

I propose to use its alias `deepMerge` (already defined in utils.js)

lib/model-builder.js
```diff
@@ -348,13 +387,14 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
+   * @options {Object} subclassProperties child model properties, added to base model
+   *   properties.
+   * @options {Object} subclassSettings child model settings such as relations and acls,
+   *   merged with base model settings.
    */
   ModelClass.extend = function(className, subclassProperties, subclassSettings) {
-    var properties = ModelClass.definition.properties;
-    var settings = ModelClass.definition.settings;
+    var baseclassProperties = ModelClass.definition.properties;
+    var baseclassSettings = ModelClass.definition.settings;
```
ebarault: prefixing properties and settings with `baseclass` to improve code readability

lib/model-builder.js
```diff
@@ -369,28 +409,81 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
         // don't inherit id properties
         continue;
       }
       if (subclassProperties[key] === undefined) {
-        var baseProp = properties[key];
+        var baseProp = baseclassProperties[key];
         var basePropCopy = baseProp;
         if (baseProp && typeof baseProp === 'object') {
-          // Deep clone the base prop
-          basePropCopy = mergeSettings(null, baseProp);
```
ebarault: `mergeSettings(baseProp)` is fully equivalent to` mergeSettings(null, baseProp)` and operates a deep clone.
`mergeSettings(baseProp)` is more readable

lib/utils.js
```diff
@@ -367,52 +368,56 @@ function parseSettings(urlStr) {
...
+ * NOTE: The function operates as a deep clone when called with a single object
+* argument.
+ *
+ * @param {Object} base The base object
+ * @param {Object} obj The object to merge with base object
+ * @returns {Object} The merged object
  */
-function mergeSettings(target, src) {
-  var array = Array.isArray(src);
+function deepMerge(base, obj) {
```
ebarault: - renaming mergeSettings to deepMerge to actually reflect the function's purpose in a generic way
- renaming `target` to `base` and `src` to `obj` to improve code readability

ebarault:
@raymondfeng, @jannyHou, @superkhau, @ssh24 ping, could you PTAL so i can move on with this one?

lib/model-builder.js
```diff
@@ -168,6 +169,9 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
     return ModelBaseClass.extend(className, properties, settings);
   }
 
+  // initializing base model inheritance rank if not set already
+  ModelBaseClass.__rank = ModelBaseClass.__rank || 1;
```
bajtos: It makes me wonder: is is possible to compute this property dynamically by crawling the inheritance chain? The following code seems to work well for me (Node.js versions 4 and 6):

```js
function computeInheritanceDepth(ctor, rootCtor) {
  if (!rootCtor) rootCtor = Object;
  let depth = 0;
  while (ctor && ctor !== rootCtor) {
    depth++;
    ctor = ctor.prototype &&
      ctor.prototype.__proto__ &&
      ctor.prototype.__proto__.constructor;
  }
  return depth;
}

// usage:
const rank = computeInheritanceDepth(ModelClass, DefaultModelBaseClass)
// => 0 for ModelBaseClass
// => 1 for User based on ModelBaseClass
// => 2 for Customer based on User
// etc.
```

lib/model-builder.js
```diff
@@ -332,6 +339,38 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
+      __default: {replace: true},
+    };
+
+    // make sure we take settings of current model with `this`
+    // model might have no settings in case it is an instance of builtin ModelBaseClass
+    return (this.settings || {}).mergePolicy || mergePolicy;
+  };
+
+  // ModelClass.base might not have getMergePolicy() if it is an instance of builtin ModelBaseClass
+  ModelClass.base.getMergePolicy = ModelClass.base.getMergePolicy || ModelClass.getMergePolicy;
```
bajtos: If I am reading the code correctly, then assigning to `ModelClass.base.getMergePolicy` is changing the merge policy of the parent (base) model, not the model we are creating right now. In other words, when creating `Customer` model extending the `User` model, the merge policy of the `User` mode is modifying. I find that confusing.

Is this intentional? Perhaps you wanted to assign to `ModelClass.getMergePolicy` instead?

lib/model-builder.js
bajtos: I like the idea of prefixing to make the intent more clear. Can we use `baseClass` case please? Or perhaps just `base` or `parent`, e.g. `baseProperties` or `parentProperties`. Or maybe `inherited`?

lib/model-builder.js
```diff
@@ -369,28 +409,81 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
         subclassProperties[key] = basePropCopy;
       }
     }
 
-    // Merge the settings
+    // Merging the settings
     var originalSubclassSettings = subclassSettings;
-    subclassSettings = mergeSettings(settings, subclassSettings);
+
+    if (!subclassSettings.enableOptionatedModelMerge) {
```
bajtos: `enableOptionatedModelMerge` - looks like a typo to me, did you mean "opinionated"? I find the flag name too long, IMO "enable" is redundant.

```js
if (!subclassSettings.opinionatedModelMerge) {
// ...
```

lib/model-builder.js
```diff
@@ -369,28 +409,81 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
+      // legacy behaviour
+      subclassSettings = deepMerge(baseclassSettings, subclassSettings);
+    } else {
+      // optionated model merge behaviour if option is set
```
bajtos: I find this block of code quite long, could you please extract it into a new function?

lib/utils.js
bajtos: `target -> base` make sense to me. I find the name `obj` too generic, I had to read jsdoc to understand what's that argument. Can we use a different name, e.g. something like `extensions` or `extra`?

lib/utils.js
```diff
@@ -423,6 +428,24 @@ function mergeSettings(target, src) {
...
+  if (!Array.isArray(array) || !Number.isFinite(rank))
+    return array;
+
+  array.forEach(function(el) {
+    // only apply ranking on objects {} in array
+    if (!el || typeof el != 'object' || Array.isArray(el))
+      return;
+
+    // initialize element rank if not set already
+    el.__rank = el.__rank || rank;
```
bajtos: Should we define this property as non-enumerable and read-only?

test/model-builder.test.js
```diff
@@ -0,0 +1,373 @@
...
+        relations: {
+          children: {
+            type: 'hasMany',
+            model: 'anotherChild',
+          },
+        },
+      });
+      var baseChild = modelBuilder.define('baseChild');
+      baseChild.attachTo(memory);
+    // the name of this must begin with a letter < b
```
bajtos: the indentation is off here

test/model-builder.test.js
```diff
@@ -0,0 +1,373 @@
...
+                __rank: 4,
+              },
+            ],
+          };
+          should.deepEqual(childRank3.settings.acls, expectedSettings.acls);
+        });
+
+        it('`{replace: true}` replaces base model array with sub model matching ' +
+        'array', function() {
+        // merge policy of settings.description is {replace: true}
```
bajtos: the indentation is off here

test/model-builder.test.js
```diff
@@ -0,0 +1,373 @@
...
+          var expectedSettings = {
+            unknownObjectParam: {anotherKey: 'this should remain after merge'},
+          };
+
+          should.deepEqual(child.settings.description, expectedSettings.description);
+        });
+
+        it('`{replace: false}` adds distinct members of matching arrays from ' +
+        'base model and sub model', function() {
+        // merge policy of settings.hidden is {replace: false}
```
bajtos: the indentation is off here

test/model-builder.test.js
```diff
@@ -0,0 +1,373 @@
...
+          var expectedSettings = {
+            hidden: ['firstProperty', 'secondProperty', 'thirdProperty'],
+          };
+
+          should.deepEqual(child.settings.hidden, expectedSettings.hidden);
+        });
+
+        it('`{patch: true}` adds distinct inner properties of matching objects ' +
+        'from base model and sub model', function() {
+        // merge policy of settings.relations is {patch: true}
```
bajtos: the indentation is off here

test/model-builder.test.js
```diff
@@ -0,0 +1,373 @@
...
+              },
+            },
+          };
+
+          should.deepEqual(child.settings.relations, expectedSettings.relations);
+        });
+
+        it('`{patch: true}` replaces baseClass inner properties with matching ' +
+        'subClass inner properties', function() {
+        // merge policy of settings.relations is {patch: true}
```
bajtos: the indentation is off here

test/util.test.js
```diff
@@ -138,7 +138,7 @@ describe('util.parseSettings', function() {
   });
 });
 
-describe('mergeSettings', function() {
+describe('util.mergeSettings', function() {
```
bajtos: `util.deepMerge`?

lib/model-builder.js
raymondfeng: Recalculating ranks is not a bad idea. We probably will need to have `mixins` which don't have an explicit chain.

lib/model-builder.js
raymondfeng: Agree with @bajtos. @ebarault Please clarify.

lib/utils.js
raymondfeng: Ideally yes. `Object.defineProperty` is pretty expensive though. If it's only called during bootstrapping, we can afford that.

lib/model-builder.js
ebarault: ModelClass.base.getMergePolicy is assigned only if not set already. ModelClass.getMergePolicy is assigned by .extend().
 
Assigning on ModelClass.base instead of ModelClass allows to have getMergePolicy also on ModelBaseClass (model.js) when defining a class with this base class (which is default and doesn't not call .extend() internally). 
And thus extending .getMergePolicy() is possible on top of base.getMergePolicy() (see the related test).

lib/model-builder.js
ebarault: Optionated : with options
I will change however

lib/model-builder.js
ebarault: https://github.com/strongloop/loopback-datasource-juggler/pull/1289/files#diff-869306f3e081589c5746ae11e6ac21a9R114

lib/model-builder.js
kjdelisle: +1 for `baseClass` instead of `baseclass`

lib/utils.js
kjdelisle: We should avoid `obj` at all costs; I would only ever use that name in a single parameter function where the input types are numerous, but the purpose is singular (since that doesn't sacrifice clarity).

lib/model-builder.js
kjdelisle: Let's say we had three models, a parent `A` and two children `B` and `C`.
With this line, if `A` had no instance of `.getMergePolicy`, then whichever of the children has its `.getMergePolicy` called first will alter `A` to share the same policy. If `B` and `C` each have different merge policies, would this not have unintended side-effects for one of the two children?

lib/utils.js
kjdelisle: If it only happened during bootstrapping, then we wouldn't have any guarantees about rank's immutability on models defined at runtime, correct?

lib/model-builder.js
kjdelisle: TBH, if we're going with a flag, I would recommend something like `customModelMerge` or `overrideModelMerge` to imply that we as the user are modifying the behaviour.

lib/utils.js
ebarault: actually I did applied @bajtos 's proposal, it's safer to prevent people interacting with this new internal property as it will be used in ACL scoring at runtime

lib/model-builder.js
ebarault: for now i changed for `configurableModelMerge`

lib/model-builder.js
ebarault: I went for baseClass and subClass prefixes

lib/utils.js
ebarault: yeah... i wasn't much inspired that day... I changed for `extras`

lib/model-builder.js
ebarault: I agree with @raymondfeng, I prefer calculating the model rank at bootstrapping time

ebarault:
@kjdelisle @bajtos @raymondfeng thanks for reviewing. I've added one more commit that addresses most of your comments, as well as adds missing tests for the new utils functions `rankArrayElements()`.

I'm waiting for more comments on the thread [here](https://github.com/strongloop/loopback-datasource-juggler/pull/1289#discussion_r114820143) before moving on with the assignment of `.getMergePolicy()` on `ModelClass` or `ModelClass.base`.

---

As for using a feature flag or not: @kjdelisle I fully agree that it's dismiss the original purpose of this improvement to require a feature flag config for users to benefit from it. The thing is that:
the PR provides at the same time bug fixes and improvements that change the original behavior under certain conditions. And some existing app where users have found ways to make it work despite of the bugs, may react in unexpected ways (I suppose) with this new implementation as the default.

> having a default mergeSettings implementation that provides the current behaviour

could you specify whether **current** refers to the current implementation in that PR, or the currently existing implementation (simple deep merge) ?

**My point of view**: I'd go without a feature flag and with the merge policy as proposed in this PR. Letting people fallback to the former merge policy by overriding `.getMergePolicy()` (with the correct explanations provided in the docs).

What do you think?

lib/model-builder.js
ebarault: Given my knowledge and practice of the inheritance behavior defined in both `.define` and `.extend`, I can't come up with such a situation where `A` is not fully loaded with `.getMergePolicy` assigned when `B` loads with `A` as base.

Would you provide a test case you have in mind that would fail with the proposed implementation?

bajtos: > how do we make sure `base` always have `.getMergePolicy` defined?

I think it should be enough to define `ModelBaseClass.getMergePolicy()` in `lib/model.js`, since all models should be always inheriting from `ModelBaseClass` (or a child model that has already transitively inherited from `ModelBaseClass`).

Alternatively, if there is no easy way for achieving that, then I'd rather fix the code that's currently requiring an existing `.getMergePolicy` method to support the case when such method is not defined.

Regarding the example with `A`, `B` and `C`: I think the problem is that in the current implementation, if `B` and `C` have different merge policies, then the merge policy used by `base.getMergePolicy` will depend on the order in which `B` and `C` are defined. Assuming that models are defined in lexicographical order, when you rename model `B` to model `D`, suddenly the default merge policy will change from what model `B`/`D` provides to what model `C` provides.

What if we changed this line to `assert` that `base.getMergePolicy` is already defined? If all existing tests are passing, then we should be good. If there are some tests that start failing, then that should give us more insights into possible inheritance hierarchies.

```diff
- ModelClass.base.getMergePolicy = ModelClass.base.getMergePolicy || ModelClass.getMergePolicy;
+ assert(ModelClass.base.getMergePolicy, `The base class ${ModelClass.modelName}` does not provide getMergePolicy(). Most likely it is not inheriting from juggler's DefaultModelBaseClass, which is an incorrect usage of the framework.`);
```

bajtos:
@kjdelisle are you sure the new merging behaviour won't break any existing applications out there? From what I've seen so far, some of our users go pretty far in relying on undocumented quirks of the framework and finding creative ways how to address LoopBack shortcomings. I am concerned that changing the default merge behaviour will break some applications, in which case such change must be released either as semver-major (juggler@4, loopback@4), or hidden behind a feature flag (as I proposed here).

kjdelisle:
@bajtos Is there no default merge policy that can sufficiently emulate the original behaviour? If we can disable the new functionality with a flag, surely we can provide the same "default" behavior with a set of merge defaults that match current expectations.

ebarault:
@bajtos :
> I think it should be enough to define ModelBaseClass.getMergePolicy() in lib/model.js

👍  I was trying to avoid that as I initially believed that `this.settings` would then return the base model settings and not the current model settings, but after reviewing the code it does reference the current model settings (ModelClass, as returned by model-builder). So yes, of course this is the way to go.

> What if we changed this line to assert that base.getMergePolicy is already defined?

👍  Again, i was pushing too hard here... I wanted to handle the case where a full custom model would be provided by the user, but this makes more sense to prevent this case, given the insanity of the required level of customization.

@kjdelisle 
> Is there no default merge policy that can sufficiently emulate the original behaviour? If we can disable the new functionality with a flag, surely we can provide the same "default" behavior with a set of merge defaults that match current expectations.

The original behaviour is fully emulated by having `getMergePolicy()` returning `{}`, this will apply a deep merge on each top-level property the settings object.

Still i am not 100% sure of what you are suggesting:
- do you propose that `getMergePolicy()` returns `{}`by default and the user can customize it to reach the behavior from this PR? Then I prefer the feature flag as it allows to settle in the code the expected new merge policy and does not require the user to copy-paste it from the doc
- or do you propose that `getMergePolicy()` returns the new policy defined in this PR by default and we specify the config for the original behavior in the doc? Then some apps could crash as mentioned by Miroslav.

Surely we can go with a feature flag and clean this when landing LB4/DSJ4

kjdelisle:
@ebarault I was proposing the first option. It's mostly a personal preference to avoid config flags if possible, but there's already precedent for it in the project and I won't hold the PR hostage over it. :P

ebarault:
@kjdelisle : actually now that I fully get your point, I believe it's a better approach since there are already some fixes and improvement that can be provided by default without compromising any existing apps. So I transitioned in my latest commit.

Now instead of having one flag `configurableModelMerge` and one setting object `mergePolicy`, we have a **single** `configureModelMerge` object with the following behavior:
-  **no setting** object provided: emulates the legacy (current) deep merge, plus performs the ACLs ranking and one fix on the `description` setting when it is an array (minor).
- `configureModelMerge: true` : leverages the recommended (this PR) merge policy
- `configureModelMerge: {/*..*/}`: leverages the merge policy provided as an object

@bajtos : I know the tests coverage on the existing deep merge is far from perfect. But the new method mostly relies on utils.deepMerge() when `configureModelMerge` is not set. Plus all the existing tests pass, incl. a lot that leverage model inheritance.

lib/model-builder.js
```diff
@@ -404,15 +405,8 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
-    if (!subClassSettings.configurableModelMerge) {
-      // legacy behaviour applying a deep merge of parent and child model settings
-      subClassSettings = deepMerge(baseClassSettings, subClassSettings);
-    } else {
-      // configurable model merge behaviour if option is set
-      let mergePolicy = ModelClass.getMergePolicy();
-      subClassSettings = mergeSettings(baseClassSettings, subClassSettings, mergePolicy);
-    }
+    let mergePolicy = ModelClass.getMergePolicy(subClassSettings);
+    subClassSettings = mergeSettings(baseClassSettings, subClassSettings, mergePolicy);
```
ebarault: no more feature flag. The merge policy returned by `getMergePolicy()` is not influenced by the provided sub class setting `configureModelMerge`

lib/model-builder.js
```diff
@@ -482,7 +477,7 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
 
       // in case no merge policy matched, apply a deep merge
       // this for example handles {replace: false} and {rank: true}
-      mergedSettings[key] = deepMerge(baseClassSettings[key], subClassSettings[key]);
+      mergedSettings[key] = deepMergeProperty(baseClassSettings[key], subClassSettings[key]);
```
ebarault: I had to introduce this new helper which is a safer version of deepMerge to work with object properties of any type.
`utils.deepMerge` actually turns a provided extra to `{}` if not an Array, which is fine when passing an object {} as extras, but is not wanted when passing a simple type as a string.

lib/model.js
```diff
@@ -741,43 +758,69 @@ ModelBaseClass.prototype.setStrict = function(strict) {
...
-    scope: {replace: true}, // object
-    scopes: {patch: true}, // object
     acls: {rank: true}, // array
-    // this setting controls which child model property's value allows deleting
-    // a base model's property
-    __delete: null,
-    // this setting controls the default merge behaviour for settings not defined
-    // in the mergePolicy specification
-    __default: {replace: true},
   };
```
ebarault: legacy merge policy + fix for `description` arrays + ACLs ranking

lib/model.js
```diff
@@ -741,43 +758,69 @@ ModelBaseClass.prototype.setStrict = function(strict) {
...
 
-  // make sure we take settings of current model with `this`
-  // current model might not have settings in case it is built-in ModelBaseClass
-  return (this.settings || {}).mergePolicy || mergePolicy;
+  let config = (options || {}).configureModelMerge;
```
ebarault: fetch `configureModelMerge` options from sub model settings

lib/utils.js
```diff
@@ -430,6 +431,20 @@ function deepMerge(base, extras) {
...
+ * @param {Object} base The base property
+ * @param {Object} extras The property to merge with base
+ * @returns {Object} The merged property
+ */
+function deepMergeProperty(base, extras) {
+  let mergedObject = deepMerge({key: base}, {key: extras});
+  let mergedProperty = mergedObject.key;
+  return mergedProperty;
+}
+
```
ebarault: what the method does is to wrap base and extras in an object to make sure deepMerge operates as expected when running on simple types

test/model-builder.test.js
```diff
@@ -3,65 +3,75 @@
...
+      const grandChild = child.extend('grand-child');
       assert.equal('child', grandChild.base.modelName);
       assert(grandChild.prototype instanceof child);
     });
 
-    describe('merge policy with flag `configurableModelMerge`', function() {
+    describe('merge policy WITH flag `configureModelMerge: true`', function() {
       describe('ModelBaseClass.getMergePolicy()', function() {
-        const defaultMergePolicy = {
+        const legacyMergePolicy = {
```
ebarault: I will actually move the `ModelBaseClass.getMergePolicy()` block of test upper in the tests nesting levels as it is no more specific to a flag, but i kept this as is to ease your diff reading

I had to adapt a bit the tests when removing the feature flag

ebarault:
@slnode test please

ebarault:
@kjdelisle @raymondfeng @superkhau as Miroslav is OOO right now, could you take a final look to this PR so we can land it? Thank you.

kjdelisle:
Love the idea of the no-flag default mixed with the default merge policy for the opt-in. Best of both approaches!

kjdelisle:
@slnode test please

ebarault:
Thank you all for reviewing! 💯  x 👏 

I have a final edit+squash to push before we can land. 

Could anyone check for me the issue with the CI? I do not have access to CI logs.

superkhau:
Travis

```
/home/travis/build/strongloop/loopback-datasource-juggler/lib/model.js:782

  let mergePolicy = {

  ^^^

SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```

ebarault:
could one of you check why one of CI's downstream is failing ? Thanks

`[cis-jenkins] downstream: loopback-connector-cloudant@master — Failed!`

ssh24:
@slnode test please

ebarault:
@slnode test please

kjdelisle:
@slnode test please

kjdelisle:
The last failure in Cloudant is a known issue; we're currently working on our setup for that connector. Merging!

@ebarault Thank you so much for all your effort, this PR is awesome. :D

ebarault:
🎉 Excellent 🎉, at last! 
Thanks all for the fruitful discussion! 💯  x 🙇

bajtos:
🎉 Woot 🎉 

Thank you @ebarault for this awesome contribution!

@kjdelisle @raymondfeng @superkhau @ssh24 thanks for keeping the discussion going even while I was away 🙇

PR #1299, opened 2017-03-31
Revert "Fix #1079 - Polymorphic hasMany inverse relation in include filter"

### LINKED ISSUE #1079: Polymorphic hasMany inverse relation doesn't work on include filter

jannyHou:
Reverts strongloop/loopback-datasource-juggler#1080
It should not be checked into 2.x-latest, but 2.x

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -325,14 +325,10 @@ Inclusion.include = function(objects, include, options, cb) {
         inq: uniq(sourceIds),
       };
       if (polymorphic) {
-        var throughModel = polymorphic.invert ?
-          relation.modelTo :
-          relation.modelFrom;
-
         //handle polymorphic hasMany (reverse) in which case we need to filter
         //by discriminator to filter other types
         throughFilter.where[polymorphic.discriminator] =
-            throughModel.definition.name;
+          relation.modelFrom.definition.name;
       }
       /**
        * 1st DB Call of 2 step process. Get through model objects first
@@ -931,3 +927,4 @@ Inclusion.include = function(objects, include, options, cb) {
     }
   }
 };
+
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -5503,32 +5503,4 @@ describe('relations', function() {
       }).should.throw('Invalid relation name: trigger');
     });
   });
-
-  describe('polymorphic hasMany', function() {
-    before(function(done) {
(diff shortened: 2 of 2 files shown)


ssh24:
@jannyHou Is this PR still needed to be reverted?

jannyHou:
@ssh24 yes it's checked in to a wrong branch

jannyHou:
merged, thank you @ssh24 :)

PR #1300, opened 2017-04-01
Fix in-mem connector file operation racing condition

raymondfeng:
### Description

Fix race condition for memory connector file read/write 

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -13,6 +13,7 @@ var geo = require('../geo');
 var utils = require('../utils');
 var fs = require('fs');
 var async = require('async');
+var debug = require('debug')('loopback:connector:memory');
 
 /**
  * Initialize the Memory connector against the given data source
@@ -104,44 +105,91 @@ Memory.prototype.collectionSeq = function(model, val) {
   return this.ids[model];
 };
 
-Memory.prototype.loadFromFile = function(callback) {
+/**
+ * Create a queue to serialize file read/write operations
+ * @returns {*} The file operation queue
+ */
+Memory.prototype.setupFileQueue = function() {
   var self = this;
+  if (!this.fileQueue) {
+    // Create a queue for writes
+    this.fileQueue = async.queue(function(task, done) {
+      var callback = task.callback || function() {};
+      var file = self.settings.file;
+      if (task.operation === 'write') {
+        // Flush out the models/ids
+        var data = JSON.stringify({
+          ids: self.ids,
+          models: self.cache,
+        }, null, '  ');
+        debug('Writing cache to %s: %s', file, data);
+        fs.writeFile(file, data, function(err) {
+          debug('Cache has been written to %s', file);
+          done(err);
+          callback(err, task.data);
+        });
+      } else if (task.operation === 'read') {
+        debug('Reading cache from %s: %s', file, data);
+        fs.readFile(file, {
+          encoding: 'utf8',
+          flag: 'r',
+        }, function(err, data) {
+          if (err && err.code !== 'ENOENT') {
+            done(err);
+            callback(err);
+          } else {
+            debug('Cache has been read from %s: %s', file, data);
+            self.parseAndLoad(data, function(err) {
+              done(err);
+              callback(err);
+            });
+          }
+        });
+      } else {
+        var err = new Error('Unknown type of task');
+        done(err);
+        callback(err);
+      }
+    }, 1);
+  }
+  return this.fileQueue;
+};
+
+Memory.prototype.parseAndLoad = function(data, callback) {
+  if (data) {
+    try {
+      data = JSON.parse(data.toString());
+    } catch (e) {
+      return callback && callback(e);
+    }
+
+    this.ids = data.ids || {};
+    this.cache = data.models || {};
+  } else {
+    if (!this.cache) {
+      this.ids = {};
+      this.cache = {};
+    }
+  }
+  callback && callback();
+};
+
+Memory.prototype.loadFromFile = function(callback) {
   var hasLocalStorage = typeof window !== 'undefined' && window.localStorage;
   var localStorage = hasLocalStorage && this.settings.localStorage;
 
-  if (self.settings.file) {
-    fs.readFile(self.settings.file, {encoding: 'utf8', flag: 'r'}, function(err, data) {
-      if (err && err.code !== 'ENOENT') {
-        callback && callback(err);
-      } else {
-        parseAndLoad(data);
-      }
+  if (this.settings.file) {
+    debug('Queueing read %s', this.settings.file);
+    this.setupFileQueue().push({
+      operation: 'read',
+      callback: callback,
     });
   } else if (localStorage) {
     var data = window.localStorage.getItem(localStorage);
     data = data || '{}';
-    parseAndLoad(data);
+    this.parseAndLoad(data, callback);
   } else {
     process.nextTick(callback);
-  }
-
-  function parseAndLoad(data) {
-    if (data) {
-      try {
-        data = JSON.parse(data.toString());
-      } catch (e) {
-        return callback(e);
-      }
-
-      self.ids = data.ids || {};
-      self.cache = data.models || {};
-    } else {
-      if (!self.cache) {
-        self.ids = {};
-        self.cache = {};
-      }
-    }
-    callback && callback();
   }
 };
 
@@ -150,36 +198,22 @@ Memory.prototype.loadFromFile = function(callback) {
  * @param {Function} callback
  */
 Memory.prototype.saveToFile = function(result, callback) {
-  var self = this;
   var file = this.settings.file;
   var hasLocalStorage = typeof window !== 'undefined' && window.localStorage;
   var localStorage = hasLocalStorage && this.settings.localStorage;
   if (file) {
-    if (!self.writeQueue) {
-      // Create a queue for writes
-      self.writeQueue = async.queue(function(task, cb) {
-        // Flush out the models/ids
-        var data = JSON.stringify({
-          ids: self.ids,
-          models: self.cache,
-        }, null, '  ');
-
-        fs.writeFile(self.settings.file, data, function(err) {
-          cb(err);
-          task.callback && task.callback(err, task.data);
-        });
-      }, 1);
-    }
+    debug('Queueing write %s', this.settings.file);
     // Enqueue the write
-    self.writeQueue.push({
+    this.setupFileQueue().push({
+      operation: 'write',
       data: result,
       callback: callback,
     });
   } else if (localStorage) {
     // Flush out the models/ids
     var data = JSON.stringify({
-      ids: self.ids,
-      models: self.cache,
+      ids: this.ids,
+      models: this.cache,
     }, null, '  ');
     window.localStorage.setItem(localStorage, data);
     process.nextTick(function() {
@@ -506,7 +540,7 @@ function applyFilter(filter) {
       // Support referencesMany and other embedded relations
       // Also support array types. Mongo, possibly PostgreSQL
       if (Array.isArray(value)) {
-        var matcher  = where[key];
+        var matcher = where[key];
         // The following condition is for the case where we are querying with
         // a neq filter, and when the value is an empty array ([]).
         if (matcher.neq !== undefined && value.length <= 0) {
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -365,36 +365,59 @@ DataSource.prototype.setup = function(name, settings) {
     }
   }
 
-  dataSource.connect = function(cb) {
-    var dataSource = this;
-    if (dataSource.connected || dataSource.connecting) {
-      process.nextTick(function() {
-        cb && cb();
-      });
-      return;
+  dataSource.connect = function(callback) {
+    callback = callback || utils.createPromiseCallback();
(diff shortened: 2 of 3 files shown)


lib/connectors/memory.js
```diff
@@ -54,10 +55,38 @@ Memory.prototype.getTypes = function() {
...
+      if (!err) self.connected = true;
+      self.connecting = false;
+      // Invoke all pending callbacks
+      async.each(self.pendingConnectCallbacks, function(cb, done) {
+        process.nextTick(function() {
+          cb(err);
+          done();
+        });
+      }, function(err) {
+        self.pendingConnectCallbacks = [];
```
bajtos: What if somebody pushed into pendingConnectCallbacks while we were processing the queue? I think the following approach should be safer:

```js
self.connecting = false;
const pendingCallbacks = self.pendingConnectCallbacks;
self.pendingConnectCallbacks = [];

async.each(pendingCallbacks, ...);
```

lib/connectors/memory.js
```diff
@@ -54,10 +55,38 @@ Memory.prototype.getTypes = function() {
...
-    this.loadFromFile(callback);
+    this.loadFromFile(function(err) {
+      if (!err) self.connected = true;
+      self.connecting = false;
+      // Invoke all pending callbacks
+      async.each(self.pendingConnectCallbacks, function(cb, done) {
+        process.nextTick(function() {
+          cb(err);
+          done();
+        });
```
bajtos: Isn't `async.each` already inserting `process.nextTick` calls? Would it make sense to do the following?

```js
function(cb, done) {
  process.nextTick(cb, err);
  done();
}
```

lib/connectors/memory.js
```diff
@@ -104,45 +133,93 @@ Memory.prototype.collectionSeq = function(model, val) {
...
+        var data = JSON.stringify({
+          ids: self.ids,
+          models: self.cache,
+        }, null, '  ');
+        debug('Writing cache to %s: %s', file, data);
+        fs.writeFile(file, data, function(err) {
+          debug('Cache has been written to %s', file);
+          cb(err);
+          callback && callback(err, task.data);
+        });
```
bajtos: Please extract the read and write branches into two methods to keep this async.queue callback shorter.

lib/connectors/memory.js
```diff
@@ -237,7 +301,8 @@ Memory.prototype.create = function create(model, data, options, callback) {
       return process.nextTick(function() {
         callback(err);
       });
-    };
+    }
+    ;
```
bajtos: This looks like a typo, I think you meant `};` to be on a single line.

lib/connectors/memory.js
```diff
@@ -259,34 +324,36 @@ Memory.prototype.updateOrCreate = function(model, data, options, callback) {
...
+    var nodes = self._findAllSkippingIncludes(model, filter);
+    if (nodes.length === 0) {
+      return self._createSync(model, data, function(err, id) {
+        if (err) return process.nextTick(function() {
+          callback(err);
+        });
+        self.saveToFile(id, function(err, id) {
+          self.setIdValue(model, data, id);
+          callback(err, self.fromDb(model, data), {isNewInstance: true});
+        });
```
bajtos: This diff is too large because of whitespace changes. Are there any significant changes here? Can you revert the whitespace changes to make the diff easier to understand please?

lib/connectors/memory.js
```diff
@@ -259,34 +324,36 @@ Memory.prototype.updateOrCreate = function(model, data, options, callback) {
...
+        callback(err, data, {isNewInstance: false});
+      });
+    } else {
+      process.nextTick(function() {
+        var error = new Error('There are multiple instances found.' +
+          'Upsert Operation will not be performed!');
+        error.statusCode = 400;
+        callback(error);
+      });
+    }
```
bajtos: Ditto - it's hard to tell what has changed because of whitespace noise.

lib/connectors/memory.js
```diff
@@ -788,7 +855,9 @@ Memory.prototype.replaceById = function(model, id, data, options, cb) {
   var self = this;
   if (!id) {
     var err = new Error(g.f('You must provide an {{id}} when replacing!'));
-    return process.nextTick(function() { cb(err); });
+    return process.nextTick(function() {
```
bajtos: `process.nextTick(cb, err)` while you are at changing this (why is the change needed in the first place?)

lib/connectors/memory.js
```diff
@@ -799,7 +868,9 @@ Memory.prototype.replaceById = function(model, id, data, options, cb) {
     var msg = 'Could not replace. Object with id ' + id + ' does not exist!';
     var error = new Error(msg);
     error.statusCode = error.status = 404;
-    return process.nextTick(function() { cb(error); });
+    return process.nextTick(function() {
```
bajtos: `process.nextTick(cb, err)` while you are at changing this (why is the change needed in the first place?)

lib/connectors/memory.js
```diff
@@ -830,7 +901,9 @@ Memory.prototype.replaceOrCreate = function(model, data, options, callback) {
     // Calling _createSync to update the collection in a sync way and
     // to guarantee to create it in the same turn of even loop
     return self._createSync(model, data, function(err, id) {
-      if (err) return process.nextTick(function() { callback(err); });
+      if (err) return process.nextTick(function() {
```
bajtos: `process.nextTick(callback, err)` while you are at changing this (why is the change needed in the first place?)

lib/datasource.js
```diff
@@ -365,36 +365,60 @@ DataSource.prototype.setup = function(name, settings) {
...
   }
 
-  dataSource.connect = function(cb) {
-    var dataSource = this;
-    if (dataSource.connected || dataSource.connecting) {
-      process.nextTick(function() {
-        cb && cb();
-      });
+  dataSource.connect = function(callback) {
+    var self = this;
```
bajtos: AFAIK, `=>` functions are much more performant than `var self = this`. Since this code will run on Node.js 4.x and newer, we should be using arrow functions instead of `self` as much as we can.

lib/datasource.js
```diff
@@ -365,36 +365,60 @@ DataSource.prototype.setup = function(name, settings) {
...
+    if (this.connected) {
+      // It's already connected or being connected
+      // Do NOT run loadFromFile again as it will override the cache
```
bajtos: I find this comment misleading in the context of `dataSource.connect`.

lib/datasource.js
```diff
@@ -365,36 +365,60 @@ DataSource.prototype.setup = function(name, settings) {
...
+    if (typeof dataSource.connector.connect !== 'function') {
+      // Connector doesn't have the connect function
+      // Assume no connect is needed
       process.nextTick(function() {
-        dataSource.connected = true;
-        dataSource.connecting = false;
-        dataSource.emit('connected');
-        cb && cb();
+        self.connected = true;
+        self.connecting = false;
```
bajtos: This two flags should be set on the first tick. Otherwise `dataSource.connect` with a connection-less connector may fire the `connected` event twice.

```js
ds.on('connected', () => console.log('connected'));
ds.connect();
ds.connect();
```

lib/datasource.js
```diff
@@ -365,36 +365,60 @@ DataSource.prototype.setup = function(name, settings) {
...
+        self.emit('connected');
+        callback && callback();
```
bajtos: Instead of checking whether `callback` is all over the place, it's better to move this check to the top of the method.


```js
callback = callback || function() {};
```

We should consider returning a Promise when no callback was specified, but that's probably out of scope of this patch.

lib/datasource.js
```diff
@@ -365,36 +365,60 @@ DataSource.prototype.setup = function(name, settings) {
...
+      } else {
+        self.emit('error', err);
+      }
+      // Invoke all pending callbacks
+      async.each(cbs, function(cb, done) {
+        process.nextTick(function() {
+          try {
+            cb(err);
+          } catch (e) {
+            // Ignore error to make sure all callbacks are invoked
```
bajtos: This will silently swallow errors that may be important for troubleshooting. I am proposing the following implementation:

```js
async.each(cbs, (cb, next) => {
  process.nextTick(cb, err);
  next();
});
```

That way any errors thrown by the callback trigger "unhandled error" handling, which typically crashes the application (as expected for an unhandled error).

lib/datasource.js
```diff
@@ -365,36 +365,60 @@ DataSource.prototype.setup = function(name, settings) {
...
+          } finally {
+            done();
+          }
+        });
+      }, function(err) {
+        // Ignore
```
bajtos: Please log the error via `debug()` at least.

lib/connectors/memory.js
```diff
@@ -54,10 +55,38 @@ Memory.prototype.getTypes = function() {
...
+  var self = this;
+  if (this.connected) {
+    // It's already connected or being connected
+    // Do NOT run loadFromFile again as it will override the cache
+    if (callback) process.nextTick(callback);
+    return;
+  }
+  if (callback) {
+    // Queue the callback
+    this.pendingConnectCallbacks = this.pendingConnectCallbacks || [];
```
bajtos: Could you please elaborate a little bit on the reasons for repeating `pendingConnectCallbacks` implementation in both the memory connector and datasource? I also noticed that the implementation is slightly different in each place, is that intentional?

test/memory.test.js
```diff
@@ -76,6 +85,35 @@ describe('Memory connector', function() {
       }, done);
     });
 
+    it('should not have out of sequence read/write', function(done) {
```
bajtos: Can you move this test at the end of the `describe` block, so that the existing tests depending on the exact number of `User` instances can remain unchanged?

lib/connectors/memory.js
```diff
@@ -104,45 +133,93 @@ Memory.prototype.collectionSeq = function(model, val) {
   return this.ids[model];
 };
 
+/**
+ * Create a queue to serialize file read/write operations
+ * @returns {*} The file operation queue
+ */
+Memory.prototype.setupFileQueue = function() {
+  var self = this;
```
bajtos: Please use `=>` functions instead of `self`, they perform better.

lib/connectors/memory.js
raymondfeng: I started with the memory connector and then realized it should be handled at datasource level so that all connectors don't have to worry about it. I'm removing the logic from memory connector.

lib/connectors/memory.js
raymondfeng: Noted and fixed.

lib/connectors/memory.js
raymondfeng: Code will be removed

lib/connectors/memory.js
raymondfeng: I'll leave it to future PRs. Use `self` now to keep it consistent.

lib/connectors/memory.js
raymondfeng: The read/write requires/shares quite a bit context. I'm leaving the function as-is for simpler code.

lib/connectors/memory.js
raymondfeng: Fixed.

lib/connectors/memory.js
raymondfeng: Reverted

lib/datasource.js
raymondfeng: Fixed

lib/datasource.js
raymondfeng: Fixed

lib/datasource.js
raymondfeng: Fixed

lib/datasource.js
raymondfeng: Fixed

test/memory.test.js
raymondfeng: I'm using a side effect here to make sure existing records in the file (produced by `create`) won't be overridden.

bajtos: That was not clear at all. In general, it's better to write self-contained tests, any side-effects should be triggered by the test code.

At minimum, please add a comment explaining why it is important to keep this particular test in the current position (i.e. what you wrote here on GH).

lib/connectors/memory.js
```diff
@@ -54,6 +55,7 @@ Memory.prototype.getTypes = function() {
 };
 
 Memory.prototype.connect = function(callback) {
+  var self = this;
```
bajtos: Unused `self`?

lib/connectors/memory.js
```diff
@@ -104,45 +106,93 @@ Memory.prototype.collectionSeq = function(model, val) {
...
+        // Flush out the models/ids
+        var data = JSON.stringify({
+          ids: self.ids,
+          models: self.cache,
+        }, null, '  ');
+        debug('Writing cache to %s: %s', file, data);
+        fs.writeFile(file, data, function(err) {
+          debug('Cache has been written to %s', file);
+          done(err);
+          callback(err, task.data);
```
bajtos: Here you call `done` and then `callback`. When performing reads (see few lines below), you call `callback` and then `done`. Should we keep the order consistent?

raymondfeng:
@bajtos PTAL

PR #1304, opened 2017-04-03
Fix forceId bug for updateOrCreate

loay:
### Description
Fix forcedId bug validation error for updateOrCreate

#### Related issues

connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1185

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -123,6 +123,13 @@ function isWhereByGivenId(Model, where, idValue) {
   return where[pk] === idValue;
 }
 
+function errorModelNotFound(idValue) {
+  var msg = g.f('Could not update attributes. {{Object}} with {{id}} %s does not exist!', idValue);
+  var error = new Error(msg);
+  error.statusCode = error.status = 404;
+  return error;
+}
+
 DataAccessObject._forDB = function(data) {
   if (!(this.getDataSource().isRelational && this.getDataSource().isRelational())) {
     return data;
@@ -509,6 +516,35 @@ DataAccessObject.upsert = function(data, options, cb) {
   if (id === undefined || id === null) {
     return this.create(data, options, cb);
   }
+  var doValidate = undefined;
+  if (options.validate === undefined) {
+    if (Model.settings.validateUpsert === undefined) {
+      if (Model.settings.automaticValidation !== undefined) {
+        doValidate = Model.settings.automaticValidation;
+      }
+    } else {
+      doValidate = Model.settings.validateUpsert;
+    }
+  } else {
+    doValidate = options.validate;
+  }
+
+  var forceId = this.settings.forceId;
+  if (forceId) {
+    options = Object.create(options);
+    options.validate = !!doValidate;
+    if (doValidate) {
+      Model.findById(id, options, function(err, model) {
+        if (err) return cb(err);
+        if (!model) return cb(errorModelNotFound(id));
+        model.updateAttributes(data, options, cb);
+      });
+    } else {
+      const model = new Model({id: id}, {persisted: true});
+      model.updateAttributes(data, options, cb);
+    }
+    return;
+  }
 
   var context = {
     Model: Model,
@@ -545,19 +581,6 @@ DataAccessObject.upsert = function(data, options, cb) {
         Model = Model.lookupModel(update);
 
         var connector = self.getConnector();
-
-        var doValidate = undefined;
-        if (options.validate === undefined) {
-          if (Model.settings.validateUpsert === undefined) {
-            if (Model.settings.automaticValidation !== undefined) {
-              doValidate = Model.settings.automaticValidation;
-            }
-          } else {
-            doValidate = Model.settings.validateUpsert;
-          }
-        } else {
-          doValidate = options.validate;
-        }
 
         if (doValidate === false) {
           callConnector();
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -913,60 +913,71 @@ describe('manipulation', function() {
     });
   });
 
-  describe('updateOrCreate when forceId is true', function() {
-    var Post;
-    before(function definePostModel(done) {
-      var ds = getSchema();
-      Post = ds.define('Post', {
-        title: {type: String, length: 255},
-        content: {type: String},
-      }, {forceId: true});
-      ds.automigrate('Post', done);
-    });
-
-    it('fails when id does not exist in db & validate is true', function(done) {
-      var unknownId = uid.fromConnector(db) || 123;
-      var post = {id: unknownId, title: 'a', content: 'AAA'};
-      Post.updateOrCreate(post, {validate: true}, (err) => {
-        err.statusCode.should.equal(404);
-        done();
+  bdd.describeIf(connectorCapabilities.supportForceId !== false,
+    'updateOrCreate when forceId is true', function() {
+      var Post;
+      before(function definePostModel(done) {
+        var ds = getSchema();
+        Post = ds.define('Post', {
+          title: {type: String, length: 255},
+          content: {type: String},
+        }, {forceId: true});
+        ds.automigrate('Post', done);
       });
-    });
 
-    it('fails when id does not exist in db & validate is false', function(done) {
-      var unknownId = uid.fromConnector(db) || 123;
-      var post = {id: unknownId, title: 'a', content: 'AAA'};
-      Post.updateOrCreate(post, {validate: false}, (err) => {
-        err.statusCode.should.equal(404);
-        done();
-      });
-    });
-
-    it('works on create if the request does not include an id', function(done) {
-      var post = {title: 'a', content: 'AAA'};
-      Post.updateOrCreate(post, (err, p) => {
-        if (err) return done(err);
-        p.title.should.equal(post.title);
-        p.content.should.equal(post.content);
-        done();
-      });
-    });
-
-    it('works on update if the request includes an existing id in db', function(done) {
-      Post.create({title: 'a', content: 'AAA'}, (err, post) => {
-        if (err) return done(err);
-        post = post.toObject();
-        delete post.content;
-        post.title = 'b';
-        Post.updateOrCreate(post, function(err, p) {
-          if (err) return done(err);
-          p.id.should.equal(post.id);
-          p.title.should.equal('b');
+      it('fails when id does not exist in db & validate is true', function(done) {
+        var unknownId = uid.fromConnector(db) || 123;
+        var post = {id: unknownId, title: 'a', content: 'AAA'};
+        Post.updateOrCreate(post, {validate: true}, (err) => {
+          err.statusCode.should.equal(404);
           done();
         });
       });
+
+      it('fails when id does not exist in db & validate is false', function(done) {
+        var unknownId = uid.fromConnector(db) || 123;
+        var post = {id: unknownId, title: 'a', content: 'AAA'};
+        Post.updateOrCreate(post, {validate: false}, (err) => {
+          err.statusCode.should.equal(404);
+          done();
+        });
+      });
+
+      it('fails when id does not exist in db & validate is false when using updateAttributes',
+    function(done) {
+      var unknownId = uid.fromConnector(db) || 123;
+      var post = new Post({id: unknownId});
+      post.updateAttributes({title: 'updated title', content: 'AAA'}, {validate: false}, (err) => {
+        err.statusCode.should.equal(404);
+        done();
+      });
     });
-  });
+
+      it('works on create if the request does not include an id', function(done) {
+        var post = {title: 'a', content: 'AAA'};
+        Post.updateOrCreate(post, (err, p) => {
+          if (err) return done(err);
+          p.title.should.equal(post.title);
+          p.content.should.equal(post.content);
+          done();
+        });
+      });
+
+      it('works on update if the request includes an existing id in db', function(done) {
+        Post.create({title: 'a', content: 'AAA'}, (err, post) => {
+          if (err) return done(err);
+          post = post.toObject();
+          delete post.content;
+          post.title = 'b';
+          Post.updateOrCreate(post, function(err, p) {
+            if (err) return done(err);
+            p.id.should.equal(post.id);
+            p.title.should.equal('b');
+            done();
+          });
+        });
+      });
+    });
 
   if (!getSchema().connector.replaceById) {
     describe.skip('replaceById - not implemented', function() {});
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -752,7 +752,18 @@ describe('manipulation', function() {
       Todo = db.define('Todo', {
         content: String,
       });
-      db.automigrate(['Post', 'Todo'], done);
+      // Here `Person` model overrides the one outside 'updataOrCreate'
+      // with forceId: false. Related test cleanup see issue:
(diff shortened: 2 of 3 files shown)


lib/dao.js
```diff
@@ -507,6 +507,11 @@ DataAccessObject.upsert = function(data, options, cb) {
     return this.create(data, options, cb);
   }
 
+  var forceId = this.settings.forceId;
+  if (forceId) {
+    return Model.replaceById(id, data, options, cb);
```
bajtos: `upsert` (`patchOrCreate`) needs to call `prototype.patchAttributes`, not `replaceById`.

(Ideally, it should call `patchById`, but that function is not implemented yet.)

test/manipulation.test.js
```diff
@@ -882,6 +882,9 @@ describe('manipulation', function() {
     });
 
     it('should allow save() of the created instance', function(done) {
+      Person.create(
+        {name: 'a-name', gender: undefined, id: 999});
+
```
bajtos: This is against the intent of the test as I understand that intent - call `updateOrCreate` in such way that a new instance is created via the connector method, then call `save` on that method.

I think you need a model with `forceId: false` to test that scenario.

test/optional-validation.test.js
```diff
@@ -226,7 +226,7 @@ describe('optional-validation', function() {
 
       // backwards compatible with validateUpsert
       it('should NOT throw on updateOrCreate(id) with invalid data', function(done) {
-        callUpdateOrCreateWithExistingUserId(null, expectChangeSuccess(INVALID_DATA, done));
+        callUpdateOrCreateWithExistingUserId(null, expectValidationError(done));
```
bajtos: I find this change suspicious, isn't it a sign that this pull request may break existing applications?

bajtos:
I'd like @Amir-61 and @jannyHou to review these changes too, they are more familiar with this part of juggler from what I remember.

test/manipulation.test.js
```diff
@@ -891,6 +894,51 @@ describe('manipulation', function() {
     });
   });
 
+  describe('updateOrCreate when forceId is true', function() {
+    var Post;
+    before(function(done) {
```
superkhau: no lambdas for mocha hooks -- see http://loopback.io/doc/en/contrib/style-guide.html#layout-of-test-files

test/manipulation.test.js
```diff
@@ -891,6 +894,51 @@ describe('manipulation', function() {
...
+      Post = ds.define('Post', {
+        title: {type: String, length: 255},
+        content: {type: String},
+      }, {forceId: true});
+      ds.automigrate('Post', done);
+    });
+
+    it('fails when id does not exist in db', function(done) {
+      var post = {id: 123, title: 'a', content: 'AAA'};
+      Post.updateOrCreate(post, function(err, p) {
```
superkhau: don't need `, p` since you don't use it

test/manipulation.test.js
```diff
@@ -891,6 +894,51 @@ describe('manipulation', function() {
...
+               if (err) return done(err);
+               post = post.toObject();
+               delete post.content;
+               post.title = 'b';
+               Post.updateOrCreate(post, function(err, p) {
+                 if (err) return done(err);
+                 p.id.should.equal(post.id);
+                 done();
+               });
+             });
```
superkhau: indentation is way off here

loay: It was not caught by the linter command, but I know what you mean. I am fixing in the next commit.

loay:
@bajtos @superkhau @jannyHou 
PTAL. Thanks.

lib/dao.js
```diff
@@ -509,6 +509,35 @@ DataAccessObject.upsert = function(data, options, cb) {
...
+  } else {
+    doValidate = options.validate;
+  }
+
+  var forceId = this.settings.forceId;
+  if (forceId) {
+    options = Object.create(options);
+    options.validate = doValidate;
+    if (doValidate) {
+      Model.findById(id, function(err, model) {
```
```js
findById(id, options, function(err, model) {
```

lib/dao.js
```diff
@@ -509,6 +509,35 @@ DataAccessObject.upsert = function(data, options, cb) {
...
+        if (err) return cb(err);
+        model.updateAttributes(data, options, cb);
+      });
+    } else {
+      const model = new Model({id: id}, {persisted: true});
+      options.validate = false;
```
bajtos: This is redundant - `options.validate` was already set to `doValidate` which should be `false`. If it is false-y but not false, then please change L528 instead:

```js
options.validate = !!doValidate;
```

test/manipulation.test.js
```diff
@@ -743,7 +743,15 @@ describe('manipulation', function() {
...
         content: String,
       });
-      db.automigrate(['Post', 'Todo'], done);
+      Person = db.define('Person', {
+        name: String,
+        gender: String,
+        married: Boolean,
+        age: {type: Number, index: true},
+        dob: Date,
+        createdAt: {type: Date, default: Date},
```
bajtos: Are all of these properties required by the newly added tests? Please keep the size of test models as small as feasible, define only properties that are used by the tests.

bajtos: On the second look, the Person model does not seem to be used at all, can you remove it please?

test/manipulation.test.js
```diff
@@ -889,6 +897,51 @@ describe('manipulation', function() {
...
+      var ds = getSchema();
+      Post = ds.define('Post', {
+        title: {type: String, length: 255},
+        content: {type: String},
+      }, {forceId: true});
+      ds.automigrate('Post', done);
+    });
+
+    it('fails when id does not exist in db', function(done) {
+      var post = {id: 123, title: 'a', content: 'AAA'};
```
bajtos: Hard-coded ids like `123` don't work with the Cassandra connector, please check the recent changes my by @Setogit and do the same here.

test/manipulation.test.js
```diff
@@ -889,6 +897,51 @@ describe('manipulation', function() {
...
+        if (err) return done(err);
+        p.title.should.equal(post.title);
+        p.content.should.equal(post.content);
+        done();
+      });
+    });
+
+    it('works on update if the request includes an existing id in db', function(done) {
+      Post.create({title: 'a', content: 'AAA'},
+      function(err, post) {
```
bajtos: Can you move this to the previous line?

```js
Post.create({title: 'a', content: 'AAA'}, function(err, post) {
```

Also since this is on `master` which supports only Node.js 4 and newer, it's usually better to use an arrow function (same applies to the other newly added tests):

```js
Post.create({title: 'a', content: 'AAA'}, (err, post) => {
```

FWIW, I would personally use Promises instead, but I understand it may be better to keep consistency with the existing callback-based coding style in this file.

test/manipulation.test.js
```diff
@@ -889,6 +897,51 @@ describe('manipulation', function() {
...
+        if (err) return done(err);
+        post = post.toObject();
+        delete post.content;
+        post.title = 'b';
+        Post.updateOrCreate(post, function(err, p) {
+          if (err) return done(err);
+          p.id.should.equal(post.id);
```
bajtos: Please add a check that `p.title.should.equal('b')`, just to be sure.

lib/dao.js
jannyHou: `options.validate = !!doValidate;` looks like the solution, this redundant is added for `options.validate === undefined`

test/manipulation.test.js
jannyHou: @bajtos It is created to override the `Person` defined outside the describe('updateOrCreate'), because [test case in updateOrCreate](https://github.com/strongloop/loopback-datasource-juggler/blob/67fd2121e3a32521f18d6db2939860199121b9b8/test/manipulation.test.js#L833) needs a `Person` model with `forceId: false`

bajtos: Makes sense. Unfortunately, this is not clear from the code. Can you please add a comment explaining why the Person model is changed for these tests?

It would be great to eventually clean up these tests so that we don't turn on/turn off `forceId` flag in a seemingly random way...

jannyHou: Could not agree more, filed an issue for it: https://github.com/strongloop/loopback-datasource-juggler/issues/1317

test/manipulation.test.js
jannyHou: let's move forward to make callback-based style consistent with new Pormises style XD

loay:
@bajtos PTAL. Thanks.

lib/dao.js
```diff
@@ -509,6 +509,41 @@ DataAccessObject.upsert = function(data, options, cb) {
...
+    doValidate = options.validate;
+  }
+
+  var forceId = this.settings.forceId;
+  if (forceId) {
+    options = Object.create(options);
+    options.validate = !!doValidate;
+    if (doValidate) {
+      Model.findById(id, options, function(err, model) {
+        if (err || !model) return cb(err || errorModelNotFound(id));
```
bajtos: I think is line is trying to be too clever. I am proposing to rewrite this bit to make it easier to understand:

```js
if (err) return cb(err);
if (!model) return cb(errorModelNotFound(id));
```

lib/dao.js
```diff
@@ -509,6 +509,41 @@ DataAccessObject.upsert = function(data, options, cb) {
...
+        model.updateAttributes(data, options, cb);
+      });
+    } else {
+      const model = new Model({id: id}, {persisted: true});
+      model.updateAttributes(data, options, cb);
+    }
+    return;
+  }
+
+  function errorModelNotFound(idValue) {
```
bajtos: Isn't there a similar method already defined somewhere in this huge file, a method we can reuse? If it isn't, them please make one - i.e. move this function to the root level of this source file, so that other methods can use it too.

loay:
@slnode test please

loay:
@slnode test please

loay:
@slnode test please

loay:
@slnode test please

loay:
@slnode test please

test/manipulation.test.js
```diff
@@ -902,6 +913,72 @@ describe('manipulation', function() {
...
+        }, {forceId: true});
+        ds.automigrate('Post', done);
+      });
+
+      it('fails when id does not exist in db & validate is true', function(done) {
+        var unknownId = uid.fromConnector(db) || 123;
+        var post = {id: unknownId, title: 'a', content: 'AAA'};
+        Post.updateOrCreate(post, {validate: true}, (err) => {
+          err.statusCode.should.equal(404);
+          done();
```
kjdelisle: 1:
Use the message you hoisted out from Comment 1 to assert against the error message.

Something like `err.message.should.contain('Could not update attributes');`

test/manipulation.test.js
```diff
@@ -902,6 +913,72 @@ describe('manipulation', function() {
...
+        });
+      });
+
+      it('fails when id does not exist in db & validate is false', function(done) {
+        var unknownId = uid.fromConnector(db) || 123;
+        var post = {id: unknownId, title: 'a', content: 'AAA'};
+        Post.updateOrCreate(post, {validate: false}, (err) => {
+          err.statusCode.should.equal(404);
```
kjdelisle: See 1

test/manipulation.test.js
```diff
@@ -902,6 +913,72 @@ describe('manipulation', function() {
...
+          done();
+        });
+      });
+
+      it('fails when id does not exist in db & validate is false when using updateAttributes',
+    function(done) {
+      var unknownId = uid.fromConnector(db) || 123;
+      var post = new Post({id: unknownId});
+      post.updateAttributes({title: 'updated title', content: 'AAA'}, {validate: false}, (err) => {
+        err.statusCode.should.equal(404);
```
kjdelisle: See 1

test/manipulation.test.js
loay: I can add it, but that error message was tested already. The goal was to test the error code `404`. I am adding it in the next commit.

kjdelisle:
Apparently the error messages aren't consistent between connectors (as demonstrated by mongodb's failure in CI just now). For now, let's remove the extra check I asked for and LGTM.

lib/dao.js
```diff
@@ -509,6 +516,35 @@ DataAccessObject.upsert = function(data, options, cb) {
...
+      Model.findById(id, options, function(err, model) {
+        if (err) return cb(err);
+        if (!model) return cb(errorModelNotFound(id));
+        model.updateAttributes(data, options, cb);
+      });
+    } else {
+      const model = new Model({id: id}, {persisted: true});
+      model.updateAttributes(data, options, cb);
+    }
+    return;
```
bajtos: I think this will break promise mode, you need to `return cb.promise`. @loay could you please check?

A fix should include a unit-test to verify.

test/manipulation.test.js
```diff
@@ -902,6 +913,72 @@ describe('manipulation', function() {
...
+          content: {type: String},
+        }, {forceId: true});
+        ds.automigrate('Post', done);
+      });
+
+      it('fails when id does not exist in db & validate is true', function(done) {
+        var unknownId = uid.fromConnector(db) || 123;
+        var post = {id: unknownId, title: 'a', content: 'AAA'};
+        Post.updateOrCreate(post, {validate: true}, (err) => {
+          err.statusCode.should.equal(404);
```
bajtos: This will fail with `TypeError` when `updateOrCreate` passes and `err` is undefined.

Here is better assert:

```js
should(err).have.property('statusCode', 404);
```

test/manipulation.test.js
bajtos: ditto, what if `err` is not defined?

test/manipulation.test.js
bajtos: ditto, what if `err` is not defined?

test/manipulation.test.js
```diff
@@ -902,6 +913,72 @@ describe('manipulation', function() {
...
+        var unknownId = uid.fromConnector(db) || 123;
+        var post = {id: unknownId, title: 'a', content: 'AAA'};
+        Post.updateOrCreate(post, {validate: false}, (err) => {
+          err.statusCode.should.equal(404);
+          done();
+        });
+      });
+
+      it('fails when id does not exist in db & validate is false when using updateAttributes',
+    function(done) {
```
bajtos: The indentation is off - we never ident the second line of a multi-line function call with less space than the first line.

test/manipulation.test.js
```diff
@@ -902,6 +913,72 @@ describe('manipulation', function() {
...
+      var unknownId = uid.fromConnector(db) || 123;
+      var post = new Post({id: unknownId});
+      post.updateAttributes({title: 'updated title', content: 'AAA'}, {validate: false}, (err) => {
+        err.statusCode.should.equal(404);
+        done();
+      });
+    });
+
+      it('works on create if the request does not include an id', function(done) {
```
bajtos: The indentation does not look right to me.

test/manipulation.test.js
```diff
@@ -902,6 +913,72 @@ describe('manipulation', function() {
...
+        var post = {title: 'a', content: 'AAA'};
+        Post.updateOrCreate(post, (err, p) => {
+          if (err) return done(err);
+          p.title.should.equal(post.title);
+          p.content.should.equal(post.content);
+          done();
+        });
+      });
+
+      it('works on update if the request includes an existing id in db', function(done) {
```
bajtos: ditto, please check the indentation

PR #1320, opened 2017-04-09
Allow selection and filtering on null DateType

pastcompute:
### Description

This lets you use a filter such as `{ where : { some_date_property: null } } `

Testing:

Before - a query like the following failed with a 500:
```
http://localhost:3000/api/MyThing?filter={"where":{"some_date_property":"null"}}
http://localhost:3000/api/MyThing?filter={"where":{"some_date_property":{"neq":"null"}}}
```

After - now it works

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1537,6 +1537,9 @@ DataAccessObject._normalize = function(filter, options) {
 };
 
 function DateType(arg) {
+  if (arg == null || arg == "null") {
+    return null;
+  }
   var d = new Date(arg);
   if (isNaN(d.getTime())) {
     throw new Error(g.f('Invalid date: %s', arg));
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1537,7 +1537,7 @@ DataAccessObject._normalize = function(filter, options) {
 };
 
 function DateType(arg) {
-  if (arg == null || arg == "null") {
+  if (arg == null || arg == 'null') {
     return null;
   }
   var d = new Date(arg);


slnode:

slnode:

slnode:

slnode:

lib/dao.js
```diff
@@ -1537,6 +1537,9 @@ DataAccessObject._normalize = function(filter, options) {
 };
 
 function DateType(arg) {
+  if (arg == null || arg == "null") {
```
ssh24: I wonder how a nested date type filter with a null will work with this code: 
`http://localhost:3000/api/MyThing?filter={"where":{"some_date_property":{"neq":"null"}}}`?

pastcompute: It works for me - or do you mean something else?

ssh24: @pastcompute Could you add some test cases to verify your changes?

pastcompute: OK, I'll give it a go. Might be a few days though

ssh24: Sounds good. Thank you!

pastcompute:
Attempting to write a test now.
Well, l this _is_ fun... I'm learning way too much about the quirks in Javascript date while trying to understand what the loopback behaviour was before my proposed patch...
1. From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date there is no specification as to what `new Date(null)` should return
2. Node existing behaviour seems to return `new Date(null) --> 1970-01-01 00:00`
3. Node existing behaviour seems to return `new Date('null') --> null`
4. Thus loopback `DateType()` existing behaviour was `1970-01-01 00:00` in first case and an Error in the second. This Error is what was causing my where query to fail.

Turns out my proposed patch was a red herring, I do apologise for that. Will be committing a different proposed change shortly...

PR #1325, opened 2017-04-13
Alter relations test case

loay:
### Description
This Pr https://github.com/strongloop/loopback-connector/pull/100  is failing on the three connectors: `mssql`, `oracle` and  `postgresql`.
EmbedsMany  relation model does not really work with connectors `mssql`, `oracle`, `postgresql`
and the id cannot be found and it generates an error: ` Error: Could not update attributes. Object with id 5 does not exist!`. 

The test case can only see one query where the id is `link.id`, however that id does not exist in the db. 

#### Related issues

connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1331
connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1185

#### Related PRS:
https://github.com/strongloop/loopback-datasource-juggler/pull/1304
https://github.com/strongloop/loopback-connector/pull/100

### DIFF
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -4987,7 +4987,10 @@ describe('relations', function() {
       Category.findById(category.id, function(err, cat) {
         if (err) return done(err);
         var link = cat.items.at(0);
-        link.updateAttributes({notes: 'Updated notes...'}, function(err, link) {
+        // use 'updateById' instead as a replacement as it is one of the embedsMany methods,
+        // that works with all connectors. `updateAttributes` does not recognize the query done on
+        // the Category Model, resulting with an error in three connectors: mssql, oracle, postgresql
+        cat.items.updateById(link.id, {notes: 'Updated notes...'}, function(err, link) {
           if (err) return done(err);
           link.notes.should.equal('Updated notes...');
           done();


test/relations.test.js
```diff
@@ -4987,7 +4987,7 @@ describe('relations', function() {
       Category.findById(category.id, function(err, cat) {
         if (err) return done(err);
         var link = cat.items.at(0);
-        link.updateAttributes({notes: 'Updated notes...'}, function(err, link) {
+        cat.items.updateById(link.id, {notes: 'Updated notes...'}, function(err, link) {
```
bajtos: My understanding of embedded relations is rather limited. However, by looking at the test name, test code and surrounding tests, I think the very purpose of this test is to verify how `link.updateAttributes` behaves. Changing `link.updateAttributes` to `cat.items.updateById` is making the test pointless IMO.

loay: `Model1.Model2.updateById` is one of the operations of embedsMany, and it is giving the same behavior but with a workaround for the issue we have with the failing connectors. The purpose was just to update and that's why the next failing is depending on this one.

jannyHou: `link.destroy` still works in https://github.com/strongloop/loopback-datasource-juggler/blob/124b1d48839830f1f69b68fb90217790700b3ae6/test/relations.test.js#L5012 , I am looking into why `link.updateAttributes` returns error with postgresql

jannyHou:
@loay I test with this branch and also `upsert-forceid` branch with postgresql on local, they both pass this test case, could you provide more details how did you run into that error?

loay:
@jannyHou  use `loopback-connector` -> branch `upsert-forceid`

test/relations.test.js
```diff
@@ -4987,7 +4987,7 @@ describe('relations', function() {
       Category.findById(category.id, function(err, cat) {
         if (err) return done(err);
         var link = cat.items.at(0);
-        link.updateAttributes({notes: 'Updated notes...'}, function(err, link) {
```
jannyHou: I am confused that calling `connector.updateAttributes()` and callback the returned result is not the last step in `DAO.prototype.updateAttributes()`:
https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L3229
So in this test case, `link.updateAttributes()` still wants to execute the code in `updateAttributesCallback()`after L3229 **to do the real update**  after `connector.updateAttributes()` returns result, which is null.

This makes it difficult to return id not found error in `SQL.prototype.update()`  from loopback-connector/sql.js, because sometimes when `SQL.prototype.update()` get called, the caller expects error returned when `count===0`, while sometimes the caller tolerant it, and will do the real update afterwards.

ssh24:
Postgres should be fixed. 

@slnode test please

loay:
@slnode test please

loay:
@slnode test please

test/relations.test.js
kjdelisle: ☝️

loay: A slight change, now it is `SQL.prototype.updateAttribute()` in pr https://github.com/strongloop/loopback-connector/pull/100 instead of `SQL.prototype.update()` in pr https://github.com/strongloop/loopback-connector/pull/97. That does not change much, but just for the record. 

There is the `Category` that embedsMany `items`, so there should be 2 queries, but  It seems like only 3 connectors `mssql`, `oracle`, `postgresql` are not recognizing the 2nd query. They only execute the one on `items` and obviously they do not see it in the database, so there are 2 options:
1) skip the test for the three connectors
2) this solution implemented in this pr, which is to use `updateById` which is one of the methods of `embedsMany` to fix the situation. https://loopback.io/doc/en/lb2/Embedded-models-and-relations.html#embedsmany

jannyHou: > this solution implemented in this pr, which is to use updateById which is one of the methods of embedsMany to fix the situation

If I understand correctly, there are two apis:
- category.items.updateById
- item.updateAttributes

Probably they both can update the embedded instance, but this test case is supposed to test the second one, not "user has a way to update the embedded instance" so whether the first one works or not doesn't matter. 

Actually I am fine to skip this test case if it turns out the embedded instance update strategy has conflict with connector's update/updateAttributes function....

Issue #1332, opened 2017-04-18
Geo query fails with null property


garrows:
Since updating to v3.5.0 I started encountering this crash. Code was introduced here https://github.com/strongloop/loopback-datasource-juggler/commit/64f64caaff70898281b09d8469db4e7a89fa160f#diff-b367a815d81f15531e0d0553ea8aaa5aR28

I dont have time right now to give you a proper reproduction case but hopefully this will help you and others encountering the problem.

For now I have rolled back to `loopback-datasource-juggler@3.4.1`

```
TypeError: Cannot read property 'hasOwnProperty' of null
    at /Users/glen/api/node_modules/loopback-datasource-juggler/lib/geo.js:28:30
    at Array.forEach (native)
    at nearSearch (/Users/glen/api/node_modules/loopback-datasource-juggler/lib/geo.js:21:25)
    at Object.nearFilter (/Users/glen/api/node_modules/loopback-datasource-juggler/lib/geo.js:44:3)
    at Function.find (/Users/glen/api/node_modules/loopback-datasource-juggler/lib/dao.js:1891:27)
    at /Users/glen/api/node_modules/loopback/common/models/user.js:322:15
    at notifySingleObserver (/Users/glen/api/node_modules/loopback-datasource-juggler/lib/observer.js:104:22)
    at /Users/glen/api/node_modules/loopback-datasource-juggler/node_modules/async/dist/async.js:3025:16
    at replenish (/Users/glen/api/node_modules/loopback-datasource-juggler/node_modules/async/dist/async.js:881:17)
    at /Users/glen/api/node_modules/loopback-datasource-juggler/node_modules/async/dist/async.js:885:9
    at eachLimit$1 (/Users/glen/api/node_modules/loopback-datasource-juggler/node_modules/async/dist/async.js:3114:22)
    at Object.<anonymous> (/Users/glen/api/node_modules/loopback-datasource-juggler/node_modules/async/dist/async.js:917:16)
    at doNotify (/Users/glen/api/node_modules/loopback-datasource-juggler/lib/observer.js:101:11)
    at doNotify (/Users/glen/api/node_modules/loopback-datasource-juggler/lib/observer.js:99:49)
    at doNotify (/Users/glen/api/node_modules/loopback-datasource-juggler/lib/observer.js:99:49)
    at doNotify (/Users/glen/api/node_modules/loopback-datasource-juggler/lib/observer.js:99:49)
[...]

PR #1334, opened 2017-04-19
Fix/geo null

paulussup:
### Description

Add check for null and undefined.

#### Related issues
#1332
connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1332

- None

### Checklist

### DIFF
--- a/lib/geo.js
+++ b/lib/geo.js
@@ -19,12 +19,13 @@ exports.nearFilter = function nearFilter(where) {
     parentKeys = parentKeys || [];
 
     Object.keys(clause).forEach(function(clauseKey) {
+      if (typeof clause[clauseKey] !== 'object' || !clause[clauseKey]) return;
       if (Array.isArray(clause[clauseKey])) {
         clause[clauseKey].forEach(function(el, index) {
           var ret = nearSearch(el, parentKeys.concat(clauseKey).concat(index));
           if (ret) return ret;
         });
-      } else if (clause[clauseKey] !== null && clause[clauseKey] !== undefined) {
+      } else {
         if (clause[clauseKey].hasOwnProperty('near')) {
           var result = clause[clauseKey];
           nearResults.push({
@@ -37,9 +38,7 @@ exports.nearFilter = function nearFilter(where) {
             key: clauseKey,
           });
         }
-      } else {
-        return;
-      }
+      };
     });
   }
   var nearResults = [];
--- a/lib/geo.js
+++ b/lib/geo.js
@@ -24,7 +24,7 @@ exports.nearFilter = function nearFilter(where) {
           var ret = nearSearch(el, parentKeys.concat(clauseKey).concat(index));
           if (ret) return ret;
         });
-      } else if (typeof clause[clauseKey] === 'object') {
+      } else if (clause[clauseKey] !== null && clause[clauseKey] !== undefined) {
         if (clause[clauseKey].hasOwnProperty('near')) {
           var result = clause[clauseKey];
           nearResults.push({
--- a/lib/geo.js
+++ b/lib/geo.js
@@ -24,7 +24,7 @@ exports.nearFilter = function nearFilter(where) {
           var ret = nearSearch(el, parentKeys.concat(clauseKey).concat(index));
           if (ret) return ret;
         });
-      } else {
+      } else if (typeof clause[clauseKey] === 'object') {
         if (clause[clauseKey].hasOwnProperty('near')) {
           var result = clause[clauseKey];
           nearResults.push({
@@ -37,6 +37,8 @@ exports.nearFilter = function nearFilter(where) {
             key: clauseKey,
           });
         }
+      } else {
+        return;
       }
     });
   }
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -153,9 +153,9 @@ describe('basic-querying', function() {
           done();
         });
     });
-    it('should query by ids and null condition', function(done) {
+    it('should query by ids to check null/undefined property', function(done) {
       // due to some connectors not being assigning null value by default
-      // the query is filtered by the fields which may consist null value
+      // the query is filtered by the fields which may consist in null value
       // or not exist at all
       User.findByIds([
         createdUsers[0].id,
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

lib/geo.js
```diff
@@ -24,7 +24,7 @@ exports.nearFilter = function nearFilter(where) {
           var ret = nearSearch(el, parentKeys.concat(clauseKey).concat(index));
           if (ret) return ret;
         });
-      } else {
+      } else if (clause[clauseKey] !== null && clause[clauseKey] !== undefined) {
```
ebarault: we usually prefer early return to keep the code simpler to read, i suggest:

```
Object.keys(clause).forEach(function(clauseKey) {
  if (typeof clause[clauseKey] !== 'object' || !clause[clauseKey]) return;

  if (Array.isArray(clause[clauseKey])) {
    // array
  // object different than null, undefined, 0
  } else {
    // object other than null, undefined, 0
  }
}
```

test/basic-querying.test.js
```diff
@@ -153,6 +153,27 @@ describe('basic-querying', function() {
           done();
         });
     });
+    it('should query by ids and null condition', function(done) {
```
ebarault: is it required for this test to search among these 4 users? unless i'm missing something could you simplify with only 2 different users?

test/basic-querying.test.js
```diff
@@ -153,6 +153,18 @@ describe('basic-querying', function() {
...
+      User.findByIds([
+        createdUsers[0].id,
+        createdUsers[1].id],
+        {where: {vip: null}}, function(err, users) {
+          should.exist(users);
+          should.not.exist(err);
```
ssh24: @paulussup Could you please swap L161 with L162? It is always a good idea to check `err` first.

ssh24:
@slnode test please

test/basic-querying.test.js
paulussup: fixed @ssh24

ssh24:
@slnode test please

ssh24:
@slnode test please

test/basic-querying.test.js
```diff
@@ -153,6 +153,18 @@ describe('basic-querying', function() {
           done();
         });
     });
+    it('should query by ids and null condition', function(done) {
+      User.findByIds([
+        createdUsers[0].id,
+        createdUsers[1].id],
+        {where: {vip: null}}, function(err, users) {
```
ssh24: @paulussup @ebarault I am not sure if this test case checks for the fix in the right away. The fix is however correct.

Shouldn't the filter do `near` query with a null value and see if the null value it is taken care of? Something like this test case but with a `null` value filter: https://github.com/strongloop/loopback-datasource-juggler/commit/64f64caaff70898281b09d8469db4e7a89fa160f#diff-8fae24fa21c82e23f7f43ccdae4fcf90R599

paulussup:
To be honest I copy pasted the test above written by @raymondfeng and made the required edits. The reason that I chose to just test the null value is that the documentation uses querying a field as null as an example. Because this was not tested the error was able to make its way in to master.

ssh24:
@paulussup That is fine. We can still keep this test case and just add that extra test case.

The test case is also failing on mongodb and cassandra connector so need to look into that.

paulussup:
I guess that is because those datasources don’t default the value to null when persisting like the sql datasources.

ssh24:
@slnode test please

ssh24:
@ebarault This fix looks good to me. Could you do me a favour and review the changes I committed, i.e fixing and adding new test cases?

@slnode test please

ssh24:
@slnode test please

test/basic-querying.test.js
```diff
@@ -153,6 +153,26 @@ describe('basic-querying', function() {
           done();
         });
     });
+    it('should query by ids and null condition', function(done) {
+      // due to some connectors not being assigning null value by default
+      // the query is filtered by the fields which may consist null value
```
ebarault: typo: "which may consist **in** null value ..."

test/basic-querying.test.js
```diff
@@ -153,6 +153,26 @@ describe('basic-querying', function() {
...
+      // or not exist at all
+      User.findByIds([
+        createdUsers[0].id,
+        createdUsers[1].id],
+        {fields: {name: true, vip: true}}, function(err, users) {
```
ebarault: i'm afraid i don't understand, the test is named: 'should query by ids and null condition' and we are not querying by null value here. I suspect it should be `vip: null` here?

test/basic-querying.test.js
```diff
@@ -593,6 +613,19 @@ describe('basic-querying', function() {
 
     describe('geo queries', function() {
       describe('near filter', function() {
+        it('supports a near query with null condition', function(done) {
```
ebarault: could you move this test right after the 'supports a basic "near" query' test please?

test/basic-querying.test.js
ebarault: or i'm missing this: one of the user has an undefined property "vip" and the test should be named something like: `'should query by ids and condition, allowing results where conditions match null properties'`

test/basic-querying.test.js
ebarault: if i'm right in my previous comment the test name should also be revised to reflect that we are not querying on a null condition, but that results on the conditions could come with empty properties

something like:
`'supports a compound near query with conditions, allowing results where conditions match null properties'`

ssh24:
@ebarault Updated the test messages.

@slnode test please

test/basic-querying.test.js
paulussup: I have to say I don't see how this test is even testing the behaviour that caused this bug. The problem arose when traversing the where object for the near key and the value is``` {where:
 {field: null}}```. Would it not make more sense to set vip explicitly to null in this test case or create new test seed data with explicit null values.

ssh24: @paulussup Good question. Your test case was not wrong however it had a loophole where it would not check for connectors which does not put `null` values for fields which are not set (it would instead have it as undefined, for example mongodb).

Improved the test case to take care of that situation by filtering by fields (name and vip) and then get those instances back and then perform a check if they are either `null` or `undefined`. After fixing this it works with the downstream connectors such as mongodb and cassandra.

paulussup: I accept that it fixes the downstream issue. If there was a test which actually set the value explicitly to null in the test I think this too would allow the downstream connectors to query the null value and return records. Mongo and Cassandra can both store null values.

ssh24: @paulussup Yes I agree with you. Would you like to fix the test cases to go with the changes you proposed? We can just add a property `vip` with value **null** on the second user of `createdUsers` and just query that with your previous test case.

Please be sure to pull latest from your branch if you haven't already.

paulussup:
Sure no problems.

paulussup:
Haven’t had a chance yet.

ssh24:
@paulussup I have fixed the test cases myself.

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
PR ready to be landed. @paulussup Thank you again for the patch!

PR #1339, opened 2017-04-21
replace exception thrown for invalid dates

dhmlau:
### Description
Originated from PR https://github.com/strongloop/loopback-datasource-juggler/pull/718
There's no response from the user to grant me access to his fork, so I've created a separate PR.  User @hotaru355 has signed CLA. 

Excerpt from original PR: 
```
Instantiating a model with an invalid date would not throw an exception, 
but calling isValid() on it would return false, without having to add the validation method explicitly.
```

2 tests are removed because they are no longer valid because no exceptions are thrown when instantiating the model. 
- `should fail if field validation fails` in manipuliation.test.js 
- `throws an error when property of type Date is set to an invalid value` in datatype.test.js

#### Related issues
https://github.com/strongloop/loopback-datasource-juggler/pull/718

### Checklist

### DIFF
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -426,6 +426,7 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
       var requiredOptions = typeof prop.required === 'object' ? prop.required : undefined;
       ModelClass.validatesPresenceOf(propertyName, requiredOptions);
     }
+    if (DataType === Date) ModelClass.validatesDateOf(propertyName);
 
     Object.defineProperty(ModelClass.prototype, propertyName, {
       get: function() {
@@ -526,10 +527,8 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
 
 // DataType for Date
 function DateType(arg) {
+  if (arg === null) return null;
   var d = new Date(arg);
-  if (isNaN(d.getTime())) {
-    throw new Error(g.f('Invalid date: %s', arg));
-  }
   return d;
 }
 
--- a/lib/validations.js
+++ b/lib/validations.js
@@ -243,6 +243,7 @@ Validatable.validateAsync = getConfigurator('custom', {async: true});
  */
 Validatable.validatesUniquenessOf = getConfigurator('uniqueness', {async: true});
 
+Validatable.validatesDateOf = getConfigurator('date');
 // implementation of validators
 
 /*!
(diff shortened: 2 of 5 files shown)


lib/model-builder.js
```diff
@@ -426,6 +426,9 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
       var requiredOptions = typeof prop.required === 'object' ? prop.required : undefined;
       ModelClass.validatesPresenceOf(propertyName, requiredOptions);
     }
+    if (DataType === Date) {
```
ssh24: Can we do a one liner if statement here? `if (DataType === Date) ModelClass.validatesDateOf(propertyName);`

lib/model-builder.js
```diff
@@ -526,10 +529,10 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
 
 // DataType for Date
 function DateType(arg) {
-  var d = new Date(arg);
-  if (isNaN(d.getTime())) {
-    throw new Error(g.f('Invalid date: %s', arg));
+  if (arg === null) {
```
ssh24: Same here. One liner if statement.

lib/validations.js
```diff
@@ -383,6 +384,18 @@ function validateUniqueness(attr, conf, err, options, done) {
   }.bind(this));
 }
 
+/*!
+ * Date validator
+ */
+function validateDate(attr, conf, err) {
```
ssh24: What is the need of `conf` here?

lib/validations.js
```diff
@@ -383,6 +384,18 @@ function validateUniqueness(attr, conf, err, options, done) {
...
+  if (this[attr] === null || this[attr] === undefined) return;
+
+  var date = new Date(this[attr]);
+  if (isNaN(date.getTime())) {
```
ssh24: One liner if statement please :)

lib/model-builder.js
```diff
@@ -526,10 +529,10 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
+    return null;
   }
+  var d = new Date(arg);
```
b-admike: I'd wrap this with a try catch block in case there is an error here

lib/validations.js
dhmlau: It turns out that `conf` is needed, otherwise the tests failed.
It's a required attribute for similar functions defined inside `validators` (in line397), even though it's not being used.

ssh24: @dhmlau I am not sure what you mean

dhmlau: the `validators` are defined in the block around line 397.  There are `validatePresence`, `validateAbsence` and also the newly added `validateDate`.  If you go to those `validate*` functions, they all have `conf`.  This is needed when those functions are being called.
In fact, without `conf`, tests fail because it expects the 2nd parameter in validate* function to be a function.

PR #1342, opened 2017-04-21
Test juggler@2.x dependencies

jannyHou:
Test 2.x dependency
There are 14 failures, could not backport any PR before getting them pass.

PR #1343, opened 2017-04-21
CI/downstreamIgnoreList

jannyHou:
### Description

juggler@2.x branch fails a lot, details see https://github.com/strongloop/loopback-datasource-juggler/pull/1342
More PRs needed in the near future, this is the first one.

Add same downstreamIgnoreList as master branch(juggler@3.x), and `loopback-connector-cassandra`.

For cassandra, there might be better way to tell CI it only works with juggler@3.x but not juggler@2.x, appreciate it if people more familiar with module's dependency version could share their thought here. Thanks.

### Checklist

- [ ] New tests added or existing tests modified to cover all changes(get old ones pass, don't need new ones)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,5 +53,13 @@
     "traverse": "^0.6.6",
     "uuid": "^3.0.1"
   },
-  "license": "MIT"
+  "license": "MIT",
+  "ci": {
+    "downstreamIgnoreList": [
+      "loopback-connector-db2z",
+      "loopback-connector-informix",
+      "loopback-connector-mqlight",
+      "loopback-connector-cassandra"
+    ]
+  }
 }


package.json
```diff
@@ -53,5 +53,13 @@
...
     "uuid": "^3.0.1"
   },
-  "license": "MIT"
+  "license": "MIT",
+  "ci": {
+    "downstreamIgnoreList": [
+      "loopback-connector-db2z",
+      "loopback-connector-informix",
+      "loopback-connector-mqlight",
+      "loopback-connector-cassandra"
```
ssh24: I assume this is the only way to tell CI to run `loopback-connector-cassandra` with juggler3.x only since [juggler3.x](https://github.com/strongloop/loopback-datasource-juggler/blob/master/package.json#L56) does not ignore cassandra.

loay:
In juggler 3.x, the tests are written as follows in order not to conflict with Cassandra

 ``` bdd.itIf(connectorCapabilities.[name] !== false, 'test description', function(done) {```
example:

 ``` bdd.itIf(connectorCapabilities.foreceId !== false, 'updateOrCreate should .... ', function(done) {```
I am not sure if that was backported or not.

rmg:
The downstream dependencies are determined by _actual_ dependencies. The reason `loopback-connector-cassandra` doesn't show up as a downstream is because no monitored branch of that connector actually depends on loopback-datasource-juggler@2.x.

rmg:
Also, I caution **strongly** against merging this. Connectors are the **most important downstream tests** for this module.

Without them there will be no indication that a connector breaks until _after_ a PR is merged - possibly even as late as after a _release_.

rmg:
A few of the downstream connectors also appear to only be run because they have a `loopback-2.x` branch created _specifically_ for this. Why not just delete the branch instead?

jannyHou:
@rmg ok I will close this one, and list the ones don't need a `loopback-2.x` branch in PR https://github.com/strongloop/loopback-datasource-juggler/pull/1342

And I agree, after releasing 3.x and 4.x version of some connectors, juggler@2.x can just test with branch connector@2.x, there is no need to keep a loopback-2.x branch, since 3.x and 4.x could not be compatible with loopback@2.x

PR #1352, opened 2017-04-25
datatype.test: minimize rounding failures

kjdelisle:
The current implementation will rarely fail as a result of a
rounding error between both values, so switching the comparison to
ISO timestamps to reduce the likelihood of such an event in future.

#### Related issues

- None

### Checklist


kjdelisle:
So apparently, any of the datasources that store these values as UTC are truncating the fractional seconds on save.

Not sure why the original representation occasionally hits a rounding error as a result, then...

PR #1354, opened 2017-04-25
create sequence for nosql id

jannyHou:
### Description
PR created when recovering test cases in cloudant:
cloudant creates string type id which could not guarantee id are incremented. To solve the problem, I add a `sequence` property to make sure the related test case returns result ordered by sequence and will skip the first data. 

### Checklist

### DIFF
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -856,8 +856,8 @@ describe('basic-querying', function() {
           if (err) return done(err);
           users.length.should.be.equal(2);
           var expectedUsers = ['John Lennon', 'Paul McCartney'];
-          (expectedUsers.indexOf(users[0].name) > -1).should.be.ok();
-          (expectedUsers.indexOf(users[1].name) > -1).should.be.ok();
+          expectedUsers.indexOf(users[0].name).should.not.equal(-1);
+          expectedUsers.indexOf(users[1].name).should.not.equal(-1);
           done();
         });
       });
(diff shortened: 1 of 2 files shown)


test/basic-querying.test.js
```diff
@@ -341,8 +341,9 @@ describe('basic-querying', function() {
         should.not.exist(err);
         should.exist(users);
         users.should.have.property('length', 2);
-        users[0].name.should.equal('John Lennon');
-        users[1].name.should.equal('Paul McCartney');
+        var expectedNames = ['John Lennon', 'Paul McCartney'];
+        expectedNames.indexOf(users[0].name).should.be.above(-1);
```
dhmlau: I guess we're checking the index is not -1... would it be `should.not.equal(-1)`?

test/relations.test.js
```diff
@@ -832,8 +835,16 @@ describe('relations', function() {
               if (err) return done(err);
               should.exist(ch);
               ch.should.have.lengthOf(2);
-              var resultIdArr = [ch[0].id, ch[1].id];
-              assert.deepEqual(resultIdArr, idArr);
+              if (typeof idArr[0] === 'object') {
+                // mongodb returns `id` as an object
+                idArr[0] = idArr[0].toString();
+                idArr[1] = idArr[1].toString();
+                idArr.indexOf(ch[0].id.toString()).should.be.above(-1);
```
dhmlau: same comment.. if we're checking index should not be -1, perhaps better to use `should.not.equal(-1)`.

jannyHou:
@dhmlau thanks applied.

jannyHou:
@slnode test please

jannyHou:
@slnode test please

jannyHou:
@slnode test please

Issue #271, opened 2014-08-31
polymorphic belongsTo along with another standard belongsTo may cause error


clark0x:
I am trying define a model with two belongsTo, one of them is polymorphic. And I got an error on boot:

```
node_modules/loopback-datasource-juggler/lib/relation-definition.js:496
      throw new Error('Could not find "' + params.as + '" relation for ' + mod
            ^
Error: Could not find "actor" relation for Favorite
    at lookupModelTo
```

Relations are defined as follow:

``` js
"target": {
      "type": "belongsTo",
      "polymorphic": true
    },
    "actor": {
      "type": "belongsTo",
      "model": "User"
    }
```

I found that, if I change the relation name `'target'` to others such as `'pet'`, everything works fine. But changing to others words such as `'abc'`, the error will come back.

PR #1356, opened 2017-04-26
Add DateString type

kjdelisle:
### Description
New type that preserves string input as a string, but ensures that
the string is a valid Date.

Additionally, provides a .toDate function to provide the Date
object representation of the string.

**NOTE**: This PR has changed from its original intent, so if you feel confused, you should know that you're not crazy. :)

### Linked Issues
https://github.com/strongloop/loopback-connector-mysql/issues/149

### Checklist

- [x] Write the Doc PR that describes this feature's usage with examples!

### DIFF
--- a/lib/date-string.js
+++ b/lib/date-string.js
@@ -1,0 +1,99 @@
+// Copyright IBM Corp. 2013,2016. All Rights Reserved.
+// Node module: loopback-datasource-juggler
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+'use strict';
+
+var inspect = require('util').inspect;
+
+module.exports = DateString;
+
+/**
+ * A String whose value is a valid representation of a Date.
+ * Use this type if you need to preserve the format of the value and still
+ * check if it's valid.
+ * Example:
+ * ```js
+ * var loopback = require('loopback');
+ * var dt = new loopback.DateString('2001-01-01');
+ *
+ * dt.toString();
+ * // '2001-01-01'
+ * dt._date.toISOString();
+ * // '2001-01-01T00:00:00.000Z'
+ * ```
+ *
+ * You can use this definition on your models as well:
+ * ```json
+ * {
+ *   "name": "Person",
+ *   "base": "PersistedModel",
+ *   "properties": {
+ *     "name": {
+ *       "type": "string"
+ *     },
+ *     "dob": {
+ *       "type": "DateString",
+ *       "required": true
+ *     },
+ *   },
+ *   "validations": [],
+ *   "relations": {},
+ *   "acls": [],
+ *   "methods": {}
+ * }
+ * ```
+ * @class DateString
+ * @param {String} value
+ * @constructor
+ */
+function DateString(value) {
+  if (!(this instanceof DateString)) {
+    return new DateString(value);
+  }
+
+  if (typeof(value) !== 'string') {
+    throw new Error('Input must be a string');
+  }
+
+  Object.defineProperty(this, 'when', {
+    get: () => { return this._when; },
+    set: (val) => {
+      var d = new Date(val);
+      if (isNaN(d.getTime())) {
+        throw new Error('Invalid date');
+      } else {
+        this._when = val;
+        this._date = d;
+      }
+    },
+  });
+
+  this.when = value;
+};
+
+/**
+ * Returns the value of DateString in its original form.
+ * @returns {String} The Date as a String.
+ */
+DateString.prototype.toString = function() {
+  return this.when;
+};
+
+/**
+ * Returns the JSON representation of the DateString object.
+ * @returns {String} A JSON string.
+ */
+DateString.prototype.toJSON = function() {
+  return JSON.stringify({
+    when: this.when,
+  });
+};
+
+DateString.prototype.inspect = function(depth, options) {
+  return 'DateString ' + inspect({
+    when: this.when,
+    _date: this._date,
+  });
+};
--- a/test/date-string.test.js
+++ b/test/date-string.test.js
@@ -1,0 +1,81 @@
+// Copyright IBM Corp. 2014,2016. All Rights Reserved.
+// Node module: loopback-datasource-juggler
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+/* global describe,it */
+/* jshint expr:true */
+
+'use strict';
+
+require('should');
+
+var DateString = require('../lib/date-string');
+var fmt = require('util').format;
+var inspect = require('util').inspect;
+var os = require('os');
+
+describe('DateString', function() {
+  describe('constructor', function() {
+    it('should support a valid date string', function() {
+      var theDate = '2015-01-01';
+      var date = new DateString(theDate);
+      date.should.not.eql(null);
+      date.when.should.eql(theDate);
+      date.toString().should.eql(theDate);
+    });
+
+    testValidInput('should allow date with time', '2015-01-01 02:00:00');
+    testValidInput('should allow full UTC datetime', '2015-06-30T20:00:00.000Z');
+    testValidInput('should allow date with UTC offset', '2015-01-01 20:00:00 GMT-5');
+
+    testInvalidInput('should throw on non-date string', 'notadate', 'Invalid date');
+    testInvalidInput('should throw on incorrect date-like value',
+      '2015-01-01 25:00:00', 'Invalid date');
+    testInvalidInput('should throw on non-string input', 20150101,
+      'Input must be a string');
+    testInvalidInput('should throw on null input', null, 'Input must be a string');
+
+    it('should update internal date on set', function() {
+      var date = new DateString('2015-01-01');
+      date.when = '2016-01-01';
+      date.when.should.eql('2016-01-01');
+      var d = new Date('2016-01-01');
+      // The internal date representation should also be updated!
+      date._date.toString().should.eql(d.toString());
+    });
+    it('should return custom inspect output', function() {
+      var date = new DateString('2015-01-01');
+      var result = inspect(date);
+      result.should.not.eql(null);
+      result.should.eql(fmt('DateString ' + inspect({
+        when: date.when,
+        _date: date._date,
+      })));
+    });
+
+    it('should return JSON output', function() {
+      var date = new DateString('2015-01-01');
+      var result = date.toJSON();
+      result.should.eql(JSON.stringify({when: date.when}));
+    });
+
+    function testValidInput(msg, val) {
+      it(msg, function() {
+        var theDate = new DateString(val);
+        theDate.when.should.eql(val);
+        var d = new Date(val);
+        theDate._date.toString().should.eql(d.toString());
+      });
+    }
+
+    function testInvalidInput(msg, val, err) {
+      it(msg, () => {
+        var fn = () => {
+          var theDate = new DateString(val);
+        };
+        fn.should.throw(err);
+      });
+    }
+  });
+});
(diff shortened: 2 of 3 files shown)


kjdelisle:
I know there's been quite a bit of spitballing regarding this MySQL date issue, but it got me thinking about the idea of "DATE" type fields in databases, and how there's easily an argument for not wanting timezones to interfere with their use. If you were attempting to insert a birthday into a database, you wouldn't want the timezone offset to push that day forwards or backwards, right?

This change would give users the ability to fall back on a string representation that still acts as a Date, and is still validated by the JS Date constructor, but without being converted to UTC. It would be on a per-property basis.

(I know, I know, _another config flag_ 🤢 )

lib/model-builder.js
```diff
@@ -526,10 +527,13 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
 };
 
 // DataType for Date
-function DateType(arg) {
+function DateType(arg, props) {
```
ssh24: What was the reasoning behind have this function both in `dao` and `modelBuilder`?

kjdelisle:
Still a work in progress.

lib/date-string.js
```diff
@@ -0,0 +1,39 @@
...
+function DateString(value) {
+  if (!(this instanceof DateString)) {
+    return new DateString(value);
+  }
+
+  if (typeof(value) !== 'string') {
+    throw new Error('Input must be a string');
+  }
+  // Preserve the reference for return via .toDate
+  this.date = new Date(value);
```
bajtos: This allows users to include the time in the value too, e.g. `2015-01-01T10:00:00.0Z`.

If we want to make this new type for dates only, then I am proposing to accept a single format `yyyy-MM-dd` only.

lib/date-string.js
```diff
@@ -0,0 +1,39 @@
...
+    throw new Error('Invalid date');
+  }
+  this.value = value;
+};
+
+/**
+ * Return the equivalent Date object for this DateString.
+ * @returns {Date} A JavaScript Date object
+ */
+DateString.prototype.toDate = function() {
```
bajtos: IMO, this function needs to be parameterized by the timezone we want to get the value in. Otherwise you can end up with the same problem when somebody calls `toLocaleString` as you are trying to fix: _if you were attempting to insert a birthday into a database, you wouldn't want the timezone offset to push that day forwards or backwards, right?_

```
$ node
> new Date('2017-04-27').toLocaleString(undefined, {timeZone:'America/Toronto'})
'4/26/2017, 8:00:00 PM'
```

The snippet above shows what happens to value `2017-04-27` when it's displayed via `toLocaleString()` by a computer located in Toronto (think of  `new DateString('2017-04-27').toDate().toLocaleString()`). Because my timezone is `Europe/Prague`, I had to explicitly tell `toLocaleString` to use a different one.

I think it may be better to not convert `DateTime` values to `Date` at all and let each connector to decide how to represent a date value.

bajtos:
@kjdelisle While I think it makes sense to have a special type representing values containing a date only, and then let each database connector to store them using a database-specific format, I personally don't think it's a fix for https://github.com/strongloop/loopback-connector-mysql/issues/149.

bajtos:
BTW you should probably add at least `toJSON()` method to customize the way how a `DateTime` value is serialized to JSON payload (think of HTTP responses).

You may want to consider adding `inspect()` and `toString()` too. The former is used by `util.inspect` used by `console.log`.

lib/date-string.js
kjdelisle: I was calling it DateString to reflect that it's like a JS Date (which is actually a Date and Time), but in string form.

It's mostly meant to allow the passing of unmodified date/time strings to an underlying driver, which in MySQL's case just happens to be helpful for DATE.

lib/date-string.js
kjdelisle: If we avoid parsing the string into a Date object (at least internal to this structure) we then get saddled with the responsibility of validating the date string ourselves, don't we? I mostly added the .toDate function so that if anyone wants to get at the underlying date object, they can do so. In hindsight, it's probably an unnecessary gesture, since they can just say `{instance}.date` to reference it.

kjdelisle:
>While I think it makes sense to have a special type representing values containing a date only, and then let each database connector to store them using a database-specific format, I personally don't think it's a fix for strongloop/loopback-connector-mysql#149.

It's not meant to fix the issue entirely by itself; it allows bypassing the Date issue on the forward pass. 
[This PR](https://github.com/strongloop/loopback-connector-mysql/pull/271) also helps with handling the problem on the reverse pass.

lib/date-string.js
```diff
@@ -0,0 +1,43 @@
...
+ * Returns the value of DateString in its original form.
+ * @returns {String} The Date as a String.
+ */
+DateString.prototype.toString = function() {
+  return this.value;
+};
+
+DateString.prototype.toJSON = function() {
+  return this.value;
+};
```
kjdelisle: @bajtos Not sure about the best way to handle this currently. If I don't override this function with the implementation you see here, then the values that get passed along the chain end up feeding a stringified JSON object to the datasource layer (which explodes fantastically). This override does the trick, but it means no one can actually get the JSON representation of this object with this method.

kjdelisle: It feels hacky and I don't like it. :(

bbito:
Is the anticipated use case for `DateString` any type of Date - i.e. for MySQL `DATE`, `DATETIME` & `TIMESTAMP` or just `DATE` types?

kjdelisle:
@bbito I'm hoping that this will be viable for any sort of datetime format that you don't want modified by LoopBack/JavaScript. It will still carry the backing Date object underneath if you really want the Date object that corresponds with it, but the default behaviour will just be validating that the string was any sort of valid date(time).

bbito:
@kjdelisle 

> viable for any sort of datetime format

That's tough because the timeless `DATE` is so different from the time-bearing types. What I get in my test app after pulling this branch is a Midnight UTC representation of the MySQL `DATE` which I think is the only predictable js Date version of a timeless `DATE`, that really facilitates server-side comparisons of other `DATE` types, but it doesn't facilitate comparisons betwixt `DATE` and time-bearing Dates. I'm concerned with how that documentation story is told...

BTW, I pulled last night and I'm able to `GET /DateTimeTables/{id}`, but a `GET /DateTimeTables/` crashes the app and a `POST` hands the whole `DateString` object to mysqljs/mysql and lands "0000-00-00" in MySQL, here's the SQL value out of mysqljs/mysql:
`'{\\"date\\":\\"2017-04-18T00:00:00.000Z\\",\\"value\\":\\"2017-04-18\\"}\',`
- I know WIP, but I thought it may be useful info for you...

kjdelisle:
>That's tough because the timeless DATE is so different from the time-bearing types. What I get in my test app after pulling this branch is a Midnight UTC representation of the MySQL DATE which I think is the only predictable js Date version of a timeless DATE, that really facilitates server-side comparisons of other DATE types, but it doesn't facilitate comparisons betwixt DATE and time-bearing Dates. I'm concerned with how that documentation story is told...

With the current iteration that I've pushed onto the PR branch, you could still compare the internal representations of the DateString objects if you really need to see a UTC-based comparison.
```
var d1 = new DateString('2016-01-01');
var d2 = new DateString('2016-01-01T16:00:00.000Z');
d1.toString();
// '2016-01-01'
d2.toString();
// '2016-01-01T16:00:00.000Z'
d1._date.toString();
// 'Thu Dec 31 2015 19:00:00 GMT-0500 (EST)'
d2._date.toString();
// 'Fri Jan 01 2016 11:00:00 GMT-0500 (EST)'
```
In this case, `d1._date` is still the result of running `new Date('2016-01-01')` along with all of the behaviour you would expect from a JS Date. But if you're only concerned with getting your date (or datetime) to the database through LoopBack, and the only validation you want is "Is this string a date?" then this should handle it for you.

kjdelisle:
Currently struggling with this warning on app start, though:
`Swagger: skipping unknown type "DateString".`
Need to find out where we handle that.

bbito:
POST is working (sending the String) after https://github.com/strongloop/loopback-datasource-juggler/pull/1356/commits/5650d82519f108bbc07e7929f8b98eb4ca747e13, but GET without `id` filter still crashes

bbito:
@kjdelisle it is Zero dates in MySQL that are crashing GET they are currently transformed to `null` in connector at: https://github.com/bbito/loopback-connector-mysql/blob/06d610a3cfc8f8d9f6046b2ddd9ed8b33e971d30/lib/mysql.js#L412
Checking that clued me in that when I rebuilt my hosed rebase for #271 I skipped a commit that would have made that read:
```
        if (!val || val == '0000-00-00 00:00:00' || val == '0000-00-00') {
          val = null;
        }
```
I can push that, or if that Zero date handling will move, I'll leave it alone.

kjdelisle:
@bbito I think that will only apply to objects of type "Date"; this change will make them "DateString" instead.

bbito:
@kjdelisle Got it - this fixes my Zero date GET crashes, should I push it?
```
      case 'Date':
      case 'DateString':

        // MySQL allows, unless NO_ZERO_DATE is set, dummy date/time entries
        // new Date() will return Invalid Date for those, so we need to handle
        // those separate.
        if (!val || val == '0000-00-00 00:00:00' || val == '0000-00-00') {
          val = null;
        }
```

kjdelisle:
@bbito Go for it

bbito:
Pushed https://github.com/strongloop/loopback-connector-mysql/pull/271/commits/883e61f26576be1f40503eb57e09923351a0af8f

test/date-string.test.js
```diff
@@ -0,0 +1,72 @@
...
+      // The internal date representation should also be updated!
+      date._date.toString().should.eql(d.toString());
+    });
+
+    it('should allow date with time', function() {
+      var theDate = '2015-01-01 02:00:00';
+      var date = new DateString(theDate);
+      date.value.should.eql(theDate);
+      var d = new Date(theDate);
+      date._date.toString().should.eql(d.toString());
```
jannyHou: just curious: is there any difference if we compare with `getTime()` instead of `toString()` here?

kjdelisle: `getTime()` gives you the time in milliseconds; `toString()` is the full datetime stamp that a Date object will generate.

bajtos:
> was calling it DateString to reflect that it's like a JS Date (which is actually a Date and Time), but in string form.

In that case let's call it DateTimeString please. I am pretty sure I won't be the only person confused by the name `DateString`, thinking it contains only the date part.

lib/date-string.js
bajtos: I don't understand. Can you show some examples of what do you mean?

Here is my expectation:

```js
expect(JSON.stringify({ when: new DateString('2017-04-28') }))
  .to.equal('{"when": "2017-04-28T00:00:00Z"}');
```

bajtos: Please add some unit-tests to cover `.toJSON` and `.toString` methods.

kjdelisle:
@bajtos 
>In that case let's call it DateTimeString please. I am pretty sure I won't be the only person confused by the name DateString, thinking it contains only the date part.

I'm not sure why we would call it DateTime string if it's just the String representation of a JS Date object (literally, Date-String). Originally, I was going to call it DateAsString, but that seemed wordy and clumsy.

bbito:
I like DateAsString!
I feel primary use case is for providing reassurance that timeless DATE types will not go through undocumented transforms into time-bearing types that can force it across a day boundary. I feel DateAsString conveys that!

kjdelisle:
@slnode test please

kjdelisle:
@slnode test please

crandmck:
@kjdelisle To add JSdocs (apidocs.strongloop.com), you need to add `lib/date-string.js` to `docs.json`.  

Then just be aware that the API doc will not be published until this module is published to npm.

PR #1360, opened 2017-04-28
datatype.test: use predefined date

kjdelisle:
Some databases are storing the value in UTC (not ISO) format,
and truncating the values. Even though this is technically
unavoidable, it does fail strict string equality.


kjdelisle:
@slnode test please

Issue #103, opened 2014-04-22
GeoPoint property validation throw AssertionError


labkode:
When we have a schema like this

<pre>
{
    "location": {
        type:"GeoPoint",
        required: true
    }
}
</pre>

and we post a value for location which is not a GeoPoint representation of lat and lgn, like this

<pre>
{
    "age": "this is not a lat and lng"
}
</pre>

Instead of returning a "user friendly" like other validations, something  like this

<pre>
{
  "error": {
    "name": "ValidationError",
    "status": 422,
    "message": "The Model instance is not valid. See `details` property of the error object for more info.",
    "statusCode": 422,
    "details": {
      "context": "cancha",
      "codes": {
        "age": [
          "presence"
        ]
      },
      "messages": {
        "age": [
          "can't be blank"
        ]
      }
    },
    "stack": "ValidationError: The Model instance is not valid... "
  }
}
</pre>

Instead it throws an AssertionError killing the Node.js process 

<pre>
AssertionError: lat must be a number when creating a GeoPoint
    at new GeoPoint (/home/foo/bar/changa/node_modules/loopback-datasource-juggler/lib/geo.js:134:3)
</pre>

This force us to do the validation manually, thing  I don't know how to do. I tried with this without success

<pre>
try{
    Cancha.create({
        "location": "wrong geopoint"
    }, function(err, cancha){
        if(err)
            return console.log(err);
        return console.log(cancha.isValid());
    })
}catch(ev){
    console.log(ev);
}
</pre>

superkhau:
Please create a test project on Github to reproduce the issue as per https://github.com/strongloop/loopback/wiki/Issues#bug-report

loay:
Closing due to inactivity. If you are still running into problems, feel free to leave a comment and I will reopen the issue.
Also, it looks like few solutions have been posted. 
Thanks.

daamsie:
Yeah, this still needs a solution. I just ran into it with a latitude being entered that was over 90. Actually, just someone accidentally putting lat/lng the wrong way around. This should really be handled as a validation error inside the model type. 

The above are not solutions - they are workarounds, mostly avoiding GeoPoint altogether. It seems a shame to have a built in type like this that can't handle validation errors cleanly.

peek4y:
@daamsie Exactly!

loay:
re-opening issue

ron137:
It is happen to me too, when I'm trying to enter wrong GeoPoint.
[...]

PR #1362, opened 2017-05-01
Support include rework for C* connector

Setogit:
#### Related issues

- https://github.com/strongloop/loopback-connector-cassandra/issues/8

### Checklist

### DIFF
--- a/test/include.test.js
+++ b/test/include.test.js
@@ -416,7 +416,8 @@ describe('include', function() {
               where: {
                 and: [
                   {id: createdPosts[0].id},
-                  {userId: createdPosts[0].userId},
+                  // Remove the duplicate userId to avoid Cassandra failure
+                  // {userId: createdPosts[0].userId},
                   {title: 'Post A'},
                 ],
               },
@@ -689,7 +690,8 @@ describe('include', function() {
               where: {
                 and: [
                   {id: createdPosts[0].id},
-                  {userId: createdPosts[0].userId},
+                  // Remove the duplicate userId to avoid Cassandra failure
+                  // {userId: createdPosts[0].userId},
                   {title: createdPosts[0].title},
                 ],
               },
--- a/test/include.test.js
+++ b/test/include.test.js
@@ -5,14 +5,20 @@
 
 'use strict';
 
-/* global getSchema:false */
-var should = require('./init.js');
-var async = require('async');
+/* global getSchema:false, connectorCapabilities:false */
 var assert = require('assert');
+var async = require('async');
+var bdd = require('./helpers/bdd-if');
+var should = require('./init.js');
 
 var DataSource = require('../').DataSource;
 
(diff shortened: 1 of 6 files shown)


test/include.test.js
```diff
@@ -887,6 +1270,7 @@ describe('include', function() {
                   },
                   include: 'parts',
                 }, function(err, result) {
+                  console.log('=================== H-4:', err);
```
raymondfeng: The `console.log` should be removed from tests.

test/include.test.js
```diff
@@ -53,14 +55,15 @@ describe('include', function() {
 
         u.__cachedRelations.should.have.property('posts');
         u.__cachedRelations.posts.forEach(function(p) {
-          p.userId.should.eql(u.id);
+          p.userId.toString().should.eql(u.id.toString());
```
raymondfeng: Why do we need this change?

test/include.test.js
```diff
@@ -40,7 +41,8 @@ describe('include', function() {
     });
   });
 
-  it('should fetch hasMany relation', function(done) {
+  bdd.itIf(connectorCapabilities.supportNonPrimaryKeyIN !== false,
```
raymondfeng: I think we can fix the inclusion implementation to allow `=` instead of `IN`.

raymondfeng:
We can fix https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/include.js#L255 to avoid `inq` if the length of FKs is 1.

test/include.test.js
Setogit: done.  and several more console.log...

test/include.test.js
Setogit: obsolete.  deleted.

Setogit: That's a left-over from trying the two designs back and force and there are many other similar cases.  At one point, made all the adhocSort === false cases to use string-based comparison with `should.be.equalOneOf` .

Did the following exercise: Changed `equalOneOf` to `oneOf` (`eql` equivalent) and removed all `toString()` except `toString() is required.

test/include.test.js
Setogit: like this?
```
      newFilter.where[fkName] = foreignKeys.length === 1 ? foreignKeys[0] : {
        inq: foreignKeys,
      };
```

Setogit: For the full fledged lusion implementation to allow `=` , an issue has been filed as:
https://waffle.io/strongloop-internal/scrum-asteroid/cards/59135f6e0fe1c5009245818f

Setogit:
Rebased/squasshed.  With inqLimit=1 change on the C* connector side, `include` cases are mostly passing.

Setogit:
The removal of redundant `userId`restriction by @raymondfeng looks good to me; that's the right fix in the test case, I agree.  The related loopbck-connector PR: https://github.com/strongloop/loopback-connector/pull/103 has a bad side effect and with the test case fix, the #103 PR is not needed.  Closed the #103 PR.

PR #1365, opened 2017-05-02
docs: add DateString definition

kjdelisle:
### Description
Add the DateString to the docs.json so that they end up in our API docs.


crandmck:
LGTM!

PR #1367, opened 2017-05-05
Fix assert as array does not guarantee order

0candy:
### Description
There is no guarantee on the order that the array values comes back so sort before asserting value.

#### Related issues

- None

### Checklist

### DIFF
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -467,7 +467,7 @@ describe('basic-querying', function() {
       return User.find({where: {seq: {inq: [0, 1, 100]}}})
         .then(function(result) {
           var seqsFound = result.map(function(r) { return r.seq; });
-          should(seqsFound).eql([0, 1]);
+          should(seqsFound.sort()).eql([0, 1]);
         });
     });
 


test/basic-querying.test.js
```diff
@@ -467,7 +467,7 @@ describe('basic-querying', function() {
       return User.find({where: {seq: {inq: [0, 1, 100]}}})
         .then(function(result) {
           var seqsFound = result.map(function(r) { return r.seq; });
-          should(seqsFound).eql([0, 1]);
+          should(seqsFound.sort()).containDeep([0, 1]);
```
bajtos: I think `containDeep` allows the array to have extra elements, e.g. `100` which should not be included. Is there any particular reason why `.eql` cannot be used now that the array items are sorted?

0candy: ah sorry, forgot to remove that from my testing from before.

kjdelisle:
@slnode test please

0candy:
Waiting for fixes to downstream CI for this PR to become green before merging.

0candy:
@slnode test please

PR #1377, opened 2017-05-16
Allow retrieval of _rev property

ssh24:
connect to https://github.com/strongloop/loopback-connector-cloudant/issues/55

Allow retrieval of `_rev` property for Cloudant connector. Make it a readonly prop, similar to `_id`.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -369,7 +369,14 @@ DataAccessObject.create = function(data, options, cb) {
             defineReadonlyProp(obj, _idName, id);
           }
           if (rev) {
-            obj._rev = rev;
+            // if cloudant, send out the `_rev` in `__data`
+            // and specify it to be a readonly prop
+            if (connector.name === 'cloudant') {
+              obj.__data._rev = rev;
+              defineReadonlyProp(obj, '_rev', rev);
+            }  else {
+              obj._rev = rev;
+            }
           }
           if (err) {
             return cb(err, obj);


ssh24:
@jannyHou It will only attach a `_rev` if a rev exists. I am afraid no other connector sends out a `_rev` value other than connector, or does it? :\

jannyHou:
@ssh24 connector sends what a model has as properties, for example:
model `Test`(attached to MySQL) has a property `rev`, and it's just a common property, here you inject a readonly property called `_rev` into data, and later on mysql connector will have it and complain the table `Test` doesn't have such a property

ssh24:
@jannyHou Why did we previously assign `_rev` value to obj outside the `__data` level? I understand your concern, but if it is not in `__data` level, the cloudant connector does not fetch it as a property. Interestingly, no other connector will ever fetch that value in the `__data` level. :\

ssh24:
@jannyHou Added some logic to check if the call is from cloudant, then only add the `_rev` property, otherwise do what it previously used to do.

jannyHou:
@ssh24 I think before we go further, a general question from me is, why do we want to modify data in juggler? 
Correct me if I am wrong, from my quick look, your change modifies data before juggler sends data to connector, user should define a property called `_rev` not `rev` in model, if we decide to define it as `rev`, then cloudant connector should do the convert, not juggler.

ssh24:
@jannyHou 
> Correct me if I am wrong, from my quick look, your change modifies data before juggler sends data to connector

That is not correct. My change adds back the `_rev` to the object returned upon create. Create, for some reason does not add it back to the `__data` level property which is passed onto the user. If this fix is not applied, the user does not get back `__rev` when doing a create operation.

jannyHou:
Sorry I misunderstood your change, thank you for correcting me!

> Create, for some reason does not add it back to the __data level property which is passed onto the user

hmm, I am interested in the reason, is it because loopback consider any property `_rev` as private property and hide it from the user? 
I am not familiar with this part and willing to dig more tomorrow(busy with polymorphic relation tonight), feel free to ignore my comments atm.

ssh24:
I have no idea why this was done in the first place. I guess it was because someone wanted to hide the `_rev` for cloudant and decided to do it globally for all connectors?

ssh24:
Closing this PR. Talked with @jannyHou and this patch is not needed.

PR #1378, opened 2017-05-17
Use correct data on replace callback

ssh24:
connected to https://github.com/strongloop/loopback-connector-cloudant/issues/55

Previously, it would just pass the old data. Need to pass the updated data because cloudant changes the `_rev` property on CRUD call.

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -3059,7 +3059,9 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
 
         function replaceCallback(err, data) {
           if (err) return cb(err);
-
+          if (typeof connector.generateContextData === 'function') {
+            context = connector.generateContextData(context, data);
+          }
           var ctx = {
             Model: Model,
             hookState: hookState,


lib/dao.js
```diff
@@ -3059,7 +3059,10 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
 
         function replaceCallback(err, data) {
           if (err) return cb(err);
-
+          // if connector is cloudant, add the `_rev` property back before passing it to the user
+          if (connector.name === 'cloudant') {
+            context.data._rev = data._rev;
```
b-admike: not sure what the difference is between `context.data` and `data`. Could there be logic somewhere that removes `._rev` property either at this or the connector layer?

ssh24: It does not remove anything. If you scroll few lines down, it uses the **old** data in a sense it sends out the __**old rev**__ value back to the user upon replace. So we should update the `context.data._rev` point to the new `_rev` value  before sending it out.

b-admike: Ok I get it now, and agree that we should update the `._rev`. However, it doesn't seem like we were even using the `data` returned in `replaceCallback(err, data)` at all before this change. Why don't we override the entire `context.data` with `data` since other keys besides the `._rev` might have changed?

ssh24: Other keys should not change. This on `replace` method and if the data the user passes on to be replaced with changes back from the driver then there are our connector does not do the right job. Only `_rev` of an existing instance will be updated upon `replace`/`update`.

jannyHou: agree with @b-admike I have been confused by this for long time: those CRUD functions return `data` that user passes in, instead of the real data returned from request to database....Not a block of this PR, just a general question :p

ssh24: I do not agree. If the user wants to replace {"name": "a", "age": 1} where `id : 3`, then the database will return back that `OK` I was able to replace the instance with the data you gave me and return the new data which should be the same as the data user initially passed.

In cloudant's case, we just need to replace the new `_rev` back or else the user gets back the old `_rev` value.

jannyHou: IMO database returns `ok` doesn't mean anything, for example, we can modify the data in connector, ideally we know what we are doing, we guarantee we don't pollute user's data, while, but if some code by accident add/delete something, database will still return ok(it doesn't care what your new data is, as long as the replacement operation executes successfully), and juggler still returns user passed in data, finally no place could catch the mistake...

Especially for cloudant(fault-tolerant database), only by checking the returned result we can know what data it actually stores :(

jannyHou:
@ssh24 If we have to do a connector specific check here, I would suggest we add a function in connector:
```javascript

Cloudant.prototype.generateContextData(context, data) {
   context.data._rev = data._rev;
   return context;
}

```

And in juggler call it: 
```javascript
if (typeof connector.generateContextData === 'function') {
  context = connector.generateContextData(context, data);
}

```

ssh24:
@jannyHou I like that idea. Will update the PR for the juggler part and have the `generateContextData` on the cloudant PR https://github.com/strongloop/loopback-connector-cloudant/pull/123

jannyHou:
@ssh24 cassandra and `postgresql` fails, could you double check? Other than that LGTM

ssh24:
@jannyHou Timeout error :P 

@slnode test please

jannyHou:
Hmm, wait could you also add a test case to verify it? There is a function you can use to create a mock connector: https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/mock-connectors.js

ssh24:
@jannyHou I am not sure how to do that. How would the mock connector test this functionality?

jannyHou:
@ssh24 Sorry, I wasn't aware that it will be that complicated to test that logic, I thought the we can just add `replaceById(id, newdata) { newdata._rev = '1'; return newdata;}` to the mock connector then define a model attached to it, call `model.replaceById` then make sure user get the data with `_rev` involved. While the test shouldn't need to override `dao.js` functions...

If it's so complicated then that's fine, ignore the test, we can still test the `getContextData` function in cloudant connector, don't have to waste time in juggler.

Have to say sorry again :( I don't know you spend so much effort on getting this done, if you feel my comments is not reasonable, feel free to refuse it, it's just a suggestion, not mandatory :(

ssh24:
@jannyHou No worries. I will revert that commit.

PR #1380, opened 2017-05-19
Return promise for batch create

raymondfeng:
### Description

The issue was discovered in loopback-next. When an array of values is passed into Model.create() without a callback function, it does NOT return a promise. The PR fixes the issue.

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -265,7 +265,7 @@ DataAccessObject.create = function(data, options, cb) {
 
   data = data || {};
   options = options || {};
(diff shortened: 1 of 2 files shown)


bajtos:
FWIW, it was intentional decision to not support batch + promise mode in `create()` - see https://github.com/strongloop/loopback/issues/2164#issuecomment-215645590.

Related issues - I'll add them to PR description:

 - https://github.com/strongloop/loopback/issues/2164
 - https://github.com/strongloop/loopback/issues/1275

PR #1383, opened 2017-05-23
hooks: add JSDoc for .trigger

kjdelisle:
### Description
Add JSDoc for the deprecated trigger function in Hookable.

#### Related issues
https://github.com/strongloop-internal/scrum-apex/issues/178

### DIFF
--- a/lib/hooks.js
+++ b/lib/hooks.js
@@ -36,7 +36,17 @@ Hookable.afterUpdate = null;
 Hookable.beforeDestroy = null;
 Hookable.afterDestroy = null;
 
-// TODO: Evaluate https://github.com/bnoguchi/hooks-js/
+/**
+ * @deprecated
+ * Setup a Model-based hook to trigger when the specified action occurs.
+ * The trigger is broken up into three segments: `beforeHook`, `work` and
+ * `afterHook`.
+ * @param {string} actionName The name of the action that triggers the hook.
+ * @param {Function} work The 2nd phase of the trigger.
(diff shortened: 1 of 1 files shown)


lib/hooks.js
```diff
@@ -36,7 +36,17 @@ Hookable.afterUpdate = null;
...
-// TODO: Evaluate https://github.com/bnoguchi/hooks-js/
+/**
+ * @deprecated
+ * Setup a Model-based hook to trigger when the specified action occurs.
+ * The trigger is broken up into three segments: `beforeHook`, `work` and
+ * `afterHook`.
+ * @param {string} actionName The name of the action that triggers the hook.
+ * @param {Function} work The 2nd phase of the trigger.
+ * @param {*} data The value(s) to provide to the 1st phase (`beforeHook`) call.
+ * @callback
```
ssh24: Do we need this if we have `@param {Function} callback` the next line?

kjdelisle: It's a special decorator that marks the function as a callback.
For humans, it's superfluous, but it tells parsers that the specific parameter at that index is a callback (so if the param was named `asdf`, it would still know that it's a callback).

jannyHou: +1, `@callback {Function} callback The callback function` is used in current juggler code, defined in this repo https://github.com/strongloop/strong-docs

b-admike: is `@param` implied if we do the above ^?

kjdelisle: Sorry, late reply.
`@callback {Function} callback` is invalid. Please see the JSDoc spec.

PR #1384, opened 2017-05-23
Fix count of properties

ssh24:
Please note there is an injected `id` property from User prototype. We do not care about that property since the `customer` model extends the `User` model and has its own **customerId**. 

connect to https://github.com/strongloop-internal/scrum-apex/issues/209

### DIFF
--- a/test/model-inheritance.test.js
+++ b/test/model-inheritance.test.js
@@ -174,9 +174,9 @@ describe('Model class inheritance', function() {
       customer.should.have.property('customerId', 'c01');
       customer.should.have.property('bio', undefined);
 
-        // The properties are defined at prototype level
+      // The properties are defined at prototype level
       assert.equal(Object.keys(customer).filter(function(k) {
-          // Remove internal properties
+        // Remove internal properties
         return k.indexOf('__') === -1;
       }).length, 0);
       var count = 0;
@@ -184,13 +184,15 @@ describe('Model class inheritance', function() {
         if (p.indexOf('__') === 0) {
           continue;
         }
-        if (typeof customer[p] !== 'function') {
+        // Please note there is an injected id from User prototype which is a function
+        // On node >= 7, this property does not show up
+        if (typeof customer[p] !== 'function' && p !== 'id') {
           count++;
         }
       }
-      assert.equal(count, 7); // Please note there is an injected id from User prototype
+      assert.equal(count, 6);
       assert.equal(Object.keys(customer.toObject()).filter(function(k) {
-          // Remove internal properties
+        // Remove internal properties
         return k.indexOf('__') === -1;
       }).length, 6);
 


ssh24:
@slnode test please

test/model-inheritance.test.js
```diff
@@ -174,23 +174,25 @@ describe('Model class inheritance', function() {
...
       }).length, 0);
       var count = 0;
       for (var p in customer) {
         if (p.indexOf('__') === 0) {
           continue;
         }
-        if (typeof customer[p] !== 'function') {
+        // Please note there is an injected id from User prototype which is a function
+        // On node >= 7, this property does not show up
+        if (typeof customer[p] !== 'function' && p !== 'id') {
```
b-admike: Hmm curious why the User model `id` prop is not dropped from the Customer model which extended it and has its own id prop. It's annoying for it to still be around if it is not used at all.

jannyHou: With the model inheritance PR checked in, wouldn't `customerId` override `id` as a primary key? I would expect the behavior in node7 is correct, since `id` shouldn't show up in `ChildModel`

ssh24: I do not why it does not override it properly. It looks like this whole test was written by @ebarault.

@ebarault Do you know why it still consists of the `id` property even after extending the model?

ebarault:
@ssh24 actually i did not write this test. I moved this existing test in this new file to group it with tests on model inheritance. 
The test was failing for me too before starting my last PR with node 7. On Miroslav's indications I moved on with node 6 and tests passed.

ssh24:
@ebarault Ah sorry. I looked at the blame for that file and saw your name. Thought you wrote it upon the PR for merging relations/ACL etc.

Anyways, this fix makes the test consistent across all node versions. We do not care about the `User` model id property as the `customer` model has its own id value so we can safely ignore that count.

@slnode test please

ssh24:
@slnode test please

ssh24:
@slnode test please

ssh24:
@slnode test please

test/model-inheritance.test.js
```diff
@@ -174,23 +174,25 @@ describe('Model class inheritance', function() {
...
         return k.indexOf('__') === -1;
       }).length, 0);
       var count = 0;
       for (var p in customer) {
         if (p.indexOf('__') === 0) {
           continue;
         }
-        if (typeof customer[p] !== 'function') {
+        // Please note there is an injected id from User prototype which is a function
+        // On node >= 7, this property does not show up
```
kjdelisle: :-1:
This is total nonsense; there has to be a sane root cause for why this is happening.

ssh24: The behaviour on node >=7 is the correct behaviour, where the default id from User model does not show up. The test case checks for the Customer model `id` (**customerId** in this case)

ssh24: Found a better way to objectify the model and iterating through the properties on it. PR: https://github.com/strongloop/loopback-datasource-juggler/pull/1392. Tests pass on both node >=7 and node < 7.

PR #1391, opened 2017-05-26
Remove node 7 support in 2.x

jannyHou:
### Description

IMO, the LB@2.x should support any node versions before the release of LB@3.x, but for new released versions like node 7, we may want to stop supporting them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "posttest": "npm run lint"
   },
   "engines": {
-    "node": ">= 0.6"
+    "node": ">= 0.6 <7"
   },
   "devDependencies": {
     "async-iterators": "^0.2.2",


jannyHou:
@slnode test please

jannyHou:
thank you @ssh24 and @bajtos , I chatted with Raymond on slack apex channel and he agrees, will involve Ritchie to make decision too.

jannyHou:
> juggler never worked on Node.js 7+, this patch is IMO not changing anything, just communicating the real state of affairs.

@bajtos juggler is actually testing node.js 7+, we fixed juggler@master branch, but juggler@2.x still fails. Backporting the PR could be a solution, while I am thinking that, there might be not that much sense/benefit to continue supporting node@7+  :)

rmg:
> there might be not that much sense/benefit to continue supporting node@7+

If anyone is _currently_ still on 0.10/0.12, then their most likely upgrade target is probably going to be node v8.x since it will be the next LTS line and will be supported after v4 and v6 are EOL'd.

Of course, if this was done 6 months ago the same could be said for supporting node@5+ in order to support v6.

crandmck:
> it would be great to capture this information in a prominent place in our LoopBack 2.x documentation.

Added https://github.com/strongloop-internal/scrum-loopback/issues/1539 to track this and related doc items.  Please add any pertinent info there.

jannyHou:
Let's do the change in PR https://github.com/strongloop/loopback-datasource-juggler/pull/1342
Closing this one.

PR #1393, opened 2017-05-29
Fix updateAttributes cb

ssh24:
Currently, the `updateAttributes` callback does not pass through the updated data from the database to juggler.

Allow pass through of the new data from the connector itself and update the context data to have the new `_rev` value got after the update.

Similar to the [replaceCallback in dao](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L3062)

connect to https://github.com/strongloop/loopback-connector-cloudant/issues/35


ssh24:
@slnode test please

ssh24:
@slnode test please

PR #1395, opened 2017-06-01
Fix mixin and validatable docs

ssh24:
Docs checklist:
 - [x] RelationMixin
 - [x] ObserverMixin
 - [x] TransactionMixin
 - [x] Validatable

connect to https://github.com/strongloop-internal/scrum-apex/issues/180

### DIFF
--- a/lib/observer.js
+++ b/lib/observer.js
@@ -10,7 +10,7 @@ var utils = require('./utils');
 module.exports = ObserverMixin;
 
 /**
- * ObserverMixin class.  Use to add observe/notifyObserversOf APIs to other
+ * ObserverMixin class. Use to add observe/notifyObserversOf APIs to other
  * classes.
  *
  * @class ObserverMixin
@@ -20,12 +20,30 @@ function ObserverMixin() {
 
 /**
  * Register an asynchronous observer for the given operation (event).
+ *
+ * Example:
+ *
+ * Registers a `before save` observer for a given model.
+ *
+ * ```javascript
+ * MyModel.observe('before save', function filterProperties(ctx, next) {
+  if (ctx.options && ctx.options.skipPropertyFilter) return next();
+  if (ctx.instance) {
+    FILTERED_PROPERTIES.forEach(function(p) {
+      ctx.instance.unsetAttribute(p);
+    });
+  } else {
+    FILTERED_PROPERTIES.forEach(function(p) {
+      delete ctx.data[p];
+    });
+  }
+  next();
+});
+ * ```
+ *
  * @param {String} operation The operation name.
  * @callback {function} listener The listener function. It will be invoked with
  * `this` set to the model constructor, e.g. `User`.
- * @param {Object} context Operation-specific context.
- * @param {function(Error=)} next The callback to call when the observer
- *   has finished.
  * @end
  */
 ObserverMixin.observe = function(operation, listener) {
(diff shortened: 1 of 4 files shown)


rmg:
Node 0.10/0.12 have been removed from the config on the Windows CI. Next time the branch is rebased those statuses should go away.

lib/observer.js
```diff
@@ -20,12 +20,28 @@ function ObserverMixin() {
 
 /**
  * Register an asynchronous observer for the given operation (event).
+ *
+ * Example:
+ *
```
b-admike: If you can add a description before the code, it'll help users understand what it does.

lib/observer.js
```diff
@@ -65,9 +100,28 @@ ObserverMixin.clearObservers = function(operation) {
...
+    if (err) return cb(err);
+    // user can specify the logic after the observers have been notified
+ });
+ * ```
+ *
  * @param {String|String[]} operation The operation name(s).
  * @param {Object} context Operation-specific context.
- * @param {function(Error=)} callback The callback to call when all observers
+ * @callback {function(Error=)} callback The callback to call when all observers
  *   has finished.
```
b-admike: `when all observers have finished`

lib/observer.js
```diff
@@ -123,19 +177,41 @@ ObserverMixin._notifyBaseObservers = function(operation, context, callback) {
 };
 
 /**
- * Run the given function with before/after observers. It's done in three serial
- * steps asynchronously:
+ * Run the given function with before/after observers.
+ *
+ * It's done in three serial steps asynchronously:
```
b-admike: nitpick: I suggest `serial asynchronous steps`

lib/observer.js
kjdelisle: Oh, cool. Didn't know you could specify the type of the parameter it expects like that.

lib/observer.js
ssh24: Fixed. The description is below.

lib/relations.js
```diff
@@ -50,25 +50,33 @@ function RelationMixin() {
...
  *  });
  *
  *   // Query chapters for the book
- *   book.chapters(function(err, chapters) {  // all chapters with bookId = book.id
+ *   book.chapters(function(err, chapters) {
+ *     // all chapters with bookId = book.id
  *     console.log(chapters);
  *   });
  *
+ *   // Query chapters for the back with a filter
```
b-admike: `for the book`

lib/relations.js
```diff
@@ -123,15 +131,20 @@ RelationMixin.hasMany = function hasMany(modelTo, params) {
  * @param {Class|String} modelTo Model object (or String name of model) to which you are creating the relationship.
  * @options {Object} params Configuration parameters; see below.
  * @property {String} as Name of the property in the referring model that corresponds to the foreign key field in the related model.
+ * @property {String} primaryKey Property name of primary key field.
  * @property {String} foreignKey Name of foreign key property.
- *
+ * @property {Object|Function} scope Explictly define additional scopes.
```
b-admike: Typo: `Explicitly`

lib/relations.js
```diff
@@ -155,28 +168,432 @@ RelationMixin.belongsTo = function(modelTo, params) {
...
+ * Examples:
+ * ```
+ * Supplier.hasOne(Account, {as: 'account', foreignKey: 'supplierId'});
+ * ```
+ *
+ * If the target model doesn’t have a foreign key property, LoopBack will add a property with the same name.
+ *
+ * The type of the property will be the same as the type of the target model’s id property.
+ *
+ * Please note the foreign key property is defined on the target model (for example, Account).
```
b-admike: I suggest `in this example, Account`

lib/relations.js
```diff
@@ -155,28 +168,432 @@ RelationMixin.belongsTo = function(modelTo, params) {
...
+ * Create a new account for the supplier. If there is already an account, an error will be reported.
+ * ```js
+ * supplier.account.create(data, function(err, account) {
+ *  ...
+ * });
+ * ```
+ *
+ * Find the supplier's account model.
+ * ```js
+ * supplier.account(function(err, account) {
```
b-admike: I reckon this is same as doing `supplier.account.find`?

ssh24: yes.

PR #1399, opened 2017-06-06
Add a bulkUpdate function

ssh24:
### Description

Add a static method for `bulkUpdate`. Connectors now can call `bulkUpdate` to perform a bulk update on data.

connect to https://github.com/strongloop/loopback-connector-cloudant/issues/35

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2737,6 +2737,108 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
   return cb.promise;
 };
 
+/**
+ * Perform a bulk update on a model instance.
+ *
+ * Supported in Cloudant connector only.
+ *
+ * Example:
+ *
+ *```js
+ * Employee.bulkUpdate([{_id: 1, _rev: 2-c23df06fcd7509dc8b70788c7af88a2c, managerId: 'x001'}], function(err, updates) {
+ *     ...
+ * });
+ * ```
+ *
+ * @param {Object} datas Datas to be updated
+ * @param {Object} [options] Options for bulk update
+ * @param {Function} cb Callback, called with (err, info)
+ */
+DataAccessObject.bulkUpdate = function(datas, options, cb) {
+  var connectionPromise = stillConnecting(this.getDataSource(), this, arguments);
+  if (connectionPromise) {
+    return connectionPromise;
+  }
+
+  assert(arguments.length >= 1, 'The datas argument is required');
+
+  if (cb === undefined && typeof options === 'function') {
+    cb = options;
+    options = {};
+  }
+
+  datas = datas || [];
+  options = options || {};
+  cb = cb || utils.createPromiseCallback();
+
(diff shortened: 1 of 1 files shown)


lib/dao.js
```diff
@@ -2737,6 +2737,108 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+    var err = new Error(g.f(
+      'The connector %s does not support {{bulkUpdate}} operation. This is not a bug in LoopBack. ' +
+      'Please contact the authors of the connector, preferably via GitHub issues.',
+      connector.name));
+    return cb(err);
+  }
+
+  var context = {
+    Model: Model,
+    options: options,
```
bajtos: This context is not honouring the fields provided by other DAO methods for `access` hook. See https://loopback.io/doc/en/lb3/Operation-hooks.html#access

lib/dao.js
```diff
@@ -2737,6 +2737,108 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+  };
+
+  Model.notifyObserversOf('access', context, function(err, ctx) {
+    if (err) return cb(err);
+    var context = {
+      Model: Model,
+      datas: datas,
+      options: options,
+    };
+    Model.notifyObserversOf('before save', context,
```
bajtos: This context is not honouring the fields provided by other DAO methods for access hook, see https://loopback.io/doc/en/lb3/Operation-hooks.html#before-save

Also "before save" is not a correct hook name considering that bulkUpdate performs delete operations too.

lib/dao.js
```diff
@@ -2737,6 +2737,108 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+      datas = removeUndefined(datas);
+    } catch (err) {
+      return process.nextTick(function() {
+        cb(err);
+      });
+    }
+    var context = {
+      Model: Model,
+      datas: datas,
+      options: options,
```
bajtos: See https://loopback.io/doc/en/lb3/Operation-hooks.html#persist

Again, we shouldn't be invoking "persist" for instances that are going to be deleted.

lib/dao.js
```diff
@@ -2737,6 +2737,108 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+
+  function bulkUpdateCallback(err, info) {
+    if (err) return cb(err);
+    var context = {
+      Model: Model,
+      datas: datas,
+      options: options,
+      info: info,
+    };
+    Model.notifyObserversOf('after save', context, function(err, ctx) {
```
bajtos: See https://loopback.io/doc/en/lb3/Operation-hooks.html#after-save

Also "after save" shouldn't be invoked for records that are going to be deleted.

ssh24:
Closing this PR. Found another way to call just the cloudant implementation of the `bulkUpdate` function by `db.connector.bulkUpdate`.

Issue #1381, opened 2017-05-22
Simple filter now broken in v3.7.0+


devpascoe:
I have a url...pretty simple and been working fine up until the 3.7.0 update
the url is a simple /items?filter[where][columnname]=20

Get the following error in console...
```
/loopback-datasource-juggler/lib/scope.js:119
          var idsB = qWhere[idKey].inq ? qWhere[idKey].inq : [qWhere[idKey]];
                                  ^

TypeError: Cannot read property 'inq' of undefined
    at smartMerge (/loopback-datasource-juggler/lib/scope.js:119:35)
```
If i revert to version 3.6.1 things work perfectly again. I'm running against a postgres db.
Anyone come across this?

dhmlau:
@devpascoe , I just tried to do a filter.  It seems to be working for me.  I'm using v3.9.0. 
Could you please update to see if you still encounter this issue?  Thanks.

robtassl:
Im having this issue as well. It seems to only effect my models with relations.

dhmlau:
I see. I was just using a very simple model.  Mind to share your models or application? 
https://github.com/strongloop/loopback/wiki/Reporting-issues#bug-report.

matteo-harpoon:
@dhmlau here's the full log. I tried with `v3.7.0` and `v3.9.0`, `v3.6.1` works fine.
```
     Uncaught TypeError: Cannot read property 'inq' of undefined
      at smartMerge (node_modules/loopback-datasource-juggler/lib/scope.js:119:35)
      at buildWhere (node_modules/loopback-datasource-juggler/lib/scope.js:129:29)
      at node_modules/loopback-datasource-juggler/lib/scope.js:133:32
      at node_modules/loopback-datasource-juggler/lib/dao.js:2093:9
      at node_modules/loopback-datasource-juggler/node_modules/async/dist/async.js:1012:9
      at node_modules/loopback-datasource-juggler/node_modules/async/dist/async.js:359:16
      at iteratorCallback (node_modules/loopback-datasource-juggler/node_modules/async/dist/async.js:935:13)
      at node_modules/loopback-datasource-juggler/node_modules/async/dist/async.js:843:16
      at node_modules/loopback-datasource-juggler/node_modules/async/dist/async.js:1009:13
      at buildResult (node_modules/loopback-datasource-juggler/lib/dao.js:2059:11)
      at node_modules/loopback-datasource-juggler/lib/dao.js:2073:13
[...]

PR #1401, opened 2017-06-07
Fix the case where qWhere[idKey] is null

Setogit:
### Description
Fix the case where qWhere[idKey] is null
#### Related issues

https://github.com/strongloop/loopback-datasource-juggler/issues/1381

### Checklist

### DIFF
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -819,6 +819,15 @@ describe('relations', function() {
             done();
           });
         });
+        it('returns empty result when filtering with wrong id key', function(done) {
+          var wrongWhereFilter = {where: {wrongIdKey: samplePatientId}};
+          physician.patients(wrongWhereFilter, function(err, ch) {
+            if (err) return done(err);
+            should.exist(ch);
+            ch.should.have.lengthOf(0);
+            done();
+          });
+        });
         it('returns patients where id in an array', function(done) {
           var idArr = [];
           var whereFilter;
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -130,7 +130,13 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
           return mergedWhere;
         };
         if (queryRelated.where !== undefined) {
-          queryRelated.where = buildWhere();
+          if (!queryRelated.where[IdKey]) {
+            // If it's null, don't bother calling smartMerge or model.find.
+            // Simply return an empty result to the client.
+            return cb(null, []);
+          } else {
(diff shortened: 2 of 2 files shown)


raymondfeng:
Is it possible to add a simple test case?

Setogit:
test case added

test/relations.test.js
```diff
@@ -819,6 +819,18 @@ describe('relations', function() {
             done();
           });
         });
+        it('error case: filtering not done with wrong id key', function(done) {
```
dhmlau: How about making the description consistent with the rest? something like:
`should return empty result when filtering with wrong id key`

test/relations.test.js
```diff
@@ -819,6 +819,18 @@ describe('relations', function() {
...
+          var wrongWhereFilter = {where: {wrongIdKey: samplePatientId}};
+          physician.patients(wrongWhereFilter, function(err, ch) {
+            if (err) return done(err);
+            should.exist(ch);
+            ch.should.have.lengthOf(3);
```
dhmlau: It seems to me that when nothing matches the filter condition, it should return empty result.  According to @jannyHou, juggler or some connectors act differently.  Adding her to review.

dhmlau:
/cc @jannyHou 
Per our discussion, pls see my review comment.  Thanks! 
>It seems to me that when nothing matches the filter condition, it should return empty result. According to @jannyHou, juggler or some connectors act differently. Adding her to review.

test/relations.test.js
jannyHou: >  According to @jannyHou, juggler or some connectors act differently.
 
Sorry ignore what I said, it seems at least memory and cloudant have same behaviour.

@dhmlau I give a quick try with both memory and cloudant connector:
- {where: {wrongKey: "wrong"}} - returns [ ]
- {wrongKey: "wrong"}  - returns all records (DANGEROUS IMO)(not related to this PR though)

So I am afraid the test case here is not correct. Even it's a relation api, juggler generates the filter to targetModel as 
`where: {and: [{id: {inq: [fkValue_array]}}, {wrongKey: 'somevalue'}]}`, which should not return any result.

Setogit:
@dhmlau  Thanks for the comment.  I re-examined the code before `smartMerge` was introduced and made the test case work with the old code.  Then, made the fix pass the test for all connectors specifically Cassandra. `smartMerge' was introduced only for Cassandra connector.

The end result is like you proposed: `returns empty result when filtering with wrong id key`  Changed the description of the test case accordingly.  Thanks, again.

PR #1413, opened 2017-06-21
Support for IBM Object Storage

hacksparrow:
### Description

Add support for `object-storage` connector.

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -348,6 +348,14 @@ DataSource.prototype.setup = function(name, settings) {
     if (settings.initialize) {
       connector = settings;
     } else if (settings.connector) {
+      // some connectors are aliases to loopback-component-storage
+      if (['object-storage'].indexOf(settings.connector) > -1) {
+        settings.provider = 'openstack';
+        settings.connector = 'loopback-component-storage';
+        settings.useServiceCatalog = true;
+        settings.useInternal = false;
+        settings.keystoneAuthVersion = 'v3';
+      }
       connector = settings.connector;
     } else if (settings.adapter) {
       connector = settings.adapter;
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "depd": "^1.0.0",
     "inflection": "^1.6.0",
     "lodash": "^4.17.4",
+    "loopback-component-storage": "^3.2.0",
     "loopback-connector": "^4.0.0",
     "minimatch": "^3.0.3",
     "qs": "^6.3.0",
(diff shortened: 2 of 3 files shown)


hacksparrow:
@bajtos @raymondfeng 

CI failing with "No identified problem".

And, what to do about "security/snyk" checks? Open a new PR, fix it in this PR?

package.json
```diff
@@ -45,6 +45,7 @@
     "depd": "^1.0.0",
     "inflection": "^1.6.0",
     "lodash": "^4.17.4",
+    "loopback-component-storage": "^3.2.0",
```
bajtos: This looks wrong to me. It should be the application who is deciding what version of loopback-component-storage it wants to use, not loopback-datasource-juggler.

Perhaps you should move this line to "devDependencies", to allow you to load the connector from your new test?

lib/datasource.js
```diff
@@ -348,6 +348,14 @@ DataSource.prototype.setup = function(name, settings) {
     if (settings.initialize) {
       connector = settings;
     } else if (settings.connector) {
+      // some connectors are aliases to loopback-component-storage
+      if (['object-storage'].indexOf(settings.connector) > -1) {
```
bajtos: I have mixed feelings about handling aliases inside juggler. I am concerned that such solution won't scale long term as more and more aliases are added. I am also worried that 3rd-party connectors maintained by our community cannot use this mechanism without waiting for us to review & approve such addition.

lib/datasource.js
```diff
@@ -348,6 +348,14 @@ DataSource.prototype.setup = function(name, settings) {
...
+        settings.provider = 'openstack';
+        settings.connector = 'loopback-component-storage';
+        settings.useServiceCatalog = true;
+        settings.useInternal = false;
+        settings.keystoneAuthVersion = 'v3';
```
bajtos: Should we allow `datasources.json` to override these settings? I.e. treat the values above only as "reasonable defaults" and allow the user/application to specify a different value of `useInternal` (or any other field) if they wish so.

package.json
hacksparrow: Makes sense.

lib/datasource.js
hacksparrow: Makes sense.

lib/datasource.js
hacksparrow: It definitely make sense to keep juggler as loosely coupled from the LoopBack ways, as possible.

The other option would be to implement this in `loopback`. Will it be acceptable there? Aliases have to be handled at some level, though.

hacksparrow:
This feature should be at a higher level, will open a PR on `loopback`, instead. Will keep this PR open for a while.

hacksparrow:
Made the changes suggested by @bajtos and opened PR at https://github.com/strongloop/loopback/pull/3457.

PR #1418, opened 2017-07-13
Bug fix "uniq" function

BlueRoboto:
Did not work properly, every element inside the array "a" is a bsonType and indexof fails.

### Description
The _uniq(a)_ function in _utils.js_ script presents a little bug. All the elements in the array are 
bsonType: `['_bsontype', 'id']` = `['ObjectID', <Buffer 59 52 2e d8 ab 84 77 0a a9 61 59 a9>]` and the follow control fails :

`
 if (a.indexOf(a[i]) === i) {
      uniqArray.push(a[i]);
    }
`

The control should be done _string_ type otherwise _if_ condition returns true at each check up and the array result isn't unique.

This implementation presents **another problem** (bigger in this case).
I do a simple test with this configuration: MongoDB v3.4.5, Loopback v3.8.0, loopback-connector-mongodb v3.2.0 and loopback-datasource-juggler v3.9.2
Two models A and B; **B -> hasMany -> A**.
Inside B I putted :
`  "scope": {
    "include": {
      "relation": "relationA"
    }
  }`
I putted one document in A and 257 documents in B and for each B-document I added a reference of A-document.
In a find query on B-collection the scope clause generates an anomaly when resolve _relationA_. In fact in this case find result generates two related instance of A in each B-document, but there should be only one instance of A in each B-document .
This anomaly vanishes when I use a _limit_ clause at 256 (_inqLimit_) documents or if I remove one document from db (max 256 documents).
If I triple document number (768 B-documents), the find result generate three related instances of A in each B-document.

In my opinion the problem is due to the function _uniq(a)_ that does not work properly and generates wrong results in _include.js_ script.
```
    async.each(listOfFKs, function(foreignKeys, done) {
      var newFilter = {};
      for (var f in filter) {
        newFilter[f] = filter[f];
      }
      if (filter.where) {
        newFilter.where = {};
        for (var w in filter.where) {
          newFilter.where[w] = filter.where[w];
        }
      }
      newFilter.where[fkName] = foreignKeys.length === 1 ? foreignKeys[0] : {
        inq: foreignKeys,
      };
      model.find(newFilter, options, function(err, results) {
        if (err) return done(err);
        items = items.concat(results);
        done();
      });
    }, function(err) {
      if (err) return cb(err);
      cb(null, items);
    });
  }
```
In fact, (line 259 in include.js) `items = items.concat(results);` generates wrong results because the _foreignKeys_ array hasn't unique elements.

On my branch I made a fix adding the following check to _uniq(a)_ function:
```
 a = a.map(element => {
    if (typeof element === 'object')
      return String(element);
    else return element;
  });
```
In this way, I convert all Objects (bsontype) to string type and indexOf function finally correctly works.

**Please analyze my solution or even consider the reported problem and find a suitable solution.**

Thanks.

#### Related issues

- None

### Checklist


slnode:

slnode:

slnode:

slnode:

loay:
Hi @BlueRoboto 
Please add test cases so we can catch any future regression. 
Also, please fix the linting problems .. You can run this command `./node_modules/.bin/eslint --fix .` that will either fix the linting problems or will point you to where the fix should occur .. 
Thanks.

package-lock.json
```diff
@@ -0,0 +1,1246 @@
+{
+  "name": "loopback-datasource-juggler",
```
ssh24: Please don't commit this file.

BlueRoboto: Ok, I will ignore this file in the next push.

lib/utils.js
```diff
@@ -565,12 +565,18 @@ function uniq(a) {
   if (!a) {
     return uniqArray;
   }
+  a = a.map(element => {
+    if (typeof element === 'object')
+      return String(element);
```
kjdelisle: Won't this return `[object Object]`?

BlueRoboto: If i remove String(element) i get bsonType: ['_bsontype', 'id']. In this way i get a simple id String.
I'll put the tests as soon as possible, meantime I have implemented a simple test on this repo: https://github.com/BlueRoboto/loopback-sandbox.

slnode:

slnode:

slnode:

Issue #1127, opened 2016-10-05
add getAsync to documentation and clarify others


ludicruz:
I only found the Promise getAsync function from reading the juggler tests, could this be added to the documentation?

also the documentation says customer orders takes a where clause but it should be a filter correct?

in the documentation

```
customer.orders([where], function(err, orders) {
  ...
});
```

should this be:

```
customer.orders([filter], function(err, orders) {
  ...
});
```

and for the Promise interface

```
customer.orders.getAsync([filter]).then(function(orders) {
  ...
});
```

???

found on this test:
https://github.com/strongloop/loopback-datasource-juggler/blob/2cb0d042e4936270540ee8fdfa04939430936b9c/test/relations.test.js#L197

the new documentation does not show anything about getAsync:
http://loopback.io/doc/en/lb2/HasMany-relations.html

nor does the old:
https://docs.strongloop.com/display/public/LB/HasMany+relations

Amir-61:
>also the documentation says customer orders takes a where clause but it should be a filter correct?

That's right; this is a documentation issue; Please see code [here](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/scope.js#L177-L205) which shows we support`filter`.


>the new documentation does not show anything about getAsync
>nor does the old

Yes you are right; it needs to be added in documentation.

@crandmck

crandmck:
I updated http://loopback.io/doc/en/lb2/HasMany-relations.html#methods-added-to-the-model to change it to `filter`.

Re `getAsync`: 
- Is this just for HasMany relations, or also for others?
- Is it only when using Promises?  In general, we haven't really documented the Promises API because it's incompletely implemented....

ludicruz:
> Is this just for HasMany relations, or also for others?

I have been using the getAsync consistently for a while now.
I looked through where I've been using it and it's being used with `BelongsTo` and `HasMany`
FWIW: the `HasMany` relationships I've been using are normal (whatever that means...), through, and polymorphic.

> Is it only when using Promises? In general, we haven't really documented the Promises API because it's incompletely implemented....

This is in a few others as well.

Hope this helps.. I would have been using this sooner if I saw it in the documentation! 😃

gunjpan:
@ludicruz : `getAsync()` was added in this commit: https://github.com/strongloop/loopback-datasource-juggler/commit/29eb3434c764903562e68f135826560649361700 

> getAsync(condition, cb)
[...]

PR #1420, opened 2017-07-14
Skip imcompatible tests

jannyHou:
connect to strongloop/loopback-connector-cloudant#33
related to https://github.com/strongloop/loopback-connector-cloudant/pull/159

PR #1424, opened 2017-07-14
Rename getAsync() methods to find() and get()

lehni:
### Description

The `getAsync()` is defined in 3 places: `BelongsTo`, `HasOne` and `HasMany`, which is implemented as a scope. Therefore, it is also present on all scopes.

I find its naming really strange though:

- It is the only function with `Async()` in its name
- Like all other functions on relationships (`create()`, `update()`, `count()`, etc), it accepts a callback function and returns a promise if none is provided.

I would like to suggest to rename it to `find()` for scopes and `HasMany` relationships to mirror `Model.find()`, and `get()` for `BelongsTo` and `BelongsTo` relationships, while keeping the `getAsync()` around for backward compatibility for now. 

@bajtos has pointed me to previous discussions around why `get()` wasn't deemed a good name before here: https://github.com/strongloop/loopback-datasource-juggler/issues/1127#issuecomment-314116414

But after looking more into `EmbedsMany`, I found out a few additional things that weren't considered back then:

`get()` on `EmbedsMany` is a reference to `findById()`, which takes a callback or returns a value synchronously in the absence of the callback. But in all other cases I have found, those versions of `findById()` return asynchronously, so the same problem that was pointed out when the name `get()` was proposed applies here as well, and apparently wasn't seen as a problem...

Here all the examples of methods I could find where `EmbedsMany` returns synchronously, while the others (`HasMany`, `HasManyThrough`, `ReferencesMany`, `DataAccessObject`, ...) return promises (if they define them at all):

`at()`
`exists()`
`updateById()`
`updateById()`
`destroyById()`

These can easily be found by searching for `// sync` in the source code. I don't think this has caused concern or confusion so far?

Based on this observation, I think that naming these methods simply `get()` is the best we can do, and the most aligned with how things are currently named and implemented, including this naming discrepancy that's hard to fix at this point. Except for scopes of course, where I think we should name it `find()` as mentioned above.

@bajtos asked me to create a PR and mention him in it, so here it is. Please review.

### Related Issues

- #1127

### Checklist

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -20,6 +20,7 @@ var idEquals = utils.idEquals;
 var ModelBaseClass = require('./model.js');
 var applyFilter = require('./connectors/memory').applyFilter;
 var ValidationError = require('./validations.js').ValidationError;
+var deprecated = require('depd')('loopback-datasource-juggler');
 var debug = require('debug')('loopback:relations');
 
 var RelationTypes = {
@@ -1338,7 +1339,11 @@ RelationDefinition.belongsTo = function(modelFrom, modelToRef, params) {
     get: function() {
       var relation = new BelongsTo(definition, this);
       var relationMethod = relation.related.bind(relation);
-      relationMethod.getAsync = relation.getAsync.bind(relation);
+      relationMethod.get = relation.get.bind(relation);
+      relationMethod.getAsync = function() {
+        deprecated(g.f('BelongsTo method "getAsync()" is deprecated, use "get()" instead.'));
+        return this.get.apply(this, arguments);
+      };
       relationMethod.update = relation.update.bind(relation);
       relationMethod.destroy = relation.destroy.bind(relation);
       if (!polymorphic) {
@@ -1580,9 +1585,8 @@ BelongsTo.prototype.related = function(condOrRefresh, options, cb) {
  * @param {Function} cb Callback of the form function (err, inst)
  * @returns {Promise | Undefined} returns promise if callback is omitted
  */
-BelongsTo.prototype.getAsync = function(options, cb) {
+BelongsTo.prototype.get = function(options, cb) {
   if (typeof options === 'function' && cb === undefined) {
-    // order.customer.getAsync(cb)
     cb = options;
     options = {};
   }
@@ -1712,7 +1716,11 @@ RelationDefinition.hasOne = function(modelFrom, modelToRef, params) {
     get: function() {
       var relation = new HasOne(definition, this);
       var relationMethod = relation.related.bind(relation);
-      relationMethod.getAsync = relation.getAsync.bind(relation);
+      relationMethod.get = relation.get.bind(relation);
+      relationMethod.getAsync = function() {
+        deprecated(g.f('HasOne method "getAsync()" is deprecated, use "get()" instead.'));
+        return this.get.apply(this, arguments);
+      };
       relationMethod.create = relation.create.bind(relation);
       relationMethod.build = relation.build.bind(relation);
       relationMethod.update = relation.update.bind(relation);
@@ -1991,15 +1999,13 @@ HasOne.prototype.related = function(condOrRefresh, options, cb) {
 
 /**
  * Define a Promise-based method for the hasOne relation itself
- * - order.customer.getAsync(cb): Load the target model instance asynchronously
+ * - order.customer.get(cb): Load the target model instance asynchronously
  *
  * @param {Function} cb Callback of the form function (err, inst)
  * @returns {Promise | Undefined} Returns promise if cb is omitted
  */
-
-HasOne.prototype.getAsync = function(options, cb) {
+HasOne.prototype.get = function(options, cb) {
   if (typeof options === 'function' && cb === undefined) {
-    // order.profile.getAsync(cb)
     cb = options;
     options = {};
   }
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -6,6 +6,7 @@
 
 var _ = require('lodash');
 var i8n = require('inflection');
+var g = require('strong-globalize')();
 var utils = require('./utils');
 var defineCachedRelations = utils.defineCachedRelations;
 var setScopeValuesFromWhere = utils.setScopeValuesFromWhere;
@@ -13,6 +14,7 @@ var mergeQuery = utils.mergeQuery;
 var DefaultModelBaseClass = require('./model.js');
 var collectTargetIds = utils.collectTargetIds;
 var idName = utils.idName;
+var deprecated = require('depd')('loopback-datasource-juggler');
 
 /**
  * Module exports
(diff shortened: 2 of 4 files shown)


slnode:

slnode:

slnode:

slnode:

lib/relation-definition.js
```diff
@@ -1338,7 +1338,8 @@ RelationDefinition.belongsTo = function(modelFrom, modelToRef, params) {
     get: function() {
       var relation = new BelongsTo(definition, this);
       var relationMethod = relation.related.bind(relation);
-      relationMethod.getAsync = relation.getAsync.bind(relation);
+      relationMethod.get = relation.get.bind(relation);
+      relationMethod.getAsync = relationMethod.get; // deprecated alias
```
bajtos: Can you add a deprecation warning please?

```js
// at the top of this file
const deprecated = require('depd')('loopback-datasource-juggler');

// here
relationMethod.getAsync = function() {
  deprecated(g.f('BelongsTo method "getAsync()" is deprecated, use "get()" instead.'));
  return this.get(...arguments);
}
```

lib/relation-definition.js
```diff
@@ -1580,9 +1581,10 @@ BelongsTo.prototype.related = function(condOrRefresh, options, cb) {
  * @param {Function} cb Callback of the form function (err, inst)
  * @returns {Promise | Undefined} returns promise if callback is omitted
  */
-BelongsTo.prototype.getAsync = function(options, cb) {
+BelongsTo.prototype.getAsync = // deprecated alias
```
bajtos: Is this method actually used anywhere? I believe `BelongsTo` class is internal, not a part of our public API.

lib/relation-definition.js
```diff
@@ -1712,7 +1714,8 @@ RelationDefinition.hasOne = function(modelFrom, modelToRef, params) {
     get: function() {
       var relation = new HasOne(definition, this);
       var relationMethod = relation.related.bind(relation);
-      relationMethod.getAsync = relation.getAsync.bind(relation);
+      relationMethod.get = relation.get.bind(relation);
+      relationMethod.getAsync = relationMethod.get; // deprecated alias
```
bajtos: Can you add a deprecation warning please?

lib/relation-definition.js
```diff
@@ -1991,15 +1994,15 @@ HasOne.prototype.related = function(condOrRefresh, options, cb) {
...
  * Define a Promise-based method for the hasOne relation itself
- * - order.customer.getAsync(cb): Load the target model instance asynchronously
+ * - order.customer.get(cb): Load the target model instance asynchronously
  *
  * @param {Function} cb Callback of the form function (err, inst)
  * @returns {Promise | Undefined} Returns promise if cb is omitted
  */
-
-HasOne.prototype.getAsync = function(options, cb) {
+HasOne.prototype.getAsync = // deprecated alias
```
bajtos: I think this alias is not needed because `HasOne` class is not part of the public API.

lib/scope.js
```diff
@@ -272,22 +272,23 @@ function defineScope(cls, targetClass, name, params, methods, options) {
...
-          // customer.orders.getAsync(condOrRefresh, cb);
+          // customer.orders.find(condOrRefresh, cb);
           cb = options;
           options = {};
         }
         options = options || {};
         return definition.related(self, f._scope, condOrRefresh, options, cb);
       };
 
+      f.getAsync = f.find; // deprecated alias
```
bajtos: Can you add a deprecation warning please?

bajtos:
@slnode ok to test

lib/relation-definition.js
```diff
@@ -1580,9 +1581,10 @@ BelongsTo.prototype.related = function(condOrRefresh, options, cb) {
  * @param {Function} cb Callback of the form function (err, inst)
  * @returns {Promise | Undefined} returns promise if callback is omitted
  */
-BelongsTo.prototype.getAsync = function(options, cb) {
+BelongsTo.prototype.getAsync = // deprecated alias
+BelongsTo.prototype.get = function(options, cb) {
   if (typeof options === 'function' && cb === undefined) {
-    // order.customer.getAsync(cb)
+    // order.customer.get(cb)
```
kjdelisle: Let's remove this comment as well, since it's of no particular use.

lib/scope.js
```diff
@@ -272,22 +272,23 @@ function defineScope(cls, targetClass, name, params, methods, options) {
...
         if (typeof condOrRefresh === 'function' &&
             options === undefined && cb === undefined) {
-          // customer.orders.getAsync(cb)
+          // customer.orders.find(cb)
           cb = condOrRefresh;
           options = {};
           condOrRefresh = {};
         } else if (typeof options === 'function' && cb === undefined) {
-          // customer.orders.getAsync(condOrRefresh, cb);
+          // customer.orders.find(condOrRefresh, cb);
```
kjdelisle: Remove unnecessary comment.

lib/scope.js
```diff
@@ -272,22 +272,23 @@ function defineScope(cls, targetClass, name, params, methods, options) {
         f._targetClass = rel && rel.modelTo && rel.modelTo.modelName || i8n.camelize(f._scope.collect);
       }
 
-      f.getAsync = function(condOrRefresh, options, cb) {
+      f.find = function(condOrRefresh, options, cb) {
         if (typeof condOrRefresh === 'function' &&
             options === undefined && cb === undefined) {
-          // customer.orders.getAsync(cb)
+          // customer.orders.find(cb)
```
kjdelisle: Remove unnecessary comment.

lib/relation-definition.js
lehni: Taking care of all this now, but I'll adhere to ES5 style JS, as in the rest of the file.

lehni:
@kjdelisle @bajtos I've addressed all requested changes now in e5d9e6bae99af6431e88dc55029b408bd3ac607f

I hope this works for you! All tests pass.

lib/scope.js
```diff
@@ -272,22 +274,26 @@ function defineScope(cls, targetClass, name, params, methods, options) {
...
         options = options || {};
         return definition.related(self, f._scope, condOrRefresh, options, cb);
       };
 
+      f.getAsync = function() {
+        deprecated(g.f('Scope method "getAsync()" is deprecated, use "find()" instead.'));
+        return this.find.apply(this, arguments);
+      };
+
+      f.getAsync = f.find; // deprecated alias
```
bajtos: I think you forgot to remove this line, it's overriding the version of `f.getAsync` that's printing a deprecation warning.

lehni: Oops you're right! Fixing it right now.

lehni:
I've just pushed it with all changes squashed into one commit. I hope this works!

bajtos:
The CI reported a build failure in Cloudant connector. I checked with @kjdelisle, we think it's not related to the changes made here. To be more confident, I have rebased the patch on top of the latest master to bring in https://github.com/strongloop/loopback-datasource-juggler/pull/1420.

bajtos:
All CI is green now, landed.

Thank you @lehni for the contribution! 🎉

lehni:
@bajtos I just had a thought for a related change, and before I open a new issue I figured I'd ask here:

For symmetry reasons I think we should provide `set()` where we've added `get()` now.

Wouldn't it be much nicer now if I could do this:

```js
var relatedInstance = await instance.relation.get()
...
 instance.relation.set(relatedInstance)
```

I think design-wise, it was rather strange to make the relations themselves functions (and then to add more functions on them). It would probably have been a better choice if they were instances of some relation class that represent the relation and have these methods on them, don't you think? 

By adding `set()`, we could smooth the path towards a cleaner API here, perhaps in LB 4?

bajtos:
> For symmetry reasons I think we should provide `set()` where we've added `get()` now.

I like this idea 👍

Issue #1388, opened 2017-05-25
missing the option argument in find (scope.js)


dmellonch:
# Description/Steps to reproduce

In have a model Azienda with a has-many-through relation, like:

`
{
  "name": "Azienda",
  "relations": {
      "mansioni": {
          "type": "hasMany",
          "model": "Mansione",
          "through": "MansioneAzienda"
        }
    }
}
`

When i perform Azienda .mansioni({}, option,  cb), the find function over related model, missing the option argument (scope.js , row 139)

`relatedModel.find(queryRelated, cb);`
      
# Additional information

win32 x64 6.10.0

loopback@3.8.0
loopback-boot@2.24.0
loopback-component-explorer@4.2.0
loopback-connector-mongodb@3.1.0

dhmlau:
@dmellonch, thanks for catching that.  Would you like to submit a PR?

dmellonch:
ok @dhmlau, what should I do?

dhmlau:
@dmellonch, you can see the instructions in here to create a pull request: https://help.github.com/articles/creating-a-pull-request/

There are some more info about how to contributing to this repo: 
https://github.com/strongloop/loopback-datasource-juggler/blob/master/CONTRIBUTING.md

dhmlau:
@dmellonch, did you have a chance to see how you can submit the PR from the instructions I've provided above?  Thanks.

dmellonch:
Hi @dhmlau , i have already carried out the PR.

https://github.com/strongloop/loopback-datasource-juggler/pull/1394

I hope i followed the procedures correctly.

ssh24:
Please create a new PR since #1394 has had issues with rebasing and the tests added initially were broken. Read more [here](https://github.com/strongloop/loopback-datasource-juggler/pull/1394#issuecomment-314155237)

dhmlau:
connected to strongloop/loopback-datasource-juggler/pull/1426

PR #1426, opened 2017-07-18
Missing the option argument

dmellonch:
### Description
In scope.js, missing the option argument in find.relatedmodel
#### Related issues
- #1388

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -85,9 +85,10 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
     }
   }
   cb = cb || utils.createPromiseCallback();
-
-  if (!self.__cachedRelations || self.__cachedRelations[name] === undefined ||
-      actualRefresh) {
+  const refreshIsNeeded = !self.__cachedRelations ||
+    self.__cachedRelations[name] === undefined ||
+    actualRefresh;
+  if (refreshIsNeeded) {
     // It either doesn't hit the cache or refresh is required
     var params = mergeQuery(actualCond, scopeParams, {nestedInclude: true});
     var targetModel = this.targetModel(receiver);
@@ -96,8 +97,8 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
     // run another query to apply filter on relatedModel(targetModel)
     // see github.com/strongloop/loopback-datasource-juggler/issues/166
     var scopeOnRelatedModel = params.collect &&
-        params.include.scope !== null &&
-        typeof params.include.scope === 'object';
+      params.include.scope !== null &&
+      typeof params.include.scope === 'object';
     if (scopeOnRelatedModel) {
       var filter = params.include;
       // The filter applied on relatedModel
@@ -144,7 +145,7 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
           queryRelated.where[IdKey] = collectTargetIds(data, IdKey);
         }
 
-        relatedModel.find(queryRelated, cb);
+        relatedModel.find(queryRelated, options, cb);
       } else {
         cb(err, data);
       }
@@ -238,8 +239,10 @@ function defineScope(cls, targetClass, name, params, methods, options) {
             return self.__cachedRelations[name];
           }
         } else {
-          if (typeof condOrRefresh === 'function' &&
-              options === undefined && cb === undefined) {
+          const condOrRefreshIsCallBack = typeof condOrRefresh === 'function' &&
+            options === undefined &&
+            cb === undefined;
+          if (condOrRefreshIsCallBack) {
             // customer.orders(cb)
             cb = condOrRefresh;
             options = {};
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -950,7 +950,7 @@ describe('relations', function() {
         });
 
         function verify(physician) {
-          physician.patients({}, {realm: 'test'}, function(err, records) {
+          physician.patients({order: 'name ASC'}, {realm: 'test'}, function(err, records) {
             if (err) return done(err);
             should.exist(records);
             records.length.should.eql(2);
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -572,7 +572,7 @@ describe('relations', function() {
 
     before(function(done) {
       Physician = db.define('Physician', {name: String});
-      Patient = db.define('Patient', {name: String, age: Number,
+      Patient = db.define('Patient', {name: String, age: Number, realm: String,
         sequence: {type: Number, index: true}});
       Appointment = db.define('Appointment', {date: {type: Date,
         default: function() {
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

dmellonch:
@slnode test please

dmellonch:
@slnode test please

ssh24:
@dmellonch I rebased the branch for you correctly. Please **git pull** before committing further changes if needed.

@slnode test please

dmellonch:
ok @ssh24. thanks!

test/relations.test.js
```diff
@@ -918,6 +918,75 @@ describe('relations', function() {
...
+            });
+          });
+        });
+
+        function verify(physician) {
+          physician.patients({}, {realm: 'test'}, function(err, records) {
+            if (err) return done(err);
+            should.exist(records);
+            records.length.should.eql(2);
+            records[0].realm.should.eql('test');
```
ssh24: Can we check for `name` instead of the `realm` field?

test/relations.test.js
```diff
@@ -918,6 +918,75 @@ describe('relations', function() {
...
+                return verify(physician);
+              });
+          }).catch(done);
+
+        function verify(physician) {
+          return physician.patients({}, {realm: 'test'}, function(err, records) {
+            if (err) return done(err);
+            should.exist(records);
+            records.length.should.eql(2);
+            records[0].realm.should.eql('test');
```
ssh24: Ditto. Check for `name` rather than `realm`

dmellonch: Of course, i do it.  I make a **git pull**, but I dont see any change.  
So where I should make these changes?

ssh24: @dmellonch I have the added the extra assertions.

lib/scope.js
```diff
@@ -87,7 +87,7 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
   cb = cb || utils.createPromiseCallback();
 
   if (!self.__cachedRelations || self.__cachedRelations[name] === undefined ||
-      actualRefresh) {
+    actualRefresh) {
```
bajtos: This change makes it difficult to see where the condition ends and where the body of the `if` branch starts. It's also unrelated to fixing a missing option argument, I am proposing to revert it, together with other white-space only changes below.

Alternatively, if we want to fix this, then let's do it properly please - see http://loopback.io/doc/en/contrib/style-guide.html#indentation-of-multi-line-expressions-in-if

test/relations.test.js
```diff
@@ -918,6 +918,79 @@ describe('relations', function() {
...
+
+        function beforeAccessFn(ctx, next) {
+          ctx.query.where.realm = ctx.options.realm;
+          next();
+        }
+      });
+      it('should find be filtered from option', function(done) {
+        var id;
+        Physician.create(function(err, physician) {
+          physician.patients.create({name: 'a', realm: 'test'}, function(err, ch) {
```
bajtos: Unhandled `err`. Please add the following line as the first thing in the callback:

```js
if(err) return done(err);
```

test/relations.test.js
```diff
@@ -918,6 +918,79 @@ describe('relations', function() {
...
+            id = ch.id;
+            physician.patients.create({name: 'z', realm: 'test'}, function() {
```
bajtos: ditto, missing error handler

test/relations.test.js
```diff
@@ -918,6 +918,79 @@ describe('relations', function() {
...
+              physician.patients.create({name: 'c', realm: 'anotherRealm'}, function() {
```
bajtos: and here too

test/relations.test.js
bajtos: I think the assertion failure message produced by these checks will be rather unhelpful. I am proposing to use a different approach:

```js
should.exist(records);
const expected = ['a:test', 'z:test'];
const actual = records.map(function(r) { return r.name + ':' + r.realm; });
actual.should.eql(expected);
```

test/relations.test.js
```diff
@@ -918,6 +918,79 @@ describe('relations', function() {
...
+            records[0].realm.should.eql('test');
+            records[0].name.should.eql('a');
+            records[1].realm.should.eql('test');
+            records[1].name.should.eql('z');
+            done();
+          });
+        }
+      });
+
+      it('should find be filtered from option with promises', function(done) {
```
bajtos: IMO, there is no need to write this test twice (once for callback, second time for promises). The code fixing the issue is independent on the way how `find` was invoked. Let's keep only one of these tests.

dmellonch:
@ssh24 I added the suggested changes by @bajtos. 
But I still have problems with the merge. I dont understand why. Can you help me please.

bajtos:
@slnode ok to test

lib/scope.js
```diff
@@ -85,9 +85,10 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
     }
   }
   cb = cb || utils.createPromiseCallback();
-
-  if (!self.__cachedRelations || self.__cachedRelations[name] === undefined ||
-      actualRefresh) {
+  const _conditions = !self.__cachedRelations ||
```
bajtos: Please use a more descriptive name instead of `_conditions`, e.g. `refreshIsNeeded` or `shouldFetchFreshData`.

lib/scope.js
```diff
@@ -184,7 +185,7 @@ function defineScope(cls, targetClass, name, params, methods, options) {
   // are same
   if (cls === targetClass) {
     cls._scopeMeta[name] = params;
-  } else if (targetClass) {
+  }  else if (targetClass) {
```
bajtos: Two spaces after `}`, there should be only one. Just revert this line back please.

lib/scope.js
```diff
@@ -239,7 +240,7 @@ function defineScope(cls, targetClass, name, params, methods, options) {
           }
         } else {
           if (typeof condOrRefresh === 'function' &&
-              options === undefined && cb === undefined) {
+            options === undefined && cb === undefined) {
```
bajtos: Please revert this whitespace change or extract the condition into a variable.

ssh24:
@dmellonch Looks like you rebased it correctly 👍 

Could you please squash your commits into one with a reasonable message that describes the patch? Please follow the 50/72 rule:
```
- The first line of the commit message should not be longer than 50 characters long.
- The second line should be empty.
- Every subsequent lines should not be longer than 72 characters long.
```

dmellonch:
@slnode test please

lib/scope.js
```diff
@@ -144,7 +145,7 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
           queryRelated.where[IdKey] = collectTargetIds(data, IdKey);
         }
 
-        relatedModel.find(queryRelated, cb);
+        relatedModel.find(queryRelated, options, cb);
```
jannyHou: Thank you dmellonch 
There is a similar PR created for solving same issue, details please check 
https://github.com/strongloop/loopback-datasource-juggler/pull/1255#issuecomment-290157667

Briefly, I feel the `options` should be applied to either `relatedModel.find` or `targetModel.find` but **NOT** both, I prefer the changes made in this PR, just we should remove the `options` in [Line 109](https://github.com/strongloop/loopback-datasource-juggler/pull/1426/files#diff-c2a679b29102d5574291e884531cb8aeR109)

dmellonch: HI @jannyHou,
why we should remove the options in line 109? *.
Let's assume that I want to use the options in every **.observe('access', ..)  (every model)
This would no longer be possible.

bajtos: I agree that `options` should be passed to both `relatedModel.find` and `targetModel.find` methods, because we are using `options` to propagate context data like current access token, current (SQL) transaction, etc.

jannyHou: OK then let's keep it in both functions :)

dmellonch:
I do not understand the reason why the two tasks above fall.
Can someone help me?
Thx

ssh24:
@slnode test please

ssh24:
@dmellonch Could you please rebase your branch? After you are done rebasing, I can trigger the CI again.

kjdelisle:
Ugh. The cloudant test run looks like it's failing due to order of the array:
```
  1) cloudant imported features relations hasMany through find over related model with options should find be filtered from option:

      Uncaught AssertionError: expected Array [ 'z:test', 'a:test' ] to equal Array [ 'a:test', 'z:test' ] (at '0', A has 'z:test' and B has 'a:test')
      + expected - actual

       [
      +  "a:test"
         "z:test"
      -  "a:test"
       ]
      
      at Assertion.fail (node_modules/should/lib/assertion.js:180:17)
      at Assertion.prop.value (node_modules/should/lib/assertion.js:65:17)
      at node_modules/loopback-datasource-juggler/test/relations.test.js:959:27
      at node_modules/loopback-datasource-juggler/lib/dao.js:2097:9
```

test/relations.test.js
```diff
@@ -918,6 +918,51 @@ describe('relations', function() {
...
+              physician.patients.create({name: 'c', realm: 'anotherRealm'}, function(err) {
+                if (err) return done(err);
+                verify(physician);
+              });
+            });
+          });
+        });
+
+        function verify(physician) {
+          physician.patients({}, {realm: 'test'}, function(err, records) {
```
b-admike: Can we specify an order filter i.e.`{order: 'ASC'}` here so that the actual values will always be in ascending order? This will help fix the CI failures.

kjdelisle:
@bajtos Is this okay to merge?

ssh24:
PR landed. @dmellonch Thank you again for the contribution!

slnode:

slnode:

slnode:

slnode:

dmellonch:
Thanks to you for the experience!

PR #1427, opened 2017-07-18
Fix tests failing when overridden for cloudant.

virkt25:
Fixed short-circuiting of test cases preventing then from being run
standalone.

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -1870,7 +1870,7 @@ describe('manipulation', function() {
     });
 
     bdd.itIf(connectorCapabilities.supportStrictDelete !== false &&
-      connectorCapabilities.cloudantCompatible,
+      connectorCapabilities.cloudantCompatible !== false,
       'should allow delete(id) - ' + 'fail with error', function(done) {
         // Please recover when the issue above get fixed
         Person.settings.strictDelete = true;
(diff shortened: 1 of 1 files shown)


test/manipulation.test.js
```diff
@@ -1870,7 +1870,7 @@ describe('manipulation', function() {
     });
 
     bdd.itIf(connectorCapabilities.supportStrictDelete !== false &&
-      connectorCapabilities.cloudantCompatible,
+      connectorCapabilities.cloudantCompatible !== false,
```
ssh24: Should the operation be `||` instead?

test/manipulation.test.js
```diff
@@ -2214,7 +2214,7 @@ describe('manipulation', function() {
         });
 
     bdd.itIf(connectorCapabilities.updateWithoutId !== false &&
-      connectorCapabilities.cloudantCompatible,
+      connectorCapabilities.cloudantCompatible !== false,
```
ssh24: Ditto.

test/manipulation.test.js
jannyHou: it should be `&&` I think, otherwise the connector(not cloudant) which doesn't support strict delete will still run this test due to `connectorCapabilities.cloudantCompatible !== false` is `true`

ssh24: Got it.

virkt25: `&&` is used because if either `connectorCapabilities.supportStrictDelete` OR `connectorCapabilities.cloudantCompatible` is set to `false` the test case should be unsupported. Using `||` would mean both would have to be `false` for the test to be unsupported. They're unrelated compatibility flags.

Issue #111, opened 2014-05-14
Model errors when using relations with the MongoDB connector cause crashes


zackbloom:
I have relations setup with a model which includes a property similar to this:

```
versions: [
  {
    id:         String
  }
]
```

If the `versions` property is actually an object, rather than an array (due to an error in saving), loopback crashes when it attempts to resolve relations involving this model.

This is the error I get:

```
/Users/zackbloom/s/API/node_modules/mongodb/lib/mongodb/connection/base.js:245
        throw message;
              ^
Error: Items must be an array: [object Object]
  at new List (/Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/lib/list.js:25:11)
  at /Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/lib/model.js:178:39
  at Array.forEach (native)
  at Function.ModelClass.forEachProperty (/Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/lib/model-builder.js:271:51)
  at ModelConstructor.ModelBaseClass._initProperties (/Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/lib/model.js:164:8)
  at /Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/lib/dao.js:557:13
  at Array.forEach (native)
  at /Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/lib/dao.js:554:12
  at /Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/lib/include.js:59:11
  at /Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:113:21
  at /Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:24:16
  at /Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/lib/include.js:124:9
  at /Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:111:13
  at Array.forEach (native)
  at _each (/Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:32:24)
  at Object.async.each (/Users/zackbloom/s/API/node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:110:9)
[...]

PR #1428, opened 2017-07-18
WIP: Should not return error when property of type array is set to object …

mateuspontes:
Convert object to array on datajungler and not throw error with wrong type of data

### Description

If we create a model with object and we pass a object on a request, datajungler can't convert it and crashes the server.
The solution here is convert the object to array and return the keys.

#### Related issues

- #111

### DIFF
--- a/lib/list.js
+++ b/lib/list.js
@@ -32,9 +32,13 @@ function List(items, itemType, parent) {
 
   items = items || [];
   if (!Array.isArray(items)) {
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

MaximeTroiano:
This is still an issue.

PR #1431, opened 2017-07-18
Fix create list only if type of item is Array

mateuspontes:
Should not convert objects to a List

### Description

If we create a model with object and we pass a object on a request, datasource-juggler can't convert it and crashes the server.
The solution here is not convert the object to a List.

#### Related issues

- #111

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -397,39 +397,41 @@ describe('manipulation', function() {
         .catch(done);
     });
 
-    it('should save existing object', function(done) {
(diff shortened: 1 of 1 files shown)


slnode:

slnode:

slnode:

slnode:

slnode:

slnode:

slnode:

slnode:

PR #1433, opened 2017-07-20
Catch errors using cb

loay:
### Description
Catch uncaught errors using cb wrapping. 

#### Related issues
connect to https://github.com/strongloop/loopback/issues/1364

test cases here: https://github.com/strongloop/loopback/pull/3481

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -283,15 +283,19 @@ DataAccessObject.create = function(data, options, cb) {
   var obj;
   var idValue = getIdValue(this, data);
 
+  try {
   // if we come from save
-  if (data instanceof Model && !idValue) {
-    obj = data;
-  } else {
-    obj = new Model(data);
-  }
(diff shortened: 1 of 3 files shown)


bajtos:
> [cis-jenkins] downstream: loopback@2.x — Failed! (90967a9)

I think this failure is expected, because we changed the way how password validation errors are reported. However, we need to fix loopback@2.x ASAP to avoid further CI failures. Can you take care of that please?

bajtos:
> However, we need to fix loopback@2.x ASAP to avoid further CI failures. Can you take care of that please?

Ah, I think this problem will be addressed by https://github.com/strongloop/loopback/pull/3481, so we should be good.

loay:
@bajtos 
Yes I will do the same thing for Master, after I make sure that it works properly, as well as I will create an issue for all the other dao methods to ensure they all work properly. I will mention you when I create the issue. 

Thanks

PR #1435, opened 2017-07-24
Catch errors using cb

loay:
### Description

Catch uncaught errors using cb wrapping.
Forwardport  of https://github.com/strongloop/loopback-datasource-juggler/pull/1433
co-dependent of https://github.com/strongloop/loopback/pull/3511

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -42,6 +42,7 @@ describe('manipulation', function() {
   before(function setupStubUserModel(done) {
     StubUser = db.createModel('StubUser', {password: String}, {forceId: true});
     StubUser.setter.password = function(plain) {
+      if (plain.length === 0) throw new Error('password cannot be empty');
       var hashed = false;
       if (!plain) return;
       var pos = plain.indexOf('-');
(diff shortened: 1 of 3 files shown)


test/manipulation.test.js
```diff
@@ -542,6 +543,22 @@ describe('manipulation', function() {
...
 
+    it('should reject created StubUser with empty password', function(done) {
+      StubUser.create({email: 'b@example.com', password: ''}, function(err, createdUser) {
+        (err.message).should.match(/password cannot be empty/);
+        done();
+      });
+    });
+
+    it('should reject updated empty password with updateAttribute', function(done) {
+      StubUser.create({password: 'abc123'}, function(err, createdUser) {
```
dhmlau: in the `2.x` PR, the json contains the email as well.  Do we still need it? 
`StubUser.create({email: 'b@example.com', password: 'abc123'}, `

loay: it is not needed in 3.x, as the user instance is designed without `email` .. as well as it will generate an error since the email is not defined.

dhmlau:
Force merge, because there is circular dependencies on loopback ( strongloop/loopback#3511)

Issue #1355, opened 2017-04-25
Problem with validatesLengthOf and validatesFormatOf with numbers


cryptofuture:
Datamodule property:
 ``` json
 "CountAvailable": {
      "type": "number"
    },
```
`Product.validatesLengthOf('CountAvailable', {min: 1, max: 10000, message: {min: 'Should be >= 1', max: 'Maximum 10000 for one item'}});`
Allows pass 0 or 10001

```
'use strict';
module.exports = function(Product) {
var = CountValidate = /^([1-10000]|10000)$/;
Product.validatesFormatOf('CountAvailable', {with: CountValidate, message: 'Count should be > 1'});
```
Blocks everything  
I tried `/^([1-10000])$/;` and `/^[1-10000]$/;` also. As for `validatesLengthOf` looks like it not works for numbers at all.

cryptofuture:
`validateAsync` with custom function works for me. Seems like `validatesLengthOf` and `validatesFormatOf` accept string only.

PR #1438, opened 2017-07-26
fix: support numbers in validatesFormatOf

### LINKED ISSUE #1437: Unable to validate numbers using validatesFormatOf

Create a model with a property that is a `Number` type and try to use `validatesFormatOf` to validate the format of the number. For example, try to validate that the number is a positive integer using something like this:

```javascript
return Model.validatesFormatOf('someProperty', { with: /^\+?(0|[1-9]\d*)$/ });
```
This issue was also noted in https://github.com/strongloop/loopback-datasource-juggler/issues/1355 in the past and in that it is mentioned that validatesLengthOf and validatesFormatOf accept string only. The issue was closed without any action.

I can't see a good reason why this type of validation should only work with strings. I would think it a pretty common use case to want to validate the format of numbers as well as strings.

# Expected result

As per the documentation at http://apidocs.strongloop.com/loopback-datasource-juggler/#validatable-validatesformatof I would expect this validation to work regardless of wether the property is a string or a number.

mrfelton:
### Description
Fixes #1437.

This fixes validatesFormatOf so that it van validate numbers as well as strings by extending the `typeof` check to support numbers and casting the value to a string using `toString` prior to running the regex check.

#### Related issues
#1437 
#1355 

### Checklist

### DIFF
--- a/lib/validations.js
+++ b/lib/validations.js
@@ -361,8 +361,8 @@ function validateExclusion(attr, conf, err, options) {
 function validateFormat(attr, conf, err, options) {
   if (nullCheck.call(this, attr, conf, err)) return;
 
-  if (typeof this[attr] === 'string') {
-    if (!this[attr].match(conf['with'])) {
+  if (typeof this[attr] === 'string' || typeof this[attr] === 'number') {
+    if (!conf['with'].test(this[attr])) {
       err();
     }
   } else {
--- a/test/validations.test.js
+++ b/test/validations.test.js
@@ -78,40 +78,40 @@ describe('validations', function() {
         User.validatesPresenceOf('pendingPeriod', {if: 'createdByAdmin'});
         var user = new User;
         user.createdByAdmin = true;
-        user.isValid().should.be.false;
+        user.isValid().should.be.false();
         user.errors.pendingPeriod.should.eql(['can\'t be blank']);
         user.pendingPeriod = 1;
-        user.isValid().should.be.true;
+        user.isValid().should.be.true();
       });
 
       it('should skip when `if` is NOT fulfilled', function() {
         User.validatesPresenceOf('pendingPeriod', {if: 'createdByAdmin'});
         var user = new User;
         user.createdByAdmin = false;
-        user.isValid().should.be.true;
-        user.errors.should.be.false;
+        user.isValid().should.be.true();
+        user.errors.should.be.false();
         user.pendingPeriod = 1;
-        user.isValid().should.be.true;
+        user.isValid().should.be.true();
       });
 
       it('should NOT skip when `unless` is fulfilled', function() {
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

mrfelton:
Additionally, I noted that in the Validations test suite there were a number of incorrect calls to verify the value of `isValid`.

Some of the checks were written as:

```javascript
user.isValid().should.be.false
```

Which will always pass regardless of the value of `user.isValid()`. `should.be.false` is a method and needs to be called in order to function. The above should be written as:

```javascript
user.isValid().should.be.false()
```

I have included a second commit in this PR that fixes up the broken tests.

test/validations.test.js
```diff
@@ -616,31 +616,53 @@ describe('validations', function() {
   });
 
   describe('format', function() {
-    it('should validate format');
-    it('should overwrite default blank message with custom format message');
+    it('should validate the format of strings', function() {
+      User.validatesFormatOf('name', {with: /[a-z][A-Z]*$/});
+      var u = new User({name: 'valid name'});
+      u.isValid().should.be.true();
```
b-admike: It'd be great to have a negative test here where you give it an invalid value.

test/validations.test.js
```diff
@@ -616,31 +616,53 @@ describe('validations', function() {
...
+    });
 
-    it('should skip missing values when allowing null', function() {
-      User.validatesFormatOf('email', {with: /^\S+@\S+\.\S+$/, allowNull: true});
+    it('should validate the format of numbers', function() {
+      User.validatesFormatOf('age', {with: /^\d+$/});
+      var u = new User({age: 30});
```
b-admike: Negative test here would be good as well.

lib/validations.js
```diff
@@ -361,8 +361,8 @@ function validateExclusion(attr, conf, err, options) {
 function validateFormat(attr, conf, err, options) {
   if (nullCheck.call(this, attr, conf, err)) return;
 
-  if (typeof this[attr] === 'string') {
-    if (!this[attr].match(conf['with'])) {
+  if (typeof this[attr] === 'string' || typeof this[attr] === 'number') {
```
b-admike: I prefer we use `conf[with].test(this[attr])` here instead of coercing numbers to string and using `match`. [RegExp test method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test).

mrfelton:
@b-admike I've updated the regex and added some additional tests to verify negative results. Please can you re-review? Thanks.

b-admike:
@slnode test please

b-admike:
@slnode test please

b-admike:
PR landed. Thank you for your contribution @mrfelton.

PR #1361, opened 2017-04-28
Fix/2.x/nested properties

nVitius:
### Description
Currently, it is impossible to search on nested date fields due to an oversight in the `_coerce` step of the `dao`. We are not coercing recursively for nested properties. This patch introduces code to fix this issue as well as another similar problem specific to the memory connector. I have added tests for both the `dao` and `memory`. Removing the code changes in the files will prove that these tests fail with the current implementation.  
I have confirmed that this issue persists in `3.x`. After this is merged, I can port the fix over there as well.

#### Related issues
- https://github.com/strongloop/loopback-connector-mongodb/issues/176#issuecomment-176690613
- https://github.com/strongloop/loopback/issues/303#issuecomment-140473221

### Checklist
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

Issue #1434, opened 2017-07-22
TypeError: Cannot read property 'properties' of undefined using @2.55.1


yshing:
# Description/Steps to reproduce
Running loopback in docker node-apline 6.11

Working fine with 2.54.1. And throws following error when using 2.55.1.

# Expected result
Should be working without throwing error
```
TypeError: Cannot read property 'properties' of undefined
    at Function.DataAccessObject._coerce (/usr/src/app/node_modules/loopback-datasource-juggler/lib/dao.js:1615:60)
    at Function.DataAccessObject._normalize (/usr/src/app/node_modules/loopback-datasource-juggler/lib/dao.js:1511:8)
    at Function.find (/usr/src/app/node_modules/loopback-datasource-juggler/lib/dao.js:1855:10)
    at prepareForTokenInvalidation (/usr/src/app/node_modules/loopback/common/models/user.js:889:15)
    at notifySingleObserver (/usr/src/app/node_modules/loopback-datasource-juggler/lib/observer.js:104:22)
    at iterate (/usr/src/app/node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:181:13)
    at Object.async.eachSeries (/usr/src/app/node_modules/loopback-datasource-juggler/node_modules/async/lib/async.js:197:9)
    at doNotify (/usr/src/app/node_modules/loopback-datasource-juggler/lib/observer.js:101:11)
    at doNotify (/usr/src/app/node_modules/loopback-datasource-juggler/lib/observer.js:99:49)
    at doNotify (/usr/src/app/node_modules/loopback-datasource-juggler/lib/observer.js:99:49)
    at doNotify (/usr/src/app/node_modules/loopback-datasource-juggler/lib/observer.js:99:49)
    at Function.ObserverMixin._notifyBaseObservers (/usr/src/app/node_modules/loopback-datasource-juggler/lib/observer.js:122:5)
    at Function.ObserverMixin.notifyObserversOf (/usr/src/app/node_modules/loopback-datasource-juggler/lib/observer.js:97:8)
    at Function.ObserverMixin._notifyBaseObservers (/usr/src/app/node_modules/loopback-datasource-juggler/lib/observer.js:120:15)
    at Function.ObserverMixin.notifyObserversOf (/usr/src/app/node_modules/loopback-datasource-juggler/lib/observer.js:97:8)
[...]

PR #1441, opened 2017-07-27
castPropertyValue: throw on malformed types

kjdelisle:
### Description
Property types defined using empty arrays or objects is not documented behaviour: 
http://loopback.io/doc/en/lb2/LoopBack-types.html#overview
http://loopback.io/doc/en/lb2/LoopBack-types.html#array-types
http://loopback.io/doc/en/lb2/LoopBack-types.html#object-types

This unexpected usage, combined with changes made to the Memory connector in #1361 throws unhandled errors when attempting to parse results from the Memory datasource.

This PR fixes this by clarifying these errors, as well as providing additional test coverage.

### Checklist

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -350,9 +350,17 @@ Memory.prototype.fromDb = function(model, data) {
   if (!data) return null;
   data = deserialize(data);
   var props = this._models[model].properties;
-  for (var key in data) {
-    data[key] = this._castPropertyValue(key, data[key], props);
+  try {
+    for (var key in data) {
+      data[key] = this._castPropertyValue(key, data[key], props);
+    }
+  } catch (err) {
+    // Modify error message and re-throw
+    err.message = g.f('Unable to convert to instance of "%s": %s', model,
+      err);
+    throw err;
   }
+
   return data;
 };
 
@@ -370,7 +378,15 @@ Memory.prototype._castPropertyValue = function(prop, val, props) {
 
   var isArray = Array.isArray(props[prop].type);
   var propType = isArray ? props[prop].type[0] : props[prop].type;
-
+  if (!propType || !propType.name) {
+    if (isArray)
+      throw new Error(g.f(
+        'Property definition "%s" did not specify any sub-types!', prop));
+    else
+      throw new Error(g.f(
+        'Property definition "%s" was null or undefined!', prop
+      ));
+  }
   switch (propType.name) {
     case 'Date':
       val = new Date(val.toString().replace(/GMT.*$/, 'GMT'));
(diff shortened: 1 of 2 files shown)


lib/connectors/memory.js
```diff
@@ -383,7 +399,9 @@ Memory.prototype._castPropertyValue = function(prop, val, props) {
       break;
     case 'ModelConstructor':
       for (var subProp in val) {
-        val[subProp] = this._castPropertyValue(subProp, val[subProp], propType.definition.properties);
```
rmg: Is this checking for something different from the above checks? Is this suppressing cases that would normally trigger the checks in `_castPropertyValue`?

test/memory.test.js
```diff
@@ -889,6 +890,109 @@ describe('Memory connector', function() {
         });
     });
   });
+
+  describe.only('_castPropertyValue', function() {
```
b-admike: Remove `.only` :P

kjdelisle: Gah! This is why I prefer node-tap...

lib/connectors/memory.js
kjdelisle: It's just a recursive check, since the property could itself be a model in need of having its own properties populated.

kjdelisle: Also, this PR isn't meant to address #1434

test/memory.test.js
```diff
@@ -889,6 +890,109 @@ describe('Memory connector', function() {
...
+          type: undefined,
+        },
+        age: {
+          type: {},
+        },
+        hobbies: {
+          type: [],
+        },
+      };
+      it(testName, function() {
```
b-admike: I don't know if it is standard to have a cb in the `it` test description and call it at the end, but if it is, I suggest you do it as well `function(done)` -> and at the end `} finally { done(); }` block.

kjdelisle: It's only required if your test is asynchronous, so that it can "know" when the test is done.
Without one, mocha will handle any unhandled Errors as the reason for test failure. Since none of the code under test is async, this is normal.

kjdelisle: Side-note: another thing I like about tap, is the non-magic behaviour of using `t.end()` or `t.plan()` from the parameter provided to the test function `t.test('thing', function(t) { ... });`

rmg: Using `done` implies the test is async. These tests are all synchronous, aren't they?

b-admike: Ah right that makes sense.

test/memory.test.js
```diff
@@ -889,6 +890,109 @@ describe('Memory connector', function() {
...
+        try {
```
rmg: I _think_ the should.js way would be to replace this entire try/catch with a single assertion:
```js
(function() { mem._castPropertyValue(prop, val, props); }).should.throw(expected);
```

rmg: Probably formatted more like this, though :P
```js
it(testName, function() {
  (function() {
    mem._castPropertyValue(prop, val, props);
  }).should.throw(expected);
});
```

Issue #771, opened 2015-11-23
PersistedModel.updateAll() ignores validation


florianheinze:
I have this model:

``` javascript
{
  "name": "Foo",
  "strict": "validate",
  "options": { "validateUpsert": true },
  "properties": {
    "name": {
      "type": "String",
      "min": 5
    }
  }
}
```

If I create a new entry with a three character name, I get, as expected, a validation error. However if I create a new entry with a longer name and then update that name to three characters, there is no validation error and the change is written to the DB.

As far as I can tell the whole validation is not triggered when `updateAll()` is used.
##### Steps to reproduce:
- Clone repository https://github.com/onBill/loopback-sandbox-validation
- Run `node .`
- The essential code is found in `server/boot/foo.js`.

Amir-61:
Yes thats seems a bug to me; the validation is not implemented for updateAll/ update

dmitru:
Any updates on the issue?

Amir-61:
@dmitru IIRC `updateAll/update` do not have validation; would you like to submit a patch?

ahmed-abdulmoniem:
This is still not implemented?

Also, updateAttributes the same?

tvdstaaij:
Would like to note that this is a very serious issue, this makes it very easy to corrupt instances through the API. These instances will then fail to validate when they are partially updated with a different set of properties than the one(s) that were corrupted. It's even worse with NoSQL backends: I strongly suspect this issue prevents strict mode from being enforced with updateAll, and adding random properties outside of the model spec would prevent instances from ever being instance-updateable again (due to #759).

I would even consider this a security issue since it allows users to sabotage app features for other users.

vkruoso:
This is really a consistency problem that will actually allow a malicious user to overcome the developer assertions about its data. Is there any update on this?

Amir-61:
@kjdelisle PTAL ☝️

ssh24:
The issue should be fixed on the upcoming version of juggler. Thank you for your patience!

bajtos:
Hello, I think the solution landed via #771 is a breaking change, see https://github.com/strongloop/loopback-datasource-juggler/pull/1445#issuecomment-319889733, and we should rework it to preserve backwards compatibility, before the new version of juggler is released.

bajtos:
[...]

PR #1445, opened 2017-07-31
Add validation on update

ssh24:
### Description
Currently, `update`/`updateAll` function does not perform any validation if a validation is specified on the model. This fix adds the validation check on dao for `update`/`updateAll` function and tests with all the [validatable functions](https://loopback.io/doc/en/lb2/Validating-model-data.html#using-validation-methods).

fixes https://github.com/strongloop/loopback-datasource-juggler/issues/771

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2689,7 +2689,21 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
     Model.notifyObserversOf('before save', context,
       function(err, ctx) {
         if (err) return cb(err);
-        doUpdate(ctx.where, ctx.data);
+
+        data = ctx.data;
+        var inst = data;
+        if (!(data instanceof Model)) {
+          inst = new Model(data, {applyDefaultValues: false});
+        }
+
+        // validation required
+        inst.isValid(function(valid) {
+          if (valid) {
+            doUpdate(ctx.where, ctx.data);
+          } else {
+            cb(new ValidationError(inst), inst);
+          }
+        }, options);
       });
   });
 
--- a/test/validations.test.js
+++ b/test/validations.test.js
@@ -26,7 +26,7 @@ function getValidAttributes() {
 }
 
 describe('validations', function() {
-  var User, Entry;
+  var User, Entry, Employee;
 
   before(function(done) {
     db = getSchema();
@@ -47,8 +47,16 @@ describe('validations', function() {
       id: {type: 'string', id: true, generated: false},
       name: {type: 'string'},
     });
+    Employee = db.define('Employee', {
+      id: {type: Number, id: true, generated: false},
+      name: {type: String},
+      age: {type: Number},
+    });
     Entry.validatesUniquenessOf('id');
-    db.automigrate(done);
+    db.automigrate(function(err) {
+      should.not.exist(err);
+      Employee.create(empData, done);
+    });
   });
 
   beforeEach(function(done) {
@@ -58,8 +66,8 @@ describe('validations', function() {
     });
   });
 
-  after(function() {
-    // db.disconnect();
+  after(function(done) {
+    Employee.destroyAll(done);
   });
 
   describe('commons', function() {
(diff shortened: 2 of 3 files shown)


lib/dao.js
```diff
@@ -2689,7 +2689,26 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+        data = ctx.data;
+        var update = data;
+        var inst = data;
+        if (!(data instanceof Model)) {
+          inst = new Model(data, {applyDefaultValues: false});
+        }
+        update = inst.toObject(false);
+
+        Model.applyProperties(update, inst);
+        Model = Model.lookupModel(update);
```
b-admike: Is this line needed to do the validation?

ssh24: No it is not. Good catch.

ssh24:
@slnode test please

ssh24:
@slnode test please

test/validations.test.js
```diff
@@ -1013,3 +1384,17 @@ describe('validations', function() {
...
+  age: 1,
+}, {
+  id: 2,
+  name: 'Bar',
+  age: 2,
+}, {
+  id: 3,
+  name: 'Baz',
+  age: 3,
+}];
```
virkt25: nitpick: newline.

ssh24: I don't think so. It does not have that red arrow thing.

ssh24:
@slnode test please

lib/dao.js
```diff
@@ -2689,7 +2689,23 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
     Model.notifyObserversOf('before save', context,
       function(err, ctx) {
         if (err) return cb(err);
-        doUpdate(ctx.where, ctx.data);
+
+        data = ctx.data;
+        var update = data;
```
loay: why do you need to declare `update` here?
I mean if you continue the code while removing the `update` variable, it will work just fine !
Also, I see it as a redundant use of data, which is not needed

b-admike: @loay `update` variable is used on line 2708

loay: true and 2699 as well .. any use for adding this variable?

ssh24: The whole point is to keep the `data` variable clean. If you scroll down along the lines, after making a copy of `data` into `update`, we modify `update` so that the data returned from the **before save** hook is not changed.
```js
        data = ctx.data;
        var update = data;
        var inst = data;
        if (!(data instanceof Model)) {
        }
        update = inst.toObject(false);

        // validation required
        inst.isValid(function(valid) {
          if (valid) {
            doUpdate(ctx.where, ctx.data);
          } else {
            cb(new ValidationError(inst), inst);
          }
        }, update, options);
      });
```

ssh24:
@slnode test please

lib/dao.js
loay: Try with this:
```js
        data = ctx.data;
        var inst = data;
        if (!(data instanceof Model)) {
        }

        // validation required
        inst.isValid(function(valid) {
          if (valid) {
            doUpdate(ctx.where, ctx.data);
          } else {
            cb(new ValidationError(inst), inst);
          }
        }, options);
      });
```
You will see that it is not affected and works fine, because `inst` variable is assigned to be equal to `data` as well .. so it is either no needed at all, or the test coverage is not covering this part .. I think it is not needed, because `inst` variable is doing the job.

ssh24: Good catch. Fixed.

ssh24:
@slnode test please

bajtos:
@ssh24 IMO, the tests you have modified in LoopBack (https://github.com/strongloop/loopback/pull/3541) are a valid usage of `update`/`updateAll` API. If they are failing after your change here in strongloop/loopback-datasource-juggler#1445, then it's very likely that the change in juggler is actually a breaking change.

In order to support validations of partial updates (either via `patchAttributes` or `updateAll`), we need to implement partial validations that will validate only properties that are being changed.

If you think it still makes sense to run full validation in `updateAll`, then please add a feature flag to enable it. For example, many DAO options have `options.validate` that can be used to control whether validation should be run or skipped. You can also introduce model-level setting, similar to `validateUpsert`.

Regardless of the approach you take, all existing tests here in juggler and in loopback must keep passing without any changes. (The feature "validate updateAll" must be disabled by default for backwards compatibility.)

cc @kjdelisle

mrfelton:
This change breaks https://github.com/fullcube/loopback-ds-cascade-update-mixin which uses updateAll to make partial updates to multiple models.

+1 on @bajtos comments - this is a breaking change.

PR #1450, opened 2017-08-02
Fix update validation callback

ssh24:
### Description

Previously it would return the object to be updated as a part of the callback even though the data was not updated. This PR makes sure only the **error** is returned if a validation error is caught.


ssh24:
@slnode test please

PR #1451, opened 2017-08-03
Honor backwards compatibility with validate update

ssh24:
### Description

This PR is to honor backwards compatibility with validation on update.

connect to https://github.com/strongloop/loopback-datasource-juggler/pull/1445

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2669,6 +2669,19 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
   this.applyScope(query);
   this.applyProperties(data);
 
+  var doValidate = false;
+  if (options.validate === undefined) {
+    if (Model.settings.validateUpdate === undefined) {
+      if (Model.settings.automaticValidation !== undefined) {
+        doValidate = Model.settings.automaticValidation;
+      }
+    } else {
+      doValidate = Model.settings.validateUpdate;
+    }
+  } else {
+    doValidate = options.validate;
+  }
+
   where = query.where;
 
   var context = {
@@ -2696,14 +2709,17 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
           inst = new Model(data, {applyDefaultValues: false});
         }
 
-        // validation required
-        inst.isValid(function(valid) {
-          if (valid) {
+        if (doValidate === false) {
+          doUpdate(ctx.where, ctx.data);
+        } else {
+          // validation required
+          inst.isValid(function(valid) {
+            if (!valid) {
+              return cb(new ValidationError(inst));
+            }
             doUpdate(ctx.where, ctx.data);
-          } else {
-            cb(new ValidationError(inst));
-          }
-        }, options);
+          }, options);
+        }
       });
   });
 
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -2259,8 +2259,6 @@ describe('manipulation', function() {
     });
 
     it('should not coerce invalid values provided in where conditions', function(done) {
-      // remove the validations since update/updateAll validates the model
-      delete Person.validations;
       Person.update({name: 'Brett Boe'}, {dob: 'notadate'}, function(err) {
         should.exist(err);
         err.message.should.equal('Invalid date: notadate');
--- a/test/validations.test.js
+++ b/test/validations.test.js
@@ -51,6 +51,8 @@ describe('validations', function() {
       id: {type: Number, id: true, generated: false},
       name: {type: String},
       age: {type: Number},
+    }, {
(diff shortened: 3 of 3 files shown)


lib/dao.js
```diff
@@ -2669,6 +2669,19 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
   this.applyScope(query);
   this.applyProperties(data);
 
+  var doValidate = undefined;
+  if (options.validate === undefined) {
+    if (Model.settings.validateUpdate === undefined) {
```
bajtos: Would you mind using `validateUpdateAll` as the option name? I am personally never sure whether `update` is for a single or multiple records.

lib/dao.js
```diff
@@ -2696,14 +2709,22 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
+        if (doValidate === false) {
+          doUpdate(where, data);
+        } else {
+          // validation required
+          inst.isValid(function(valid) {
+            if (!valid) {
+              if (doValidate) {
+                return cb(new ValidationError(inst));
+              } else {
+                g.warn('Ignoring validation errors in {{update()}}:');
```
bajtos: I think we shouldn't be printing these warnings, because they will usually make no sense. (E.g. when a model has a required "name" property, but we are updating "count" property only, then a warning about missing "name" does not make much sense.)

lib/dao.js
ssh24: `update` and `updateAll` are aliases. As the jsdoc for the function says:
```
/**
 * Update multiple instances that match the where clause
 *
 * Example:
 *
 *```js
 * Employee.update({managerId: 'x001'}, {managerId: 'x002'}, function(err) {
 *     ...
 * });
 * ```
 *
 * @param {Object} [where] Search conditions (optional)
 * @param {Object} data Changes to be made
 * @param {Object} [options] Options for update
 * @param {Function} cb Callback, called with (err, info)
 */
```

I think we it would be the best to allow the user specify either `validateUpdate` or `validateUpdateAll` flag. In that case we need to take care of both the flags.

bajtos: I know they are aliases. It's just that when I see `update` in code, I have to always remember that it's an alias for `updateAll`. I am trying to say that the name `update` is not clear enough, while `updateAll` makes the intent much more clear.

bajtos: Not a big deal, I don't mind your current solution.

lib/dao.js
```diff
@@ -2696,14 +2709,19 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
...
-        // validation required
-        inst.isValid(function(valid) {
-          if (valid) {
+        if (doValidate === false) {
+          doUpdate(where, data);
+        } else {
+          // validation required
+          inst.isValid(function(valid) {
+            if (!valid) {
+              if (doValidate) {
```
bajtos: So if `doValidate` is not true, then we don't need to run `isValid` at all, because any validation errors will be silently discarded anyways.

Considering that some of our built-in validations (like "unique") are pretty expensive, I think we should avoid running `isValid` if it's not necessary.

bajtos:
How about the cases where #1445 is calling `delete XX.validations` in newly added tests? (Search for "delete" in `test/validations.test.js`.) I find it really weird that our tests are setting up a model with validations only to delete those validations later. Are those `delete` calls necessary? If they are, then shouldn't we use a different model, one that has only the validations needed by the test?

bajtos:
Adding to my previous comment, why are we setting `validateUpdate` to true when we delete all validations next? I find it too complex and unexpected.

ssh24:
@bajtos We set `validateUpdate` to true enable validation on the update function. On the other hand, we delete the previous validations on each of the validator tests to test for the **success** of that specific validator. This is done because two validators may cause conflict with one another. it is done as part of clean up before the test. See [here](https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/validations.test.js#L403)

lib/dao.js
```diff
@@ -2696,14 +2709,19 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
           inst = new Model(data, {applyDefaultValues: false});
         }
 
-        // validation required
-        inst.isValid(function(valid) {
-          if (valid) {
-            doUpdate(ctx.where, ctx.data);
-          } else {
-            cb(new ValidationError(inst));
+        if (doValidate === false) {
```
jannyHou: `if (doValidate) {}` and `if (doValidate === false) {` ---> `if (doValidate !== true)`?

ssh24: Ah, good catch. Fixed.

Issue #966, opened 2016-06-08
Implement static patchById and replaceById methods in PersistedModel


Nagarjuna-S:
The associated connector method for updating the existing record(PUT /model/{id}) is “updateAttributes” method. How ever I noticed that when ever we invoke PUT with ID operation, 'all' method is invoked first and on successful response of all method control is going to 'updateAttributes' method. I am thinking this is intentional. I have few concerns here.
1. There are many endpoints charged based on number api's called. As we are invoking internally two endpoint api calls(one in all method and another in updateAttributes method) for single operation, number of endpoint api's used will be doubled for Update operation(PUT with ID). Users and connector developers may surprise on number of calls.
2. Response time will be more which is crucial in API world. 

As we are expecting the connector developer to handle updating the existing record in updateAttributes method, why do we need to call all method from framework. I could think reasons behind this are
1. We want to make sure the record existence in endpoint before making the updateAttributes method.
2. Few endpoints expects full record for update operation.

But logic for above scenarios should be part of connector updateAttributes method. If connector developer needs to verify the record existence or for full record, they can call 'all method' from connector updateAttributes method.

Maybe I am missing something which is essential for framework. Can you please provide your comments
### Raymond Comments

I tend to agree with you. We should introduce static patchById and replaceById methods to PersistedModel so that:

PATCH /api/<model>/:id --> patchById
PUT /api/<model>/:id --> replaceById

It will be consistent with findById and deleteById.

Amir-61:
Hey @Nagarjuna-S ,

I'm afraid we decided to put implementation of `patchById` out of scope for now for some difficulties we noticed.
For more info Please see:
https://github.com/strongloop/loopback-datasource-juggler/pull/910/files#r61268157
and https://github.com/strongloop-internal/scrum-loopback/issues/752#issuecomment-215112749

Connect to https://github.com/strongloop/loopback-datasource-juggler/pull/910
Connect to https://github.com/strongloop-internal/scrum-loopback/issues/752

CC: @bajtos 

Thanks!

bajtos:
*Cross-posting from strongloop-internal/scrum-loopback#752*

Let me elaborate a bit more on my last comment.

Ideally, patchById should have these two features:

    It provides the updated model instance to the callback
    It validates the subset of model properties which is being updated.

The first feature require connectors to return data of the updated instance, which is not the case in the current versions of our SQL connectors.

[...]

PR #1453, opened 2017-08-04
Flag id as updateOnly when forceId is in effect

rashmihunt:
@bajtos  Please take a look. I will sync up with you on the test regressions.

Other PRs involved in the solution of issue https://github.com/strongloop/loopback/issues/2924
 - https://github.com/strongloop/loopback-swagger/pull/92
 - https://github.com/strongloop/loopback/pull/3548

Changes in this PR

- Added updatedOnly in memory flag for generated ID in model-builder.js. I decided to go with property level setting as you suggested rather than top level setting updateOnly= [ ]
- Modified an existing test case to cover this.
- forceID is set back to settings[] in datasource.js which can now correctly be used in all parts of the code
- added new test case to check  forceId in settings[]
- Implemented getUpdateOnlyProperties() in model.js
- Added test case for this 
- 6 existing test cases are failing due to the fact the fact that forceId now being 'true' for generated id cases unless user has passed in forceID = false while creating the model, 

In dao.js

```
  var forceId = this.settings.forceId;
  if (forceId) {
    options = Object.create(options);
    options.validate = !!doValidate;
    if (doValidate) {
      Model.findById(id, options, function(err, model) {
        if (err) return cb(err);
        if (!model) return cb(errorModelNotFound(id));
        model.updateAttributes(data, options, cb);
      });
    } else {
      const model = new Model({id: id}, {persisted: true});
      model.updateAttributes(data, options, cb);
    }
    return cb.promise;
  }
```

if `forceId`  is true in most failing test cases and `doValidate` is false, which means it ends up in 

```
else {
      const model = new Model({id: id}, {persisted: true});
      model.updateAttributes(data, options, cb);
    }
```

which creates a new model and with the new set of attributes passed in. In one case 
`validateUpsert` is calculated and set into 'doValidate' in logic prior to above block of code. Looking through the documentation and closed issues in 3.x , I noticed that in 3.x, validateUpsert is changed to 'true'  by default, which means 'undefined' should be considered as 'true' ?  Should there be logic just before -  `if (forceId) {..}`

```
  if (doValidate === undefined) {
    doValidate = true;
  }
```

Anyway, I will sync up with you on Monday morning on this.

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -628,14 +628,6 @@ DataSource.prototype.setupDataAccess = function(modelClass, settings) {
       idProp.type = idType;
       modelClass.definition.rawProperties[idName].type = idType;
       modelClass.definition.properties[idName].type = idType;
-      var forceId = settings.forceId;
-      if (idProp.generated && forceId !== false) {
-        forceId = true;
-      }
-      settings.forceId = forceId;
-      if (forceId) {
-        modelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
-      }
     }
     if (this.connector.define) {
       // pass control to connector
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -628,15 +628,6 @@ DataSource.prototype.setupDataAccess = function(modelClass, settings) {
       idProp.type = idType;
       modelClass.definition.rawProperties[idName].type = idType;
       modelClass.definition.properties[idName].type = idType;
-      var forceId = settings.forceId;
-      if (idProp.generated && forceId !== false) {
-        forceId = true;
-      }
-      // set the calculated forceId back to settings
-      settings.forceId = forceId;
-      if (forceId) {
-        modelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
-      }
     }
     if (this.connector.define) {
       // pass control to connector
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -632,6 +632,8 @@ DataSource.prototype.setupDataAccess = function(modelClass, settings) {
       if (idProp.generated && forceId !== false) {
         forceId = true;
       }
+      // set the calculated forceId back to settings
+      settings.forceId = forceId;
       if (forceId) {
         modelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
       }
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -628,6 +628,14 @@ DataSource.prototype.setupDataAccess = function(modelClass, settings) {
       idProp.type = idType;
       modelClass.definition.rawProperties[idName].type = idType;
       modelClass.definition.properties[idName].type = idType;
+      var forceId = settings.forceId;
+      if (idProp.generated && forceId !== false) {
+        forceId = true;
+      }
+      settings.forceId = forceId;
+      if (forceId) {
+        modelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
+      }
     }
     if (this.connector.define) {
       // pass control to connector
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -349,7 +349,13 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
     var idName = modelDefinition.idName();
     idProp = ModelClass.definition.rawProperties[idName];
     if (idProp.generated && forceId !== false) {
-      forceId = true;
+      forceId = 'auto';
+    } else if (!idProp.generated && forceId === 'auto') {
+      // One of our parent models has enabled forceId because
+      // it uses an auto-generated id property. However,
+      // this particular model does not use auto-generated id,
+      // therefore we need to disable `forceId`.
+      forceId = false;
     }
 
     if (forceId) {
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -308,9 +308,7 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
   // Add the id property
   if (idInjection) {
     // Set up the id property
-    // updateOnly property is added to indicate that this property will appear in the model for update/updateorcreate
-    // operations but and not for create operation.
-    ModelClass.definition.defineProperty('id', {type: Number, id: 1, generated: true, updateOnly: true});
+    ModelClass.definition.defineProperty('id', {type: Number, id: 1, generated: true});
   }
 
   idNames = modelDefinition.idNames(); // Reload it after rebuild
@@ -342,6 +340,24 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
       configurable: true,
       enumerable: false,
     });
+  }
+
+  // updateOnly property is added to indicate that this property will appear in
+  // the model for update/updateorcreate operations but and not for create operation.
+  var forceId = ModelClass.definition.settings.forceId;
+  if (idNames.length > 0) {
+    var idName = modelDefinition.idName();
+    idProp = ModelClass.definition.rawProperties[idName];
+    if (idProp.generated && forceId !== false) {
+      forceId = true;
+    }
+    // set forceId back to into the settings
+    ModelClass.settings.forceId = forceId;
+    if (forceId) {
+      ModelClass.definition.properties[idName].updateOnly = true;
+      ModelClass.definition.rawProperties[idName].updateOnly = true;
+      ModelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
+    }
   }
 
   // A function to loop through the properties
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -344,7 +344,7 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
 
   // updateOnly property is added to indicate that this property will appear in
   // the model for update/updateorcreate operations but and not for create operation.
-  var forceId = ModelClass.definition.settings.forceId;
+  var forceId = ModelClass.settings.forceId;
   if (idNames.length > 0) {
     var idName = modelDefinition.idName();
     idProp = ModelClass.definition.rawProperties[idName];
@@ -354,6 +354,9 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
     if (forceId) {
       ModelClass.definition.properties[idName].updateOnly = true;
       ModelClass.definition.rawProperties[idName].updateOnly = true;
+    } else {
+      ModelClass.definition.properties[idName].updateOnly = false;
+      ModelClass.definition.rawProperties[idName].updateOnly = false;
     }
   }
 
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -351,12 +351,9 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
     if (idProp.generated && forceId !== false) {
       forceId = true;
     }
-    // set forceId back to into the settings
-    ModelClass.settings.forceId = forceId;
     if (forceId) {
       ModelClass.definition.properties[idName].updateOnly = true;
       ModelClass.definition.rawProperties[idName].updateOnly = true;
-      ModelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
     }
   }
 
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -308,7 +308,9 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
   // Add the id property
   if (idInjection) {
     // Set up the id property
-    ModelClass.definition.defineProperty('id', {type: Number, id: 1, generated: true});
+    // updateOnly property is added to indicate that this property will appear in the model for update/updateorcreate
+    // operations but and not for create operation.
+    ModelClass.definition.defineProperty('id', {type: Number, id: 1, generated: true, updateOnly: true});
   }
 
   idNames = modelDefinition.idNames(); // Reload it after rebuild
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -351,13 +351,15 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
     if (idProp.generated && forceId !== false) {
       forceId = true;
     }
+
     if (forceId) {
-      ModelClass.definition.properties[idName].updateOnly = true;
-      ModelClass.definition.rawProperties[idName].updateOnly = true;
-    } else {
-      ModelClass.definition.properties[idName].updateOnly = false;
-      ModelClass.definition.rawProperties[idName].updateOnly = false;
+      ModelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
     }
+
+    ModelClass.definition.properties[idName].updateOnly = !!forceId;
+    ModelClass.definition.rawProperties[idName].updateOnly = !!forceId;
+
+    ModelClass.settings.forceId = forceId;
   }
 
   // A function to loop through the properties
--- a/lib/model.js
+++ b/lib/model.js
@@ -827,6 +827,27 @@ ModelBaseClass.getMergePolicy = function(options) {
   return mergePolicy;
 };
 
+/**
+ * Gets properties defined with 'updateOnly' flag set to true from the model. This flag is also set to true
+ * internally for the id property, if this property is generated and IdInjection is true.
+ * @returns {updateOnlyProps} List of properties with updateOnly set to true.
+ */
+
+ModelBaseClass.getUpdateOnlyProperties = function() {
+  var Model = this;
+  var props = Model.definition.properties;
+  var property;
+  var updateOnlyProps = [];
+
+  for (var key in props) {
+    property = props[key];
+    if (property.updateOnly) {
+      updateOnlyProps.push(key);
+    }
+  }
+  return updateOnlyProps;
+};
+
 // Mixin observer
 jutil.mixin(ModelBaseClass, require('./observer'));
 
--- a/lib/model.js
+++ b/lib/model.js
@@ -835,17 +835,8 @@ ModelBaseClass.getMergePolicy = function(options) {
 
 ModelBaseClass.getUpdateOnlyProperties = function() {
   var Model = this;
-  var props = Model.definition.properties;
-  var property;
-  var updateOnlyProps = [];
-
-  for (var key in props) {
-    property = props[key];
-    if (property.updateOnly) {
-      updateOnlyProps.push(key);
-    }
-  }
-  return updateOnlyProps;
+  const props = this.definition.properties;
+  return Object.keys(props).filter(key => props[key].updateOnly);
 };
 
 // Mixin observer
--- a/test/default-scope.test.js
+++ b/test/default-scope.test.js
@@ -208,7 +208,7 @@ describe('default scope', function() {
       var data = {id: ids.productA, description: 'Anything...', kind: 'ingored'};
       Tool.updateOrCreate(data, function(err, p) {
         should.not.exist(err);
-        p.name.should.equal('Product A'); // TODO [rashmi] Sync up with Miroslav about this. propertyly setting the forceId back to settings has this side affect
+        p.name.should.equal('Product A');
         p.kind.should.equal('Tool');
         p.description.should.equal('Anything...');
         done();
--- a/test/default-scope.test.js
+++ b/test/default-scope.test.js
@@ -208,7 +208,7 @@ describe('default scope', function() {
       var data = {id: ids.productA, description: 'Anything...', kind: 'ingored'};
       Tool.updateOrCreate(data, function(err, p) {
         should.not.exist(err);
-        // p.name.should.equal('Product A'); // TODO [rashmi] Sync up with Miroslav about this. propertyly setting the forceId back to settings has this side affect
+        p.name.should.equal('Product A'); // TODO [rashmi] Sync up with Miroslav about this. propertyly setting the forceId back to settings has this side affect
         p.kind.should.equal('Tool');
         p.description.should.equal('Anything...');
         done();
--- a/test/default-scope.test.js
+++ b/test/default-scope.test.js
@@ -208,7 +208,7 @@ describe('default scope', function() {
       var data = {id: ids.productA, description: 'Anything...', kind: 'ingored'};
       Tool.updateOrCreate(data, function(err, p) {
         should.not.exist(err);
-        p.name.should.equal('Product A');
+        // p.name.should.equal('Product A'); // TODO [rashmi] Sync up with Miroslav about this. propertyly setting the forceId back to settings has this side affect
         p.kind.should.equal('Tool');
         p.description.should.equal('Anything...');
         done();
--- a/test/loopback-dl.test.js
+++ b/test/loopback-dl.test.js
@@ -1931,3 +1931,61 @@ describe('ModelBuilder options.models', function() {
     assert.deepEqual(codes.number, ['unknown-property']);
   });
 });
+
+describe('updateOnly', function() {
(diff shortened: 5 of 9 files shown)


lib/model-builder.js
```diff
@@ -308,7 +308,9 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
   // Add the id property
   if (idInjection) {
     // Set up the id property
-    ModelClass.definition.defineProperty('id', {type: Number, id: 1, generated: true});
+    // updateOnly property is added to indicate that this property will appear in the model for update/updateorcreate
+    // operations but and not for create operation.
+    ModelClass.definition.defineProperty('id', {type: Number, id: 1, generated: true, updateOnly: true});
```
bajtos: I think `updateOnly` should be set only when `forceId` is in effect. With `forceId:false`, clients can provide custom `id` value even when the `id` column was injected by the datasource.

I think `updateOnly` flag should be set in `lib/datasource.js`:

```js
if (forceId) {
  modelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
  // set updateOnly here
  // something along the following lines, I am not sure if we need to change rawProperties too
  modelClass.definition.properties[idName].updateOnly = true;
}
```

lib/model.js
```diff
@@ -827,6 +827,27 @@ ModelBaseClass.getMergePolicy = function(options) {
...
+ * @returns {updateOnlyProps} List of properties with updateOnly set to true.
+ */
+
+ModelBaseClass.getUpdateOnlyProperties = function() {
+  var Model = this;
+  var props = Model.definition.properties;
+  var property;
+  var updateOnlyProps = [];
+
+  for (var key in props) {
```
bajtos: Let's use [Array.prototype.filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) to simplify this code.

```js
const props = this.definition.properties;
return Object.key(props).filter(key => props[key].updateOnly);
```

test/loopback-dl.test.js
```diff
@@ -645,8 +645,28 @@ describe('DataSource define model', function() {
...
+      date: {type: Date, default: function() {
+        return new Date();
+      }},
+    });
+
+    // check if forceId is added as true in ModelClass's settings[] explicitly, if id a generated (default) and forceId
+    // in from the model is true(unspecified is 'true' which is the default).
+    Post.settings.should.have.property('forceId').eql(true);
 
+    // check if method getUpdateOnlyProperties exist in ModelClass and check if the Post has 'id' in updateOnlyProperties list
```
bajtos: One assert per test please. There should be two tests: one checking `Post.settings.forceId`, the other verifying `getUpdateOnlyProperties()`. We may want to write more tests for `getUpdateOnlyProperties` in the future, it's probably a good idea to create a new `describe` block for tests of this method.

test/loopback-dl.test.js
```diff
@@ -645,8 +645,28 @@ describe('DataSource define model', function() {
...
+    Post.should.have.property('getUpdateOnlyProperties');
+    var updateOnlyProps = Post.getUpdateOnlyProperties();
+    updateOnlyProps.should.containEql('id');
```
bajtos: I think in this test we know exactly what properties should be marked as `updateOnly`, don't we? In which case I think it's better to assert the exact content of the returned array:

```js
updateOnlyProps.should.eql(['id']);
```

test/default-scope.test.js
```diff
@@ -208,7 +208,7 @@ describe('default scope', function() {
       var data = {id: ids.productA, description: 'Anything...', kind: 'ingored'};
       Tool.updateOrCreate(data, function(err, p) {
         should.not.exist(err);
-        p.name.should.equal('Product A');
+        // p.name.should.equal('Product A'); // TODO [rashmi] Sync up with Miroslav about this. propertyly setting the forceId back to settings has this side affect
```
bajtos: What are the side effects? Could you post the outcome of the test here please?

bajtos: Please ignore my comment, I see you have already analyzed the issue and posted learning in the pull request description 👍

bajtos:
@rashmihunt I pushed two commits to fix the test failures you have discovered - see d6716ae83 and 023439bc0

bajtos:
I am cross-posting the commit message from d6716ae for posterity:

The contract of `updateOrCreate` is expecting a full object instance to be passed to the callback.

The current implementation was creating an empty instance and calling updateAttributes under the hood. As a result, the callback was called with the attributes being updated only.

In order to preserve existing behaviour, we have to always build a full initial instance by calling `findById`.

See the following discussion for more context: #966

jannyHou:
I feel `_rev` in cloudant connector probably needs this feature too, and I will review it asap, thank you @rashmihunt

lib/model-builder.js
rashmihunt: @bajtos We can't set updateOnly property in the above place, since  LoopBack repository's PersistentModel.setupRemoting() --> getUpdateOnlyProperties() gets invoked before above logic in datasource-gluggler's datasource.js code gets executed. Since getUpdateOnlyProperties() gets invoked before setting 'updateOnly' model as you suggested above, getUpdateOnlyProperties() will not return id property even if forceId is true. Any other parts of the logic we can set this property?

rashmihunt: @bajtos As alternate approach, I tried this in model builder

  // Add the id property
  if (idInjection) {
    // Set up the id property
    // updateOnly property is added to indicate that this property will appear in the model for update related operations
    // but and not for create operation.
    if (ModelClass.definition.settings.forceId !== false) {
      ModelClass.definition.defineProperty('id', {type: Number, id: 1, generated: true, updateOnly: true});
    } else {
      ModelClass.definition.defineProperty('id', {type: Number, id: 1, generated: true});
    }
  }
}
Unfortunately, this doesn't seem to work either if user explicitly sets forceId = false in model file, for e.g  book.json file.  ModelClass.definition.settings.forceId logic seems to look at the base model's (Model) setting rather than the book model's setting.

bajtos: > Unfortunately, this doesn't seem to work either if user explicitly sets forceId = false in model file, for e.g book.json file. ModelClass.definition.settings.forceId logic seems to look at the base model's (Model) setting rather than the book model's setting.

I think you should be using `ModelClass.settings.forceId` here.

There are two things I am concerned about:

 - We should keep the logic deciding how to treat`forceId: undefined` in a single place.
 - When `forceId` is effectively `false`, then the `id` property should not be marked as `updateOnly`.

I did not realize that `setupDataAccess` is called too late in our setup process. In that case, I am proposing to move all logic determining `forceId` from `setupDataAccess` to this method (`ModelBuilder.prototype.define`). I am talking about the following lines:

```js
if (idProp.generated && forceId !== false) {
  forceId = true;
}
// set the calculated forceId back to settings
settings.forceId = forceId;
```

I believe we should have all required information already available here. The only thing that can change in `setupDataAccess` is the type of the `id` property, but that is not affecting `forceId` (at least I think so).

Thoughts?

bajtos:
@jannyHou 
> I feel `_rev` in cloudant connector probably needs this feature too, and I will review it asap, thank you 

Yes, I was keeping the use case of `_rev` property in mind while guiding @rashmihunt's work on these changes.

@rashmihunt would you mind splitting this pull request into two - one fixing `updateOrCreate` for `forceId === undefined`, another adding `updateOnly`? It will make it easier for us in the future to understand the context of the changes being made and distinguish between changes needed for `updateOnly` from changes needed to fix `forceId`.

lib/model-builder.js
jannyHou: Agree ^^

I tried
```
  var forceId = ModelClass.settings.forceId;

  if (idNames.length > 0) {
    // id already exists
    idInjection = false;
    // Does method `idName` also works for composite Id at this phase?
    var idName = modelDefinition.idName();
    var idProp = ModelClass.definition.rawProperties[idName];
    if (idProp.generated && forceId !== false) {
      forceId = true;
    }
  } else {
    if (forceId !== false) forceId = true;
  }
  ModelClass.settings.forceId = forceId;
```
received same errors as @rashmihunt is fixing in this PR, other tests pass.

rashmihunt: @bajtos what you are proposing is to move only the forceId determining logic from setUpDataAccess and leave rest of the logic as is..  for e.g setting idType and calling modelClass.validatesAbsenceOf(..) based on 
forceId from the settings.
```
DataSource.prototype.setupDataAccess = function(modelClass, settings) {
  if (this.connector) {
    // Check if the id property should be generated
    var idName = modelClass.definition.idName();
    var idProp = modelClass.definition.rawProperties[idName];
    if (idProp && idProp.generated && this.connector.getDefaultIdType) {
      // Set the default id type from connector's ability
      var idType = this.connector.getDefaultIdType() || String;
      idProp.type = idType;
      modelClass.definition.rawProperties[idName].type = idType;
      modelClass.definition.properties[idName].type = idType;
      var forceId = settings.forceId;
      if (forceId) {
        modelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
      }
    }
....
```

rashmihunt: I did try your proposed solution and it does work.

bajtos: > what you are proposing is to move only the forceId determining logic from setUpDataAccess and leave rest of the logic as is.. for e.g setting idType and calling modelClass.validatesAbsenceOf(..) based on
forceId from the settings.
 
> I did try your proposed solution and it does work.

Cool!

I think it probably makes sense to move `modelClass.validatesAbsenceOf` over too, while we are at cleaning up this code. Having said that, what you are proposing in your comment above looks reasonable too.

test/loopback-dl.test.js
rashmihunt: @bajtos done

test/loopback-dl.test.js
rashmihunt: @bajtos done PTAL

lib/model-builder.js
rashmihunt: @bajtos I saw lot of tests failed in CI in downstream component 'loopback' if calculated forceId is set back into model class's settings[]  in mode-builder.define (..) method instead of datasource.setupDataAccess(). See in this commit 
 https://github.com/strongloop/loopback-datasource-juggler/pull/1453/commits/bb9be2d4c50d2ab99aecce12655db44cc6779a56#diff-546e03515f85748fe5b4d0db11cd81b7R355

I am bit puzzled what's the difference between setting it in model-builder.js vs datasource.js. 

TO avoid the CI failues, I reverted the refactoring of model-builder.js and datasource.js with my latest commit. PTAL.

rashmihunt:
@bajtos PTAL

lib/datasource.js
```diff
@@ -632,6 +632,7 @@ DataSource.prototype.setupDataAccess = function(modelClass, settings) {
       if (idProp.generated && forceId !== false) {
         forceId = true;
       }
+      settings.forceId = forceId;
```
bajtos: I believe the lines 632-635 can be removed now, because `forceId` is already calculated in `lib/model-builder`. Unless I am missing something?

lib/model-builder.js
```diff
@@ -342,6 +342,21 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
   }
 
+  // updateOnly property is added to indicate that this property will appear in
+  // the model for update/updateorcreate operations but and not for create operation.
+  var forceId = ModelClass.definition.settings.forceId;
+  if (idNames.length > 0) {
+    var idName = modelDefinition.idName();
+    idProp = ModelClass.definition.rawProperties[idName];
+    if (idProp.generated && forceId !== false) {
+      forceId = true;
```
bajtos: Please update `ModelClass.settings.forceId` too.

lib/model-builder.js
```diff
@@ -342,6 +342,21 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
     });
   }
 
+  // updateOnly property is added to indicate that this property will appear in
+  // the model for update/updateorcreate operations but and not for create operation.
+  var forceId = ModelClass.definition.settings.forceId;
```
bajtos: Shouldn't this be `ModelClass.settings.forceId`?

test/loopback-dl.test.js
```diff
@@ -1911,3 +1911,61 @@ describe('ModelBuilder options.models', function() {
     assert.deepEqual(codes.number, ['unknown-property']);
   });
 });
+
+describe('updateOnly', function() {
+  it('verify forceId', function(done) {
```
bajtos: We are using BDD style, where the test names should read like an English sentence, e.g. `it verifies forceId`. They should say what is being tested (that's usually specified in `describe`), what is the observed behaviour and what were the conditions of the test.

In this case, I think a better test name is `it('sets forceId to true when model id is generated')`.

Feel free to use different test names as long as they follow the BDD style.

test/loopback-dl.test.js
```diff
@@ -1911,3 +1911,61 @@ describe('ModelBuilder options.models', function() {
...
+      }},
+    });
+    // check if forceId is added as true in ModelClass's settings[] explicitly,
+    // if id a generated (default) and forceId in from the model is
+    // true(unspecified is 'true' which is the default).
+    Post.settings.should.have.property('forceId').eql(true);
+    done();
+  });
+
+  it('verify getUpdateOnlyProperties with forceId undefined', function(done) {
```
bajtos:
```js
it('flags id as updateOnly when forceId is undefined')
```

test/loopback-dl.test.js
```diff
@@ -1911,3 +1911,61 @@ describe('ModelBuilder options.models', function() {
...
+      }},
+    });
+    // check if method getUpdateOnlyProperties exist in ModelClass and check if
+    // the Post has 'id' in updateOnlyProperties list
+    Post.should.have.property('getUpdateOnlyProperties');
+    Post.getUpdateOnlyProperties().should.eql(['id']);
+    done();
+  });
+
+  it('verify getUpdateOnlyProperties with forceId = false', function(done) {
```
bajtos:
```js
it('does not flag id as updateOnly when forceId is false')
```

test/loopback-dl.test.js
```diff
@@ -1911,3 +1911,61 @@ describe('ModelBuilder options.models', function() {
...
+      gender: String,
+    }, {forceId: false});
+    // id should not be there in updateOnly properties list if forceId is set
+    // to false
+    Person.should.have.property('getUpdateOnlyProperties');
+    Person.getUpdateOnlyProperties().should.eql([]);
+    done();
+  });
+
+  it('verify getUpdateOnlyProperties with forceId = true', function(done) {
```
bajtos:
```js
it('flags id as updateOnly when forceId is true')
```

bajtos:
> CI for 'loopback' is passing - https://cis-jenkins.swg-devops.com/job/ds/job/loopback~master/1082/ Not sure why the results show as failing

Looks like a (temporary?) problem of our CI infrastructure, see the [console log](https://cis-jenkins.swg-devops.com/job/ds/job/loopback~master/1082/console):

```
v4.8.4 is already installed.
Now using node v4.8.4 (npm v2.15.11)
Now using node v4.8.4 (npm v2.15.11)
npm ERR! Linux 4.4.0-83-generic
npm ERR! argv "/home/jenkins/.nvm/versions/node/v4.8.4/bin/node" "/home/jenkins/.nvm/versions/node/v4.8.4/bin/npm" "install" "-g" "semver"
npm ERR! node v4.8.4
npm ERR! npm  v2.15.11
npm ERR! code ECONNREFUSED
npm ERR! errno ECONNREFUSED
npm ERR! syscall connect

npm ERR! Error: connect ECONNREFUSED 10.91.203.59:40975
```

lib/datasource.js
rashmihunt: @bajtos I had written a comment on we can't forceId calculation and setting in model-builder. Looks like it was wiped out after I did a commit with force push option, after I had rebased my PR. 

If we set  forceId in model-builder.js instead of datasource.js, 60 tests in loopback are failing. Hence I left the forceId back the settings[] in datasource.js. I will sync up with you on this tomorrow.

lib/model-builder.js
rashmihunt: @bajtos ModelClass.definition.settings.forceId and ModelClass.settings.forceId are exactly the same. Anyway changing to ModelClass.settings.forceId

lib/model-builder.js
rashmihunt: See my above comments on 60 failing tests in 'loopback' repository

lib/datasource.js
bajtos: I find this very suspicious, please let me try to investigate this myself to understand better what's going on, before we move forward.

lib/model-builder.js
```diff
@@ -354,6 +354,9 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
     if (forceId) {
       ModelClass.definition.properties[idName].updateOnly = true;
       ModelClass.definition.rawProperties[idName].updateOnly = true;
+    } else {
+      ModelClass.definition.properties[idName].updateOnly = false;
+      ModelClass.definition.rawProperties[idName].updateOnly = false;
```
bajtos: I think these 6 lines can be simplified as follows:

```js
ModelClass.definition.properties[idName].updateOnly = !!forceId;
ModelClass.definition.rawProperties[idName].updateOnly = !!forceId;
```

lib/datasource.js
bajtos: @rashmihunt I made the changes I had in my mind in 4ccef202d, there were only two tests to fix. Perhaps we misunderstood each other? Anyhow, please take a look at my code and feel free to modify it further as needed.

bajtos:
> > If we set forceId in model-builder.js instead of in datasource.js, 60 tests in loopback are failing. I debugged couple of failing 60 tests and the reason seems to be below logic in dao.js

>  made the changes I had in my mind in 4ccef20, there were only two tests to fix. Perhaps we misunderstood each other? Anyhow, please take a look at my code and feel free to modify it further as needed.

Oh, you mean the tests in the `loopback` repository. I'll check them out.

bajtos:
Here is my understanding:

 - The base `Model` class has a generated `id`. Because this model is never attached to any datasource (only inherited from), `forceId` was never enabled before our changes.
 - `AccessToken` inherits from `Model` and defines a custom `id` property that is not generated, therefore `forceId` should not be enabled (and we have code that's relying on that).
 - With our change in place, `Model.settings.forceId` is set to `true`, because the model has an auto-generated id-property. When `AccessToken` inherits from `Model`, it will also inherit `forceId: true`.

I think we need to find a way how to allow sub-models to detect the situation where the parent model does not specify explicit `strictId` value, so that we can run the id-based check ourselves. E.g. use `'auto'` instead of `true` as the `forceId` value.

bajtos:
> I think we need to find a way how to allow sub-models to detect the situation where the parent model does not specify explicit strictId value, so that we can run the id-based check ourselves. E.g. use 'auto' instead of true as the forceId value.

@rashmihunt see 3f3e2d5. The tests in loopback are passing on my machine now. However, I am little bit concerned about the complexity/hackiness of this solution. What do you think? Do we mind?
 Can we find a cleaner solution?

rashmihunt:
@bajtos seems like a reasonable solution for now.  Looks like 5 tests are failing with this change. I will take a look into it.

rashmihunt:
@slnode test please

rashmihunt:
@bajtos I had to modify 2 tests based on your 'auto' flag logic.  https://github.com/strongloop/loopback-datasource-juggler/pull/1453/commits/e32b656a113539cfaea88d2debdbe4a1f8c68984 

1) User ->  inherits base model   Customer doesn't have id property or forceId set.. it's inherited from base model, which means id property is 'id', generated is 'true' and forceId != false. Hence customer  model has forceId set to 'auto' in memory.

2) Customer --> extends User. Customer has defined id property as customerId: {type: String, id: true}} Since customer is derived from User, forceId is what User has which is 'forceId' to begin with. Notice that 'generated' is not set for customerId, hence it's 'undefined' during runtime. Hence this enters else if (!idProp.generated && forceId === 'auto') { forceId = false;} setting forceId for Customer to 'false'.  here it's assumed that 'generated' as 'undefined' means it's value is 'false' which in end results in 'updateOnly' to 'false' which means user can send ID for create() operation without the validation in place. If  'generated' is 'undefined', is this what you expect? 
I tried to search in lb3 documentation, on user should assume if 'generated' is undefined for the ID,  and I don't see any documentation.  On the other hand, forceId as 'undefined' we consider this as 'true' in lb3.

rashmihunt:
@slnode test please

test/model-inheritance.test.js
```diff
@@ -239,7 +239,8 @@ describe('Model class inheritance', function() {
         );
 
       assert.deepEqual(User.settings, {
-        forceId: true,
+        // forceId is set to 'auto' in memory if idProp.generated && forceId !== false
+        forceId: 'auto',
```
bajtos: This change makes sense to me. We are basically treating "auto" as "forceId was not specified by the user, but it should be enabled because id is auto-generated".

test/model-inheritance.test.js
```diff
@@ -258,7 +259,7 @@ describe('Model class inheritance', function() {
       });
 
       assert.deepEqual(Customer.settings, {
-        forceId: true,
+        forceId: false,
```
bajtos: Since the Customer has specified `customerId: {type: String, id: true}` as a non-generated property, `forceId` it makes sense to me that `forceId` should be disabled, even though the parent `User` model has it enabled. This is the behaviour we are relying on in LoopBack, isn't it?

test/loopback-dl.test.js
```diff
@@ -1911,3 +1911,61 @@ describe('ModelBuilder options.models', function() {
...
+  });
+
+  it('flags id as updateOnly when forceId is true', function(done) {
+    var ds = new DataSource('memory');
+    var Person = ds.define('Person', {
+      name: String,
+      gender: String,
+    }, {forceId: true});
+    // id should be there in updateOnly properties list if forceId is set
+    // to false
```
bajtos: forceId is set to `true` here, please fix the comment

rashmihunt:
@slnode test please

jannyHou:
Cloudant fails with same reason in https://github.com/strongloop/loopback-connector-cloudant/pull/161#issuecomment-323395370

There is a big refactor happen in that repo and two PRs block each other, so I have to check in one(with the failure) then the other one(contains the fix), it's not caused by this PR.

rashmihunt:
@slnode test please

lib/model-builder.js
```diff
@@ -342,6 +342,32 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
+    } else if (!idProp.generated && forceId === 'auto') {
+      // One of our parent models has enabled forceId because
+      // it uses an auto-generated id property. However,
+      // this particular model does not use auto-generated id,
+      // therefore we need to disable `forceId`.
+      forceId = false;
+    }
+
+    if (forceId) {
+      ModelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
```
simonbrunel: @rashmihunt @bajtos shouldn't we also reject any requests containing an empty string ID when `forceId` is `true`? [`validatesAbsenceOf`](/strongloop/loopback-datasource-juggler/blob/master/lib/validations.js#L298) considers an [empty string value](/strongloop/loopback-datasource-juggler/blob/master/lib/validations.js#L742) valid, which generates a record with no id (at least with MongoDB) when calling a POST endpoint with an explicit empty string ID even if `forceId` is `true` (I'm new to LoopBack, so I might have missed something).

bajtos: For posterity, I responded here: https://github.com/strongloop/loopback-datasource-juggler/issues/1519#issuecomment-364513700

PR #1454, opened 2017-08-07
Fix forceId

jannyHou:
### Related issue

https://github.com/strongloop/loopback-connector-couchdb2/issues/13

Test Case TBD in loopback-connector-couchdb2 and juggler

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -900,7 +900,9 @@ DataAccessObject.replaceOrCreate = function replaceOrCreate(data, options, cb) {
   }
 
   var forceId = this.settings.forceId;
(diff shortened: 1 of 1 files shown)


bajtos:
+1, let's fix the issue via #1453. It may be better to split #1453 into two pull requests, one fixing `forceId` and the second adding `updateOnly` property flag. That way the forceId fix can be landed quicker.

jannyHou:
The fix in https://github.com/strongloop/loopback-datasource-juggler/pull/1453/files#diff-9e9565730a5d5c0bdf028ccde9dd3a4eR636 is much better. Closing this one.

PR #1416, opened 2017-07-03
Fix undefined properties in where

raymondfeng:
https://github.com/strongloop/loopback/issues/2364

#### Related issues

- None

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #1455, opened 2017-08-09
Fix undefined properties in where

loay:
### Description

backport of https://github.com/strongloop/loopback-datasource-juggler/pull/1416

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -731,8 +731,11 @@ DataAccessObject.upsertWithWhere = function(where, data, options, cb) {
 
         function callConnector() {
           try {
-            ctx.where = removeUndefined(ctx.where);
-            ctx.where = Model._coerce(ctx.where);
+            // Support an optional where object
+            var handleUndefined =  Model._getSetting('normalizeUndefinedInQuery');
+            // alter configuration of how removeUndefined handles undefined values
+            ctx.where = removeUndefined(ctx.where, handleUndefined);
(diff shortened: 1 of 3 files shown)


virkt25:
Not just mongo, cloudant, soap and 2 others are failing.

loay:
@slnode test please

loay:
@slnode test please

lib/dao.js
```diff
@@ -2187,7 +2187,7 @@ DataAccessObject.destroyAll = function destroyAll(where, options, cb) {
         var handleUndefined =  Model._getSetting('normalizeUndefinedInQuery');
         // alter configuration of how removeUndefined handles undefined values
         where = removeUndefined(where, handleUndefined);
-        where = Model._coerce(where, options);
+        where = Model._coerce(where);
```
jannyHou: hmm...why some functions accept `options` but these two not?

loay: The `options` object are empty, therefore the test case failed in 2 occurrences.. 
once in juggler
```
 should count using date string
```
and once in Mongo 

```
mongodb default id type should be able to delete by string id
```
Where the `option` variable is not even affection the current fix as it is not needed..
The reason why , it seems that once the `Option` object is empty, it causes  an assertion failure, then the test case fail.

loay:
PR cannot be backported to 2.x for the simple reason that `._coerce` function signature has changed in 3.x, therefore it is not compatible.

PR #1463, opened 2017-08-16
Fix include util functions

kobaska:
### Description
Fix include util functions to only include objects when key exists on the object

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/include_utils.js
+++ b/lib/include_utils.js
@@ -23,7 +23,10 @@ function buildOneToOneIdentityMapWithOrigKeys(objs, idName) {
   for (var i = 0; i < objs.length; i++) {
     var obj = objs[i];
     var id = obj[idName];
-    kvMap.set(id, obj);
+
+    if (id) {
+      kvMap.set(id, obj);
+    }
   }
   return kvMap;
 }
@@ -33,9 +36,12 @@ function buildOneToManyIdentityMapWithOrigKeys(objs, idName) {
   for (var i = 0; i < objs.length; i++) {
     var obj = objs[i];
     var id = obj[idName];
-    var value = kvMap.get(id) || [];
-    value.push(obj);
-    kvMap.set(id, value);
+
+    if (id) {
+      var value = kvMap.get(id) || [];
+      value.push(obj);
+      kvMap.set(id, value);
+    }
   }
   return kvMap;
 }
--- a/test/include_util.test.js
+++ b/test/include_util.test.js
@@ -49,6 +49,19 @@ describe('include_util', function() {
       result.getKeys().should.have.lengthOf(2);  // no additional properties
     });
   });
+
+  describe('#buildOneToOneIdentityMapWithOrigKeys', function() {
+    it('should return an object with keys only if key exist on the object', function() {
+      var objs = [
+        {id: 11, 'fk_id': '11', letter: 'A'},
+        {id: 22, 'fk_id': null, letter: 'B'},
+      ];
+      var result = includeUtils.buildOneToOneIdentityMapWithOrigKeys(objs, 'fk_id');
+      result.get(11).should.be.ok;
+      result.getKeys().should.have.lengthOf(1);  // no additional properties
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

ssh24:
@kobaska Could you please fix the commit message? The commit message needs follow the convention:
```
1. The first line of the commit message should not be longer than 50 characters long.
2. The second line should be empty.
3. Every consecutive line after should not be longer than 72 characters long.
```

kobaska:
@ssh24 I have fixed the commit message

ssh24:
@slnode test please

jannyHou:
hi @kobaska I commented in your PR https://github.com/strongloop/loopback-datasource-juggler/pull/1465 and let's keep discussion here:

Sorry I don't find any issues open for this PR, and wondering do you fix it due to `kvMap.set()` fails when `id` is `null` or `undefined`?

Since when you assign `null` or `undefined` as property to an object, e.g.
```
var foo = {};
foo[null] = 1;
foo[undefined] = 2;
```

They are interpreted as string `'null'` and `'undefined'` and being added to `foo` instead of being ignored:
```
foo: {
  null: 1,
  undefined: 2
}
```

Could you elaborate more what's the context of your fix? Thanks!

kjdelisle:
@jannyHou I think it's not about the keys themselves, but the values.

Working off your example, you could have something like this:
```
var item = {};
item[null] = null;
item[undefined] = undefined;

util.inspect(foo);
// returns '{ null: null, undefined: undefined }'
```

Since there's no particular worth to those items, they should probably be excluded.

kobaska:
yes exactly what @kjdelisle mentions.

The specific use case we are using it for:

An Event could have 0 or many recommended options.

Example events:

- Treat paddock for lice event has 2 options:
1) Fertilise with Avenge product
2) Fertilise with Piranha product

So the fertilise event has optionId set to FERTILISE, which maps to Options model as a foreign key to give the two options.

- Move a mob event has no options to select from so option Id is set to null. 

In this case, the relation would fail for Move mob event, as the key is going to be null. Hence it should exclude that from the lookup.

model configs from loopback:

```
{
  "name": "ProgramEvent",
  "properties": {
     "optionGroup": {
        "type": "string"
    }
  },
  "relations": {
      "recommendationOptions": {
        "type": "hasMany",
        "model": "RecommendationOption",
        "foreignKey": "optionGroup",
        "primaryKey": "optionGroup"
    }
}

{
  "name": "RecommendationOption",
  "base": "RegionalTemplateBaseModel",
  "properties": {
    "orderIndex": {
      "type": "number"
    },
    "optionGroup": {
      "type": "string"
    }
  }
}

```

kjdelisle:
@slnode test please

kjdelisle:
@slnode test please
@kobaska Sorry for the wait, I'm going to run the tests again. You might need to rebase, but the PR still LGTM.

slnode:

slnode:

slnode:

kobaska:
@kjdelisle I have done the rebase

jannyHou:
@slnode test please

kobaska:
@jannyHou @kjdelisle  Can we get this merged please?

kjdelisle:
@slnode test please

kjdelisle:
@kobaska If you're online around this time, then please rebase this and I'll run the tests again. If not, would you be able to give me access to your fork so I can rebase and land this?

kjdelisle:
@kobaska It looks like there are some failing tests:

loopback-connector-postgresql downstream has 3 test failures and loopback-connector-oracle has 20 failures.

I've included a gist with the console output.
https://gist.github.com/kjdelisle/4b4de151541bb85b411a7d0c53b4186a

kjdelisle:
@slnode test please

PR #1470, opened 2017-08-23
package: use qs@6.5.0

kjdelisle:
### Description

Fix for advisory: https://snyk.io/vuln/npm:qs:20170213

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "inflection": "^1.6.0",
     "loopback-connector": "^2.1.0",
     "minimatch": "^3.0.3",
-    "qs": "^3.1.0",
+    "qs": "^6.5.0",
     "shortid": "^2.2.6",
     "strong-globalize": "^2.6.2",
     "traverse": "^0.6.6",


kjdelisle:
A similar PR will be made for 3.x as well.

kjdelisle:
@slnode test please

bajtos:
I vaguely remember that `qs` changed the way how deeply nested values are coerced - think of `?foo[0]=bar&foo[1]=baz`, which directly affects how `filter`/`where` parameters are handled.  By upgrading to a newer version, we may be breaking backwards compatibility.

Could you please post a list of breaking changes between qs versions 3.1.0 and 6.5.0, so that we have a better understanding of possible impacts?

kjdelisle:
@slnode test please

kjdelisle:
@bajtos Strangely enough, I don't see anything in the way of breaking changes beyond the end of support for < Node 4 (which was restored in qs@6.0): https://github.com/ljharb/qs/blob/master/CHANGELOG.md

kjdelisle:
@slnode test please

kjdelisle:
The CI system keeps losing connection to the registry...

bajtos:
I may be wrong. Here is the related pull request - please read through the "outdated" comments to build a better understanding.

https://github.com/strongloop/loopback-datasource-juggler/pull/1143

@doublemarked you are the author of that pull request, do you happen to know if it's safe to upgrade `qs` from `3.x` to `6.x` in respect to how `where` queries are/will be treated?

doublemarked:
@bajtos Looks like those breaking changes still haven't been made in `qs`. Currently they are scheduled for 7.x: https://github.com/ljharb/qs/issues/107

bajtos:
@doublemarked thanks for chiming in! Good to know the breaking changes haven't been made yet.

In which case this change LGTM.

bajtos:
@kjdelisle IIUC, this is effectively dropping support for Node 0.x versions in our 2.x release series. When this change is released to npm, we should write a post to our announcement mailing list, to let people know about this breaking change.

kjdelisle:
@bajtos The support change was reverted:
6.0.0
  - Revert ES6 requirement and restore support for node down to v0.8.

bajtos:
> The support change was reverted in 6.x:

Even better for us 👍

kjdelisle:
@slnode test please

PR #1472, opened 2017-08-23
Add a better way to handle transactions by binding them to models

lehni:
### Description

Based on the way [Objection.js](https://github.com/Vincit/objection.js) is [handling transactions by binding them to models](http://vincit.github.io/objection.js/#binding-models-to-a-transaction), I was looking for a way to do the same in LoopBack. At first I thought I would have to make heavy use of `Proxy` objects, but luckily it turned out that there was already a mechanism in place, but only useable for `memory` and `transient` datasources:

https://github.com/strongloop/loopback-datasource-juggler/blob/6c6df15286400e0dfbc2a458e0a074acc1abeecd/lib/datasource.js#L2068-L2101

It works by making use of `DataSource`'s `copyModel()` which creates a slave model bound to a given data source.

This PR extends this existing infrastructure and adds support for database transactions through the use of `beginTransaction()` and `Transaction.commit()` / `Transaction.rollback()`.

Below the documentation I wrote about its functioning.

I believe this is pretty complete and sturdy ~~, but I need to write unit tests for it still~~.

~~I couldn't find any unit tests for `beginTransaction()` and `Transaction.commit()` / `Transaction.rollback()`, so I think I could use some help with that...~~

And finally, an example of how this can be used in ES6 syntax with `async` / `await`:

```js
// Rollback:
try {
  await app.dataSources.db.transaction(async models => {
    const {MyModel} = models
    console.log(await MyModel.count()) // 0
    await MyModel.create({foo: 'bar'})
    console.log(await MyModel.count()) // 1
    throw new Error('Oops')
  })
} catch (e) {
  console.log(e) // Oops
}
console.log(await app.models.MyModel.count()) // 0

// Commit:
await app.dataSources.db.transaction(async models => {
  const {MyModel} = models
  console.log(await MyModel.count()) // 0
  await MyModel.create({foo: 'bar'})
  console.log(await MyModel.count()) // 1
})
console.log(await app.models.MyModel.count()) // 1

```

```js
/**
 * Run a transaction against the DataSource.
 *
 * This method can be used in different ways based on the passed arguments and
 * type of underlying data source:
 *
 * If no `execute()` function is provided and the underlying DataSource is a
 * database that supports transactions, a Promise is returned that resolves to
 * an EventEmitter representing the transaction once it is ready.
 * `transaction.models` can be used to receive versions of the DataSource's
 * model classes which are bound to the created transaction, so that all their
 * database methods automatically use the transaction. At the end of all
 * database transactions, `transaction.commit()` can be called to commit the
 * transactions, or `transaction.rollback()` to roll them back.
 *
 * If no `execute()` function is provided on a transient or memory DataSource,
 * the EventEmitter representing the transaction is returned immediately. For
 * backward compatibility, this object also supports `transaction.exec()`
 * instead of `transaction.commit()`, and calling `transaction.rollback()` is
 * not required on such transient and memory DataSource instances.
 *
 * If an `execute()` function is provided, then it is called as soon as the
 * transaction is ready, receiving `transaction.models` as its first
 * argument. `transaction.commit()` and `transaction.rollback()` are then
 * automatically called at the end of `execute()`, based on whether exceptions
 * happen during execution or not. If no callback is provided to be called at
 * the end of the execution, a Promise object is returned that is resolved or
 * rejected as soon as the execution is completed, and the transaction is
 * committed or rolled back.
 *
 * @param {Function} execute The execute function, called with (models). Note
 *     that the instances in `models` are bound to the created transaction, and
 *     are therefore not identical with the models in `app.models`, but slaves
 *     thereof (optional).
 * @options {Object} options The options to be passed to `beginTransaction()`
 *     when creating the transaction on database sources (optional).
 * @param {Function} cb Callback called with (err)
 * @returns {Promise | EventEmitter}
 */
```

#### Related issues

- none

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -130,6 +130,45 @@ function errorModelNotFound(idValue) {
   return error;
 }
 
+function invokeConnectorMethod(connector, method, Model, args, options, cb) {
+  var dataSource = Model.getDataSource();
+  // If the DataSource is a transaction and no transaction object is provide in
+  // the options yet, add it to the options, see: DataSource#transaction()
+  var opts = dataSource.isTransaction && !options.transaction ? Object.assign(
+    options, {transaction: dataSource.currentTransaction}) : options;
+  var optionsSupported = connector[method].length >= args.length + 3;
+  var transaction = opts.transaction;
+  if (transaction) {
+    if (!optionsSupported) {
+      return process.nextTick(function() {
+        cb(new Error(g.f(
+          'The connector does not support {{method}} within a transaction', method)));
+      });
+    }
+    // transaction isn't always a Transaction instance. Some tests provide a
+    // string to test if options get passed through, so check for ensureActive:
+    if (transaction.ensureActive && !transaction.ensureActive(cb)) {
+      return;
+    }
+  }
+  var modelName = Model.modelName;
+  var fullArgs;
+  if (!optionsSupported && method === 'count') {
+    // NOTE: The old count() signature is irregular, with `where` coming last:
+    // [modelName, cb, where]
+    var where = args[0];
+    fullArgs = [modelName, cb, where];
+  } else {
+    // Standard signature: [modelName, ...args, (opts, ) cb]
+    fullArgs = [modelName].concat(args);
+    if (optionsSupported) {
+      fullArgs.push(opts);
+    }
+    fullArgs.push(cb);
+  }
+  connector[method].apply(connector, fullArgs);
+}
+
 DataAccessObject._forDB = function(data) {
   if (!(this.getDataSource().isRelational && this.getDataSource().isRelational())) {
     return data;
@@ -365,7 +404,6 @@ DataAccessObject.create = function(data, options, cb) {
     obj.trigger('create', function(createDone) {
       obj.trigger('save', function(saveDone) {
         var _idName = idName(Model);
-        var modelName = Model.modelName;
         var val = removeUndefined(obj.toObject(true));
         function createCallback(err, id, rev) {
           if (id) {
@@ -431,12 +469,8 @@ DataAccessObject.create = function(data, options, cb) {
         };
         Model.notifyObserversOf('persist', context, function(err) {
           if (err) return cb(err);
-
-          if (connector.create.length === 4) {
-            connector.create(modelName, obj.constructor._forDB(context.data), options, createCallback);
-          } else {
-            connector.create(modelName, obj.constructor._forDB(context.data), createCallback);
-          }
+          invokeConnectorMethod(connector, 'create', Model, [obj.constructor._forDB(context.data)],
+            options, createCallback);
         });
       }, obj, cb);
     }, obj, cb);
@@ -579,8 +613,6 @@ DataAccessObject.upsert = function(data, options, cb) {
         Model.applyProperties(update, inst);
         Model = Model.lookupModel(update);
 
-        var connector = self.getConnector();
-
         if (doValidate === false) {
           callConnector();
         } else {
@@ -611,11 +643,7 @@ DataAccessObject.upsert = function(data, options, cb) {
           };
           Model.notifyObserversOf('persist', context, function(err) {
             if (err) return done(err);
-            if (connector.updateOrCreate.length === 4) {
-              connector.updateOrCreate(Model.modelName, update, options, done);
-            } else {
-              connector.updateOrCreate(Model.modelName, update, done);
-            }
+            invokeConnectorMethod(connector, 'updateOrCreate', Model, [update], options, done);
           });
         }
         function done(err, data, info) {
@@ -728,7 +756,6 @@ DataAccessObject.upsertWithWhere = function(where, data, options, cb) {
   var self = this;
   var Model = this;
   var connector = Model.getConnector();
-  var modelName = Model.modelName;
   var query = {where: where};
   var context = {
     Model: Model,
@@ -795,7 +822,7 @@ DataAccessObject.upsertWithWhere = function(where, data, options, cb) {
           };
           Model.notifyObserversOf('persist', context, function(err) {
             if (err) return done(err);
-            connector.upsertWithWhere(modelName, ctx.where, update, options, done);
+            invokeConnectorMethod(connector, 'upsertWithWhere', Model, [ctx.where, update], options, done);
           });
         }
         function done(err, data, info) {
@@ -944,8 +971,6 @@ DataAccessObject.replaceOrCreate = function replaceOrCreate(data, options, cb) {
           Model.applyProperties(update, inst);
           Model = Model.lookupModel(update);
 
-          var connector = self.getConnector();
-
           if (options.validate === false) {
             return callConnector();
           }
@@ -972,7 +997,7 @@ DataAccessObject.replaceOrCreate = function replaceOrCreate(data, options, cb) {
             };
             Model.notifyObserversOf('persist', context, function(err) {
               if (err) return done(err);
-              connector.replaceOrCreate(Model.modelName, context.data, options, done);
+              invokeConnectorMethod(connector, 'replaceOrCreate', Model, [context.data], options, done);
             });
           }
           function done(err, data, info) {
@@ -1097,7 +1122,6 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
   var connector = Model.getConnector();
 
   function _findOrCreate(query, data, currentInstance) {
-    var modelName = self.modelName;
     function findOrCreateCallback(err, data, created) {
       if (err) return cb(err);
       var context = {
@@ -1155,12 +1179,8 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
 
     Model.notifyObserversOf('persist', context, function(err) {
       if (err) return cb(err);
-
-      if (connector.findOrCreate.length === 5) {
-        connector.findOrCreate(modelName, query, self._forDB(context.data), options, findOrCreateCallback);
-      } else {
-        connector.findOrCreate(modelName, query, self._forDB(context.data), findOrCreateCallback);
-      }
+      invokeConnectorMethod(connector, 'findOrCreate', Model, [query, self._forDB(context.data)],
+        options, findOrCreateCallback);
     });
   }
 
@@ -1776,7 +1796,7 @@ DataAccessObject._coerce = function(where, options) {
       }
     } else {
       if (val != null) {
-        const allowExtendedOperators = self._allowExtendedOperators(options);
+        var allowExtendedOperators = self._allowExtendedOperators(options);
         if (operator === null && val instanceof RegExp) {
           // Normalize {name: /A/} to {name: {regexp: /A/}}
           operator = 'regexp';
@@ -1952,8 +1972,8 @@ DataAccessObject.find = function find(query, options, cb) {
             cb(err);
           } else if (Array.isArray(data)) {
             memory.define({
-              properties: self.dataSource.definitions[self.modelName].properties,
-              settings: self.dataSource.definitions[self.modelName].settings,
+              properties: self.dataSource.definitions[modelName].properties,
+              settings: self.dataSource.definitions[modelName].settings,
               model: self,
             });
 
@@ -1993,11 +2013,7 @@ DataAccessObject.find = function find(query, options, cb) {
         }
 
         var geoCallback = options.notify === false ? geoCallbackWithoutNotify : geoCallbackWithNotify;
-        if (connector.all.length === 4) {
-          connector.all(self.modelName, {}, options, geoCallback);
-        } else {
-          connector.all(self.modelName, {}, geoCallback);
-        }
+        invokeConnectorMethod(connector, 'all', self, [{}], options, geoCallback);
       }
       // already handled
       return cb.promise;
@@ -2100,11 +2116,7 @@ DataAccessObject.find = function find(query, options, cb) {
   };
 
   if (options.notify === false) {
-    if (connector.all.length === 4) {
-      connector.all(self.modelName, query, options, allCb);
-    } else {
-      connector.all(self.modelName, query, allCb);
-    }
+    invokeConnectorMethod(connector, 'all', self, [query], options, allCb);
   } else {
     var context =  {
       Model: this,
@@ -2114,10 +2126,7 @@ DataAccessObject.find = function find(query, options, cb) {
     };
     this.notifyObserversOf('access', context, function(err, ctx) {
       if (err) return cb(err);
-
-      connector.all.length === 4 ?
-        connector.all(self.modelName, ctx.query, options, allCb) :
-        connector.all(self.modelName, ctx.query, allCb);
+      invokeConnectorMethod(connector, 'all', self, [ctx.query], options, allCb);
     });
   }
   return cb.promise;
@@ -2258,11 +2267,7 @@ DataAccessObject.destroyAll = function destroyAll(where, options, cb) {
     };
 
     if (whereIsEmpty(where)) {
-      if (connector.destroyAll.length === 4) {
-        connector.destroyAll(Model.modelName, {}, options, done);
-      } else {
-        connector.destroyAll(Model.modelName, {}, done);
-      }
+      invokeConnectorMethod(connector, 'destroyAll', Model, [{}], options, done);
     } else {
       try {
         // Support an optional where object
@@ -2276,11 +2281,7 @@ DataAccessObject.destroyAll = function destroyAll(where, options, cb) {
         });
       }
 
-      if (connector.destroyAll.length === 4) {
-        connector.destroyAll(Model.modelName, where, options, done);
-      } else {
-        connector.destroyAll(Model.modelName, where, done);
-      }
+      invokeConnectorMethod(connector, 'destroyAll', Model, [where], options, done);
     }
 
     function done(err, info) {
@@ -2444,17 +2445,7 @@ DataAccessObject.count = function(where, options, cb) {
   };
   this.notifyObserversOf('access', context, function(err, ctx) {
     if (err) return cb(err);
-    where = ctx.query.where;
-
-    if (connector.count.length <= 3) {
-      // Old signature, please note where is the last
-      // count(model, cb, where)
-      connector.count(Model.modelName, cb, where);
-    } else {
-      // New signature
-      // count(model, where, options, cb)
-      connector.count(Model.modelName, where, options, cb);
-    }
+    invokeConnectorMethod(connector, 'count', Model, [ctx.query.where], options, cb);
   });
   return cb.promise;
 };
@@ -2507,7 +2498,6 @@ DataAccessObject.prototype.save = function(options, cb) {
 
   var inst = this;
   var connector = inst.getConnector();
-  var modelName = Model.modelName;
 
   var context = {
     Model: Model,
@@ -2591,12 +2581,8 @@ DataAccessObject.prototype.save = function(options, cb) {
 
           Model.notifyObserversOf('persist', context, function(err) {
             if (err) return cb(err);
-
-            if (connector.save.length === 4) {
-              connector.save(modelName, inst.constructor._forDB(data), options, saveCallback);
-            } else {
-              connector.save(modelName, inst.constructor._forDB(data), saveCallback);
-            }
+            invokeConnectorMethod(connector, 'save', Model, [Model._forDB(data)],
+              options, saveCallback);
           });
         }, data, cb);
       }, data, cb);
@@ -2770,12 +2756,7 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
     };
     Model.notifyObserversOf('persist', context, function(err, ctx) {
       if (err) return cb(err);
-
-      if (connector.update.length === 5) {
-        connector.update(Model.modelName, where, data, options, updateCallback);
-      } else {
-        connector.update(Model.modelName, where, data, updateCallback);
-      }
+      invokeConnectorMethod(connector, 'update', Model, [where, data], options, updateCallback);
     });
   }
   return cb.promise;
@@ -2907,11 +2888,7 @@ DataAccessObject.prototype.remove =
             });
           }
 
-          if (connector.destroy.length === 4) {
-            connector.destroy(inst.constructor.modelName, id, options, destroyCallback);
-          } else {
-            connector.destroy(inst.constructor.modelName, id, destroyCallback);
-          }
+          invokeConnectorMethod(connector, 'destroy', Model, [id], options, destroyCallback);
         }, null, cb);
       }
       return cb.promise;
@@ -3040,7 +3017,6 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
   if (isPKMissing(Model, cb))
     return cb.promise;
 
-  var model = Model.modelName;
   var hookState = {};
 
   if (id !== data[pkName]) {
@@ -3155,8 +3131,8 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
           options: options,
         };
         Model.notifyObserversOf('persist', ctx, function(err) {
-          connector.replaceById(model, id,
-            inst.constructor._forDB(context.data), options, replaceCallback);
+          invokeConnectorMethod(connector, 'replaceById', Model, [id, Model._forDB(context.data)],
+            options, replaceCallback);
         });
       }
     }
@@ -3216,7 +3192,6 @@ function(data, options, cb) {
 
   var allowExtendedOperators = Model._allowExtendedOperators(options);
   var strict = this.__strict;
-  var model = Model.modelName;
   var hookState = {};
 
   // Convert the data to be plain object so that update won't be confused
@@ -3352,13 +3327,9 @@ function(data, options, cb) {
               options: options,
             };
             Model.notifyObserversOf('persist', ctx, function(err) {
-              if (connector.updateAttributes.length === 5) {
-                connector.updateAttributes(model, getIdValue(inst.constructor, inst),
-                  inst.constructor._forDB(context.data), options, updateAttributesCallback);
-              } else {
-                connector.updateAttributes(model, getIdValue(inst.constructor, inst),
-                  inst.constructor._forDB(context.data), updateAttributesCallback);
-              }
+              invokeConnectorMethod(connector, 'updateAttributes', Model,
+                [getIdValue(Model, inst), Model._forDB(context.data)],
+                options, updateAttributesCallback);
             });
           }, data, cb);
         }, data, cb);
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -27,6 +27,7 @@ var traverse = require('traverse');
 var g = require('strong-globalize')();
 var juggler = require('..');
 var deprecated = require('depd')('loopback-datasource-juggler');
+var Transaction = require('loopback-connector').Transaction;
 
 if (process.env.DEBUG === 'loopback') {
   // For back-compatibility
@@ -2102,10 +2103,51 @@ DataSource.prototype.copyModel = function copyModel(Master) {
 
 /**
  * Run a transaction against the DataSource.
- * @returns {EventEmitter}
- * @private
+ *
+ * This method can be used in different ways based on the passed arguments and
+ * type of underlying data source:
+ *
+ * If no `execute()` function is provided and the underlying DataSource is a
+ * database that supports transactions, a Promise is returned that resolves to
+ * an EventEmitter representing the transaction once it is ready.
+ * `transaction.models` can be used to receive versions of the DataSource's
+ * model classes which are bound to the created transaction, so that all their
+ * database methods automatically use the transaction. At the end of all
+ * database transactions, `transaction.commit()` can be called to commit the
+ * transactions, or `transaction.rollback()` to roll them back.
+ *
+ * If no `execute()` function is provided on a transient or memory DataSource,
+ * the EventEmitter representing the transaction is returned immediately. For
+ * backward compatibility, this object also supports `transaction.exec()`
+ * instead of `transaction.commit()`, and calling `transaction.rollback()` is
+ * not required on such transient and memory DataSource instances.
+ *
+ * If an `execute()` function is provided, then it is called as soon as the
+ * transaction is ready, receiving `transaction.models` as its first
+ * argument. `transaction.commit()` and `transaction.rollback()` are then
+ * automatically called at the end of `execute()`, based on whether exceptions
+ * happen during execution or not. If no callback is provided to be called at
+ * the end of the execution, a Promise object is returned that is resolved or
+ * rejected as soon as the execution is completed, and the transaction is
+ * committed or rolled back.
+ *
+ * @param {Function} execute The execute function, called with (models). Note
+ *     that the instances in `models` are bound to the created transaction, and
+ *     are therefore not identical with the models in `app.models`, but slaves
+ *     thereof (optional).
+ * @options {Object} options The options to be passed to `beginTransaction()`
+ *     when creating the transaction on database sources (optional).
+ * @param {Function} cb Callback called with (err)
+ * @returns {Promise | EventEmitter}
  */
-DataSource.prototype.transaction = function() {
+DataSource.prototype.transaction = function(execute, options, cb) {
+  if (cb === undefined && typeof options === 'function') {
+    cb = options;
+    options = {};
+  } else {
+    options = options || {};
+  }
+
   var dataSource = this;
   var transaction = new EventEmitter();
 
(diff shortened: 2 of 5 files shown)


slnode:

lib/datasource.js
```diff
@@ -2102,39 +2103,122 @@ DataSource.prototype.copyModel = function copyModel(Master) {
...
+  transactionSource.models = modelBuilder.models;
+  transactionSource.definitions = modelBuilder.definitions;
 
-  for (var i in dataSource.modelBuilder.models) {
-    dataSource.copyModel.call(transaction, dataSource.modelBuilder.models[i]);
+  var connector = dataSource.connector;
+  cb = cb || utils.createPromiseCallback();
+  if (connector.beginTransaction) { // Database
+    Transaction.begin(transactionSource.connector, options,
+      function(err, transaction) {
```
bajtos: Can you extract this long callback into a standalone function, to make the code easier to read?

```js
Transaction.begin(transactionSource.connector, options, onTransactionCreated);

// ...

function onTransactionCreated(err, transaction) {
  // ...
}
```

lib/dao.js
```diff
@@ -433,6 +439,7 @@ DataAccessObject.create = function(data, options, cb) {
           if (err) return cb(err);
 
           if (connector.create.length === 4) {
+            options = addTransaction(Model, options);
             connector.create(modelName, obj.constructor._forDB(context.data), options, createCallback);
           } else {
             connector.create(modelName, obj.constructor._forDB(context.data), createCallback);
```
bajtos: When the connector does not support transactions, but the `options` or Model properties indicate that a transaction should be used, then I think we should throw an error. Otherwise our users will think their code is wrapped in a transaction while it is not. This applies to all DAO methods.

lib/datasource.js
```diff
@@ -2102,39 +2103,122 @@ DataSource.prototype.copyModel = function copyModel(Master) {
...
-    transaction.connector.exec(cb);
-  };
+  // we can only copy the models after transaction.connector is set up
+  var models = dataSource.modelBuilder.models;
+  for (var i in models) {
+    dataSource.copyModel.call(transactionSource,  models[i]);
+  }
 
-  return transaction;
+  return connector.transaction ? transactionSource : cb.promise;
```
bajtos: Uff, a return type that allows both a transaction and a promise is pretty confusing to me. Are we expecting all users of `ds.transaction` to check what has been returned? I agree backwards compatibility is important, can we perhaps use the number of arguments to decide whether this function was called in the old way (and the transaction should be returned), or in the new way (and the promise should be returned)?

lib/datasource.js
lehni: Yes!

lib/datasource.js
lehni: Yes, that's definitely possible. Will get on it.

lib/dao.js
lehni: `Model.getDataSource().isTransaction / transaction` will only be set if transactions are supported and `dataSource.transaction(execute)` was used. So I don't think there is a need for exceptions?

slnode:

slnode:

slnode:

lehni:
@bajtos I've addressed your feedback and pushed a new version now.

lehni:
I've worked some more on this now, and fully decoupled the dealing with transient/memory sources VS database sources from the handling of `execute()`. All combinations of these scenarios are now supported and should behave the same. When no `execute()` function is provided, the returned `transaction` object supports `exec()`, `commit()` and `rollback()` in either case.

lehni:
Oh and I forgot to mention: I optimised the generation of these 'slave-models' that are bound to the transaction object by the use of getters: Only the models that are actually accessed are now created as bound versions through `copyModel()`, and only on first access. This should be much faster now and apps with a large amount of models.

lehni:
I've further improved the code now to handle more scenarios of `execute()` and data-sources, and have improved the documentation of it to better describe these possible scenarios.

I've also used `TransactionMixin.beginTransaction()` instead of `Transaction.begin()` to share code that handles timeouts and transaction ids, but I think it would be better to do what I wrote in this TODO:

```js
    // Reuse TransactionMixin.beginTransaction() through a mock model that
    // returns the connector.
    // TODO: Move timeout/id handling from TransactionMixin.beginTransaction()
    // to loopback-connector's Transaction class instead?
```

I consider the code ready again for review now.

lib/dao.js
bajtos: My concern is that the connection between the fact whether connector supports transactions and the number of argument of CRUD methods like `connector.create` is not obvious and easy to unintentionally break in the future. I personally prefer a defensive approach - if we assume that transactions are enabled only when connectors support `options` argument, then it's better to capture/assert that assumption.

For example:

```js
options = addTransaction(Model, options);
if (connector.create.length === 4) {
   connector.create(modelName, ..., options, createCallback);
} else if (options.transaction) {
  throw new Error('The connector does not support create() within a transaction');
} else {
   connector.create(modelName, ..., createCallback);
}
```

lehni: Ok, that makes sense. I could easily move that to `addTransaction()` and call the function `handleTransaction()` instead, so there wouldn't any replication of code. The name of the calling function (`create`) in this example) could be passed to it as a 2nd argument.

lehni: I've just addressed this now in e3d47d8ebb1b77e77ea638e10a512dfadb0aa066 
Please have a look and let me know your thoughts.

lib/dao.js
```diff
@@ -130,6 +130,20 @@ function errorModelNotFound(idValue) {
...
 }
 
+function determineOptions(Model, options, method, numArguments) {
+  var dataSource = Model.getDataSource();
+  var opts = dataSource.isTransaction && !options.transaction ? Object.assign(
+    options, {transaction: dataSource.transaction}) : options;
+  if (dataSource.connector[method].length < numArguments) {
+    if (opts.transaction) {
+      throw new Error(
+        g.f('The connector does not support {{method}} within a transaction', method));
```
bajtos: This will most likely crash the process on an unhandled exception. Can we report the error via the callback please?

It may be easier (and probably more robust too) to change this method to take care of method invocation to.

Here is an example to illustrate my idea:

```js
function invokeConnectorMethod(method, Model, args, options, cb) {
  const dataSource = Model.getDataSource();
  const numArguments = args.length + 3; 
  options = // add transaction if needed
  const optionsSupported = dataSource.connector[method].length >= numArguments;
  if (options.transaction && !optionsSupported) {
    return cb(new Error(/*...*/));
  }
  const connector = dataSource.connector;
  const fullArgs = [Model.modelName].concat(args);
  if (optionsSupported) fullArgs.push(options);
  fullArgs.push(cb);
  connector[method].apply(connector, fullArgs);
}
```

Thoughts?

bajtos:
@slnode ok to test

bajtos:
>  couldn't find any unit tests for beginTransaction() and Transaction.commit() / Transaction.rollback(), so I think I could use some help with that...

I found this test that's doing a very minimal test of the transaction feature: https://github.com/strongloop/loopback-datasource-juggler/blob/2ab4a26396691949627f582998153bc10ea3470d/test/schema.test.js#L36-L63

IMO, we should ideally test all DAO methods to verify that they correctly propagate the transaction. I am proposing to create a new file `test/transactions.test.js` and place all new tests there. Let's keep the existing test unchanged as a guard for backwards compatibility.

Could you please rebase your patch on top of the latest master? I added support for code coverage via `nyc` and integration with coveralls in #1481, it will help us to better asses what test cases we may be missing.

lib/datasource.js
```diff
@@ -2115,26 +2157,125 @@ DataSource.prototype.transaction = function() {
...
+        // thenable object and also calls the passed `cb` function.
+        done = function() {};
+      };
+      try {
+        var result = execute(slaveModels, done);
+        if (result && typeof result.then === 'function') {
+          result.then(function() { done(); }, done);
+        } else if (execute.length < 2) {
+          // execute didn't return a thenable and doesn't receive a callback.
+          // It must executed synchronously, so call done() now.
```
kjdelisle: Minor nitpick:
// It must've executed synchronously, so call done now.

kjdelisle:
Forgot to add:
+9001 for the documentation-quality explanation of the purpose and function of the pull request!

lehni:
@kjdelisle thanks for the feedback and the kind words! I'll get on it shortly.

In the meantime, I'd like to suggest that we get these two PRs merged, so that my code here can rely on that being present for timeouts to work, instead of the dirty trick with `TransactionMixin.beginTransaction()` as described above (https://github.com/strongloop/loopback-datasource-juggler/pull/1472#issuecomment-324613756):

strongloop/loopback-datasource-juggler#1484
strongloop/loopback-connector#116

lib/datasource.js
lehni: Addressed!

lehni:
@bajtos, @kjdelisle regarding tests: I believe that testing the functionality of actual transaction handling os the business of `loopback-connector` and outside of the scope of this RP. What we need to test here is that the various scenarios outlined in the function's doc comment work as described, that the transaction's models are actually bound to the transaction, that `commit()` and `rollback()` get called in the right situations, etc. I am not quite sure how I'll do it yet, but it'll most probably be with a mock data-source that has a `beginTransaction()` method for database sources, and another that has `transaction()` for memory and transient sources. I'll get on it shortly!

lib/dao.js
lehni: You're probably right! I did it this way because of your suggestion to throw an Exception : )
I will look into this now, but the reason why we probably can't do this `invokeConnectorMethod()` is because of irregular method signatures, see https://github.com/lehni/loopback-datasource-juggler/blob/e3d47d8ebb1b77e77ea638e10a512dfadb0aa066/lib/dao.js#L2473-L2475

But we could of course add this as an exception directly in `invokeConnectorMethod()`?

lehni: Implemented in 7d41a8f19af2b034df3221f65e9f58e6dc6fb687

lehni:
@bajtos, @kjdelisle I've added some tests now. Once strongloop/loopback-datasource-juggler#1484 and strongloop/loopback-connector#116 land, I'll add a test for timeouts also. 

But I think generally, this is about how much I think we should test here. The rest really belongs to strongloop/loopback-connector and co, no?

bajtos:
> The rest really belongs to loopback-connector and co., no?

So far, we were keeping the shared test suite here in juggler - see [test/persistence-hooks.suite.js](https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/persistence-hooks.suite.js) for an example. Connectors then include these test files in their test suite, see e.g. [mongodb connector](https://github.com/strongloop/loopback-connector-mongodb/blob/e5b0bb99b80989fb6845268146a8b170bac5fd31/test/persistence-hooks.test.js).

I guess the situation is a bit tricky for transactions, because the only built-in connector (`memory`) does not support them, so we don't have an easy way how to run these tests as part of juggler's `npm test`.

@kjdelisle what's your take on this?

lib/datasource.js
```diff
@@ -2115,26 +2157,116 @@ DataSource.prototype.transaction = function() {
...
+  // with the actual values on first use.
+  var masterModels = dataSource.modelBuilder.models;
+  Object.keys(masterModels).forEach(function(name) {
+    Object.defineProperty(slaveModels, name, {
+      enumerable: true,
+      configurable: true,
+      get: function() {
+        // Delete getter so copyModel() can redefine slaveModels[name].
+        delete slaveModels[name];
+        return dataSource.copyModel.call(transaction, masterModels[name]);
```
bajtos: I find this suspicious - we delete `slaveModels[name]` but don't set any new value. Should we store the result of `dataSource.copyModel` to `slaveModels[name]`?

lib/datasource.js
```diff
@@ -2115,26 +2157,116 @@ DataSource.prototype.transaction = function() {
...
+        // thenable object and also calls the passed `cb` function.
+        done = function() {};
+      };
+      try {
+        var result = execute(slaveModels, done);
+        if (result && typeof result.then === 'function') {
+          result.then(function() { done(); }, done);
+        } else if (execute.length < 2) {
+          // execute didn't return a thenable and doesn't receive a callback.
+          // It must have executed synchronously, so call done() now.
```
bajtos: Is this a safe assumption? What if the function to be executed is reading `arguments` directly? Is it necessary to support the use case of synchronous `execute`?

lib/datasource.js
lehni: `dataSource.copyModel()` sets the new value by itself. I need to delete it so that it can do that, because if it's still there, it wouldn't override. That's part of the caching mechanism, and works well.

lehni: I could store it additionally for clarity, but it's not required...

lib/datasource.js
lehni: I've replicated this from the scripts in `loopback-boot`... I don't think you'd read a `done` method from arguments lists? LB already makes these kind of assumptions in lots of places.

lib/datasource.js
bajtos: Sounds good. Is there any test verifying this? (I.e. accessing the same slave model multiple times?)

lib/datasource.js
bajtos: I see. I personally prefer to keep our API as small as possible and if we don't need to support sync `execute` for backwards compatibility, then I'd rather leave the support out. The less features/code lines we have to maintain, the easier for us :)

lehni:
@bajtos I just meant that since we're already testing transaction functionality separately elsewhere, what we need to test here isn't necessarily that the transactions work, it's that the `transaction(execute)` approach calls `commit()` / `rollback()` for us, exposes the slave models, passes on the transactions to the database methods, etc... And all of that is being tested for now. I can expand it to other database methods, I could test them all through a loop,  but wonder if that's necessary? At the moment, only `create()` is being tested.

lib/datasource.js
bajtos: Having said that, this is a detail and I am fine with going with what you have proposed here.

lib/datasource.js
lehni: That's a good point! I will add a test for it.

lib/datasource.js
lehni: Ok, I agree! It doesn't make much sense to have sync code here anyway. I'll take it out.

bajtos:
> since we're already testing transaction functionality separately elsewhere

Ah, I was not aware that we already have tests verifying that transactions are correctly applied in all DAO methods. In which case I agree with you that we don't need more tests in this PR. Could you please post a link pointing to those existing transaction tests for future readers?

lehni:
I'm not sure it's true for all adapters, but here's the one for postgresql: https://github.com/strongloop/loopback-connector-postgresql/blob/f3ca6436bc4f9d2146ba1f947c9ce5fa62a0f1b9/test/postgresql.transaction.test.js

lib/datasource.js
lehni: I've addressed your feedback in 2ad198278b62aa988d0e66b80638ec970e7774cf

lehni:
One last change for today: I've added support and tests for timeouts in 407dbe4c3366be8ac3c14e47a527e143dab6e7ac

Please have a look and let me know what you think. I'm wondering especially about `err.code === 'TRANSACTION_TIMEOUT'`. I've seen this convention elsewhere but am wondering if it's still the current way of adding error codes.

lib/datasource.js
```diff
@@ -2223,6 +2224,13 @@ DataSource.prototype.transaction = function(execute, options, cb) {
       // retrieved again in determineOptions() in dao.js, as well as referenced
       // in transaction.commit() and transaction.rollback() below.
       transaction.transaction = tx;
+      // Handle timeout and pass it on as an error.
+      tx.observe('timeout', function(context, next) {
+        const err = new Error(g.f('Transaction is rolled back due to timeout'));
+        err.code = 'TRANSACTION_TIMEOUT';
+        done(err);
+        next();
```
bajtos: I think the usual way is to pass errors through the `next` callback. Can we use it here too?

```js
tx.observe('timeout', function(context, next) {
  const err = new Error(g.f('Transaction is rolled back due to timeout'));
  err.code = 'TRANSACTION_TIMEOUT';
  next(err);
});
```

bajtos:
> Merge remote-tracking branch 'upstream/master' into feature/better-tr…

Please use `git rebase master && git push -f` in the future, it makes the final cleanup of git history much easier.

kjdelisle:
@lehni Your other PRs have landed; I'm going to kick off the tests again.
@slnode test please

kjdelisle:
@slnode test please

lib/datasource.js
lehni: Unfortunately that doesn't work as expected: Calling `next(err)` will trigger this default callback in `loopback-connect`: https://github.com/strongloop/loopback-connector/blob/c7e4f5856218724425bf05f0e69bc60f014071dc/lib/transaction.js#L110-L116

And as you can see, if an error is passed to it, the rollback isn't executed there.

So there's the choice between calling:

```js
next(err);
done(err);
```
OR

```js
next();
done(err);
```

OR

```js
done(err);
```

I can't tell right now which one is right, but I'm fairly certain that we should call `done(err)` because otherwise the promise will never resolve...

bajtos: At this point you know more about this part of our codebase than I do, so I am happy to follow your judgment. Maybe call `next` first, so that the transaction rollback is started before we return control back to our caller via `done`?

lehni: I think this is the way to go:

```js
        next(err);
        done(err);
```

`Passing `err` to next()` prevents `tx.rollback()` from being called inside https://github.com/strongloop/loopback-connector/blob/c7e4f5856218724425bf05f0e69bc60f014071dc/lib/transaction.js#L110-L116 which is actually good, because we have no way of getting a callback for when it finished the rollback...

`done(err) handles the rollback in the code of this PR, and also rejects the promise with it at the end of the rollback, which is much more predictable behavior than having it execute in parallel after return.

bajtos: >  I'll add a code comment about it.

Yes please!

lehni:
@bajtos I've fixed the rebase (apologies for that), and improved the timeout handling along with tests for it. I will have another look to see if I can think of more tests, but think I'm pretty much done now otherwise.

lehni:
There is one failing test now, which will only succeed once strongloop/loopback-datasource-juggler#1484 and strongloop/loopback-connector#116 make it into a release that we can request in `package.json`... @bajtos how to go about this?

test/transaction.test.js
```diff
@@ -0,0 +1,216 @@
...
+const DataSource = require('..').DataSource;
+const EventEmitter = require('events');
+const Connector = require('loopback-connector').Connector;
+const Transaction = require('loopback-connector').Transaction;
+const should = require('./init.js');
+
+let callCount;
+let transactionPassed;
+
+function resetState() {
```
bajtos: We prefer to keep helper functions (and class definitions) at the bottom of the test file and start with tests themselves - see http://loopback.io/doc/en/contrib/style-guide.html

bajtos:
>There is one failing test now, which will only succeed once strongloop/loopback-datasource-juggler#1484 and strongloop/loopback-connector#116 make it into a release that we can request in package.json... @bajtos how to go about this?

Is loopback-datasource-juggler release needed? I mean this pull request is changing juggler too, so it should be enough to rebase your feature branch on top of the latest master. Unless the juggler release is needed by the connector?

As for releasing `loopback-connector`, I think we can publish a new semver-minor version any time, but I'd prefer to leave that for @kjdelisle to handle, since he is more familiar with that module and work that may be in progress.

lehni:
@bajtos you're of course right, it's only concerning `loopback-connector`. And I just fixed the tests now.

kjdelisle:
I'll kick off a loopback-connector release shortly.

kjdelisle:
@lehni loopback-connector@4.3.0 has been released; I'll kick off a new test run, but I'm concerned that we're having stability issues in CI.

I think there's some urgent maintenance work that's in progress which could mangle the results.
@slnode test please

kjdelisle:
@lehni Looks like the new loopback-connector build is failing tests due to a duplicate call to `done()`: https://travis-ci.org/strongloop/loopback-connector/jobs/270906536

lehni:
@kjdelisle that's very strange! I just had a similar thing happening with the timeout unit test for this PR here, which runs fine on my own machine but made problems on Travis CI. I believe it has to do with execution speed of the machine, and varies on each run. Here is how I addressed it for this PR: 30921df89626e757e6a2684f6d6534e5bf8dee0b

I think the `loopback-connector` timeout test needs a similar safeguard to prevent double execution of `done()`. It makes sense if you think about it: The timeout happening can't stop the `setTimeout()` callback that will still fire later on, and also call `done()`.

lehni:
@ ssh24 yes it is my intent to squash into one single commit before merge. I kept the separate commits for now as many of the changes related to ongoing discussions.

lehni:
I'be squashed all previous commits into 138b4bb2fb5c20dfd6106f08a630754267e7f410, but then I found out that `invokeConnectorMethod()` wasn't used for all connector methods yet, so I've fixed that in d0332fe1c4559f893413ccc6e63d6e8d0327c22a.

@kjdelisle PTAL, I believe this is what you were referring to earlier.

I consider this work finished now, and will squash again once this discussion is wrapped up.

lehni:
So shall I squash and merge this all in, then?

kjdelisle:
@lehni Sorry, it's been a hectic day today! Let me take a quick peek at the changes one more time.

kjdelisle:
>I think the loopback-connector timeout test needs a similar safeguard to prevent double execution of done(). It makes sense if you think about it: The timeout happening can't stop the setTimeout() callback that will still fire later on, and also call done().

I thought my PR (https://github.com/strongloop/loopback-connector/pull/117) was doing that; once we receive the event, we know that the event is correctly being called, so we unhook it and call done. If you have an alternative proposal for how to do that in the test, toss it into a comment in the review and I'll refactor accordingly. :)

kjdelisle:
@lehni LGTM, merge it at your discretion.

lehni:
Whoop!

lehni:
@kjdelisle Now that this is done, I think we should discuss documentation of this feature. How would I go about getting some of it up on:

https://loopback.io/doc/en/lb3/Using-database-transactions.html

And this feature, along with the discussion in https://github.com/strongloop/loopback-connector/pull/117#issuecomment-327375379 merits a new release, I believe.

bajtos:
Hey @lehni, congratulations on landing your first pull request as a collaborator 👏 

> I think we should discuss documentation of this feature. How would I go about getting a text about it included here: https://loopback.io/doc/en/lb3/Using-database-transactions.html

See http://loopback.io/doc/en/contrib/doc-contrib.html for general guidelines about contributing to our documentation. In order to find the source markdown file for the page you would like to change, you can use the "Edit this page" button/link in the top-right corner - see my screenshot below:

screen shot 2017-09-11 at 10 46 48

The rest is the usual process: create a feature branch, make your changes, open a pull request. You have write access to https://github.com/strongloop/loopback.io, so you can open the feature branch directly there if you like.

> And this feature, along with the discussion in strongloop/loopback-connector#117 (comment) merits a new release of loopback-datasource-juggler, I believe.

Based on https://github.com/strongloop/loopback-connector/pull/117#issuecomment-327829116, I think the new juggler version has been already released?

I'll leave it up to @kjdelisle to handle the release.

zbarbuto:
@lehni Is there any way to get to the actual transaction from inside `execute`? If not, do you think this is something that should probably be added?

e.g

```typescript
let instance = Model.findOne(); // instance I had loaded previously

await app.dataSources.db.transaction(async (models, cb, tx) => {
  const {MyModel} = models
  let target = await MyModel.create({foo: 'bar'})
  // I already have an instance but it isn't bound to the tx
  instance.updateAttributes({ targetId: target.id }, { transaction: tx});
})
```

lehni:
I agree, this makes a lot of sense. I would suggest to add it as the optional 2nd parameter before `cb` though. I’m away until next week, will look into it then!

lehni:
@zbarbuto I've started to look into this now. I see a few issues here currently that I'm not quite sure how to resolve yet:

1. The description of `transaction()` mentions a `transaction` object that is returned if no `execute()` is provided, giving access to `models` (the slave models bound to the transaction), `commit()`, `rollback()` etc. This isn't the actual `transaction` object that needs to be passed to the db functions though, it's a wrapper thereof. Wouldn't it be weird if `execute()` now received the actual `transaction` object, which wouldn't be this one?

2. It makes the most sense to provide this as the 2nd argument, pushing callback to the 3rd place, but this would mean we'd break compatibility with the current version. Since this just rolled out a few days ago I don't think that would be a problem, but it would merit a minor version increment I guess?

@bajtos what do you think we should do here?

bajtos:
> 2] It makes the most sense to provide this as the 2nd argument, pushing callback to the 3rd place, but this would mean we'd break compatibility with the current version. Since this just rolled out a few days ago I don't think that would be a problem, but it would merit a minor version increment I guess?

We are pretty strict about backwards compatibility. If a code working with the latest version published to npmjs.org stops working after upgrade, then such changes must be released as semver-major version.

Unfortunately, releasing a new major version of LoopBack is requires non-trivial amount of work which we are not willing to make now, considering the coming Beta release of https://github.com/strongloop/loopback-next (which we are planning to release as LoopBack 4.0).

> 3] If we decide to insert a 2nd argument there, perhaps we could choose to provide a context object instead as elsewhere in LoopBack, with the properties transaction (in this case being the "internal" transaction object) and models, allowing us to add other things to it in the future, if needed? Although I can't really think of anything else right now, other than perhaps the initial options when calling transaction()...

I am little bit confused. All CRUD methods (`lib/dao.js`) should be accepting an optional `options` argument that's typically used to provide the transaction. AFAICT, `updateAttributes` is already accepting it, see https://github.com/strongloop/loopback-datasource-juggler/blob/3a6ddf927dc7120efe9bdf149b78e4a5b6cd9d8d/lib/dao.js#L3155

IMO, we should continue using the convention of passing the `transaction` as an `options` property.

In which case, if I understand the problem correctly, we need to figure out how deal with two different kinds of "transaction" objects we can have in our app. I think `transaction.currentTransaction` (or perhaps `transaction.current`?) is a reasonable solution. Is it perhaps possible to reliably detect which kind of transaction object we have received and convert the "wrapper" to the "real" transaction automatically for the user? 

Sorry for if my comment is too terse a cryptic, I am short of time :(

zbarbuto:
What about if the proxy `models` result had a `getTransaction` method on it or something similar that returned the transaction. That’s way it stays backwards compatible:


```typescript
db.transaction(async models => {
  let tx = models.getTransaction();
  await instannce.updateAttributes(update, { transaction: tx });
})
```

lehni:
Hmm that feels like a bit like a hack, no?

zbarbuto:
Can’t really think of any other way unless a new method gets added to db with the signature that has the extra tx argument passed to the callback. But that’s just as hack-y imo. In any case I think it probably warrants its issue as it’s been a bit of a pain point for me when trying to use this (have to re-findById every instance I use and can’t pass the tx out to other helper functions, have to pass the whole `models`)

zbarbuto:
Alternative (still hack but backwards compatible):

```typescript
models.models = models
models.transaction = transaction
```

Allowing:

```typescript
db.transaction(({ models, transaction}, cb) => {
 // ...
})
```

lehni:
Yeah I was thinking of that too... I don't think it's a problem to add a 2nd argument since the feature itself just rolled out. Adoption is probably at 0.0001% right now : )

But what about the other question regarding the `transaction` object? I'm a bit split there...

zbarbuto:
I think that is more a case of documentation. If you do not pass in an `execute()` function what the promise is resolving to isn't _really_ the transaction, it's the transaction context (like you sort of hinted to before) which could have the actual `transaction` on it. So the flow would be something like:

```typescript
const transactionCtx = await db.transaction();
const { transaction, models } = transactionCtx;
await model.updateAttributes(update, { context: transactionCtx.transaction })
await models.Foo.find();
await transactionCtx.commit();
```

Although I must admit looking at this it's a bit clumsy - since you're passing in transactions in some places and not in others, based on whether or not the model was retrieved using the proxy. Perhaps a better option would be to provide some method of binding existing model instances to the transaction so they behave like ones that were fetched from the proxy. Something like:

```typescript
const transactionCtx = await db.transaction();
const { transaction, models } = transactionCtx;
transactionCtx.bindToModel(model);
await model.updateAttributes(update) // is now part of the tx
await models.Foo.find();
await transactionCtx.commit();
```

That way the first few lines of your tx execution are binding your models to the tx and then the rest behaves as-is.

lehni:
I somehow didn't see @bajtos' reply earlier. If compatibility must remain, although the docs [here](https://loopback.io/doc/en/lb3/Using-database-transactions.html#higher-level-transaction-api) don't actually mention this 2nd `callback` argument yet (they should!), the only options I see now is `models.getTransaction()` as @zbarbuto proposes (or perhaps more easily, just `models.transaction` ?), or a 2nd `ctx` argument instead of `models`, so that `ctx = {models, transaction}`. To remain backward compatible, `ctx` would also have to functions as the `models` argument, e.g. through JS-style inheritance, perhaps something like:
```js
let ctx = Object.create(models);
ctx.models = models;
ctx.transaction = transaction;
```
But I don't think this is clean though...

I think `models.transaction` is actually the best solution now...

As for the two suggestions by @zbarbuto, neither of those are easily feasible with the current code-base, and I don't have the time to investigate them:

The 1st suggestion doesn't work since `db.transaction()` has been around, and has been returning what I called above the transaction wrapper object. It's in fact a `EventEmitter` object that inherits all properties of the original data-source, and is used internally on the created slave-models as part of the mechanism that binds the models to the transaction. All this code has already been in place from earlier work for transient and memory sources, I've just leveraged it in this PR to add support for db transactions.

The 2nd suggestion isn't possible either, because binding can't be changed on existing models without breaking things elsewhere, since in order to change this binding, we need to change the model's data-source (hence these bound slave models).

@bajtos to clarify: The db methods only ever receive the "internal" transaction object. The wrapper object is created to pretend to be a data-source that is in fact bound to the transaction, but inherits from the real data-source. I'm not explaining this well, but I'm also not the author of this approach. It was in place in the original `dataSource.transaction()` method that only worked with transient and memory sources.

I've called the reference to the internal transaction `transaction.currentTransaction` because `transaction` here really is the slave model's data-source, and is accessed through it in `invokeConnectorMethod()` (look for: `{transaction: dataSource.currentTransaction}`). If you're fine with this simply being called `dataSource.current`, we could add it to the documentation and expose it... In the context of `let transaction = db.transaction(); ... {transaction: transaction.current}` this would make sense, but elsewhere it would seem strange. And currently it's considered and internal property only.

lehni:
A last option, and perhaps the most clean one, could be to make every `transaction` object a proper data-source that inherits from the underlying data-source and works in the way expected by `db.transaction()`. There would be quite a bit of work involved, but I think it may clean up quite a bit of confusion and be clearer to maintain also...

lehni:
@zbarbuto could you try this workaround for now?

```
db.transaction(async models => {
  let tx = models[Object.keys(models)[0]].getDataSource().currentTransaction;
  await instannce.updateAttributes(update, { transaction: tx });
})
```

zbarbuto:
Thanks for the workaround. Is it right right to get the transaction directly from the dataSource? What if you have multiple concurrent transactions happening asynchronously - what is `currentTransaction` on the `dataSource`? The last one?

lehni:
Since you're retrieving the model from the `models` object that is passed to `execute()` it only contains slave-models bound to this transaction. Getting the `currentTransaction` from one of the slave model's dataSource (which is the same as the transaction wrapper object as outlined above) will give you the right transaction. Note that these slave models are only ever bound to one transaction. If you're running multiple transactions concurrently, you end up with multiple versions of such bound slave models, each bound to only one transaction. 

I hope this clarifies it.

lehni:
@bajtos @zbarbuto to clarify my rambling above from yesterday:

We can speak of the *transaction instance* which is provided by `loopback-connector` and can be passed to all the database methods through their `options`, and the *transaction dataSource*, which is created by `dataSource.transaction()`when it creates a *transaction dataSource* that inherits from the main dataSource it is called on, connecting this slave dataSource with the created transaction, and uses it to bind the created slave-models to this transaction.

The only part in this that I don't think is elegant currently is that this *transaction dataSource* isn't actually a `DataSource` instance. It's a `EventEmitter` into which a bunch of properties are copied over.

I can look into creating a `TransactionDataSource` subclass that does all the right things, and then clean up the documentation. I can also look into supporting such a value in `options: { transaction: transactionDataSource }` so that it would retrieve the actual option from it.

And the reason why I didn't call the *transaction instance* `transactionDataSource.transaction` was that on `dataSoure.transaction()` is the main method that we're discussing here. But calling it on a `transactionDataSource` doesn't make sense anyway, so maybe it's ok if this property then becomes the actual *transaction instance* there?

bajtos:
Great discussion! I am afraid I don't have bandwidth to follow in details, I prefer to leave it up to you two to figure this out. From my limited understanding of the problem and your proposals, `models.transaction` looks like the best option to me, especially considering the implementation complexity of other options.

My only concern is about a possible name collision (what if the app has a model called `Transaction`, e.g. in banking or e-commerce). Perhaps `models.tx` would be less probable to collide? `models.getTransaction()` is probably the safest option, since model names usually don't start with a verb. Few more options that come to my mind: `models.$transaction`, `models.$tx`, `models._transaction`, `models._tx`.

zbarbuto:
Yeah naming collisions was exactly why I suggested a getter in the first place. `_transaction` isn’t too bad as it fits the loopback pattern of prefacing private props with underscores.

lehni:
In that case I think I prefer the getter actually, since accessing underscored properties form the outside always feels like I'm doing something I'm not supposed to be doing : /

But regarding naming collisions: `models.transaction` wouldn't collide with `models.Transaction`, and I don't think people use underscore naming for LB models?

lehni:
Some more options

1. We could inspect the arguments of the passed `execute()` function:
  ```js
  function getParameters(fn) {
    var str = fn.toString().match(/^\s*function[^\(]*\(([^\)]*)/)[1];
    return str ? str.split(/\s*,\s*/) : [];
  }

  getParameters(function(models, transaction, cb) {})
  // ["models", "transaction", "cb"]
  ```
  We could simply look at parameter count (`fn.length`) and if `fn.length === 2`, then we'd look at the 2nd parameter name, and assume it's the transaction object if it is `transaction` or `tx`, and fallback to passing the callback in any other case. If `fn.length === 3`, then it's always the transaction.

  This would guarantee backward compatibility. But in the rare case where somebody uses code minification in Node.js, it would always behave in the old way, passing callback 2nd (only if there are 2 parameters)

2. We could call `execute()` with `transactionDataSource` as `this`. You won't be able to use double-arrow functions then, but you could access the `transactionInstance` through `this.currentTransaction` inside a normal `function() {}` callback. 

3. We could move the part of the code that "wraps" the code in the `execute()` with a transaction to  a new function with the new, better behaviour, and call it something like `dataSource.executeTransaction(models, transaction, callback)`, `dataSource.wrapInTransaction(models, transaction, callback)` or `dataSource.runInTransaction(models, transaction, callback)` or something the like? This function could then be used by `db.transaction()` to handle `execute()` also in a backward compatible way, but we could drop it from the documentation there then.

I quite like the last option...

bajtos:
> But regarding naming collisions: `models.transaction` wouldn't collide with `models.Transaction`, and I don't think people use lowercased names for LB models?

Well, we are using `user` extending the built-in `User` model in our examples, I vaguely remember that I may have seen some other examples of lowercased model names in the issues people are reporting.

> accessing underscored properties form the outside always feels like I'm doing something I'm not supposed to be doing

+1

--- 

Regarding `models.getTransaction()` vs. `dataSource.executeTransaction()` or `runInTransaction()` - both approaches are fine with me. I think I like the latter slightly more.

lehni:
@bajtos what about suggestion 2, the use of `this` to run the execute method with?  And what naming do prefer for suggestion 3?

bajtos:
> what about suggestion 2

I find overloading `this` as confusing, it makes the code more difficult to read. I find the inability to use arrow functions as too limiting.

> And what naming do prefer for suggestion 3? It could also simply be runTransaction()?

I don't have any strong opinion, please pick one that looks best to you.

lehni:
Let's see. What feels the best? @zbarbuto, any suggestions?
```js
dataSource.executeTransaction(async (models, transaction) => {
  await outsideInstance.updateAttributes(update, { transaction: tx });
})
```
```js
dataSource.runTransaction(async (models, transaction) => {
  await outsideInstance.updateAttributes(update, { transaction: tx });
})
```
```js
dataSource.runInTransaction(async (models, transaction) => {
  await outsideInstance.updateAttributes(update, { transaction: tx });
})
```

zbarbuto:
My issue with adding another method is that it sort of just adds un-needed method bloat to the datasource object. You have `.transaction` and `.executeTransaction` which do effectively exactly the same thing but one is just a slightly less functional version of the other. Moreover beacuse of the nature of how the committing/rollback works it would make it hard to deprecate `transaction` and just call `.executeTransaction` under the hood. I definitely prefer adding more context to the current return value.

>If there was a case of a collision I think we'd simply not expose the transaction and give the transaction model the precedence.

Doesn't seem fair that if you want to use a `transaction` lowercase model you don't get to use this method of using db transactions.

My vote goes to leaving the method named as-is but appending `.getTransaction` or `.currentTransaction` to the models proxy. The documentation could always be updated to avoid using the variable name `models` to avoid confusion:

```
db.transaction(async context /* or 'execution' */ => {
  const tx = context.getTransaction();
  const { User, Account } = context;
  // ...
})
```

lehni:
Ok but in that case I'll suggest doing this:

We add `context.models` and `context.transaction`, and add all models that do not clash with either of these names to `context` directly as well (cheap, through thin getters that redirect to `models[name]`).

Then you can should use it like this:

```js
db.transaction(async ({models, transaction}) => {
  const { User, Account } = models
  // ...
})
```

And we're remain compatible as long as people don't deal with a model called `models` or `transaction`, so this would still work:

```js
db.transaction(async (models) => {
  const { User, Account } = models
  // ...
})
```

I really think this is fine, as this code has barely been out, and we'll remove the documentation that mentions the `models` argument in favour of `context`. It will also allow us to expose other things without changing the calling signature again, should we need them.

@bajtos do you approve?

zbarbuto:
@lehni I thought the decision was to go for a `getTransaction` getter to retrieve the `transaction` rather than `.transaction` to avoid collisions. Transaction seems like a perfectly reasonable model name to want to use.

>What if the app has a model called Transaction, e.g. in banking or e-commerce
>Naming collisions was exactly why I suggested a getter in the first place
>Doesn't seem fair that if you want to use a transaction lowercase model you don't get to use this method of using db transactions.

If you really want to allow the `db.transaction(async ({models, transaction}) =>` syntax, perhaps add that `transaction` property to the context only if there is no existing `transaction` model in the app and have it implicitly call `getTransaction` under the hood.

lehni:
No decision has been made yet. I liked the separate method best, `getTransaction()` on `models` feels like a hack to me...

`context` seems like the most future-proof approach, as it allows adding more things at a later point if required, but that means we need to start encouraging its use right away.

PR #1480, opened 2017-08-28
Fixing the query where in the related models

f0rmiga:
### Description

Fixed querying with where in the related models.

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #1497, opened 2017-10-05
Remove test

jannyHou:
Related to https://github.com/strongloop/loopback-datasource-juggler/pull/1480

filter with undefined property returns all result not empty, since connectors ignore the property when build query.

Code details see https://github.com/strongloop/loopback-connector/commit/eba7e680598705a778cd9d5abd64bd4a9962d0e4

### DIFF
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -820,15 +820,6 @@ describe('relations', function() {
             done();
           });
         });
-        it('returns empty result when filtering with wrong id key', function(done) {
-          var wrongWhereFilter = {where: {wrongIdKey: samplePatientId}};
-          physician.patients(wrongWhereFilter, function(err, ch) {
-            if (err) return done(err);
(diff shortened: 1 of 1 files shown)


jannyHou:
@kjdelisle This test case passed before not because an expected logic happened, but something like double negative makes a positive(I can elaborate more if necessary), the fix for those two negatives is in https://github.com/strongloop/loopback-datasource-juggler/pull/1491 (tl;dr the real PR is from community https://github.com/strongloop/loopback-datasource-juggler/pull/1480)

jannyHou:
I don't add new test here because the code in `scope.js` is wrong, so no sense to add coverage to it.

Well I thought it would make the community fix check in easier if I remove it in this PR(also fixes 3 connectors) but now it turns out it's better to keep all change in one PR.


Close this one.

PR #1501, opened 2017-10-10
Add nanoid

ai:
### Description

What do you think about adding [NanoID](https://github.com/ai/nanoid).

It similar to `shortid` (random unique ID generator, use full alphabet to generate more compact IDs, compare to UUID).

But it has few very important benefits over `shortid`:

1. `shortid` is not maintained anymore. The project has many issues without any answer. (Honestly, I tried to take maintenance, but current project public API didn’t allow me to fix all these issues. This is why I created `nanoid` to fix all these problems with API).
2. It has big uniformity problem (so bigger collision probability) [85](https://github.com/dylang/shortid/issues/85).
3. `nanoid` use hardware random generator ([docs](https://github.com/ai/nanoid#security)). `shortid` has some problem with RNG [70](https://github.com/dylang/shortid/issues/70).
4. `nanoid` is 10x times faster [benchmark](https://github.com/ai/nanoid#benchmark) and very small (179 bytes).

Compare to UUID, `nanoid` generates smaller ID (in 22 symbols, with the same collision probability).

@kjdelisle I am not sure, what is ID generator policy in this project. Anyway, what are your thoughts?

### Checklist


slnode:

slnode:

slnode:

slnode:

zbarbuto:
I'm not opposed to this idea but a few comments:

1. `package-lock.json` should not be committed in this PR as the repo does not use one
2. If anything I feel like maybe it should be an optional dependency since it's not really required for running loopback
3. Perhaps a better alternative for this is to allow registering custom functions for ids as suggested in the `TODO` comment a few lines below yours. That was people are free to register whatever ID generation they want (e.g. nanoids, hashids or some sort of slug generator function).

ai:
> Remove `package-lock.json`

Oops. Done.

ai:
> If anything I feel like maybe it should be an optional dependency since it's not really required for running loopback

Sure. Should I move `shortid` to optional dependency too?

zbarbuto:
>Sure. Should I move shortid to optional dependency too?

I'll let one of the longer-running loopback devs make the call on that one. But I think my preference would be option 3 - i.e. removing nanoids as a dep completely but allowing people to register an id factory function with a string key.

ai:
@zbarbuto yeap, I agree. Best way is to remove `shortid` and keep only default `uuid`. All other ID generators should be optional.

But I have no experience with this project to make this PR.

zbarbuto:
If you don't feel comfortable making the PR for it yourself make an issue for it assigned to me and I can try to look at it in the next couple of days.

ai:
Done https://github.com/strongloop/loopback-datasource-juggler/issues/1502

Issue #1507, opened 2017-10-13
No access to request context in custom validators


bmatson:
In the scope of fixing issue #984 the options from the request context are passed to all of the "built in" validators.  This was to address an issue with database transactions, but in my case I'm using a tenant ID stored in the auth token to enforce tenancy via a mixin that adds operation hooks.  Because of the fixes in #984 the built in validators "just work" for my use case as well (eg. when the uniqueness validator goes looking for dups, the access hook modifies the where so that it only looks for dups in the requested tenant's records, not all records)

I've got a validation case that can't be handled by one of the built in validators, so I started digging into the code to see how the "options" get passed to a custom validator, and determined that they make it all the way to the code that calls the custom validator, then get dropped there.  

Based on how options passing was done elsewhere, the options "should" be right before the callback, this means for an async validator the function signature becomes validate(err, options, done).  In the PR for the fix for this I'm looking at validate.length (the number of declared formal parameters for the function) to determine if an async validator was declared with three parameters, if so it assumes the validator expects options and passes them before the callback, otherwise it only passes two parameters to maintain compatibility with the documented format for the existing code.  If someone had "existing" code for an async validator that took three parameters (contrary to the documentation), their code would likely break after this change, but given a tradeoff between supporting someone who isn't following the documentation or breaking the pattern for everyone (by tacking the options on the end after the done callback), I opted for consistency in the context passing pattern.  

In the sync case I just tacked them on the end so existing code that doesn't "expect" options isn't affected.

bmatson:
Anyone????  The fix for this is implemented, just sitting there for someone on the loopback side to review?

PR #1508, opened 2017-10-13
Pass context to custom validators

bmatson:
### Description

Adds logic to pass request context to custom validators.

#### Related issues
 - connect to #1507 

### Checklist


slnode:

slnode:

slnode:

slnode:

dhmlau:
@bmatson , there's a linting problem in your commit message.  Could you please fix it?  Here is the error:
```
**  2 problems found:
**    2426075 - added unit tests for options passing to custom vali: First line should be 50 characters or less (saw 57)
**    8bc8846 - updated to allow passing options to custom validato: First line should be 50 characters or less (saw 53)
```

bmatson:
'Scuse me for being descriptive....how's that?

dhmlau:
@slnode test please

fperks:
Any chance of an update on this PR? This is a feature i would really really like to see, since it will fix a lot of issues for me.

bmatson:
Any progress on this PR, the lack of a context is going to mess up transactions for Postgress users, and it causes me headaches with my multi-tennancy code (since the tennant ID is baked into the Access Token)  Anyone???

lib/validations.js
```diff
@@ -379,7 +379,15 @@ function validateCustom(attr, conf, err, options, done) {
     done = options;
     options = {};
   }
-  conf.customValidator.call(this, err, done);
+  if (done === undefined) { // called from synch validator, stick options on end
```
bajtos:
```js
if (!done) {
  // called from a sync validator, pass options as the last argument
  conf.customValidator.call(this, err, options);
  return;
}

if (conf.customValidator.length === 3) {
  // ...
} else {
  // ...
}
```

lib/validations.js
```diff
@@ -379,7 +379,15 @@ function validateCustom(attr, conf, err, options, done) {
...
+    conf.customValidator.call(this, err, options);
+  } else {
+    if (conf.customValidator.length === 3) {  // if they declared the validator with 3 params, they are expecting options
```
bajtos: Please move the comment to the next line

```js
if (/*...*/) {
  // comment goes here
  conf.customValidator.call(/*...*/);
}
```

test/validations.test.js
```diff
@@ -322,6 +322,43 @@ describe('validations', function() {
       });
     });
 
+    it('should pass options to custom synch', function(done) {
```
bajtos:
```js
it('passes options to custom sync validator', done => {
 // ...
});
```

test/validations.test.js
```diff
@@ -322,6 +322,43 @@ describe('validations', function() {
...
+      delete User.validations;
+      User.validate('name', function(err, options) {
+        if (options.options !== 'someValue') err();
```
bajtos: `options.options` looks rather odd to me. I am proposing to rename `options.options` to something like `options.testFlag`.

If `options` object contains unexpected values, then this way of signaling the error will produce a message that's not very helpful. Unfortunately it looks like we don't have any control over the Error instance produced by our validation framework. In which case I am proposing to at least print a `console.error` message.

For example:

```js
if (options.testFlag !== 'someValue') {
  console.error(
    'Unexpected validation options: %j Expected %j',
    options, {testFlag: 'someValue'});
  err();
}
```

test/validations.test.js
```diff
@@ -322,6 +322,43 @@ describe('validations', function() {
...
+      });
+      User.create({name: 'Valid'}, {options: 'someValue'}, function(e, d) {
```
bajtos: `e` is not handled. If you are expecting this operation to pass, then:

```js
if (e) return done(e);
```

test/validations.test.js
```diff
@@ -322,6 +322,43 @@ describe('validations', function() {
...
+        d.updateAttribute('name', null, {options: 'otherValue'}, function(e) {
+          should.exist(e);
+          e.should.be.instanceOf(Error);
```
bajtos: I think this check is redundant considering that the next check is `instanceOf(ValidationError)`.

test/validations.test.js
```diff
@@ -322,6 +322,43 @@ describe('validations', function() {
...
+          e.should.be.instanceOf(ValidationError);
+          d.updateAttribute('name', 'Vasiliy', {options: 'someValue'}, function(e) {
+            should.not.exist(e);
+            done();
+          });
+        });
+      });
+    });
+
+    it('should pass options to custom async', function(done) {
```
bajtos:
```js
it('passes options to async validator', done => {
  // ...
});
```

test/validations.test.js
```diff
@@ -322,6 +322,43 @@ describe('validations', function() {
...
+      delete User.validations;
+      User.validateAsync('name', function(err, options, done) {
+        if (options.options !== 'someValue') err();
```
bajtos: Same comment as above applies here too.

Additionally, your code will call both `err` and `done`, which does not look correct to me. I think you should `return` after calling `err()`.

test/validations.test.js
```diff
@@ -322,6 +322,43 @@ describe('validations', function() {
...
+        process.nextTick(function() { done(); });
+      });
+      User.create({name: 'Valid'}, {options: 'someValue'}, function(e, d) {
```
bajtos: Missing error handling of `e`.

test/validations.test.js
```diff
@@ -322,6 +322,43 @@ describe('validations', function() {
...
+      User.validate('name', function(err, options) {
+        if (options.options !== 'someValue') err();
+      });
+      User.create({name: 'Valid'}, {options: 'someValue'}, function(e, d) {
+        d.updateAttribute('name', null, {options: 'otherValue'}, function(e) {
+          should.exist(e);
+          e.should.be.instanceOf(Error);
+          e.should.be.instanceOf(ValidationError);
+          d.updateAttribute('name', 'Vasiliy', {options: 'someValue'}, function(e) {
+            should.not.exist(e);
```
bajtos:
```js
if (err) return done(err);
```

Your way discard details and the stack trace from the original error and produces a stack trace pointing to this line, which is rather unhelpful.

test/validations.test.js
```diff
@@ -322,6 +322,43 @@ describe('validations', function() {
...
+        if (options.options !== 'someValue') err();
+        process.nextTick(function() { done(); });
+      });
+      User.create({name: 'Valid'}, {options: 'someValue'}, function(e, d) {
+        d.updateAttribute('name', null, {options: 'otherValue'}, function(e) {
+          should.exist(e);
+          e.should.be.instanceOf(Error);
+          e.should.be.instanceOf(ValidationError);
+          d.updateAttribute('name', 'Vasiliy', {options: 'someValue'}, function(e) {
+            should.not.exist(e);
```
bajtos:
```js
if (err) return done(err);
```

Your way discard details and the stack trace from the original error and produces a stack trace pointing to this line, which is rather unhelpful.

bajtos:
@slnode ok to test

test/validations.test.js
bmatson: The error object is a function, so it's not done(err), as documented here: https://apidocs.loopback.io/loopback-datasource-juggler/#validatable-validateasync  it's err(), but you were right about missing the return.

bmatson: Look here at the custom validator docs, when you want to fail validation you call err(), then done(), wonky and different from a lot of other things, but that's how it works.  https://apidocs.loopback.io/loopback-datasource-juggler/#validatable-validateasync

The docs here also should be updated to reflect the options param in the callback.

`
// custom validator which includes optionsFromContext value.
function customValidator(err, options, done) {
        process.nextTick(function () {
            if (this.name === 'bad') err();
            done();
        });
    });`

test/validations.test.js
bmatson: nope, thats correct, you call err() then done()....

test/validations.test.js
```diff
@@ -322,6 +322,47 @@ describe('validations', function() {
...
+      delete User.validations;
+      User.validate('name', function(err, options) {
+        if (options.testFlag !== 'someValue') err();
+      });
+      User.create({name: 'Valid'}, {testFlag: 'someValue'}, function(e, d) {
+        d.updateAttribute('name', null, {testFlag: 'otherValue'}, function(e) {
+          should.exist(e);
+          e.should.be.instanceOf(ValidationError);
+          d.updateAttribute('name', 'Vasiliy', {testFlag: 'someValue'}, function(e) {
+            should.not.exist(e);
```
bajtos:
```js
if (err) return done(err);
```

test/validations.test.js
```diff
@@ -322,6 +322,47 @@ describe('validations', function() {
...
+        }
+        process.nextTick(function() { done(); });
+      });
+      User.create({name: 'Valid'}, {testFlag: 'someValue'}, function(e, d) {
+        if (e) return done(e);
+        d.updateAttribute('name', null, {testFlag: 'otherValue'}, function(e) {
+          should.exist(e);
+          e.should.be.instanceOf(ValidationError);
+          d.updateAttribute('name', 'Vasiliy', {testFlag: 'someValue'}, function(e) {
+            should.not.exist(e);
```
bajtos:
```js
if (err) return done(err);
```

test/validations.test.js
bmatson: The code pattern here is just copied from all the other validation tests above.  The specific case we're concerned with is that when we pass the correct "testFlag" value we either do or do not get an exception when we do the updateAttribute.  In writing the test case it's clearer to state that we are expecting a specific exception, or we are expecting NO exception, which is what this does.  

Put another way, the "test failure" in this case is not the content of the exception, it's either its presence or absence.

bajtos:
What's missing before this patch can be landed:

 - address my comments regarding `should.not.exist(err)`
 - rebase on top of the latest `master`
 - squash the commits into a single one, make sure the commit message follows our [50/72 rule](http://loopback.io/doc/en/contrib/git-commit-messages.html)

I can do these final changes for you if you prefer.

bmatson:
@bajtos If you'd take care of the final cleanup that would be helpful, I'm not really a git guru, so squashing/rebasing is always a bit of a science project because I do it so infrequently.

bajtos:

PR #1509, opened 2017-10-13
Fix basic-querying

jannyHou:
connect to strongloop-internal/scrum-apex#331
### Description

Fix the current test failures[part 1].

### Problem

some connector like postgresql & oracle uses same model name `MyModel` and results in conflict when `automigrate`: creating existing schema `public`


jannyHou:
the mssql failure is not related to what's been fixed in this PR, I am debugging and will create another PR to fix it.
It shouldn't block this PR check in

dhmlau:
Force merging requested by @jannyHou, as the error is not related to the change.

PR #1513, opened 2017-10-27
Allow customizing embedded relation property

zbarbuto:
### Description

Having very little control over embedded properties can be frustrating. My two use-cases for this feature are demonstrated in the tests I have added but basically boils down to setting datasource options on the property.

Cleaner solution would be to make `property` support the property options I suppose but it would be more of a breaking change.

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -2045,7 +2045,10 @@ RelationDefinition.embedsOne = function(modelFrom, modelToRef, params) {
     methods: params.methods,
   });
 
-  var opts = {type: modelTo};
+  var opts = Object.assign(
+    params.options && params.options.property ? params.options.property : {},
+    {type: modelTo}
+  );
 
   if (params.default === true) {
     opts.default = function() { return new modelTo(); };
@@ -2441,9 +2444,14 @@ RelationDefinition.embedsMany = function embedsMany(modelFrom, modelToRef, param
     embed: true,
   });
 
-  modelFrom.dataSource.defineProperty(modelFrom.modelName, propertyName, {
-    type: [modelTo], default: function() { return []; },
-  });
+  var opts = Object.assign(
+    params.options && params.options.property ? params.options.property : {},
+    {
+      type: [modelTo], default: function() { return []; },
+    }
+  );
+
+  modelFrom.dataSource.defineProperty(modelFrom.modelName, propertyName, opts);
 
   if (typeof modelTo.dataSource.connector.generateId !== 'function') {
     modelFrom.validate(propertyName, function(err) {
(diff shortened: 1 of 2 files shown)


zbarbuto:
@slnode test please

zbarbuto:
This also somewhat relates to https://github.com/strongloop/loopback-connector-postgresql/pull/304 as my ultimate intention is to allow querying on nested `embedsOne` relations.

bajtos:
@zbarbuto thank you for the pull request. To be honest, my knowledge of relations (and embedded relations in particular) is very limited. Having said that, your proposal looks good to me, I don't see any obvious issues. 

I'll ask the code owners to review this patch in detail.

lib/relation-definition.js
```diff
@@ -2045,7 +2045,11 @@ RelationDefinition.embedsOne = function(modelFrom, modelToRef, params) {
     methods: params.methods,
   });
 
-  var opts = {type: modelTo};
+  var opts = Object.assign(
+    {},
```
jannyHou: nitpick: `{},` looks like a duplicate, since `params.options && params.options.property ? params.options.property : {}` would guarantee the returned value is not `undefined` nor `null`.

lib/relation-definition.js
```diff
@@ -2441,9 +2445,15 @@ RelationDefinition.embedsMany = function embedsMany(modelFrom, modelToRef, param
     embed: true,
   });
 
-  modelFrom.dataSource.defineProperty(modelFrom.modelName, propertyName, {
-    type: [modelTo], default: function() { return []; },
-  });
+  var opts = Object.assign(
+    {},
```
jannyHou: ditto

lib/relation-definition.js
```diff
@@ -2045,7 +2045,11 @@ RelationDefinition.embedsOne = function(modelFrom, modelToRef, params) {
     methods: params.methods,
   });
 
-  var opts = {type: modelTo};
+  var opts = Object.assign(
+    {},
+    params.options && params.options.property ? params.options.property : {},
```
jannyHou: A personal opinion: naming the property mapping configuration as `property` sounds a little ambiguous when I read the code, how about `propertyMapping` or `propertyConfig`?

lib/relation-definition.js
zbarbuto: The idea behind this was that your original params do not get mutated - however I am happy to change as I see your point.

lib/relation-definition.js
zbarbuto: I can see how it sounds ambiguous but on the other hand isn't `propertyMapping` and `propertyConfig` a bit redundant? Since you're effectively saying `options.propertyOptions` in that context?

zbarbuto:
Unable to see failures due to IBM firewall - though since the PR was passing before the minor change I just made I assume it is unrelated.

jannyHou:
@slnode test please

jannyHou:
@slnode test please

jannyHou:
@zbarbuto we are experiencing some connection problem of the internal registry, your code should be fine.

jannyHou:
@slnode test please

Issue #1514, opened 2017-10-27
Where filter: date comparison doesn't work in nested properties


luisfavila:
# Description

When filtering a nested property (e.g. schedule.happenedOn) which is a date, comparison operators (lt, lte, gt, gte, between) don't work and an empty array is returned instead.

# Steps to reproduce

- Create a model with a nested Date property, .eg.:
```
"properties": {
  "schedule": {
    "type": {
      "duration": {
        "type": "number",
        "required": true
      },
      "happenedOn": {
        "type": "date",
        "required": true
      }
    }
  }
}

````
- Insert record:
```
{
  "schedule": {
    "duration": 2,
    "happenedOn": "2017-10-23T11:30:00.000Z"
  }
}

````

- Query [find] with filter:
```
{
  "where": { 
    "schedule.time": {"gt": "2017-10-20T00:00:00.000Z"}
  }
}

```

# Expected result

You would expect the inserted record to be returned, but an empty array is instead.

# Additional information

win32 x64 8.3.0
+-- loopback@3.14.0
+-- loopback-boot@3.0.1
+-- loopback-component-auto-migrate@0.3.0
+-- loopback-component-explorer@5.2.0
+-- loopback-connector-mongodb@3.2.1
+-- loopback-datasource-juggler@3.12.0

luisfavila:
I have found the bug to be located on `DataAccessObject._coerce` in `dao.js`, where nested properties are not checked and so the ISO string is not converted to DateType.

luisfavila:
#1169 Seems to be related. I am working on the issue, but can't really write test cases for the Date ones since memory connector doesn't support them. How should I proceed @bajtos?

bajtos:
>  I am working on the issue, but can't really write test cases for the Date operations since memory connector doesn't support them. How should I proceed?

@luisfavila sorry for the delay. To be honest, I don't know that much about juggler, could you @kjdelisle or @jannyHou help here?

My first reaction is to add support for Date operations to the memory connector, because I don't see any reason why Date should not be supported?

luisfavila:
No problem @bajtos 

The problem was actually not the lack of support for those operations, but that nested properties were not being casted. I fixed it and also provided tests :)

bajtos:
@luisfavila 
> I fixed it and also provided tests :)

Cool! I reviewed your pull request #1515, please see my comments there.

luisfavila:
[...]

PR #1515, opened 2017-10-27
Coerce nested properties in WHERE filters

luisfavila:
### Description

Loopback doesn't coerce nested properties. This intends to fix it.

#### Related issues

- connect to strongloop/loopback#517
- connect to #1514 
- connect to #1169

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1663,7 +1663,7 @@ DataAccessObject._coerce = function(where, options) {
   }
 
   var props = self.definition.properties;
-  for (var p in where) {
+  eachprop: for (var p in where) {
     // Handle logical operators
     if (p === 'and' || p === 'or' || p === 'nor') {
       var clauses = where[p];
@@ -1683,7 +1683,27 @@ DataAccessObject._coerce = function(where, options) {
     }
     var DataType = props[p] && props[p].type;
     if (!DataType) {
-      continue;
+      // Handle nested properties
+      if (p.indexOf('.') !== -1) {
+        var parts = p.split('.');
+        var end = parts.pop();
+        var obj = props;
+        while (parts.length) {
+          var part = parts.shift();
+          if (typeof obj[part].type !== 'function' ||
+            !obj[part].type.definition ||
+            !obj[part].type.definition.properties) {
+            continue eachprop;
+          }
+          obj = obj[part].type.definition.properties;
+        }
+        DataType = obj[end] && obj[end].type;
+        if (!DataType) {
+          continue;
+        }
+      } else {
+        continue;
+      }
     }
     if (Array.isArray(DataType) || DataType === Array) {
       DataType = DataType[0];
--- a/test/nested-properties.test.js
+++ b/test/nested-properties.test.js
@@ -1,0 +1,70 @@
+// Copyright IBM Corp. 2013,2016. All Rights Reserved.
+// Node module: loopback-datasource-juggler
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+// This test written in mocha+should.js
+'use strict';
+
+/* global getSchema:false, connectorCapabilities:false */
+var should = require('./init.js');
+
+var db, Model;
+
+describe('basic-querying', function() {
+  before(function(done) {
+    var modelDef = {
+      'nested': {
+        'duration': Number,
+        'happenedOn': Date,
+      },
+    };
+
+    db = getSchema();
+    Model = db.define('Model', modelDef);
+    db.automigrate(done);
+  });
+
+  describe('ping', function() {
+    it('should be able to test connections', function(done) {
+      db.ping(function(err) {
+        should.not.exist(err);
+        done();
+      });
+    });
+  });
+
+  describe('find', function() {
+    before(function(done) {
+      db = getSchema();
+      Model.destroyAll(done);
+    });
+
+    it('should query by nested Date: found', function(done) {
+      Model.create({
+        nested: {
+          duration: 2,
+          happenedOn: '2017-10-30T06:07:31.805Z',
+        },
+      }, function(err, m) {
+        should.not.exist(err);
+        should.exist(m.id);
+        console.log(m);
+        Model.find({
+          where: {
+            'nested.happenedOn': {
+              'gt': new Date('2017-10-20T06:07:31.805Z'),
+            },
+          },
+        }, function(err, m) {
+          should.not.exist(err);
+          should.exist(m);
+          console.log(m);
+          m.should.not.be.empty();
+          m[0].should.be.an.instanceOf(Model);
+          done();
+        });
+      });
+    });
+  });
+});
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -389,7 +389,12 @@ Memory.prototype.destroy = function destroy(model, id, options, callback) {
 Memory.prototype.fromDb = function(model, data) {
   if (!data) return null;
   data = deserialize(data);
-  var props = this._models[model].properties;
+  var props;
+  if (typeof model === 'object') {
+    props = model.properties;
+  } else {
+    props = this._models[model].properties;
+  }
   for (var key in data) {
     var val = data[key];
     if (val === undefined || val === null) {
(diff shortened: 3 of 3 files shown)


slnode:

slnode:

slnode:

slnode:

jannyHou:
@slnode test please

luisfavila:
Hi @jannyHou, you're welcome! Could you check #1514? I found this bug due to not being able to query nested Dates with lt/lte/gt/gte, and have wrote tests for those, but they fail since memory connector can't handle Date comparisons. I have the patch running on production and could verify it works with mongodb connector.

luisfavila:
@jannyHou, I've added support for nested properties in memory connector and wrote tests. Let me know if anything else is needed.

fckbo:
Hi we encouter this problem on our project. We use mongodb. How can we get this fix to test it ?

luisfavila:
@fckbo Feel free to use branch `coerce-nested` of my fork: https://github.com/luisfavila/loopback-datasource-juggler/tree/coerce-nested

lib/dao.js
```diff
@@ -1663,7 +1663,7 @@ DataAccessObject._coerce = function(where, options) {
   }
 
   var props = self.definition.properties;
-  for (var p in where) {
+  eachprop: for (var p in where) {
```
bajtos: I personally find jump labels an anti-pattern. Could you please refactor this code by extracting a nested function for this outer loop body?

```js
for (var p in where) {
  this._coerceProperty(p/*, ...*/);
}

// in _coerceProperty
// ...
while (parts.length) {
  var part = parts.shift();
  if (typeof obj[part].type !== 'function' ||
      !obj[part].type.definition ||
      !obj[part].type.definition.properties) {
    return;
  }
  obj = obj[part].type.definition.properties;
}
// ...
```

lib/dao.js
```diff
@@ -1683,7 +1683,27 @@ DataAccessObject._coerce = function(where, options) {
...
+      // Handle nested properties
+      if (p.indexOf('.') !== -1) {
+        var parts = p.split('.');
+        var end = parts.pop();
+        var obj = props;
+        while (parts.length) {
+          var part = parts.shift();
+          if (typeof obj[part].type !== 'function' ||
+            !obj[part].type.definition ||
+            !obj[part].type.definition.properties) {
```
bajtos: We are trying to avoid multi-line `if` conditions, see http://loopback.io/doc/en/contrib/style-guide.html#indentation-of-multi-line-expressions-in-if

lib/dao.js
```diff
@@ -1683,7 +1683,27 @@ DataAccessObject._coerce = function(where, options) {
     }
     var DataType = props[p] && props[p].type;
     if (!DataType) {
-      continue;
+      // Handle nested properties
+      if (p.indexOf('.') !== -1) {
+        var parts = p.split('.');
```
bajtos: Please use `const` instead of `var` in the new code, see http://loopback.io/doc/en/contrib/style-guide.html#indentation-of-multi-line-expressions-in-if

lib/dao.js
```diff
@@ -1683,7 +1683,27 @@ DataAccessObject._coerce = function(where, options) {
     }
     var DataType = props[p] && props[p].type;
     if (!DataType) {
-      continue;
+      // Handle nested properties
```
bajtos: Please consider replacing this comment with calling a named function extracted from the `if` body.

```js
if (!DataType && p.indexOf('.') !== -1) {
  DataType = this._coerceNestedProperty(p/*, ...*/);
} 

if (!DataType) {
  continue;
}
```

test/nested-properties.test.js
```diff
@@ -0,0 +1,70 @@
...
+        'happenedOn': Date,
+      },
+    };
+
+    db = getSchema();
+    Model = db.define('Model', modelDef);
+    db.automigrate(done);
+  });
+
+  describe('ping', function() {
```
bajtos: I am pretty sure `ping` is already tested elsewhere, there is no need to test it again in this file. Please remove this `describe` block.

test/nested-properties.test.js
```diff
@@ -0,0 +1,70 @@
...
+'use strict';
+
+/* global getSchema:false, connectorCapabilities:false */
+var should = require('./init.js');
+
+var db, Model;
+
+describe('basic-querying', function() {
+  before(function(done) {
+    var modelDef = {
```
bajtos: Use `const` instead of `var`.

test/nested-properties.test.js
```diff
@@ -0,0 +1,70 @@
...
+      Model.destroyAll(done);
+    });
+
+    it('should query by nested Date: found', function(done) {
+      Model.create({
+        nested: {
+          duration: 2,
+          happenedOn: '2017-10-30T06:07:31.805Z',
+        },
+      }, function(err, m) {
```
bajtos: We prefer to use Promises in the new tests we are writing, could you please rewrite this test?

```js
it('should query by nested Date: found', function() {
  return Model.create({/*...*/})
    .then(created => {
      should.exist(created.id);
      return Model.find({/*...*/});
    })
    .then(found => {
      should.exist(found);
      found.should.not.be.empty();
      found[0].should.be.an.instanceOf(Model);
    });
```

test/nested-properties.test.js
```diff
@@ -0,0 +1,70 @@
...
+        should.not.exist(err);
```
bajtos: When this assertion fails, it hides the useful stack trace from `err` and shows useless stack trace pointing to this particular line instead. Here is the correct solution:

```js
if (err) return done(err);
```

(This won't be needed anymore after you switch to Promises, but I am still pointing it out for educational purposes.)

test/nested-properties.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+// This test written in mocha+should.js
+'use strict';
+
+/* global getSchema:false, connectorCapabilities:false */
+var should = require('./init.js');
+
+var db, Model;
+
+describe('basic-querying', function() {
```
bajtos: Many of the test files in juggler are executed by our connectors too, in order to verify that CRUD operations work in those connectors as expected. I am not sure if it's worth creating a new file when you are adding just a single new test (I'd personally simply add the new test at the end of the existing `describe('find')` in [test/basic-querying.test.js](https://github.com/strongloop/loopback-datasource-juggler/blob/6bd9fca080d8050f9c89df4a62cdb4f57af99e45/test/basic-querying.test.js#L171-L918), possibly wrapped in a second-level `describe('nested properties')`).

If we decide to keep this new file, then:

 1. This `describe` block should be given a unique name, e.g. `query nested properties`.
 2. The new test file should be added to [test/common.batch.js](https://github.com/strongloop/loopback-datasource-juggler/blob/6bd9fca080d8050f9c89df4a62cdb4f57af99e45/test/common.batch.js) so that connectors pick it up.

test/nested-properties.test.js
```diff
@@ -0,0 +1,70 @@
...
+    it('should query by nested Date: found', function(done) {
+      Model.create({
+        nested: {
+          duration: 2,
+          happenedOn: '2017-10-30T06:07:31.805Z',
+        },
+      }, function(err, m) {
+        should.not.exist(err);
+        should.exist(m.id);
+        console.log(m);
```
bajtos: Forgotten console log (there are more of them in this test, please remove them all).

buzzoids:
Hi, has anyone resolved this issue? Using the patch, it fails on nested Array fields now.

b-admike:
@buzzoids can you please add in more context via code snippets or logs of what failed and what you were trying to do? @luisfavila are you still interested to work on this patch and apply feedback from @jannyHou and @bajtos?

luisfavila:
@b-admike, I'm sorry for the delay, I'm studying and didn't really have the time to finish it.
I've just made the changes @bajtos suggested - just didn't migrate the test to the file suggested since this may be really behind and it'll be easier to merge.

@buzzoids send me some logs or deeper explanation and I'll look into it :)

bajtos:
> I'm not sure why but this merge request did not reflect the changes I commited to the branch and this pull request was totally forgotten.

That's weird.

@luisfavila could you please rename your feature branch and open a new pull request?

PR #1522, opened 2017-11-09
Fix query for related models

joostdebruijn:
### Description
This PR fixes the posibility to query on related models. Please note that all the credits goes to @f0rmiga and @jannyHou who prepared the fix in PR's #1480 and #1491. However, due the fact those PR's are a bit stale, I merged the changes from those two PR's to a single commit (instead of 15 merged commits), because I think this functionality is very welcome in a next release.

#### Related issues

- strongloop/loopback#3634

### Checklist

### DIFF
--- a/lib/scope.js
+++ b/lib/scope.js
@@ -116,30 +116,38 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
         var relatedModel = targetModel.relations[filter.relation].modelTo;
         var IdKey = idName(relatedModel);
 
-        var smartMerge = function(idCond, qWhere, idKey) {
+        // return {inq: [1,2,3]}}
+        var smartMerge = function(idCollection, qWhere) {
+          if (!qWhere[IdKey]) return idCollection;
           var merged = {};
-          var idsA = idCond[idKey].inq;
-          var idsB = qWhere[idKey].inq ? qWhere[idKey].inq : [qWhere[idKey]];
+
+          var idsA = idCollection.inq;
+          var idsB = qWhere[IdKey].inq ? qWhere[IdKey].inq : [qWhere[IdKey]];
+
           var intersect = _.intersectionWith(idsA, idsB, _.isEqual);
-          if (intersect.length === 1) merged[idKey] = intersect[0];
-          if (intersect.length > 1) merged[idKey] = {inq: intersect};
+          if (intersect.length === 1) merged = intersect[0];
+          if (intersect.length > 1) merged = {inq: intersect};
+
           return merged;
         };
-        // Merge queryRelated filter and targetId filter
-        var buildWhere = function() {
-          var IdKeyCondition = {};
-          IdKeyCondition[IdKey] = collectTargetIds(data, IdKey);
-          var mergedWhere = smartMerge(IdKeyCondition, queryRelated.where, IdKey);
-          return mergedWhere;
-        };
+
         if (queryRelated.where !== undefined) {
-          if (!queryRelated.where[IdKey]) {
-            // If it's null, don't bother calling smartMerge or model.find.
-            // Simply return an empty result to the client.
-            return cb(null, []);
-          } else {
-            queryRelated.where = buildWhere();
-          }
+          // Merge queryRelated filter and targetId filter
+          var IdKeyCondition = {};
+          IdKeyCondition[IdKey] = smartMerge(collectTargetIds(data, IdKey),
+            queryRelated.where);
+
+          // if the id in filter doesn't exist after the merge,
+          // return empty result
+          if (_.isObject(IdKeyCondition[IdKey]) && _.isEmpty(IdKeyCondition[IdKey])) return cb(null, []);
+
+          var mergedWhere = {
+            and: [
+              IdKeyCondition,
+              _.omit(queryRelated.where, IdKey),
+            ],
+          };
+          queryRelated.where = mergedWhere;
         } else {
           queryRelated.where = {};
           queryRelated.where[IdKey] = collectTargetIds(data, IdKey);
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -820,12 +820,13 @@ describe('relations', function() {
             done();
           });
         });
-        it('returns empty result when filtering with wrong id key', function(done) {
-          var wrongWhereFilter = {where: {wrongIdKey: samplePatientId}};
-          physician.patients(wrongWhereFilter, function(err, ch) {
+        it('returns patient where name equal to samplePatient name', function(done) {
+          var whereFilter = {where: {name: 'a'}};
+          physician.patients(whereFilter, function(err, ch) {
             if (err) return done(err);
             should.exist(ch);
-            ch.should.have.lengthOf(0);
+            ch.should.have.lengthOf(1);
+            ch[0].name.should.eql('a');
             done();
           });
         });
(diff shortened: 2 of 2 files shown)


slnode:

slnode:

slnode:

jannyHou:
@slnode test please

jannyHou:
@joostdebruijn Thank you I am totally ok with your PR, I will trigger a build.

kjdelisle:
Damn. Those two failures in the cassandra downstream test look real to me. I have no practical experience with that database, or the connector. Let me check and see if we included a docker setup script with the repository.

jannyHou:
@slnode test please

kjdelisle:
@joostdebruijn If you pull down the [cassandra connector](https://github.com/strongloop/loopback-connector-cassandra) and run `source setup.sh`, it'll pull a docker image with cassandra on it. You can then `npm it` to run the unit tests. If you link your local branch against it and try out the tests, you might be able to replicate the failure locally.

test/relations.test.js
```diff
@@ -820,12 +820,23 @@ describe('relations', function() {
...
+          var whereFilter = {where: {name: 'a'}};
+          physician.patients(whereFilter, function(err, ch) {
+            if (err) return done(err);
+            should.exist(ch);
+            ch.should.have.lengthOf(1);
+            ch[0].name.should.eql('a');
+            done();
+          });
+        });
+        it('returns patient where id equal to samplePatientId', function(done) {
```
jannyHou: @joostdebruijn This is a duplicate test case, see https://github.com/joostdebruijn/loopback-datasource-juggler/blob/301505fb27acca1fdddc0463bbd26cc847e6029a/test/relations.test.js#L813, we can remove it, see if it fixes the cassandra failure.

kjdelisle:
So the "local run of cassandra" suggestion was a bit of a crapshoot on my end, I get dozens of test failures. Not sure why...

@raymondfeng Who's our resident expert on cassandra now that Tetsuo has left?

test/relations.test.js
joostdebruijn: Ah, I see. It's removed now.

joostdebruijn:
Just remove the duplicate test case, but I don't expect that this will fix the problem. However, I cannot see the build log of the Cassandra connector at all. I can have a look at the Cassandra docker image later today, but I'm not sure if that will make any sense, given your previous comment.😉 Also, I'm not familiar with Cassandra at all.

jannyHou:
@slnode test please

jannyHou:
@joostdebruijn The cassandra connector's error is:

> imported features relations hasMany through fetch scoped instances with paging filters with filter where returns patient where id equal to samplePatientId:
     Error: id cannot be restricted by more than one relation if it includes an Equal

jannyHou:
I just did a quick search regarding the error, maybe it's caused by Cassandra doesn't support ad-hoc query, try make `name` indexable at [L575](https://github.com/joostdebruijn/loopback-datasource-juggler/blob/0b100fd33fd5c94cd7b159ef56360886f32f183b/test/relations.test.js#L575): 
```js
{name: {type: String, index: true}}
```

see https://stackoverflow.com/questions/21036143/invalidrequestexceptionwhyempid-cannot-be-restricted-by-more-than-one-relation

> CqlConfigHelper.setInputWhereClauses should only be used on indexed columns that are not part of the key.

joostdebruijn:
I just pushed a new commit with name indexed. I tried to run the Casandra docker, but I didn't manage to get it up and running, so can you run the tests again for the new commit?

jannyHou:
@slnode test please

joostdebruijn:
I've the Docker up and running and I can reproduce the two failing tests. Most interesting is that the test case at [L802](https://github.com/joostdebruijn/loopback-datasource-juggler/blob/0b100fd33fd5c94cd7b159ef56360886f32f183b/test/relations.test.js#L802) is failing and the test at [L833](https://github.com/joostdebruijn/loopback-datasource-juggler/blob/0b100fd33fd5c94cd7b159ef56360886f32f183b/test/relations.test.js#L833) not. The only difference is that the last test uses an inq-operator to select on id and the first test not. I'll have a look at it later today. 

However, maybe the functionality of the pull request is not supported by the Cassandra-connector and we should skip this tests for the Cassandra connector by putting it behind a ConnectorCapability-flag. What do you think about that, @jannyHou?

joostdebruijn:
@jannyHou The issue should now be fixed, at least it runned succesfully on my local machine! 😅

jannyHou:
@slnode test please

joostdebruijn:
Hm, this doesn't look good. I think there is a mistake in my new testcase. I'll have a look at it... However, is it possible to look into the build logs of Jenkins?

joostdebruijn:
There was another bug in the code which caused the callback to be called twice in some cases, resulting in failing tests. It did not notice it earlier because I only ran the relevant parts of the test suite. The issue should be resolved now. I have run the full test suite against Cassandra and MSSQL (both failing in the last build).

jannyHou:
@slnode test please

joostdebruijn:
Looks much better now! Only the REST-connector is failing. Locally I see 'Cannot read property 'formatted_address' of undefined'. The error occurs also on the master branche of the REST-connector so I don't think this is related to my pull request. So, I hope the PR can be merged now.

jannyHou:
@joostdebruijn I don't think the rest failure is related to your PR, the same error happened with PR#1517(only a readme typo update), @kjdelisle @dhmlau could you use admin permission to merge this PR? Thanks, the fix is delayed for long time....

lib/scope.js
```diff
@@ -116,30 +116,38 @@ ScopeDefinition.prototype.related = function(receiver, scopeParams, condOrRefres
...
-        var smartMerge = function(idCond, qWhere, idKey) {
+        // return {inq: [1,2,3]}}
+        var smartMerge = function(idCollection, qWhere) {
+          if (!qWhere[IdKey]) return idCollection;
           var merged = {};
-          var idsA = idCond[idKey].inq;
-          var idsB = qWhere[idKey].inq ? qWhere[idKey].inq : [qWhere[idKey]];
+
+          var idsA = idCollection.inq;
+          var idsB = qWhere[IdKey].inq ? qWhere[IdKey].inq : [qWhere[IdKey]];
```
jannyHou: @joostdebruijn hmm, a question, compared with my previous fix, is there any reason why taking out the validation for `typeof inq`?

```js
if (!Array.isArray(qWhere[IdKey].inq)) {
           return qWhere[IdKey];
        }
```

jannyHou: this is the code I am talking about:
https://github.com/strongloop/loopback-datasource-juggler/pull/1491/files#diff-c2a679b29102d5574291e884531cb8aeR126

joostdebruijn: Yes, the check caused that in some cases the lines [128/129](https://github.com/joostdebruijn/loopback-datasource-juggler/blob/dc20fe1d8e809e46007cf1f4f233d6fe3e095213/lib/scope.js#L128) where not executed because of the returing, this led to unexpected query results. When I removed the check, the results were as expected. Maybe we can add a test case which validates if the case you were validating is handled correctly?

jannyHou: @joostdebruijn If user provide an invalid value in `inq`, e.g. `{where: {id: {inq: 'INVALID'}}}`, my code will pass it into the query validator which will returns error, but taking it out results in the merge fails and passes `{}` as id query.....

jannyHou: @joostdebruijn Anyway don't worry I will add a test case and continue the fix in my PR :) Thank you for contribution!! :ship:

joostdebruijn: @jannyHou Ah, I understand the problem! Let me know if I can help you in some way or review the next PR or testcase.

Issue #1466, opened 2017-08-17
Duplicate results with include relation filter in a query


BlueRoboto:
# Description/Steps to reproduce

I generated two models A and B; **B -> hasMany -> A**.
Inside B I putted an include filter, for example :
`  "scope": {
    "include": {
      "relation": "relationA"
    }
  }`
I putted one document in A and 257 documents in B and for each B-document I added a reference of A-document.
In a find query on B-collection the scope clause generates an anomaly when resolve _relationA_. In fact in this case find result generates two related instance of A in each B-document, but there should be only one instance of A in each B-document .
This anomaly vanishes when I use a _limit_ clause at 256 (_inqLimit_) documents or if I remove one document from db (max 256 documents).
If I triple document number (768 B-documents), the find result generate three related instances of A in each B-document.

In my opinion the problem is due to the function _uniq(a)_ that does not work properly and generates wrong results in _include.js_ script.

I have done a solution in this pull request #1418.

**Please analyze my solution or even consider the reported problem and find a suitable solution.**

# Additional information

MongoDB v3.4.5 
Loopback v3.8.0
loopback-connector-mongodb v3.2.0
loopback-datasource-juggler v3.9.2

loay:
Can you please provide a sample repo to reproduce the issue? Thank you 
[How to report a bug](http://loopback.io/doc/en/contrib/Reporting-issues.html)

BlueRoboto:
@loay I have implemented a simple test on this repo: https://github.com/BlueRoboto/loopback-sandbox.
I have used _hasAndBelongsToMany_ instead of _hasMany_ relation.

BlueRoboto:
@loay Have you see my test ?

pbouda:
@loay I ran into the same issue, the given workaround works for me. Is anybody working on this?

pbouda:
I added PR #1526 with my solution.

PR #1526, opened 2017-11-29
Fix bug with bson objects in utils uniq function

pbouda:
### Description

The uniq function does currently not work when the database is mongodb. In the case of mongodb, the function will receive an array of bson object of bson type ObjectID. The indexOf function in uniq will return a different index for the same mongodb ID, as it is wrapped in the ObjectID. This results in an array that has duplicate mongodb IDs. This commit first transforms any ObjectID in the given array to a string representation. We can then use indexOf to return unique items.

Replaces PR: https://github.com/strongloop/loopback-datasource-juggler/pull/1418

#### Related issues

- connect to #1466

### Checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -566,8 +566,10 @@ function uniq(a) {
     return uniqArray;
   }
   assert(Array.isArray(a), 'array argument is required');
-  for (var i = 0, n = a.length; i < n; i++) {
-    if (a.indexOf(a[i]) === i) {
+  var comparableA = a.map(
+    item => item.hasOwnProperty('_bsontype') ? item.toString() : item);
+  for (var i = 0, n = comparableA.length; i < n; i++) {
+    if (comparableA.indexOf(comparableA[i]) === i) {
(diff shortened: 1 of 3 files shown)


kjdelisle:
@slnode test please

kjdelisle:
@pbouda I'm good with this PR, but only if we switch to the bson implementation of ObjectID in the mongodb connector (I've created a linked PR for this).

@raymondfeng Your thoughts?

kjdelisle:
@slnode test please

pbouda:
@kjdelisle yes, makes sense to use bson also in the mongodb connector.

kjdelisle:
@slnode test please

kjdelisle:
@pbouda The failures don't seem related to any of the changes made; once the CI run is done, I'll check any remaining failures and if they're all bogus, then I'll land this.

kjdelisle:
🎉 @pbouda Thank you for your contribution! :tada:

PR #1527, opened 2017-12-10
Also ignore empty string for parsing object/array

xinbenlv:
### Description
Some datasources such as MySQL could use empty string as default value of TEXT field, representing nothing there (same as NULL), it should be treated the same way. Without this, the datasource-juggler could crash.


slnode:

slnode:

slnode:

slnode:

jannyHou:
@slnode test please

PR #1532, opened 2018-01-02
feat: omit default fn for embedsMany

jannyHou:
Related to https://github.com/strongloop/loopback/issues/2882

## Use case

- `Person` EmbeddsMany `Address`
- create a new person `let p = new Person({name: 'Jack', id: 1})`

## Current behaviour

The generated instance is `{name: 'Jack', id: 1, addresses: []}`

## Feature

In this PR I add a flag `omitDefaultEmbeddedItem`

### Flag off (default behaviour)

Same as current behaviour, the new person instance have an empty array property `addresses` 

### Flag on(user need to config it in model.json relations)

The new person instance is `{name: 'Jack', id: 1}`
Please note it doesn't have the embedded item.

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -2471,8 +2471,10 @@ RelationDefinition.embedsMany = function embedsMany(modelFrom, modelToRef, param
 
   var opts = Object.assign(
     params.options && params.options.property ? params.options.property : {},
+    params.options && params.options.omitDefaultEmbeddedItem ? {type: [modelTo]} :
     {
-      type: [modelTo], default: function() { return []; },
+      type: [modelTo],
+      default: function() { return []; },
     }
   );
 
@@ -2920,9 +2922,8 @@ EmbedsMany.prototype.create = function(targetModelData, options, cb) {
   targetModelData = targetModelData || {};
   cb = cb || utils.createPromiseCallback();
 
-  var embeddedList = this.embeddedList();
-
   var inst = this.callScopeMethod('build', targetModelData);
+  var embeddedList = this.embeddedList();
 
   var updateEmbedded = function(callback) {
     if (modelInstance.isNewRecord()) {
@@ -2975,6 +2976,7 @@ EmbedsMany.prototype.build = function(targetModelData) {
   var modelInstance = this.modelInstance;
   var forceId = this.definition.options.forceId;
   var persistent = this.definition.options.persistent;
+  var propertyName = this.definition.keyFrom;
   var connector = modelTo.dataSource.connector;
 
   var pk = this.definition.keyTo;
@@ -3008,8 +3010,10 @@ EmbedsMany.prototype.build = function(targetModelData) {
 
   if (this.definition.options.prepend) {
     embeddedList.unshift(inst);
+    modelInstance[propertyName] = embeddedList;
   } else {
     embeddedList.push(inst);
+    modelInstance[propertyName] = embeddedList;
   }
 
   this.prepareEmbeddedInstance(inst);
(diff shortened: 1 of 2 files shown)


raymondfeng:
Array is always a bit tricky. Some expect to see `[]` while others don't want to see it if there are no items. I'm fine with @jannyHou's proposal to control the behavior with a flag.

lib/relation-definition.js
```diff
@@ -2920,9 +2922,8 @@ EmbedsMany.prototype.create = function(targetModelData, options, cb) {
   targetModelData = targetModelData || {};
   cb = cb || utils.createPromiseCallback();
 
-  var embeddedList = this.embeddedList();
-
   var inst = this.callScopeMethod('build', targetModelData);
+  var embeddedList = this.embeddedList();
```
jannyHou: Reverse the sequence of defining `inst` and `embeddedList`, so `embeddedList` won't refer to a dangle `[]`.

lib/relation-definition.js
```diff
@@ -3008,8 +3010,10 @@ EmbedsMany.prototype.build = function(targetModelData) {
 
   if (this.definition.options.prepend) {
     embeddedList.unshift(inst);
+    modelInstance[propertyName] = embeddedList;
   } else {
     embeddedList.push(inst);
+    modelInstance[propertyName] = embeddedList;
```
jannyHou: Sync parent instance's embedded property when `embeddedList` refers to a dangle `[]`

test/relations.test.js
```diff
@@ -5021,6 +5021,73 @@ describe('relations', function() {
...
+      Address = tmp.define('Address', {street: String});
+      Address.validatesPresenceOf('street');
+
+      db.automigrate(['Person'], done);
+    });
+
+    it('can be declared', function(done) {
+      Person.embedsMany(Address, {
+        options: {
+          omitDefaultEmbeddedItem: true,
```
jannyHou: set flag as true.

test/relations.test.js
```diff
@@ -5021,6 +5021,73 @@ describe('relations', function() {
     });
   });
 
+  describe('embedsMany - omit default value for embedded item', function() {
+    before(function(done) {
+      tmp = getTransientDataSource({defaultIdType: Number});
+      Person = db.define('Person', {name: String});
```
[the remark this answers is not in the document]
jannyHou: @b-admike I didn't think too much about the before function, copied most things from the existing suites, so for your question, my answer may not very insightful:

For the embedded model, we attach it to a transient datasource not a remote database, that's why you see two datasource here: `db` and `tmp`.
Person(attached to `db`) embeds many Address(attached to `tmp`)
The transient datasource is similar to memory.

jannyHou:
@b-admike 
> is the only way to create the embedded model instances is via scoped instances of the source model if the omitDefaultEmbeddedItem flag is set

The way to create embedded model is not affected by the flag, you are right on "it is via scoped property of the parent instance", there are 3 ways: 
- `person.addresses.create()`: create a new embedded item
- `person.addresses.build()`: forget it's difference from `create`...but there is a difference.
- `person.addresses.add()`: add an existing embedded item

The flag itself doesn't impact scoped method I think. It makes difference when you first create the parent instance: its embedded item doesn't get added as an empty array anymore(when flag on).

jannyHou:
The downstream connector failures are not related to this PR.

jannyHou:
@virkt25 
> Just to confirm behaviour, this only omits the embedded property if it's empty [] ... and not if it has items in it right?

Yes 👍

dhmlau:
Force merging because the change is not related to the CI failure.

PR #1534, opened 2018-01-08
Set model constructor name to model name 

bajtos:
Rework the code building model constructors to leverage `Function` class and dynamically emit a constructor function named after the model.

Before this change, all model classes were called "ModelConstructor", which made debugging difficult.

After this change, a model class for model "User" is called "User" 🎉 🎊 

Because not all valid model names are also valid JavaScript identifiers, we implement a simple sanitization technique to cover most common invalid characters (replacing "-", "." and ":" with underscore "_") and fall back to legacy "ModelConstructor" if the model name is still not a valid JS identifier.

_This issue has been itching me for years! My initial attempt to use `new Function` for emitting model class constructor failed, because I was not able to access `ModelBaseClass` from the outer closure in the emitted class constructor. Until finally last night, out of the blue, my dear unconscious mind offered a solution: use `new Function` to emit a factory function that accepts any closure variables as function arguments. Voilà, problem solved._ 🕺 🎉

### Checklist

### DIFF
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -185,16 +185,8 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
   // Create the ModelClass if it doesn't exist or it's resolved (override)
   // TODO: [rfeng] We need to decide what names to use for built-in models such as User.
   if (!ModelClass || !ModelClass.settings.unresolved) {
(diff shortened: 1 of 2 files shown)


bajtos:
@slnode test please

PR #1536, opened 2018-01-11
Prevent unauthorized alteration of records on same table

virkt25:
### Description

Example:

#### Related issues

### Checklist

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -107,6 +107,15 @@ function bindRelationMethods(relation, relationMethod, definition) {
   });
 };
 
+function preventFkOverride(inst, data, fkProp) {
+  var err;
+  if (!fkProp) return err;
+  if (data[fkProp] !== undefined && !idEquals(data[fkProp], inst[fkProp])) {
+    err = new Error(g.f('Cannot override foreign key %s from %s to %s', fkProp, inst[fkProp], data[fkProp]));
+  }
+  return err;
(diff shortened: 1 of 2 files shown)


virkt25:
@slnode test please

lib/relation-definition.js
```diff
@@ -107,6 +107,15 @@ function bindRelationMethods(relation, relationMethod, definition) {
   });
 };
 
+function preventFkOverride(inst, data, fkProp) {
+  var err;
+  if (!fkProp) return err;
```
jannyHou: I think we need to create a new Error then return it? Or maybe I missed some code that creates it?

virkt25: It returns `undefined`, the location where the function is called checks to see if err is `undefined` or not and passes the error to the callback and returns to stop execution.

test/relations.test.js
```diff
@@ -3609,7 +3648,24 @@ describe('relations', function() {
...
+              done();
+            });
+        });
+      });
+    });
+
+    it('should update the related item on scope with same foreign key', function(done) {
+      Supplier.create({name: 'Supplier 2'}, function(e, supplier) {
+        var sid = supplier.id;
+        Supplier.findById(supplierId, function(e, supplier) {
```
jannyHou: `sid` is not used, would it be `supplierId` that used in the code after?

PR #1537, opened 2018-01-11
Prevent unauthorized alteration of records on same table

virkt25:
### Description

Example:

#### Related issues

### Checklist


virkt25:
@slnode test please

PR #1538, opened 2018-01-11
fix unauthorized fk change

virkt25:
### Description

Example:

#### Related issues

### Checklist

### DIFF
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -95,6 +95,22 @@ describe('relations', function() {
         });
       });
 
+      it('should not update FK', function(done) {
+        Book.create(function(err, book) {
+          book.chapters.create({name: 'chapter 1'}, function(err, c) {
+            if (err) return done(err);
+            should.exist(c);
+            c.bookId.should.eql(book.id);
+            c.name.should.eql('chapter 1');
+            book.chapters.updateById(c.id, {name: 'chapter 0', bookId: 10}, function(err, cc) {
+              should.exist(err);
+              err.message.should.startWith('Cannot override foreign key');
+              done();
+            });
+          });
+        });
+      });
+
       it('should create record on scope with promises', function(done) {
         Book.create()
         .then(function(book) {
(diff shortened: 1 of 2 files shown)


rmg:
@slnode test please

test/relations.test.js
```diff
@@ -3600,7 +3639,7 @@ describe('relations', function() {
       .catch(done);
     });
 
-    it('should ignore the foreign key in the update', function(done) {
+    it('should error tring to change the foreign key in the update', function(done) {
```
zbarbuto: typo here - `tring` -> `trying`

zbarbuto:
While I do totally agree with this change on principle - it presents a pretty big security hole in your example - I do think there should be an opt-out for this check at a per-model basis, as there are definitely cases where this sort of behaviour might need to be allowed. For example, a teacher wants to move a student to another teacher's class.

In fact, I'd be tempted to say make this behaviour opt-in as making it on by default could be a potentially breaking change if someone relied on this functionality in their existing app.

virkt25:
I see your concern regarding the change. The reason I believe this should be the default is so new / existing users aren't exposed by default/unintentionally. If this is intentional / required for an application it can be done so by accessing the through model directly. 

The purpose of this check on related methods ensures referential integrity is maintained.

virkt25:
@slnode test please

virkt25:
@zbarbuto Does my comment address your concerns? I’d like to land this PR ASAP.

zbarbuto:
I get your argument about not making it opt-in - but I still potentially think the choice to opt-out should be there. 

Also tests are failing which I don’t have permissions to see. Will have to wait for one of the internal loopback guys to have a look.

kjdelisle:
@zbarbuto I don't think it makes sense to create a config/flag-based way of opting out of this change. Normally, I'm all about avoiding breaking changes, but this was an unintentional flaw, and shouldn't be a part of the v3 spec for loopback-datasource-juggler.

kjdelisle:
@slnode test please

virkt25:
The failing tests are in connector downstream tests which are unrelated to this PR. This can be verified by my other PR created against master with no changes to code. Same tests in same connectors fail.

virkt25:
@zbarbuto May I have your approval if you agree with @kjdelisle 's comment so I can land this PR. :)

zbarbuto:
Approved but please do make those minor tweaks in comments if you get a chance and it’s not too urgent.

PR #1542, opened 2018-01-19
Add warning for datasources with mismatched names

nitro404:
### Description

As discussed [here](https://github.com/strongloop/loopback/pull/3733), @bajtos requested I add a warning for datasource objects which have a different name from what is specified in the corresponding settings object. Please let me know if any adjustments are required. Thanks!

#### Related issues

- connect to strongloop/loopback#3732

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -375,6 +375,14 @@ DataSource.prototype.setup = function(name, settings) {
   name = name || (connector && connector.name);
   this.name = name;
 
+  if (typeof settings === 'object' && typeof settings.name === 'string' && name !== settings.name) {
+    console.warn(
+      'A datasource has a name of %j while a name of %j is specified in ' +
+        'its settings. Please adjust your configuration so these names match. ' +
+        '%j will be assigned as the actual datasource name.',
+      name, settings.name, name);
+  }
+
   if (name && !connector) {
     if (typeof name === 'object') {
       // The first argument might be the connector itself


slnode:

bajtos:
@slnode ok to test

lib/datasource.js
```diff
@@ -375,6 +375,14 @@ DataSource.prototype.setup = function(name, settings) {
   name = name || (connector && connector.name);
   this.name = name;
 
+  if (typeof settings === 'object' && typeof settings.name === 'string' && name !== settings.name) {
+    console.warn(
+      'A datasource has a name of %s while a name of %s is specified in ' +
```
kjdelisle: Minor nit: Can you wrap the formatters in quotes?
`A datasource has a name of "%s" while a name of "%s" is specified in ' + `

bajtos: +1, this is a good idea

bajtos:
> Do we want a quick unit test to validate that this is working?

That's a good question. I'd say it's not worth the effort, unless there is any easy way how to mock `console.warn` and then restore it after the test.

nitro404:
Warning message has been updated.

bajtos:
You forgot to change the third `%s` case.

In my experience, it's better to use `%j` instead of `"%s"`, as it better handles edge cases like a name containing the double-quote character itself, etc.

I amended your commit to speed up landing of this patch.

bajtos:
Landed 🎉  Thank you for the contribution! ❤️

bmatson:
@bajtos The merge that created this warning does not seem to make a whole lot of sense.  It's creating a warning about a problem that EVERYONE following the loopback documented way of setting up a datasource is going to get.  The problem its flagging is actually created by the code itself just before the warning in lines 371-373.

As an example, given a really simple, per-the-documentation mongodb datasource defined as follows:
{"db": {
    "name": "db",
    "connector": "mongodb",
    "url": "mongodb://localhost/engdb"
  }
}

The code at 348 sets connector = "mongodb"
Then the code at 371 overrides the passed in name parameter (which was equal to "db" as defined in the datasource) to be equal to "mongodb", which is what is firing your new warning?????

How exactly is someone supposed to prevent getting this warning????

bajtos:
@bmatson good catch, looks like we overlooked an obvious use case that's probably not covered by tests 🤕 Would you like to contribute a pull request fixing this problem? See http://loopback.io/doc/en/contrib/code-contrib.html to get you started. If you are new to GitHub, then you may find this guide helpful: https://www.nearform.com/blog/first-time-with-open-source/

Sorry for the inconvenience we created by landing this change!

bmatson:
Miroslav,

Not sure that I understand the implications of making the name !==
connector in the follow-on code.  It appears that the follow-on code is
handling a bunch of different oddball ways of initializing datasources that
are really not well documented.  I'm a bit leery of how long it will take
me to pick apart all the different ways that the initialization is supposed
to work based on reverse engineering the code.

Bill

bajtos:
@bmatson I understand your concerns! To be honest, I would be surprised if anybody from the LoopBack team was familiar with all the "oddball" cases the DataSource constructor support, we would have to do the same reverse engineering ourselves too.

I think it's important to preserve the most important use cases (and write unit tests to verify them):
 - when `name` argument is provided, `ds.name` is set to that name
 - when `options.name` is provided, `ds.name` is set to that name
 - when no name is provided, then `ds.name` is set to connector name

I am personally willing to take the risk of breaking more exotic edge case along the way, we can always fix them when somebody fills a complaint.

bajtos:
I think this problem should have been fixed by https://github.com/strongloop/loopback-datasource-juggler/pull/1555. @bmatson could you please verify?

bmatson:
Miroslav,

At a quick glance it looks like #1555 does address the issue.

I need to create a whole new loopback app, since the one I'm working on
mostly requires the fix in PR #1508, so I'm using a forked version.  I
noticed the issue when I accidentally updated my fork to the head....had to
revert that, then just pull in your last release and re-tag, kind of a PITA.

I'll check out the 3.15.2 release in a bit in a simple "clean" project and
let you know what I find.

Bill

bmatson:
Miroslav,

Looks like 3.15.2 fixes the issue I found....any update on my PR #1508?
I'd love to get rid of my fork.....

Bill

bajtos:
> Looks like 3.15.2 fixes the issue I found

Great, thank you for confirming!

> any update on my PR #1508?

Let me take a look.

Issue #1544, opened 2018-01-21
Yarn install throws error with 2.56.0 at current node LTS version


yshing:
# Description/Steps to reproduce

When installing the latest release of using `yarn` (1.2.1 or 1.3.1)
It would throw that. (install using npm works fine.)
`error loopback-datasource-juggler@2.56.0: The engine "node" is incompatible with this module. Expected version ">= 4 <= 6".
`
Then current `engine` setting of main Loopback 2.x branch is `"node": ">=4.0.0"`.
It expected that this package would be using the same setting and should not throw Error when installing using `yarn`.
Loopback 2.x engins states     "node": ">=4.0.0"

# Link to reproduction sandbox

# Expected result
Should not get Error when installing 2.x branch using `yarn` having node in LTS version.

ambrt:
Bumping since Git isn't removing 'stale' after thumb up.

Extarys:
Ouch, wanted to switch from Restify to Loopback, harsh start. Any idea what I should do? The project is barely started. (Is loopback 4 will be ready soon? Maybe I could try that)

joe-at-startupmedia:
Try this with the latest version of yarn. Place the following in your package.json file
```
  "resolutions": {
    "loopback-datasource-juggler": "2.54.1"
  },
```
This will make sure any requirement within nested dependencies resolve to this version (which doesn't enforce <= 6)

PR #1545, opened 2018-01-21
Should follow engines setting as Loopback 2.x

yshing:
### Description
loopback-datasource-juggler 2.x should have the same node engines settings as loopback 2.x

#### Related issues
- connect to #1544 

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "posttest": "npm run lint"
   },
   "engines": {
-    "node": ">= 4 <= 6"
+    "node": ">= 4"
   },
   "devDependencies": {
     "async-iterators": "^0.2.2",


dhmlau:
@jannyHou , saw that the `<=6` restriction is introduced in this PR https://github.com/strongloop/loopback-datasource-juggler/commit/2190f2510f42c7fe542fefd2a021edd614b92df0.  Any reason we need this?  Thanks.

bajtos:
If we decided to go with this change, then the patch must add Node.js 8 to the list of Node.js versions in `.travis.yml` file, so that we can verify that all tests are passing on Node.js 8.x before we mark that version as supported.

joe-at-startupmedia:
For what it's worth I've been running two Loopback 2.x.x application in production for several months using node version 7.10.1

billinghamj:
In what way does it "not run"? We have run Loopback v2 in production using Node 7, 8 and now 9 with no problems.

bajtos:
@joe-at-startupmedia @billinghamj thank you for chiming in and confirming that LoopBack 2.x works for you on Node.js versions newer than 6.x

I checked out our `2.x` branch and run `npm it` on Node.js 8.x. The tests fail with a single error:

```
  1) ModelBuilder processing json files allows model extension:

      AssertionError [ERR_ASSERTION]: 6 == 7
      + expected - actual

      -6
      +7

      at Context.<anonymous> (test/loopback-dl.test.js:1868:12)
```

If this turns out to be something easy to fix, then I don't mind relaxing the `engines` field to include support for Node.js 8.x. As I said in my earlier comment, such change must be accompanied by a change to `.travis.yml` file to add this new version to the platform matrix.

I don't have time to look into this myself. If @yshing does not have bandwidth to update this pull request, then it would be best if one of you could open a new one please.

joe-at-startupmedia:
So if I'm reading this correctly, in addition to the changes made to package.json engines property in the pull request, the node versions need to be added to the following file?
https://github.com/strongloop/loopback-datasource-juggler/blob/2.x/.travis.yml

Also wanted to make the point that my current version is 7 So would we not be adding versions 4,6,7,8,9 To that yaml file?

joe-at-startupmedia:
@virkt25 When I run the test and troubleshoot the failing test at question the issue is that the `id`
property is not being injected or inheritied from the User model. The User model however is not declared with such property resulting in only 6 properties in total, not the expected 7. Where in this code would the `id` be injected?
```json
    var User = modelBuilder.define('User', {
      name: String,
      bio: ModelBuilder.Text,
      approved: Boolean,
      joinedAt: Date,
      age: Number,
    });

    var Customer = User.extend('Customer', {customerId: {type: String, id: true}});

    var customer = new Customer({name: 'Joe', age: 20, customerId: 'c01'});
```

joe-at-startupmedia:
Further investigates indicates that the `customerId` property is considered an `idNames` and therefore `idInjection` is set to false. For this reason because the `customerId` property exists in the model definition, the test should be asserting for 6, not 7. See code below. Can someone please confirm this? @virkt25 @bajtos

```js
  var idInjection = settings.idInjection;
  if (idInjection !== false) {
    // Default to true if undefined
    idInjection = true;
  }

  var idNames = modelDefinition.idNames();
  console.log(idNames, idInjection)
  if (idNames.length > 0) {
    // id already exists
    idInjection = false;
  }

  // Add the id property
  if (idInjection) {
    // Set up the id property
    ModelClass.definition.defineProperty('id', {type: Number, id: 1, generated: true});
  }
```

joe-at-startupmedia:
Test test is failing in node version 4 and 6 because (as I previously expected) the id is not injected. What happens however later in the model-builder is that a non-enumerable "virtual" id propery id defined on the prototype of the the model definition. It seems to me that the test is functioning correct in node version 8 because it it using `for...in` iterator to traverse the object which should only produce non-enumerable properties which in this case `id` property is not. 

In node version 4 and 6 the `for...in` iterator is somehow able to iterate over the non-enumerable "virtual" `id` property. Still looking into this. If anyone has any insights as to why this is happening feel free to chime in.

Portion of ModelBuilder responsible for the definition of the non-enumerable `id` property.

```js
  if (idNames.length === 1) {
    var idProp = idNames[0];
    if (idProp !== 'id') {
      console.log('creating virtual property');
      Object.defineProperty(ModelClass.prototype, 'id', {
        get: function() {
          var idProp = ModelClass.definition.idNames()[0];
          console.log('idProp is: ' + idProp);
          return this.__data && this.__data[idProp];
        },
        configurable: true,
        enumerable: false,
      });
    }
  } else {
```

joe-at-startupmedia:
The issue with non-enumerable properties being iterated in `for..in` loops  in node version 4 and 6 is possibly related to the following v8 bugs reports:
https://bugs.chromium.org/p/v8/issues/detail?id=705
https://github.com/nodejs/node-v0.x-archive/issues/9352

joe-at-startupmedia:
@virkt25 Unfortunately the test cannot be modified to only check for 6 or 7. Because everything below node version 8 doesn't correctly modify the virtual id property via the ModelClass.prototype to set enumerability to false. The issue is the it's actually working correctly in node8 and nto any of the previous version.
The only way to get the test passing in all supported version of node (including 8) is to not iterate over the properties using a for...in. For what it's worth, I do not see this iteration being tested in the newer version of loopback-datasource-juggler, so I am not sure of the significance of what being tested there. 
In any case, I think it would be ashame not to move forward with supporting 8 because this one test is failing. As I've mentioned, the reason why it fails in 8, because versions of node 4 and 6 do not (correctly) honor modification of the id property to set enumerable to false. This is is an edgecase that can only be replicated under the following circumstances:
1) A model is created with an injected `id` property
2) The same model is extended with another id-type property with the intent to override the initial `id` property (e.g. `{name: customerId, id: true}`)
3) The properties are iterated using a `for...in` loop

billinghamj:
Could add a `hasOwnProperty` check perhaps? Or something equivalent which filters out non-enumerable properties.

joe-at-startupmedia:
@virkt25 Oh, I see what you're saying, sorry I misunderstood. So you're proposing to assert for 6 if it's above node version 6, otherwise assert for 7.

joe-at-startupmedia:
@billinghamj Not sure if that would work because the the virtual id property is still enumerable when it shouldn't be (in node <8), despite `Object.defineProperty(ModelClass.protoype, 'id', {is_enumable: false})` in the ModelBuilder.

billinghamj:
I haven't been able to replicate this behaviour, with it not respecting the `enumerable` property:

https://runkit.com/embed/06mr2u01fnld

Perhaps I'm testing for the issue in the wrong way.

joe-at-startupmedia:
@billinghamj Thanks for providing that tool, seems pretty useful. Please toggle node version both 6 and 8. Notice how the id property is only logged in the for...in loop on node version 6.
https://runkit.com/joe-at-startupmedia/runkit-npm-crpc

joe-at-startupmedia:
@billinghamj 
Also see this runkit where I attempt to modify the virtual id. This is slightly different from how the model builder actually works but produces the same result.
https://runkit.com/joe-at-startupmedia/modifying-virtual-id

joe-at-startupmedia:
Actually that previous revision was botched, (been a long day). It actually works when operating outside of the modelbuilder. I suspect this has something to do in the model builder and the extend function. Modification of the prototype is not being applied to the original id.
https://runkit.com/joe-at-startupmedia/modifying-virtual-id/2.0.0
@billinghamj

billinghamj:
@joe-at-startupmedia bear in mind it is `enumerable: false`, rather than `is_enumerable: false`. Though I believe the default value is false anyway

bajtos:
What a mess! I looked into the problem and to be honest, I cannot tell what is the correct (expected) behaviour, i.e. how to rewrite the test to make it pass on all Node.js versions. Considering that this test has been removed in 3.x/`master` branch, I think it's best to remove the problematic assertion in 2.x too.

I sent a pull request to make that change - see https://github.com/strongloop/loopback-datasource-juggler/pull/1584.

I am closing this one as superseded.

Issue #1547, opened 2018-01-30
Data source names are not correctly retained in some cases


nitro404:
# Description/Steps to reproduce

When a data source object is created the name provided is not always correctly assigned to the object instance and it is sometimes substituted with the connector name instead. This is directly related to a prior [issue](https://github.com/strongloop/loopback/issues/3732) and [pull request](https://github.com/strongloop/loopback/pull/3733) on [Loopback](https://github.com/strongloop/loopback) which apparently did not account for all possible situations where data source names are not correctly retained.

# Link to reproduction sandbox

https://github.com/nitro404/retain-datasource-name-sandbox

# Expected result

In the sandbox environment the MongoDB database data source should have a name attribute value of "database" after creation.

# Additional information

This is definitely a substantial bug that should be addressed to ensure that data source objects have the correct name value assigned to them in all situations. I have a [fix](https://github.com/nitro404/loopback-datasource-juggler/tree/hotfix/retain-datasource-name) on a fork of the repository.

PR #1548, opened 2018-01-30
Fix datasource not correctly retaining name value

nitro404:
### Description

Data source names are not correctly retained and assigned to the object instance in some cases.

Hopefully my changes are correct, I'm not super familiar with the code base so it would be helpful to have a second opinion. I can add extra test cases if needed, not sure what might be required, if anything.

#### Related issues

- connect to #1547
- see also https://github.com/strongloop/loopback-datasource-juggler/pull/1542

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -364,18 +364,8 @@ DataSource.prototype.setup = function(name, settings) {
     debug('Settings: %j', this.settings);
   }
 
-  // Disconnected by default
-  this.connected = false;
-  this.connecting = false;
-
-  if (typeof connector === 'string') {
-    name = connector;
-    connector = undefined;
-  }
-  name = name || (connector && connector.name);
-  this.name = name;
-
-  if (typeof settings === 'object' && typeof settings.name === 'string' && name !== settings.name) {
+  if (typeof settings === 'object' && typeof settings.name === 'string' &&
+      typeof name === 'string' && name !== settings.name) {
     console.warn(
       'A datasource has a name of %j while a name of %j is specified in ' +
         'its settings. Please adjust your configuration so these names match. ' +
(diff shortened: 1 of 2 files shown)


slnode:

nitro404:
@bajtos Tagging you here just in case you have any input on this matter, since this related to my prior pull request on Loopback.

nitro404:
Amended my commit with some new unit tests for data source names for a few different cases and added some extra checks to ensure that data sources are correctly assigned their name. I had forgotten to check the settings object for a name as well, so this is also no taken into account and tested. Please let me know if any additional tests or changes are required. Thanks!

bmatson:
@bajtos @kjdelisle @nitro404 I accidentally updated my fork to the head yesterday and started getting your new warning about datasource names.  My code still worked, however your warning was spit out.  After digging into the code in datasource.js, it appears that everyone who is using a "normal" datasource setup (at least what's documented in the LB 3 docs) is going to get your error.  For a datasource like this: 
{"db": {
    "name": "db",
    "connector": "mongodb",
    "url": "mongodb://localhost/engdb"
  }
}
When run against the latest release of loopback (which has some other fixes related to passing names through).  The name "db" and the settings are correctly passed to the setup function, however the if statement at line 348 sets connector = "mongodb", then the code at line 371 overrides the passed in name with the connector value, which triggers the warning you just added.

Here's the offending code:
```
if (typeof connector === 'string') {
    name = connector;
    connector = undefined;
  }
  name = name || (connector && connector.name);
  this.name = name;


```

nitro404:
Hi @bmatson, thanks for your comment. I too had noticed this when working with the latest changes to Loopback as well and set up this pull request as a fix for this issue. There was a use case I had not accounted for in my sandbox testing environment where the connector name was still being applied as the name of the actual data source. Test cases that were previously absent have also been added to this module to account for this as well. Try applying the changes from this pull request to your datasource juggler to see if it resolves the warning. Thanks!

bmatson:
@nitro404 The issue I'm pointing out does happen with this PR, and probably needs to be addressed in its scope before the next "official" release of datasource-juggler (since the initial version of the PR has been landed on master).

nitro404:
Oh interesting, I thought you meant you were getting this with the latest version of Loopback. Fair enough, I'll have to look into it. Sorry for the misunderstanding.

nitro404:
@bmatson I'm not able to replicate your issue with the changes in this pull request.

@kjdelisle Any chance we could get this pull request landed and versioned soon? There's currently warnings being output on the published module version that this pull request fixes.

PR #1554, opened 2018-02-07
feat(datasource): set named datasource with settings option

truongminh:
This pull request attempts to set an the appropriate name for the datasource.
In the existing code, `this.name` refers to the connector name.
In this PR, that name is changed to `this.connectorName`, while keeping the `DataSource`'s methods unchanged to the outside. 
That `.name` is now `settings.name || connector.name` and is used to find the datasource within an application context.
Related comments can be found here:
https://github.com/strongloop/loopback-next/pull/966#discussion_r166274557
@bajtos do you have any suggestion on this?

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -335,14 +335,14 @@ DataSource.prototype.connect = function(callback) {
  * @returns {*}
  * @private
  */
-DataSource.prototype.setup = function(name, settings) {
+DataSource.prototype.setup = function(connectorName, settings) {
   var dataSource = this;
   var connector;
 
   // support single settings object
-  if (name && typeof name === 'object' && !settings) {
-    settings = name;
-    name = undefined;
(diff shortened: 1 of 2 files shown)


slnode:

lib/datasource.js
```diff
@@ -335,14 +335,14 @@ DataSource.prototype.connect = function(callback) {
  * @returns {*}
  * @private
  */
-DataSource.prototype.setup = function(name, settings) {
+DataSource.prototype.setup = function(connectorName, settings) {
```
raymondfeng: The `name` argument is the `name` of the `dataSource`. It should be changed to `dsName` to make it obvious.

raymondfeng:
Close it to favor https://github.com/strongloop/loopback-datasource-juggler/pull/1555

PR #1555, opened 2018-02-08
Use more meaningful names for dataSource and connector

raymondfeng:
### Description

A clean up of https://github.com/strongloop/loopback-datasource-juggler/pull/1554

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #1557, opened 2018-02-09
fix: add more tests to verify new DataSource()

raymondfeng:
### Description

This is a follow-up to https://github.com/strongloop/loopback-datasource-juggler/pull/1555

#### Related issues

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -354,10 +354,16 @@ DataSource.prototype.setup = function(dsName, settings) {
     dsName = undefined;
   }
 
+  if (typeof dsName !== 'string') {
+    dsName = undefined;
+  }
+
   if (typeof settings === 'object') {
     if (settings.initialize) {
       // Settings is the resolved connector instance
       connector = settings;
+      // Set settings to undefined to avoid confusion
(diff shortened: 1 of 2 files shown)


lib/datasource.js
```diff
@@ -354,10 +354,15 @@ DataSource.prototype.setup = function(dsName, settings) {
...
 
+  if (typeof dsName !== 'string') {
+    dsName = undefined;
+  }
+
   if (typeof settings === 'object') {
     if (settings.initialize) {
       // Settings is the resolved connector instance
       connector = settings;
+      settings = {};
```
b-admike: Is there a reason why we're assigning settings to an empty object here?

test/datasource.test.js
```diff
@@ -46,37 +46,95 @@ describe('DataSource', function() {
     }).should.throw(/expected test error/);
   });
 
+  /**
+   * new DataSource(dsName, settings) without setting.name
```
b-admike: did you mean to say `without setting name` or `settings.name`?

raymondfeng:
@b-admike Good catch. Fixed.

PR #1560, opened 2018-02-15
fixes #1559

### LINKED ISSUE #1559: Multiple operators on one field are skipped and only one applied

When querying on a model using a composite filter on a single property *(such as **gte**, and **lte**)*, only one of the operators is applied - sent to database.
**example:**
```js
Model.find({
  where: {
    timestamp: {
      gte: '2018-02-13T08:52:30.893Z',
      lte: '2018-02-15T08:52:30.893Z'
    }
  }
})
```
end up sending to database only one of the operators on the field *(in this case the **gte** operator)*

This is a bug in the **_coerce** function

# Expected result

Both operators should be sent for the property object

# Additional information
**Related Issues:**
- https://github.com/strongloop/loopback-connector-mongodb/issues/176

linux x64 9.5.0
loopback-datasource-juggler@3.15.4 /home/mitsos/Documents/Projects/loopback-datasource-juggler
├── loopback-connector@4.4.0

mitsos1os:
### Description
Allow single field in filter to hold more than one operators at the same time. 
Flow changed a little and an extra function created in order to reuse code needed in different places.
Not sure if this behavior complies with loopback logic of filters... I just faced this incompatibility with MongoDB querying where this behavior is supported.

Also a pull request must be submitted to looback-connector-mongodb because this is also missed there...
#### Related issues

- connect to <https://github.com/strongloop/loopback-datasource-juggler/issues/1559>

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1649,7 +1649,7 @@ function coerceArray(val) {
  * @private
  */
 DataAccessObject._processValue = function processValue(val, operator, DataType, options) {
-  const self = this;
+  var self = this;
   try {
     // Coerce val into an array if it resembles an array-like object
     val = coerceArray(val);
@@ -1658,7 +1658,7 @@ DataAccessObject._processValue = function processValue(val, operator, DataType, 
   }
 
   // Coerce the array items
-  let i;
+  var i;
   if (Array.isArray(val)) {
     for (i = 0; i < val.length; i++) {
       if (val[i] !== null && val[i] !== undefined) {
@@ -1669,7 +1669,7 @@ DataAccessObject._processValue = function processValue(val, operator, DataType, 
     }
   } else {
     if (val != null) {
-      const allowExtendedOperators = self._allowExtendedOperators(options);
+      var allowExtendedOperators = self._allowExtendedOperators(options);
       if (operator === null && val instanceof RegExp) {
         // Do nothing... handle it outside this function
       } else if (operator === 'regexp' && val instanceof RegExp) {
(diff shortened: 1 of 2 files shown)


mitsos1os:
I see filing builds on NodeJS v4... Should I use ES5 only semantics?

mitsos1os:
I haven't received a reply, so I reverted to using ES5 semantics in order for the CI tests to pass... If this is the proper solution, please update the docs that encourage the usage of ES6 when writing loopback-datasource-juggler v3 code

nitro404:
@b-admike ?

b-admike:
@mitsos1os @nitro404 sorry for the late respone. I'll take a look at this PR. Thank you @mitsos1os for your contribution!

b-admike:
@slnode test please

b-admike:
hey @mitsos1os, what do you think of https://github.com/strongloop/loopback-datasource-juggler/issues/1559#issuecomment-395214653? Is what you're trying to do solved by using the `and` condition in your where filter object?

mitsos1os:
Hi there @b-admike .
Yes this is accomplished with the `and` operator. I issued this pull request in order to make it compatible to the way for example **MongoDB** works which accepts multiple operators on a field when match *(especially date matches)*
So in case you do not want this, then this pull request is not useful

nitro404:
I have to agree with https://github.com/strongloop/loopback-datasource-juggler/issues/1559#issuecomment-395214653, as nice as it would be to have this module compatible with MongoDB, it must also function in a database agnostic manner. Operators should be the preferred style of querying in this situation.

PR #1575, opened 2018-04-17
Polyfill Number.isFinite() to support PhantomJS

bajtos:
### Description

When running `npm test` of `loopback` locally, I noticed a test failure when running on PhantomJS browser:

```
PhantomJS 2.1.1 (Mac OS X 0.0.0) app 
  "before each" hook for "can be read via `app.get(key)`" FAILED
	undefined is not a constructor (evaluating 'Number.isFinite(rank)')
```

This change is fixing the problem by adding a polyfill for `Number.isFinite()` function. The API was introduced as part of ES6/ES2015 spec and is not supported by PhantomJS.

#### Related issues

- n/a

### Checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -451,6 +451,10 @@ function deepMergeProperty(base, extras) {
   return mergedProperty;
 }
 
+const numberIsFinite = Number.isFinite || function(value) {
+  return typeof value === 'number' && isFinite(value);
+};
+
 /**
  * Adds a property __rank to array elements of type object {}
  * If an inner element already has the __rank property it is not altered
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -451,6 +451,10 @@ function deepMergeProperty(base, extras) {
   return mergedProperty;
 }
 
+const numberIsFinite = Number.isFinite || function(value) {
+  return typeof value === 'number' && isFinite(value);
```
jannyHou: @bajtos where is the `isFinite()` function from? You may need to import it from somewhere.

Issue #1519, opened 2017-11-06
[BUG] Empty string id is not correctly validated by forceId


simonbrunel:
# Description/Steps to reproduce
- define a model with `forceId: true`
- associate this model to a mongodb data source
- call the API to create a new model with an explicit empty string id
```
curl -X POST --header "Content-Type: application/json" --header "Accept: application/json" -d "{
  \"id\": \"\"
}" "http://localhost:3000/api/foos"
```
- this call creates a db entry with empty id and returns: `{"id": ""}`

Seems related to #1453 (see [my comment](/strongloop/loopback-datasource-juggler/pull/1453#discussion_r144831689))

# Link to reproduction sandbox
https://github.com/simonbrunel/loopback-sandbox/tree/bug/empty-id-mongodb

# Expected result
```json
{
  "error": {
    "name": "ValidationError",
    "status": 422,
    "message": "L'instance `foo` n'est pas valide. Détails : `id` can't be set (value: \"\").",
    "statusCode": 422,
    "details": {
      "context": "foo",
      "codes": {
        "id": [
          "absence"
        ]
      },
      "messages": {
        "id": [
          "can't be set"
        ]
      }
    }
  }
}
```

# Additional information
> node v6.11.3, Windows 10 64bit
```bash
npm ls --prod --depth 0 | grep loopback
loopback-sandbox@1.0.0
+-- loopback@2.39.0
+-- loopback-boot@2.27.0
+-- loopback-component-explorer@2.7.0
+-- loopback-datasource-juggler@2.55.3
```

simonbrunel:
Sandbox sources: [loopback-sandbox-bug-empty-id-mongodb.zip](https://github.com/strongloop/loopback-datasource-juggler/files/1629261/loopback-sandbox-bug-empty-id-mongodb.zip) (repository deleted)

bajtos:
@simonbrunel thank you for reporting this issue and sorry for responding so late! 

Originally, I wanted to respond to your comment https://github.com/strongloop/loopback-datasource-juggler/pull/1453#discussion_r144831689 and say that I am not sure if your analysis of 
 our implementation is correct. 

[...]

PR #1576, opened 2018-04-28
Fix forceId to reject empty strings

Yaty:
Status : WIP

### Description

Add allowBlank and allowNull opions to validatesAbsenceOf
Make allowBlank and allowNull to false when forceId if set to true

My test are passing but I broke other tests like this one : https://github.com/strongloop/loopback-datasource-juggler/blob/1ef38a6929ae100c73030b720097ba5c4b90a7b3/test/optional-validation.test.js#L117
It seems that the validation is made before the ID is generated so it breaks.
This is the first time I'm working on this codebase, any tips where I should look :) ?

#### Related issues

#1519

### Checklist

### DIFF
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -351,7 +351,7 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
     }
 
     if (forceId) {
-      ModelClass.validatesAbsenceOf(idName, {if: 'isNewRecord'});
+      ModelClass.validatesAbsenceOf(idName, {if: 'isNewRecord', allowBlank: false, allowNull: false});
     }
 
     ModelClass.definition.properties[idName].updateOnly = !!forceId;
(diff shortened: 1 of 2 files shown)


jannyHou:
@slnode test please

nitro404:
It would probably be beneficial to squash or amend these commits together as they both seem to have the same message, and is a bit confusing, unless there is some reason to keep them separate.

bajtos:
Hello @Yaty, thank you for the pull request. The proposed changes look reasonable to me, we need to polish the edge cases now.

> My test are passing but I broke other tests like this one 

I think this is a sign that your change will probably break existing applications. IIUC, the intention is to reject `null` and empty strings. However, we must still allow `undefined` as a valid (and required!) value.

Perhaps `nullCheck` treats `undefined` the same way as `null`?

PR #1578, opened 2018-05-03
Add TypeScript declaration files

raymondfeng:
### Description

This PR adds basic TypeScript declarations for juggler related types to be referenced in LoopBack 4.

#### Related issues

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,5 +13,5 @@ npm-debug.log
 .project
 test/memory.json
 .nyc_output
-
+dist
 
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,18 @@
+{
+  "$schema": "http://json.schemastore.org/tsconfig",
+  "compilerOptions": {
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "noImplicitAny": true,
+    "strictNullChecks": true,
+
+    "lib": ["es2018", "dom"],
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "target": "es2017",
+    "sourceMap": true,
+    "declaration": true
+  },
+  "include": ["types/", "index.d.ts"],
+  "exclude": ["node_modules/**"]
+}
--- a/types/common.d.ts
+++ b/types/common.d.ts
@@ -1,0 +1,28 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: loopback-datasource-juggler
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+/**
+ * Objects with open properties
+ */
+export interface AnyObject<T = any> {
+  [property: string]: T;
+}
+
+/**
+ * Type alias for options object
+ */
+export type Options = AnyObject<any>;
+
+/**
+ * Type alias for Node.js callback functions
+ */
+export type Callback<T = any> = (err: any, result?: T) => void;
+
+/**
+ * Return export type for promisified Node.js async methods.
+ *
+ * Note that juggler uses Bluebird, not the native Promise.
+ */
+export type PromiseOrVoid<T = any> = PromiseLike<T> | void;
--- a/types/datasource.d.ts
+++ b/types/datasource.d.ts
@@ -1,0 +1,177 @@
+// Copyright IBM Corp. 2018. All Rights Reserved.
+// Node module: loopback-datasource-juggler
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+import {AnyObject, Callback, Options, PromiseOrVoid} from './common';
+import {Connector} from './connector';
+import {
+  ModelBaseClass,
+  ModelBuilder,
+  ModelDefinition,
+  PropertyDefinition,
+} from './model';
+
+/**
+ * LoopBack models can manipulate data via the DataSource object.
+ * Attaching a `DataSource` to a `Model` adds instance methods and static methods to the `Model`.
+ *
+ * Define a data source to persist model data.
+ * To create a DataSource programmatically, call `createDataSource()` on the LoopBack object; for example:
+ * ```js
+ * var oracle = loopback.createDataSource({
+ *   connector: 'oracle',
+ *   host: '111.22.333.44',
+ *   database: 'MYDB',
+ *   username: 'username',
+ *   password: 'password'
+ * });
+ * ```
+ *
+ * All classes in single dataSource share same the connector type and
+ * one database connection.
+ *
+ * For example, the following creates a DataSource, and waits for a connection callback.
+ *
+ * ```
+ * var dataSource = new DataSource('mysql', { database: 'myapp_test' });
+ * dataSource.define(...);
+ * dataSource.on('connected', function () {
+ *     // work with database
+ * });
+ * ```
+ * @class DataSource
+ * @param {String} [name] Optional name for datasource.
+ * @options {Object} settings Database-specific settings to establish connection (settings depend on specific connector).
+ * The table below lists a typical set for a relational database.
+ * @property {String} connector Database connector to use.  For any supported connector, can be any of:
+ *
+ * - The connector module from `require(connectorName)`.
+ * - The full name of the connector module, such as 'loopback-connector-oracle'.
+ * - The short name of the connector module, such as 'oracle'.
+ * - A local module under `./connectors/` folder.
+ * @property {String} host Database server host name.
+ * @property {String} port Database server port number.
+ * @property {String} username Database user name.
+ * @property {String} password Database password.
+ * @property {String} database Name of the database to use.
+ * @property {Boolean} debug Display debugging information. Default is false.
+ *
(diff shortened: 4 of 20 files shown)


types/scope.d.ts
```diff
@@ -0,0 +1,115 @@
...
+  options?: Options,
+): ScopeDefinition;
+
+/**
+ * Methods injected by scopes
+ */
+export interface ScopedMethods<T extends ModelBase = ModelBase> {
+  build(data: ModelData<T>): T;
+
+  create(
```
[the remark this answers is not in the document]
raymondfeng: @zbarbuto Thank you for chiming in. It's a good idea to use overloaded method signatures. Let's do it incrementally. I would like to use the type information for LoopBack 4 soon.

For `loopback` module, I think there is already a community module - `@types/loopback`. Do you want to check out?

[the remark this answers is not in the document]
zbarbuto: Yes have used this in the past - it does have some issues. Was only mentioning in case it was on the cards to eventually have typedefs out of the box in `loopback` as well.

Understood re: changing incrementally.

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "bracketSpacing": false,
+  "singleQuote": true,
```
bajtos: Let's not mix adding Prettier with typings. Could you please open a new PR adding Prettier to our workflow (including `npm test` executed by CI)? Alternatively, keep these prettier-related files local only to help your IDE, don't commit them to git.

tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "http://json.schemastore.org/tsconfig",
+  "compilerOptions": {
```
bajtos: I am confused why do we need a tsconfig file, when this project does not contain any TypeScript sources to compile, only `.d.ts` files. Could you please clarify?

types/common.d.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Type alias for Node.js callback functions
+ */
+export type Callback<T = any> = (err: any, result?: T) => void;
+
+/**
+ * Return export type for promisified Node.js async methods.
+ *
+ * Note that juggler uses Bluebird, not the native Promise.
+ */
+export type PromiseOrVoid<T = any> = PromiseLike<T> | void;
```
bajtos: I don't see any import or definition of `PromiseLike<T>`, perhaps an oversight on your side?

I think we need to introduce some kind of a smoke test (as part of this pull request!) to at least verify that type definitions can be parsed & loaded by `tsc`.

types/datasource.d.ts
```diff
@@ -0,0 +1,169 @@
...
+ *   - new DataSource('myDataSource', {name: 'anotherDataSource', connector: 'memory'});
+ *
+ * 2. new DataSource(settings). For example:
+ *   - new DataSource({name: 'myDataSource', connector: 'memory'});
+ *   - new DataSource({connector: 'memory'});
+ *
+ * 3. new DataSource(connectorModule, settings). For example:
+ *   - new DataSource(connectorModule, {name: 'myDataSource})
+ *   - new DataSource(connectorModule)
+ */
```
bajtos: Is it a good idea to duplicate the jsdoc comments in typing files? I am concerned that the content in `.d.ts` files will quickly get out of sync with the original comments in `.js` files.

types/model.d.ts
```diff
@@ -0,0 +1,161 @@
...
+// License text available at https://opensource.org/licenses/MIT
+
+/// <reference types="node" />
+
+import {AnyObject, Options, Callback, PromiseOrVoid} from './common';
+import {DataSource} from './datasource';
+import {EventEmitter} from 'events';
+
+// tslint:disable:no-any
+// tslint:disable:max-line-length
```
bajtos: AFAICT, we don't have `tslint` integration in juggler, therefore we should not need any `tslint` overrides.

types/persisted-model.d.ts
```diff
@@ -0,0 +1,511 @@
...
+
+export type PersistedData<
+  T extends PersistedModel = PersistedModel
+> = ModelData<T>;
+
+export interface Count {
+  count: number;
+}
+
+export declare class PersistedModel extends ModelBase {
```
bajtos: Technically speaking, there is no such thing as a `PersistedModel` at juggler level. We have `ModelBase` and `DataAccessObject` mixin that's added to `ModelBase` when a model is attached to a datasource using the default DAO implementation. Similarly, we have `KeyValueAccessObject` mixin that provides API for models connected to key-value stores.

The class `PersistedModel` is implemented in `loopback`, as a thin wrapper providing default DAO methods which throw a helpful message when the model was not attached to a datasource yet.

In the light of that, I feel the currently proposed typedefs are subtly incorrect and I would prefer if we could find a way how to make `.d.ts` files closer to the actual juggler API and behavior.

Thoughts?

types/common.d.ts
raymondfeng: `PromiseLike` is a built-in type from `lib.es5.d.ts`.

tsconfig.json
raymondfeng: The presence of `tsconfig.json` file allows VsCode to validate the syntax of `.d.ts` files.

types/datasource.d.ts
raymondfeng: When the package is used by other modules such as `@loopback/repository`, VSCode uses the tsdocs from `.d.ts` file, IIUC.

types/persisted-model.d.ts
raymondfeng: I'll check

types/scope.d.ts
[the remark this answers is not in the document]
raymondfeng: @zbarbuto I don't think I include `Model` in the declaration files. Please note that `ModelBase` is different from `Model`:

- ModelBase: a base class in `loopback-datasource-juggler` as the common root of all models
- Model: a class in `loopback`

types/persisted-model.d.ts
raymondfeng: LB3 uses mixins at runtime to build model classes. Even though the `PersistedModel` is physically from `loopback` module, we do have the assumption of `PersistedModel` class in many cases in `loopback-datasource-juggler`.

It's hard to describe such contracts in other forms such as interfaces to allow both static and prototype methods/properties.

types/common.d.ts
raymondfeng: I added `npm run tsc` to validate `*.d.ts` files.

tsconfig.json
bajtos: Cool!

I think we should have a build step to run that validations, otherwise our `.d.ts` files can quickly become invalid. I consider this as a must have for this pull request.

types/common.d.ts
bajtos: Great!

types/datasource.d.ts
bajtos: I see, that makes sense.

It would be great to find a way how to automatically sync these jsdoc/tsdoc comments, but that's out of scope of this pull request.

bajtos:
> LB3 uses mixins at runtime to build model classes. Even though the PersistedModel is physically from loopback module, we do have the assumption of PersistedModel class in many cases in loopback-datasource-juggler.
>
> It's hard to describe such contracts in other forms such as interfaces to allow both static and prototype methods/properties.

I'll see. This is an interesting problem, I may give it a try myself. For now, I am ok with a suboptimal solution that works right now.

.gitignore
```diff
@@ -13,5 +13,5 @@ npm-debug.log
 .project
 test/memory.json
 .nyc_output
-
+dist
```
bajtos: Please add `dist` to [.npmignore](https://github.com/strongloop/loopback-datasource-juggler/blob/master/.npmignore) too.

tsconfig.json
bajtos: Ah, I see you have added `npm run tsc`, we are good here 👍

bajtos:
Few more comments (see above), I think this is almost ready to be landed. My expectation is that we will be fixing any bugs as we discover them from real-world usage, thus this initial typings can be landed quickly.

I feel it's really important to clearly set the expectations about the stability of the TypeScript type definitions. Could you please add a bold comment to `index.d.ts` to mark TypeScript support as experimental and explicitly tell the users that backwards-incompatible fixes may come in semver-minor versions?

types/datasource.d.ts
```diff
@@ -0,0 +1,172 @@
...
+export declare class DataSource {
+  name: string;
+  settings: Options;
+
+  connected?: boolean;
+  connecting?: boolean;
+
+  connector?: any;
```
b-admike: Can we bring in https://github.com/strongloop/loopback-next/blob/master/packages/repository/src/connectors/connector.ts for this?

raymondfeng: Added.

raymondfeng:
@bajtos @zbarbuto @b-admike Any more comments before I merge the PR?

zbarbuto:
Agree with @b-admike - LGTM for a first-pass

Issue #1580, opened 2018-05-08
{include: true} or {include: 1} throws TypeError


spurreiter:
# Description/Steps to reproduce

On `{include: true}` or `{include: 1}` the lib crashes with 
"TypeError: includes.forEach is not a function".

# Link to reproduction sandbox

# Expected result

On `{include: true}` or `{include: 1}` the lib shall not throw. In such cases an empty array shall be returned.

# Additional information

bajtos:
Published the fix in `loopback-datasource-juggler@2.59.2` and `loopback-datasource-juggler@3.29.2`

PR #1581, opened 2018-05-08
fix: normalize include with boolean or number

spurreiter:
### Description

- On `{include: true}` or `{include: 1}` the lib crashes with
  "TypeError: includes.forEach is not a function".
- Fix by checking for boolean and number type and return empty array.

#### Related issues

- connect to #1580 {include: true} or {include: 1} throws TypeError

### Checklist

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -16,6 +16,8 @@ const uniq = utils.uniq;
 const idName = utils.idName;
 const debug = require('debug')('loopback:include');
 
+const DISALLOWED_TYPES = ['boolean', 'number', 'symbol', 'function'];
+
 /*!
  * Normalize the include to be an array
  * @param include
@@ -47,6 +49,9 @@ function normalizeInclude(include) {
       newInclude = newInclude.concat(subIncludes);
     }
     return newInclude;
+  } else if (DISALLOWED_TYPES.includes(typeof include)) {
+    debug('Ignoring invalid "include" value of type %s:', typeof include, include);
+    return [];
   } else {
     return include;
   }
--- a/test/include.test.js
+++ b/test/include.test.js
@@ -1254,6 +1254,43 @@ describe('include', function() {
     });
   });
 
+  it('should not throw on fetch User if include is boolean equals true', function(done) {
+    User.find({include: true}, function(err, users) {
+      if (err) return done(err);
+      should.exist(users);
+      users.should.not.be.empty();
+      done();
+    });
+  });
+
+  it('should not throw on fetch User if include is number', function(done) {
+    User.find({include: 1}, function(err, users) {
+      if (err) return done(err);
+      should.exist(users);
+      users.should.not.be.empty();
+      done();
+    });
+  });
+
+  it('should not throw on fetch User if include is symbol', function(done) {
+    User.find({include: Symbol('include')}, function(err, users) {
+      if (err) return done(err);
+      should.exist(users);
+      users.should.not.be.empty();
+      done();
+    });
+  });
+
+  it('should not throw on fetch User if include is function', function(done) {
+    const include = () => {};
+    User.find({include}, function(err, users) {
+      if (err) return done(err);
+      should.exist(users);
+      users.should.not.be.empty();
+      done();
+    });
+  });
+
   // Not implemented correctly, see: loopback-datasource-juggler/issues/166
   // fixed by DB optimization
   it('should support include scope on hasAndBelongsToMany', function(done) {


slnode:

test/include.test.js
```diff
@@ -1224,6 +1224,24 @@ describe('include', function() {
     });
   });
 
+  it('should not throw on fetch User if include is true', function(done) {
```
nitro404: Fairly minor, but for consistency's sake perhaps this test message could indicate that a boolean value should not throw an error, rather than just the value true.

spurreiter:
Changed the test title as recommended.
Moved the typeof statements for performance reasons (string and plain-obj should be the normal case).

spurreiter:
ping... any update on this? What can I do that cls-jenkins reports the status back?

virkt25:
@slnode ok to test

spurreiter:
PR rebased to master. Could you review, please?

test/include.test.js
```diff
@@ -1254,6 +1254,24 @@ describe('include', function() {
     });
   });
 
+  it('should not throw on fetch User if include is boolean equals true', function(done) {
+    User.find({include: true}, function(err, users) {
+      should.not.exist(err);
+      should.exist(users);
+      users.length.should.be.ok;
```
jannyHou: here we should assert that the length is 0, since it returns an empty array.

test/include.test.js
```diff
@@ -1254,6 +1254,24 @@ describe('include', function() {
...
+      done();
+    });
+  });
+
+  it('should not throw on fetch User if include is number', function(done) {
+    User.find({include: 1}, function(err, users) {
+      should.not.exist(err);
+      should.exist(users);
+      users.length.should.be.ok;
```
jannyHou: ditto, `users` should be an empty array.

jannyHou:
@slnode ok to test

test/include.test.js
spurreiter: Hi @jannyHou, the include in `./lib/include.js:51` normalizes to an empty array. `users` are returned as if there would have been no include statement in the find.  
users.length is equal 5 in this testcase.

lib/include.js
```diff
@@ -47,6 +47,8 @@ function normalizeInclude(include) {
       newInclude = newInclude.concat(subIncludes);
     }
     return newInclude;
+  } else if (typeof include === 'boolean' || typeof include === 'number') {
+    return [];
   } else {
     return include;
```
nitro404: It appears that the normalizeInclude function is expected to transform the original value into an array, and this still does not handle cases for "undefined", "symbol", "function", null or other special objects if I understand correctly? Should probably add some additional test cases to account for other data types to ensure that the behaviour is correct. Perhaps someone with some more knowledge on this code can weigh in.

test/include.test.js
bajtos: I believe `ok` must be called to actually execute the assertion.

```suggestion
      users.length.should.be.ok();
```

I find it weird to assert that a length (a number) is ok. Can we find a better assertion, e.g. "greater than zero" or `.not.empty()`?

```suggestion
      users.should.not.be.empty();
```

I don't remember Should.js APIs from the top of my head, I hope `empty()` is a valid assertion 🙄

lib/include.js
bajtos: +1 to make these checks more generic and ignore all invalid `include` values.

To make this situation easier to troubleshoot, we should at least print a `debug` statement to let the callers know about the problem.

It would be even better to reject such queries with an error, but I understand that may be too much work.

spurreiter:
Updated according to review comments. Could you please review again?

lib/include.js
```diff
@@ -47,6 +49,8 @@ function normalizeInclude(include) {
       newInclude = newInclude.concat(subIncludes);
     }
     return newInclude;
+  } else if (DISALLOWED_TYPES.includes(typeof include)) {
+    return [];
```
bajtos: Please add a debug statement to let the user know that we are discarding invalid `include` value.

```ts
debug('Ignoring invalid "include" value of type %s:', typeof include, include);
```

test/include.test.js
```diff
@@ -1254,6 +1254,43 @@ describe('include', function() {
     });
   });
 
+  it('should not throw on fetch User if include is boolean equals true', function(done) {
+    User.find({include: true}, function(err, users) {
+      should.not.exist(err);
```
bajtos:
```suggestion
      if (err) return done(err);
```

This way the original stack trace is preserved to make troubleshooting easier.

test/include.test.js
```diff
@@ -1254,6 +1254,43 @@ describe('include', function() {
...
+      should.exist(users);
+      users.should.not.be.empty();
+      done();
+    });
+  });
+
+  it('should not throw on fetch User if include is number', function(done) {
+    User.find({include: 1}, function(err, users) {
+      should.not.exist(err);
```
bajtos: Ditto.

```suggestion
      if (err) return done(err);
```

test/include.test.js
```diff
@@ -1254,6 +1254,43 @@ describe('include', function() {
...
+      should.exist(users);
+      users.should.not.be.empty();
+      done();
+    });
+  });
+
+  it('should not throw on fetch User if include is symbol', function(done) {
+    User.find({include: Symbol('include')}, function(err, users) {
+      should.not.exist(err);
```
bajtos: Ditto.

```suggestion
      if (err) return done(err);
```

test/include.test.js
```diff
@@ -1254,6 +1254,43 @@ describe('include', function() {
...
+      should.exist(users);
+      users.should.not.be.empty();
+      done();
+    });
+  });
+
+  it('should not throw on fetch User if include is function', function(done) {
+    const include = () => {};
+    User.find({include}, function(err, users) {
+      should.not.exist(err);
```
bajtos: Ditto.

```suggestion
      if (err) return done(err);
```

spurreiter:
latest review changes are added. Thanks for review.

spurreiter:
Jenkins fails on Commit Linter. Unfortunately I can't access the job page (gives me a 404). `npm run lint` passes locally here. I checked the lines. Imho all should be ok. Please provide me with feedback what needs to be solved. Thanks.

spurreiter:
ping... any updates on this?

bajtos:
Landed, thank you @spurreiter for the contribution!

PR #1585, opened 2018-05-22
Fix datasource state management

raymondfeng:
Use case:
1. Configure a datasource with lazyConnect = true
2. Do NOT start the DB
3. Start the app
4. Send first request and it fails to connnect to the DB
5. Start the DB
5. Requests are now served correctly

#### Related issues

- connect to https://github.com/strongloop/loopback-connector-mongodb/issues/268

### Checklist

### DIFF
--- a/test/datasource.test.js
+++ b/test/datasource.test.js
@@ -155,6 +155,172 @@ describe('DataSource', function() {
     dataSource.connector.should.equal(mockConnector);
   });
 
+  it('should set states correctly with eager connect', function(done) {
+    var mockConnector = {
+      name: 'loopback-connector-mock',
+      initialize: function(ds, cb) {
+        ds.connector = mockConnector;
+        this.connect(cb);
+      },
+
+      connect: function(cb) {
+        process.nextTick(function() {
+          cb(null);
+        });
+      },
+    };
+    var dataSource = new DataSource(mockConnector);
+    // DataSource is instantiated
+    // connected: false, connecting: false, initialized: false
+    dataSource.connected.should.be.false();
+    dataSource.connecting.should.be.false();
+    dataSource.initialized.should.be.false();
+
+    dataSource.on('initialized', function() {
+      // DataSource is initialized with lazyConnect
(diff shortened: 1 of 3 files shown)


test/datasource.test.js
```diff
@@ -155,6 +155,171 @@ describe('DataSource', function() {
...
+
+      connect: function(cb) {
+        process.nextTick(function() {
+          cb(null);
+        });
+      },
+    };
+    var dataSource = new DataSource(mockConnector);
+    // DataSource is instantiated
+    // connected: false, connecting: false, initialized: false
```
b-admike: I think the comments showing the status i.e. `// connected: false, connecting: false, initialized: false` can be removed since the assertions below them show what we expect.

test/datasource.test.js
```diff
@@ -155,6 +155,171 @@ describe('DataSource', function() {
...
+      // DataSource is now connected
+      // connected: true, connecting: false
+      dataSource.connected.should.be.true();
+      dataSource.connecting.should.be.false();
+    });
+
+    // Call connect() in next tick so that we'll receive initialized event
+    // first
+    process.nextTick(function() {
+      // At this point, the datasource is already connected by initialize
```
b-admike: IIUC, can we clarify the comment to be `// At this point, the datasource is already connected by the connector's (mockConnector) initialize function`

test/datasource.test.js
```diff
@@ -155,6 +155,171 @@ describe('DataSource', function() {
...
+    process.nextTick(function() {
+      dataSource.connect(function() {
+        // DataSource is now connected
+        // connected: true, connecting: false
+        dataSource.connected.should.be.true();
+        dataSource.connecting.should.be.false();
+        done();
+      });
+      // As the datasource is not connected, connecting will happen
+      // connected: true, connecting: false
```
b-admike: I think `// connected: true, connecting: false` should be the other way around `// connected: false, connecting: true`, but like I said above, the comment can be removed

PR #1587, opened 2018-05-24
Correct syntax error in dao.js script for persist hook call in updateAttributes and replaceById methods.

juliendossantos:
### Description

There is a syntaxe error in updateAttributes and replaceById methods for the Model.notifyObserversOf persist call.

The context past to the Model._forDB() method are not the context that are return by the persist hook.

There is a mixup between the variable 'ctx' returned by the 'persist' hook and the variable 'context' that correspond to the above code block.

This syntax error prevent any modification of the data in the 'persist' hook from being taken into account when updateAttributes and replaceById methods are used. 

### Checklist

- [ x] Code conforms with the [style

### DIFF
--- a/test/persistence-hooks.suite.js
+++ b/test/persistence-hooks.suite.js
@@ -1505,7 +1505,7 @@ module.exports = function(dataSource, should, connectorCapabilities) {
           if (err) return done(err);
 
           ctxRecorder.records.should.eql(aCtxForModel(TestModel, {
-            data: {name: 'changed'},
+            data: {id: existingInstance.id, name: 'changed'},
             isNewInstance: false,
           }));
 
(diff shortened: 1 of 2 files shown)


slnode:

bajtos:
Hello @juliendossantos, thank you the pull request! The fix looks good to me, but I would like to see new tests that will verify the changes and prevent regressions in the future. These test should be added to [test/persistence-hooks.suite.js](https://github.com/strongloop/loopback-datasource-juggler/blob/77f11cda3b6e9e39c1cc039808180129a23d67ff/test/persistence-hooks.suite.js), see other tests in that file for an inspiration. Look for existing tests called "it('applies updates from `persist` hook')", you need provide such a test for each of the two data-access operations you are fixing.

juliendossantos:
Hi @bajtos,

I added tests cases and also correct another context mixup from the 'updateAttributesCallback' method in 'replaceById' and 'updateAttributes'.

In the case that the connector don't provided the 'generateContextData' method, the context data past to the 'loaded' hook are not the data returned by the 'persist' hook but the context of the above block.

I'm also a little perplex about the fact that is this case the test on 'persist' hook represent the real behavior of the code...

For example the 'updateAttributes' method have already a test case for persist hook see: 

https://github.com/strongloop/loopback-datasource-juggler/blob/77f11cda3b6e9e39c1cc039808180129a23d67ff/test/persistence-hooks.suite.js#L1419-L1434

Run this test on the version 3.21.1 don't reflect the syntax error that I try to correct. The modified data in the persist 'hook' are reflected into the saved Instance, but in a real use case the modification is not persisted in datatabase.

Also Running the test that I added in the version 3.21.1 don't reflect the syntax error for the 'replaceById' method, the data modified are also reflected into the saved Instance.

I cannot figure what is the difference between the test and a real use... If you have any idea ?

I have also to add test for the correction of the context mixup from the 'updateAttributesCallback', but before to go further I think we have to find what can explain this difference of behavior between test and real code execution...

jannyHou:
@slnode test please

test/persistence-hooks.suite.js
```diff
@@ -2556,6 +2556,45 @@ module.exports = function(dataSource, should, connectorCapabilities) {
...
+          }));
+
+          // By default, the instance passed to updateAttributes callback is NOT updated
+          // with the changes made through persist/loaded hooks. To preserve
+          // backwards compatibility, we introduced a new setting updateOnLoad,
+          // which if set, will apply these changes to the model instance too.
+          TestModel.settings.forceId = true;
+          TestModel.settings.updateOnLoad = true;
+          TestModel.replaceOrCreate(
+            {name: 'a name'}, function(err, instance) {
```
bajtos: One argument per line please, see http://loopback.io/doc/en/contrib/style-guide-es6.html#one-argument-per-line

test/persistence-hooks.suite.js
```diff
@@ -2556,6 +2556,45 @@ module.exports = function(dataSource, should, connectorCapabilities) {
...
+            if (err) return done(err);
+            instance.should.have.property('extra', 'hook data');
+            TestModel.settings.updateOnLoad = false;
```
bajtos: This code will not be executed when the assertion above fails.

Fortunately, there is no need to reset this flag, because a fresh new `TestModel` class is created for each test (see `beforeEach` at the top of this test file).

Same comment applies to the other new test added below.

bajtos:
@juliendossantos ping, what's the status of this pull request?

juliendossantos:
Hi @bajtos,

* I have corrected the tests cases taking into account your comments in test/persistence-hooks.suite.js. 
* I have also added tests cases for the bug of the context mixup from the 'updateAttributesCallback' method.

Unfortunately I don't have time to implement the helper function for all other test case. I have corrected only the tests that corresponds to the scope of this pull request.

For my use case this PR corrected the problem that encounter. Let me know if there are any other steps to provide in order to merge this PR.

bajtos:
> Unfortunately I don't have time to implement the helper function for all other test case. I have corrected only the tests that corresponds to the scope of this pull request.
>
> For my use case this PR corrected the problem that encounter. Let me know if there are any other steps to provide in order to merge this PR.

Fair enough 👍 

I'll clean up the code a bit myself then, to get this landed sooner.

As I started to play with the changes you have proposed, I discovered that many of your newly added tests are already passing against the current `master` branch. It makes sense for the cases where you are adding missing tests for existing replaceById functionality. However, we also need tests that fail against the current master and pass with your fix in place. I'll try to write them myself to get this pull request merged sooner, but it will take few more days.

bajtos:
As I was investigating your patch, I discovered that the problem applies to other hooks too. I decided to fix as many problems as possible in a single go, see #1598.

PR #1588, opened 2018-05-24
Fix tests to ensure upward compatibility w/ should@10

raymondfeng:
### Description

Fix code and tests to be compatible with `should@10.x` which checks prototypes for comparison.
The extra step exposes some issues in our code:

- Invalid value for `regexp` operator error is not thrown correctly
- List (extends from Array) cannot be compared against plain array
- Validation errors 

#### Related issues

### Checklist


virkt25:
Good stuff! 👍 
Please confirm that the downstream failures aren't related to this PR before merging.

PR #1591, opened 2018-05-29
Fix DateString throwing error

klarkc:
### Description
The issue happens when using PersistedModel.upsert or PersistedModel.upsertWithWhere methods - that for some reason instance values twice - with model that uses DateString type

### Checklist

### DIFF
--- a/lib/date-string.js
+++ b/lib/date-string.js
@@ -53,6 +53,13 @@ function DateString(value) {
     return new DateString(value);
   }
 
+  if (value instanceof DateString) {
+    return Object.assign(
+      new DateString(value.when),
+      value
+    );
+  }
+
   if (typeof(value) !== 'string') {
     throw new Error('Input must be a string');
   }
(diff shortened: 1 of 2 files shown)


slnode:

dhmlau:
@slnode test please

dhmlau:
@strongloop/sq-lb-apex , could you please review? Thanks.

lib/date-string.js
```diff
@@ -53,6 +53,13 @@ function DateString(value) {
     return new DateString(value);
   }
 
+  if (value instanceof DateString) {
+    return Object.assign(
+      new DateString(value.when),
+      value
+    );
```
bajtos: I find this implementation rather complex and difficult to understand, I am worried about introducing subtle bugs. Have you considered a simpler implementation along the following lines?

```ts
function DateString(value) {
  if (!(this instanceof DateString)) {
    return new DateString(value);
  }

  if (value instanceof DateString) {
    value = value.when;
  }

  if (typeof(value) !== 'string') {
    throw new Error('Input must be a string');
  }

  // etc.
}
```

bajtos:
Closing in favor of https://github.com/strongloop/loopback-datasource-juggler/pull/1740

PR #1594, opened 2018-06-12
Upgrade dependencies

bajtos:
- Drop support for Node 4.x,  it is no longer maintained by Node.js project.
- Travis: add Node.js 10.x to the build matrix
- Disable package-lock feature of npm
- Update eslint + config to latest
- Update remaining dev-dependencies to latest (keep should at 8.x because of connectors)

The diff is best reviewed with white-space changes ignored.


test/memory.test.js
```diff
@@ -962,14 +962,6 @@ describe('Memory connector with options', function() {
     });
   });
 
-  it('should receive options from the find method', function(done) {
```
dhmlau: just curious.. we're removing this test because it's no longer applicable?

bajtos: Good question!

This test has the same name and mostly the same body as the previous test above. I don't see what value such duplicate test is adding, therefore I thought it would be better to remove it.

The removed test:

```js
  it('should receive options from the find method', function(done) {
    var opts = {transaction: 'tx2'};
    Post.find({}, opts, function(err, p) {
      savedOptions.find.should.be.eql(opts);
      done(err);
    });
  });
```

The other copy of this test that stays around.

```js
  it('should receive options from the find method', function(done) {
    var opts = {transaction: 'tx1'};
    Post.find({where: {title: 't1'}}, opts, function(err, p) {
      savedOptions.find.should.be.eql(opts);
      done(err);
    });
  });
```

PR #1597, opened 2018-06-21
handle invalid inputs for destroyAll

b-admike:
### Description
Fixes https://github.com/strongloop/loopback/issues/3094. At the moment, there is just a test to reproduce the issue (which I wasn't able to locally, so will gather more feedback).

### Checklist

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -1623,16 +1623,6 @@ describe('manipulation', function() {
 
   describe('destroy', function() {
     beforeEach(function resetFixtures(done) {
-      Person.destroyAll(function() {
-        Person.create([
-          {name: 'John'},
-          {name: 'Jane'},
-          {name: 'Mary'},
-        ], done);
-      });
-    });
-
-    after(function deleteFixtures(done) {
       Person.destroyAll(done);
     });
 
@@ -1699,74 +1689,58 @@ describe('manipulation', function() {
         .catch(done);
     });
 
-    // TODO: implement destroy with filtered set
-    it.skip('should destroy filtered set of records');
-
-    it('should not delete using condition from filter', function(done) {
-      Person.destroyAll({where: {name: 'John'}}, function(err, info) {
-        should.exist(err);
-        err.should.match(/Filter object detected/);
-        Person.find({fields: 'name'}, function(err, users) {
-          should.not.exist(err);
-          users.should.containDeep([
-            {name: 'Jane'},
-            {name: 'Mary'},
-            {name: 'John'},
-          ]);
-          done();
-        });
-      });
+    it('should destroy filtered set of records', function() {
+      return givenNamedPeople(['John', 'Jane', 'Mary'])
+        .then(() =>
+          Person.destroyAll({name: 'Jane'}))
+        .then(() => Person.find({fields: {name: true}})
+          .then(users => {
+            let userData = [];
+            users.forEach(user => userData.push(user.__data));
+            userData.should.deepEqual([
+              {name: 'John'},
+              {name: 'Mary'},
+            ]);
+          }));
     });
 
-    it('should not delete all for bad and condition', function(done) {
-      Person.destroyAll({and: [{name: 'Jane'}, {where: {name: 'John'}}]}, function(err, info) {
-        should.not.exist(err);
-        Person.find({fields: 'name'}, function(err, users) {
-          should.not.exist(err);
-          let userObjects = [];
-          users.forEach(user => userObjects.push(user.toObject()));
-          userObjects.should.containDeep([
-            {name: 'John'},
-            {name: 'Jane'},
-            {name: 'Mary'},
-          ]);
-          done();
-        });
-      });
+    it('should reject unknown properties', function() {
+      return givenNamedPeople(['John', 'Jane', 'Mary'])
+        .then(() =>
+          Person.destroyAll({names: 'John'}).should.be.rejectedWith(/Unknown property/))
+        .then(() => Person.find({fields: {name: true}, order: 'name ASC'})
+          .then(users => {
+            let userData = [];
+            users.forEach(user => userData.push(user.__data));
+            userData.should.deepEqual([
+              {name: 'Jane'},
+              {name: 'John'},
+              {name: 'Mary'},
+            ]);
+          }));
     });
 
-    it('should delete all for undefined object', function(done) {
-      Person.destroyAll(undefined, function(err, info) {
-        should.not.exist(err);
-        Person.find({fields: 'name'}, function(err, users) {
-          should.not.exist(err);
-          users.should.containDeep([]);
-          done();
-        });
-      });
+    it('should reject where object inside a filter object', function() {
+      return givenNamedPeople(['John', 'Jane', 'Mary'])
+        .then(() =>
+          Person.destroyAll({where: {name: 'John'}}).should.be.rejectedWith(/Unknown property/))
+        .then(() => Person.find({fields: {name: true}})
+          .then(users => {
+            let userData = [];
+            users.forEach(user => userData.push(user.__data));
+            userData.should.deepEqual([
+              {name: 'John'},
+              {name: 'Jane'},
+              {name: 'Mary'},
+            ]);
+          }));
     });
 
-    it('should delete all for empty object', function(done) {
-      Person.destroyAll({}, function(err, info) {
-        should.not.exist(err);
-        Person.find({fields: 'name'}, function(err, users) {
-          should.not.exist(err);
-          users.should.containDeep([]);
-          done();
-        });
-      });
-    });
-
-    it('should normalize filter with undefined property', function(done) {
-      Person.destroyAll({name: undefined}, function(err, info) {
-        should.not.exist(err);
-        Person.find({fields: 'name'}, function(err, users) {
-          should.not.exist(err);
-          users.should.containDeep([]);
-          done();
-        });
-      });
-    });
+    function givenNamedPeople(namesList) {
+      return Promise.all(
+        namesList.map(name => Person.create({name}))
+      );
+    }
   });
 
   bdd.describeIf(connectorCapabilities.reportDeletedCount !== false &&
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -1622,6 +1622,20 @@ describe('manipulation', function() {
   });
 
   describe('destroy', function() {
+    beforeEach(function resetFixtures(done) {
+      Person.destroyAll(function() {
+        Person.create([
+          {name: 'John'},
+          {name: 'Jane'},
+          {name: 'Mary'},
+        ], done);
+      });
+    });
+
+    after(function deleteFixtures(done) {
+      Person.destroyAll(done);
+    });
+
     it('should destroy record', function(done) {
       Person.create(function(err, p) {
         if (err) return done(err);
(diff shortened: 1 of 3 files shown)


lib/dao.js
```diff
@@ -2224,11 +2224,24 @@ DataAccessObject.destroyAll = function destroyAll(where, options, cb) {
...
+  if (typeof where === 'object' && Object.keys(where).length === 0) {
+    throw new Error('Cannot pass in an empty where object \'{}\'. Please pass in \'undefined\' instead.');
+  }
+
   cb = cb || utils.createPromiseCallback();
-  where = where || {};
   options = options || {};
 
-  assert(typeof where === 'object', 'The where argument must be an object');
+  // assert(typeof where === 'object', 'The where argument must be an object');
```
nitro404: I believe you will still need checks for the type of the value of where, since you can now pass through null values and such.

lib/dao.js
```diff
@@ -2224,11 +2224,24 @@ DataAccessObject.destroyAll = function destroyAll(where, options, cb) {
     }
   }
 
+  if (typeof where === 'object' && Object.keys(where).includes('where')) {
+  /** Option 1
+   *  throw new Error('This method does not accept filter objects. Please pass
+  in a where object');
```
nitro404: I like this option best, personally.

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+        Person.find({fields: 'name'}, function(err, users) {
+          should.not.exist(err);
+          users.should.containDeep([]);
+          done();
+        });
+      });
+    });
+
+    it('should delete all for empty object', function(done) {
+      Person.destroyAll({}, function(err, info) {
```
virkt25: I think this is the dangerous behaviour we want to eliminate. If a user wants to destroyAll, then it should be called as such (with `undefined` or nothin) but if an object is present then it should be a filter and not an empty object IMO.

bajtos: Wouldn't such change be backwards incompatible and require a semver-major release?

test/manipulation.test.js
```diff
@@ -1622,6 +1622,20 @@ describe('manipulation', function() {
...
+        Person.create([
+          {name: 'John'},
+          {name: 'Jane'},
+          {name: 'Mary'},
+        ], done);
+      });
+    });
+
+    after(function deleteFixtures(done) {
+      Person.destroyAll(done);
```
bajtos: Cleaning the database after the tests is an antipattern, see https://loopback.io/doc/en/lb4/Testing-your-application.html#clean-the-database-before-each-test

> Start with a clean database before each test. This may seem counter-intuitive: why not reset the database after the test has finished? When a test fails and the database is cleaned after the test has finished, then it’s difficult to observe what was stored in the database and why the test failed. When the database is cleaned in the beginning, then any failing test will leave the database in the state that caused the test to fail.

test/manipulation.test.js
```diff
@@ -1622,6 +1622,20 @@ describe('manipulation', function() {
   });
 
   describe('destroy', function() {
+    beforeEach(function resetFixtures(done) {
+      Person.destroyAll(function() {
+        Person.create([
+          {name: 'John'},
+          {name: 'Jane'},
+          {name: 'Mary'},
+        ], done);
```
bajtos: Sharing test data with so many tests is not ideal. When reading the test code and the assertions, it's difficult to understand where the "magic string" names are coming from.

A better solution is to create a helper function for creating an arbitrary list of person instances, and then call that helper from each test.

Also note that new tests can use Promise-based flow control to keep the code simpler.

```js
function givenNamedPeople(namesList) {
  return Promise.all(
    namesList.map(name => Person.create({name})
  );
}

it('should not delete using condition from filter', function() {
  return givenNamedPeople(['John', 'Jane'])
    .then(() => 
      Person.destroyAll({where: {name: 'John'}}).should.be.rejectedWith(/Filter object detected/))
    .then(() => Person.find({fields: 'name'})
    .then(users => { 
      users.should.containDeep([
        // ...
      ]);
    );
});
```

Feel free to keep using callbacks if you prefer, in which case you should use something like [`async.map`](https://github.com/caolan/async/blob/v1.5.2/README.md#map) instead of `Promise.all`.

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
     // TODO: implement destroy with filtered set
-    it('should destroy filtered set of records');
+    it.skip('should destroy filtered set of records');
+
+    it('should not delete using condition from filter', function(done) {
+      Person.destroyAll({where: {name: 'John'}}, function(err, info) {
+        should.exist(err);
+        err.should.match(/Filter object detected/);
+        Person.find({fields: 'name'}, function(err, users) {
+          should.not.exist(err);
```
bajtos: This is an anti-pattern I have been fighting for years.

When `Person.find` fails, this assertions says something along the lines "Expected {Error} to not exist" and shows a stack trace pointing to this very line, throwing the original stack trace away.

A better solution is to pass `err` to test callback. That way the test fails with the original error triggered by `Person.find` and provides all information from the original error (message, stack trace, error code, etc.)

```js
if (err) return done(err);
```

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+            {name: 'John'},
+          ]);
+          done();
+        });
+      });
+    });
+
+    it('should not delete all for bad and condition', function(done) {
+      Person.destroyAll({and: [{name: 'Jane'}, {where: {name: 'John'}}]}, function(err, info) {
+        should.not.exist(err);
```
bajtos:
```js
if (err) return done(err);
```

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+        Person.find({fields: 'name'}, function(err, users) {
+          should.not.exist(err);
```
bajtos:
```js
if (err) return done(err);
```

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+            {name: 'Mary'},
+          ]);
+          done();
+        });
+      });
+    });
+
+    it('should delete all for undefined object', function(done) {
+      Person.destroyAll(undefined, function(err, info) {
+        should.not.exist(err);
```
bajtos:
```js
if (err) return done(err);
```

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+        Person.find({fields: 'name'}, function(err, users) {
+          should.not.exist(err);
```
bajtos:
```js
if (err) return done(err);
```

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+          users.should.containDeep([]);
+          done();
+        });
+      });
+    });
+
+    it('should delete all for empty object', function(done) {
+      Person.destroyAll({}, function(err, info) {
+        should.not.exist(err);
```
bajtos:
```js
if (err) return done(err);
```

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+        Person.find({fields: 'name'}, function(err, users) {
+          should.not.exist(err);
```
bajtos:
```js
if (err) return done(err);
```

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+          users.should.containDeep([]);
+          done();
+        });
+      });
+    });
+
+    it('should normalize filter with undefined property', function(done) {
+      Person.destroyAll({name: undefined}, function(err, info) {
+        should.not.exist(err);
```
bajtos:
```js
if (err) return done(err);
```

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+        Person.find({fields: 'name'}, function(err, users) {
+          should.not.exist(err);
```
bajtos:
```js
if (err) return done(err);
```

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+        });
+      });
+    });
+
+    it('should delete all for empty object', function(done) {
+      Person.destroyAll({}, function(err, info) {
+        should.not.exist(err);
+        Person.find({fields: 'name'}, function(err, users) {
+          should.not.exist(err);
+          users.should.containDeep([]);
```
bajtos: Isn't this assertion always true as long as `users` is any array? I think you need to call `users.should.deepEqual` instead.

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+        });
+      });
+    });
+
+    it('should normalize filter with undefined property', function(done) {
+      Person.destroyAll({name: undefined}, function(err, info) {
+        should.not.exist(err);
+        Person.find({fields: 'name'}, function(err, users) {
+          should.not.exist(err);
+          users.should.containDeep([]);
```
bajtos: Isn't this assertion always true as long as `users` is any array? I think you need to call `users.should.deepEqual` instead.

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
...
+        });
+      });
+    });
+
+    it('should delete all for undefined object', function(done) {
+      Person.destroyAll(undefined, function(err, info) {
+        should.not.exist(err);
+        Person.find({fields: 'name'}, function(err, users) {
+          should.not.exist(err);
+          users.should.containDeep([]);
```
bajtos: Isn't this assertion always true as long as `users` is any array? I think you need to call `users.should.deepEqual` instead.

test/manipulation.test.js
b-admike: Yes! I found that out after debugging and was looking for an alternative. Thank you :-)

test/manipulation.test.js
b-admike: Yes. This is a good discussion point. I agree with changing the usage to use `undefined` value for the `where` object (which is also the case when it's omitted in API Explorer since it's optional) to delete all records. That means changing our default assignment to `undefined` since in `dao` we default to an empty object. Then, we need to make sure that the connector specific methods also adhere to this behaviour. I also want to implement rejection of empty objects and to advise users to use `undefined` instead. Yes that would mean a semver-major release, but perhaps we can mitigate it by introducing a flag for this new behaviour as well? Please keep in mind that this is also what the users complained about in the [original issue](https://github.com/strongloop/loopback/issues/3094).

bajtos: Please note that detecting `{}` can be tricky.

Should we treat `{userId: undefined}` the same way as `{}`? Or does the first example mean "match all records that have no `userId` set"? With SQL connectors, this can be probably implemented as a NULL check, but I don't know what's the level of support in noSQL connectors for such query.

> perhaps we can mitigate it by introducing a flag for this new behaviour as well?

Yeah, we will have to.

> Please keep in mind that this is also what the users complained about in the original issue.

As I understand the discussion in the original issue, the biggest pain point is the situation when people accidentally wrap the condition inside `where`, which is an extension of the situation when they make a typo in a property name.

The case of `where={}` is an edge case that was brought up later, as we started to discuss possible fixes.

I would personally focus on the first part (i.e. fix connectors to stop ignoring unknown properties) and leave changes in handling of the empty `where` object for later.

Baby steps FTW!

test/manipulation.test.js
```diff
@@ -1686,7 +1700,73 @@ describe('manipulation', function() {
     });
 
     // TODO: implement destroy with filtered set
-    it('should destroy filtered set of records');
+    it.skip('should destroy filtered set of records');
```
virkt25: Reminder: This should be implemented / removed to cleanup the code.

test/manipulation.test.js
virkt25: > Should we treat {userId: undefined} the same way as {}?

I don't think that `userId: undefined` is the same as an empty object. I think it makes more sense for that to be treated as `match all records that have no userId set` while an empty object should not be allowed. If the intent is to delete all object then the user should call `destroyAll()` with no arguments. 

Any reason we can't have a breaking change with this module? I wouldn't consider this a breaking change considering I think this was a design oversight that we are fixing but to adhere to semver we can do a major release (at least that's my opinion).

b-admike: I agree with making fixes in small increments. I'll remove this test case for now (thoughts?). We could definitely follow up with issues regarding how we would like to handle empty objects. I agree that `{userId: undefined}` shouldn't be the same as `{}` but I think we should either reject it or yes treat it as `match all records that have no userId set`.

bajtos:
> Any reason we can't have a breaking change with this module? I wouldn't consider this a breaking change considering I think this was a design oversight that we are fixing but to adhere to semver we can do a major release (at least that's my opinion).

LoopBack users cannot choose the juggler version they are using, juggler version is controlled by `loopback`'s dependencies specified in its `package.json`.
 
When we release a new semver-major version of juggler (4.0), we need to decide what to do with `loopback@3.x`.

Is it going to keep using juggler 3.x and not receiving this fix? To me, that sounds as not very useful to LB 3.x users.

If we upgrade `loopback` dependencies to use `loopback-datasource-juggler@4.x`, then we need to publish such change as `loopback@4.0`. Can we release `loopback@4.0` when our intention is to release loopback-next as _the 4.0 version of LoopBack_?

virkt25:
> LoopBack users cannot choose the juggler version they are using, juggler version is controlled by loopback's dependencies specified in its package.json.

That makes sense. I was thinking of `juggler` as an individual module. I think @b-admike has a fix which won't be breaking.

dhmlau:
@b-admike, are you still interested in completing this PR? Thanks.

PR #1598, opened 2018-06-26
Fix Operation Hooks to propagate data changes

bajtos:
Before this change, many Operation Hooks were not correctly propagating changes made to `ctx.data` via reassigning ctx.data to a new object.

This change modifies existing tests to account for this different scenario, adds few more tests for scenarios that were not covered by tests before and finally fixes the problem discovered.

This pull request supersedes #1587.

/cc @juliendossantos 

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -467,9 +467,10 @@ DataAccessObject.create = function(data, options, cb) {
           hookState: hookState,
           options: options,
         };
-        Model.notifyObserversOf('persist', context, function(err) {
+        Model.notifyObserversOf('persist', context, function(err, ctx) {
           if (err) return cb(err);
-          invokeConnectorMethod(connector, 'create', Model, [obj.constructor._forDB(context.data)],
+          val = ctx.data;
+          invokeConnectorMethod(connector, 'create', Model, [obj.constructor._forDB(ctx.data)],
             options, createCallback);
         });
       }, obj, cb);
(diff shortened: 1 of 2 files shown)


lib/dao.js
```diff
@@ -2581,6 +2584,7 @@ DataAccessObject.prototype.save = function(options, cb) {
 
           Model.notifyObserversOf('persist', context, function(err) {
```
virkt25: Shouldn't `ctx` be passed into the callback on this line ... and 2 lines below, `ctx.data` be used instead of `context.data`? ... If not, can you explain the difference between the two variables and why this scenario is different from the rest?

jannyHou: @virkt25 I guess `ctx` is only added to `loaded` hook but not `persisted` hook.

bajtos: Good question!

Let's take a look at the code:

```js
const context = {
  // ...
};

Model.notifyObserversOf('persist', context, function(err, ctx) {
  if (err) return cb(err);
  data = context.data;
  // etc.
});
```

`notifyObserversOf` accepts a context argument which is eventually forwarded as the second argument of the callback.

In this particular case, `ctx === context`.

The reason why I cannot use `context` in some places below is that those places are defining another `context` variable inside the callback. Conceptually:

```ts
const context = {
  // ...
};

Model.notifyObserversOf(..., function(err, ctx) {
  // ...
  const context = {
    // ...
  };
  Model.notifyObserversOf(..., context, ...);
});
```

What a mess :(

What do you think - should I change all places modified by this pull request to always use `ctx` from the second callback argument of `notifyObserversOf` for consistency?

bajtos: I modified `notifyObserversOf` to accept `ctx` as the second argument in most places.

PR #1599, opened 2018-06-28
Fix type definitions for PersistedModel API

bajtos:
- Callback's first argument is an optional Error now. Was: required `any`.

- PersistedModel methods return `PersistedModel` now. Before this change, methods were returning `PersistedData` (`PersistedModel | AnyObject`). The problem with `AnyObject` is that it does not contain any `PersistedModel` instance data and cannot be assigned to functions expecting `Partial<PersistedModel>`. As a result, consumers of this API were forced to either cast the result to `PersistedModel` (which feels wrong) or deal with the `AnyObject` case (which never happen at runtime).

- Fix definition of `ModelData<T>` to `T | Partial<T>`. Before this change, `ModelData` allowed any values not related to the actual model at all, for example arrays.

These problems were discovered while working on https://github.com/strongloop/loopback-next/pull/1475. The fixes presented in this pull request seem to be backwards compatible, at least as far as loopback-next tests are concerned.

### DIFF
--- a/types/common.d.ts
+++ b/types/common.d.ts
@@ -18,7 +18,7 @@ export type Options = AnyObject<any>;
 /**
  * Type alias for Node.js callback functions
  */
-export type Callback<T = any> = (err: any, result?: T) => void;
+export type Callback<T = any> = (err?: Error | null, result?: T) => void;
 
 /**
  * Return export type for promisified Node.js async methods.
--- a/types/common.d.ts
+++ b/types/common.d.ts
@@ -18,7 +18,7 @@ export type Options = AnyObject<any>;
 /**
  * Type alias for Node.js callback functions
  */
-export type Callback<T = any> = (err?: Error | null, result?: T) => void;
+export type Callback<T = any> = (err?: any | null, result?: T) => void;
 
 /**
  * Return export type for promisified Node.js async methods.
--- a/types/model.d.ts
+++ b/types/model.d.ts
@@ -246,4 +246,4 @@ export declare class ModelBuilder extends EventEmitter {
  * Union export type for model instance or plain object representing the model
  * instance
  */
-export type ModelData<T extends ModelBase = ModelBase> = T | AnyObject;
+export type ModelData<T extends ModelBase = ModelBase> = T | Partial<T>;
--- a/types/model.d.ts
+++ b/types/model.d.ts
@@ -243,7 +243,13 @@ export declare class ModelBuilder extends EventEmitter {
 }
 
 /**
+ * An extension of the built-in Partial<T> type which allows partial values
+ * in deeply nested properties too.
+ */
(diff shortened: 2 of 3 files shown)


types/common.d.ts
```diff
@@ -18,7 +18,7 @@ export type Options = AnyObject<any>;
 /**
  * Type alias for Node.js callback functions
  */
-export type Callback<T = any> = (err: any, result?: T) => void;
+export type Callback<T = any> = (err?: Error | null, result?: T) => void;
```
bajtos: Enable the following usage patterns:

- `callback()`
- `callback(null)`

types/model.d.ts
```diff
@@ -246,4 +246,4 @@ export declare class ModelBuilder extends EventEmitter {
  * Union export type for model instance or plain object representing the model
  * instance
  */
-export type ModelData<T extends ModelBase = ModelBase> = T | AnyObject;
+export type ModelData<T extends ModelBase = ModelBase> = T | Partial<T>;
```
bajtos: Strangely enough, `AnyObject` allows arbitrary values not related to ModelBase at all, for example an array. This is a problem in PersistedModel API, where the `data` argument of methods like `create` or `upsert` suddenly allows callers to provide an array, despite the fact that these methods don't support batch mode.

types/persisted-model.d.ts
```diff
@@ -38,8 +38,14 @@ export declare class PersistedModel extends ModelBase {
   static create(
     data: PersistedData,
     options?: Options,
-    callback?: Callback<PersistedData>,
-  ): PromiseOrVoid<PersistedData>;
+    callback?: Callback<PersistedModel>,
+  ): PromiseOrVoid<PersistedModel>;
+
+  static create(
+    data: PersistedData[],
```
bajtos: Add an overload for batch create mode, where an array of data objects is accepted and the method returns an array of model instances created.

types/common.d.ts
virkt25: Shouldn't it be `err?: any` - AFAIK normally any value passed to first parameter that's not null / undefined means an error. So ... for example, I can pass a string into the Callback?

types/model.d.ts
raymondfeng: We probably should introduce a `DeepPartial` as follows. It's used in TypeORM too.
```ts
export type DeepPartial<T> = { [P in keyof T]?: DeepPartial<T[P]>; };
```

types/common.d.ts
raymondfeng: Node.js does allow the error parameter to be `any`.

types/model.d.ts
jannyHou: wouldn't `T | Partial<T>` be the same as `Partial<T>`?

bajtos:
@raymondfeng @virkt25 @jannyHou Thank you for the feedback, I have addressed your comments in 6f3675b. LGTY now?

PR #1602, opened 2018-07-06
Fix regression in Operation Hooks

virkt25:
add check before overriding context.data - fix the regression introduced by #1598

Co-Authored-By: Kyusung Shim <kyu.shim@ibm.com>

#### Related issues

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -3087,11 +3087,15 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
         var typedData = convertSubsetOfPropertiesByType(inst, data);
         context.data = typedData;
 
-        function replaceCallback(err, data) {
+        // Depending on the connector, the database response can
+        // contain information about the updated record(s). This object
+        // has usually database-specific structure and does not match
+        // model properties. For example, MySQL returns OkPacket:
+        //   {fieldCount, affectedRows, insertId, /*...*/, changedRows}
+        function replaceCallback(err, dbResponse) {
           if (err) return cb(err);
-          context.data = data;
           if (typeof connector.generateContextData === 'function') {
-            context = connector.generateContextData(context, data);
+            context = connector.generateContextData(context, dbResponse);
           }
           var ctx = {
             Model: Model,
@@ -3138,6 +3142,8 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
         };
         Model.notifyObserversOf('persist', ctx, function(err) {
           if (err) return cb(err);
+          // apply updates made by "persist" observers
+          context.data = ctx.data;
           invokeConnectorMethod(connector, 'replaceById', Model, [id, Model._forDB(ctx.data)],
             options, replaceCallback);
         });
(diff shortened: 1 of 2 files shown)


lib/dao.js
```diff
@@ -3089,7 +3089,7 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
 
         function replaceCallback(err, data) {
           if (err) return cb(err);
-          context.data = data;
+          if (data) context.data = data;
```
shenghu: Is it better to move this line after calling `generateContextData`, e.g. 

```
if (typeof connector.generateContextData === 'function') {
   context = connector.generateContextData(context, data);	                       
} else {
  if (data) context.data = data;
}
```

In this way, `generateContextData(context, data)` will continue to get old data in `context`. Otherwise, I think implementation of `generateContextData` is likely to be broken.

lib/dao.js
```diff
@@ -3089,7 +3089,7 @@ DataAccessObject.replaceById = function(id, data, options, cb) {
...
           if (typeof connector.generateContextData === 'function') {
             context = connector.generateContextData(context, data);
```
shenghu: Is it possible to change this to be callback? Because `generateContextData ` might need query db to get result?

shenghu:
This can fix https://github.com/strongloop/loopback-connector-cassandra/issues/65. But I think https://github.com/strongloop/loopback-connector-cassandra/issues/65 is still necessary.

lib/dao.js
bajtos: I think you are right, @shenghu.

In the original pull request #1587 contributed by @juliendossantos, the patch looked like this:

```diff
           if (typeof connector.generateContextData === 'function') {
             context = connector.generateContextData(context, data);
+          } else {
+            context.data = data;
           }
```

I changed it to what we have now, because it seemed to be a simpler solution and frankly I did not really understand the difference and possible ramifications.

lib/dao.js
bajtos: Let's leave such change out of scope of this pull request please. We need to fix the regression introduced by #1598 as soon as possible.

Changing `generateContextData` into an async function would be more involved, because we need to preserve backwards compatibility with existing connectors that provide a sync version only.

test/persistence-hooks.suite.js
```diff
@@ -1529,7 +1529,7 @@ module.exports = function(dataSource, should, connectorCapabilities) {
           if (err) return done(err);
 
           ctxRecorder.records.should.eql(aCtxForModel(TestModel, {
-            data: {id: existingInstance.id, name: 'changed'},
+            data: {name: 'changed'},
```
bajtos: This test was incorrectly changed by #1598, I am reverting that change now that the implementation has been fixed.

See https://github.com/strongloop/loopback-datasource-juggler/pull/1598/files#diff-145ab49855b38984ba190919e6e7397dL1508

shimks:
@slnode test please

virkt25:
The current set of failing tests is what I would expect to fail because of known issues with the CI for those connectors (I haven't seen those green in a while) and don't think that should block this PR.

PR #1605, opened 2018-07-09
Make test output concise

bajtos:
### Description

Configure Mocha to use dot reporter. With 1965 tests, printing names of all tests is no longer useful. The first commit switches to "dot" reporter which prints a single dot per test.

Hide nyc coverage from test output. Configure nyc/istanbul to generate an HTML report that can be inspected in a browser instead of printing a long table after each test run.


b-admike:
Let's see if restart of the dashdb/db2 containers makes a difference in our builds 🤞 . @slnode test please

PR #1608, opened 2018-07-13
fix: update TypeScript declaration for KV

raymondfeng:
### Description

update TypeScript declaration for KV

#### Related issues

### Checklist

### DIFF
--- a/types/kv-model.d.ts
+++ b/types/kv-model.d.ts
@@ -5,30 +5,30 @@
 
 import {Callback, Options, PromiseOrVoid} from './common';
 import {ModelBase, ModelData} from './model';
-import {Filter} from './query';
 
 /**
  * Data object for KV models
  */
-export type KVData = ModelData<KeyValueModel>;
+export type KVData<T extends ModelBase = ModelBase> = ModelData<T>;
 
 /**
- * Key/Value model
+ * Key/Value model. Strictly speaking, KeyValueModel is not a class
+ * but a mixin into existing model classes
  */
 export declare class KeyValueModel extends ModelBase {
   /**
    * Return the value associated with a given key.
    *
-   * @param {String} key Key to use when searching the database.
+   * @param {string} key Key to use when searching the database.
    * @options {Object} options
    * @callback {Function} callback
    * @param {Error} err Error object.
-   * @param {Any} result Value associated with the given key.
+   * @param {any} result Value associated with the given key.
    * @promise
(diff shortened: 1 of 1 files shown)


types/kv-model.d.ts
```diff
@@ -164,5 +194,8 @@ export declare class KeyValueModel extends ModelBase {
    *
    * @header KeyValueModel.iterateKeys(filter)
    */
-  iterateKeys(filter?: Filter, options?: Options): Iterator<Promise<string>>;
+  static iterateKeys(
+    filter?: Filter,
+    options?: Options,
+  ): Iterator<Promise<string>>;
```
raymondfeng: This is not accurate. It should be `AsyncIterator<string>` which is only available with `esnext`. See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html.

```
Caveats
Keep in mind that our support for async iterators relies on support for Symbol.asyncIterator to exist at runtime. You may need to polyfill Symbol.asyncIterator, which for simple purposes can be as simple as: (Symbol as any).asyncIterator = Symbol.asyncIterator || Symbol.for("Symbol.asyncIterator");
You also need to include esnext in your --lib option, to get the AsyncIterator declaration if you do not already have it.
Finally, if your target is ES5 or ES3, you’ll also need to set the --downlevelIterators flag.
```

types/kv-model.d.ts
```diff
@@ -49,13 +49,43 @@ export declare class KeyValueModel extends ModelBase {
...
+  /**
+   * Delete all keys (and values) associated to the current model.
+   *
+   * @options {Object} options Unused ATM, placeholder for future options.
+   * @callback {Function} callback
+   * @param {Error} err Error object.
+   * @promise
+   */
+  static deleteAll(
+    key: string,
```
virkt25: Don't think it's supposed to take a `key` argument for `deleteAll`.

raymondfeng: Good catch. Will fix.

types/kv-model.d.ts
```diff
@@ -114,7 +144,7 @@ export declare class KeyValueModel extends ModelBase {
    *
    * @header KeyValueModel.keys(filter, cb)
    */
-  keys(
+  static keys(
     filter?: Filter,
```
bajtos: `KVFilter`, not `Filter`.

PR #1615, opened 2018-07-30
feat: omit default fn for embeds many (2.x)

bajtos:
**Backport of #1532**

_In general, we don't backport new features to LTS versions because of stability reasons. Having said that, this particular change looks more like a bug fix to me, thus I think it's ok to add it to 2.x version too._

Related to https://github.com/strongloop/loopback/issues/2882

## Use case

- `Person` EmbeddsMany `Address`
- create a new person `let p = new Person({name: 'Jack', id: 1})`

## Current behaviour

The generated instance is `{name: 'Jack', id: 1, addresses: []}`

## Feature

In this PR I add a flag `omitDefaultEmbeddedItem`

### Flag off (default behaviour)

Same as current behaviour, the new person instance have an empty array property `addresses` 

### Flag on(user need to config it in model.json relations)

The new person instance is `{name: 'Jack', id: 1}`
Please note it doesn't have the embedded item.

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -2391,9 +2391,16 @@ RelationDefinition.embedsMany = function embedsMany(modelFrom, modelTo, params) 
     embed: true,
   });
 
-  modelFrom.dataSource.defineProperty(modelFrom.modelName, propertyName, {
-    type: [modelTo], default: function() { return []; },
-  });
+  var opts = Object.assign(
+    {},
+    params.options && params.options.omitDefaultEmbeddedItem ? {type: [modelTo]} :
+    {
+      type: [modelTo],
+      default: function() { return []; },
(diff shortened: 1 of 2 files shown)


lib/relation-definition.js
```diff
@@ -2391,9 +2391,16 @@ RelationDefinition.embedsMany = function embedsMany(modelFrom, modelTo, params)
     embed: true,
   });
 
-  modelFrom.dataSource.defineProperty(modelFrom.modelName, propertyName, {
-    type: [modelTo], default: function() { return []; },
-  });
+  var opts = Object.assign(
+    {},
```
jannyHou: Since the old code doesn't get `opts` from `params.options.property`, it's good to just use `{}` here 👌

Issue #1619, opened 2018-08-23
Throw or reject with a 404 error for findById() if resource is not found.


hacksparrow:
`findById()` returns `null` or resolves with `null`, if the resource if not found. It should throw or reject with a 404 error.

PR #1620, opened 2018-08-23
[WIP] feature: 404 error for findById() if resource is not found

hacksparrow:
404 error for findById() if resource is not found

#### Related issues

- connect to https://github.com/strongloop/loopback-datasource-juggler/issues/1619

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1359,7 +1359,8 @@ DataAccessObject.findById = function findById(id, filter, options, cb) {
     }
   }
 
-  cb = cb || utils.createPromiseCallback();
+  // set strictFind to `true` for rejecting with a 404 error, else it will be resolved with a `null`
+  cb = cb || utils.createPromiseCallback(options && options.strictFind);
   options = options || {};
   filter = filter || {};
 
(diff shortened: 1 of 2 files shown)


bajtos:
I am against this change for two reasons:

- Juggler as the data-access layer should not be coupled with the specific requirements of transport layers like REST.
- In LoopBack 1.x-3.x, we are able to convert `null` result to `404` error. I am sure we can do that in LB4 too.

virkt25:
> Juggler as the data-access layer should not be coupled with the specific requirements of transport layers like REST.

+1 **BUT** I don't think it should be returning null either. Should be a `count` of the deleted object or something other than null to indicate it didn't find anything to delete. Thoughts?

bajtos:
> +1 BUT I don't think it should be returning null either. Should be a count of the deleted object or something other than null to indicate it didn't find anything to delete. Thoughts?

This is `findById` method. It returns either the model instance with the given `id`, or `null` when no such model instance was found.

I agree returning a `null` is not a great user experience, I think that's one of the reasons why `findById` in LB4 throws an error instead.

bajtos:
We decided to pursue a different approach in LB4, this pull request is no longer needed.

bajtos:
We decided to pursue a different approach in LB4, this pull request is no longer needed.

PR #1623, opened 2018-09-06
Fix patchAttributes for embedded document properties in MongoDB

### LINKED ISSUE #1622: patchAttributes stores null for undefined embedded document properties in MongoDB

If we have a model with the following properties:
```json
{
	"name": "string",
	"address": {
		"city": "string",
		"areaCode": "string"
	}
}
```
When calling `prototype.patchAttributes` on a model instance and send an update that also includes a partial doc from the embedded document property, such as:
```js
{
	name: 'Mitsos',
	address: {
		city: 'Boston'
	}
}
```
when the code execution reaches [convertSubsetOfPropertiesByType](https://github.com/strongloop/loopback-datasource-juggler/blob/251798c7115f037fbeff2d873089bdf6fdf7cd5f/lib/dao.js#L3292) it converts the embedded document propery using the function [toObject](https://github.com/strongloop/loopback-datasource-juggler/blob/251798c7115f037fbeff2d873089bdf6fdf7cd5f/lib/dao.js#L78) so the embedded model is initialized with its missing values set as `undefined`

This results an update document being sent  in mongo with undefined values, which in turn get converted to `null` from the MongoDriver. 
**update doc:**
```js
{
	name: 'Mitsos',
	address: {
		city: 'Boston',
		areaCode: undefined
	}
}
```
I find this inconsistent with the expected behavior.

# Expected result

It should not add non included properties in the model, not even `undefined` to avoid conflicts with Mongo driver behavior.

# Additional information

**node:** linux x64 10.8.0
**loopback:** loopback-datasource-juggler@3.23.0

mitsos1os:
### Description
Moved `removeUndefined` call to the proper location. This way when sending partial document updates in embedded documents, rest properties do not get converted to null in MongoDB

#### Related issues

- connect to #1622 

### Checklist

### DIFF
--- a/test/crud-with-options.test.js
+++ b/test/crud-with-options.test.js
@@ -23,6 +23,7 @@ describe('crud-with-options', function() {
       role: {type: String, index: true},
       order: {type: Number, index: true, sort: true},
       vip: {type: Boolean},
+      address: {type: {city: String, area: String}},
     });
     options = {};
     filter = {fields: ['name', 'id']};
@@ -145,7 +146,7 @@ describe('crud-with-options', function() {
 
     it('should allow promise-style findById',
       function(done) {
-        User.create({name: 'w', email: 'w@y.com'}).then(function(u) {
+        User.create({id: 15, name: 'w', email: 'w@y.com'}).then(function(u) {
           should.exist(u.id);
           return User.findById(u.id).then(function(u) {
             should.exist(u);
@@ -520,6 +521,24 @@ describe('crud-with-options', function() {
       });
     });
   });
+
+  describe('updateAttributes', function() {
+    beforeEach(seed);
+    it('preserves document properties not modified by the patch', function() {
+      return User.findOne({where: {name: 'John Lennon'}})
+        .then(function(user) {
+          return user.updateAttributes({address: {city: 'Volos'}});
+        })
+        .then(function() {
+          return User.findOne({where: {name: 'John Lennon'}}); // retrieve the user again from the db
+        })
+        .then(function(updatedUser) {
+          updatedUser.address.city.should.equal('Volos');
+          should(updatedUser.address.area).not.be.exactly(null);
+          should(updatedUser.address.area).be.undefined();
+        });
+    });
+  });
 });
 
 describe('upsertWithWhere', function() {
@@ -588,6 +607,7 @@ describe('upsertWithWhere', function() {
 function seed(done) {
   var beatles = [
     {
+      id: 0,
       seq: 0,
       name: 'John Lennon',
       email: 'john@b3atl3s.co.uk',
@@ -597,6 +617,7 @@ function seed(done) {
       vip: true,
     },
     {
+      id: 1,
       seq: 1,
       name: 'Paul McCartney',
       email: 'paul@b3atl3s.co.uk',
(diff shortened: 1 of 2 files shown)


mitsos1os:
Code coverage  decreased while also added extra tests ???

zbarbuto:
It just means that a line was added that wasn't covered since code coverage is a percentage not a count of tests.

mitsos1os:
Ok. Could we move on with the request or is this blocking it?

dhmlau:
@slnode test please
The change LGTM, but would like to get others to review. 
Kicking off CI.

bajtos:
@mitsos1os could you please use a more descriptive pull request title in the future? Seeing "fixes #1622 #1623" in the list of pull requests and email notifications is rather unhelpful.

For example: _Fix patchAttributes for embedded document properties in MongoDB_

test/crud-with-options.test.js
```diff
@@ -520,6 +521,27 @@ describe('crud-with-options', function() {
       });
     });
   });
+
+  describe('updateAttributes', function() {
+    beforeEach(seed);
+    it('shouldn\'t null not given embedded document properties', function() {
```
bajtos:
```js
it('preserves document properties not modified by the patch')
```

test/crud-with-options.test.js
```diff
@@ -520,6 +521,27 @@ describe('crud-with-options', function() {
...
+      if (db.name !== 'mongodb') {
+        console.log('No need to check except for Mongo');
```
bajtos: Please remove this console log. There may be other NoSQL databases besides MongoDB that require this fix.

test/crud-with-options.test.js
```diff
@@ -520,6 +521,27 @@ describe('crud-with-options', function() {
...
+        .then(function(user) {
+          return user.updateAttributes({address: {city: 'Volos'}});
+        })
+        .then(function() {
+          return User.findOne({where: {name: 'John Lennon'}}); // retrieve the user again from the db
+        })
+        .then(function(updatedUser) {
+          updatedUser.address.city.should.equal('Volos');
+          should(updatedUser.address.area).not.be.exactly(null);
+          should.not.exist(updatedUser.address.area);
```
bajtos: I am rather confused about this check. What is the expected `area` value? Can you assert on what the value should be?

mitsos1os:
> @mitsos1os could you please use a more descriptive pull request title in the future? Seeing "fixes #1622 #1623" in the list of pull requests and email notifications is rather unhelpful.
> 
> For example: _Fix patchAttributes for embedded document properties in MongoDB_

Yes sorry for that, I thought it would be worse referencing a pretty much similar title to the actual issue being resolved.

test/crud-with-options.test.js
mitsos1os: done

test/crud-with-options.test.js
mitsos1os: Since the address.area was not initialized it should not have the value `null` which it does without the fix. So I am checking that it strictly is not null, and below that it does not exist *(it is undefined)*. I will explicitly change the check for `undefined`

mitsos1os:
@bajtos Do the latest changes cover your requests?

mitsos1os:
Hi any news with the pull request progress?

mitsos1os:
@bajtos no worries! I understand the extra workload overhead...
Thank you for your assistance.
I would also appreciate if you could speed up the review for: https://github.com/strongloop/loopback-connector-mongodb/pull/471 :D
We are waiting for this to land in order to use evertything all together in our project.
Thanks!

PR #1624, opened 2018-09-17
non-native itemType constructor must be called with new

virkt25:
### Description

- Identified while working on `loopback4-example-shopping`. A list of models should be instantiated using the model constructor. 

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

- [ ] New tests added or existing tests modified to cover all changes
- [ ] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #1625, opened 2018-09-17
Allow List to take items as instances of a class

raymondfeng:
### Description

Supersedes #1624

#### Related issues

### Checklist


test/list.test.js
```diff
@@ -0,0 +1,119 @@
+'use strict';
+
```
Amir-61: Shouldn't we add MIT license header on new files?

PR #1629, opened 2018-10-03
fix: accelerate unique id checking

mgabeler-lee-6rs:
### Description

Saving (very) large objects with loopback can result in node spinning for a long time (>50s in my case) on a single statement checking the uniquness of a list of IDs of child objects.  This causes lots of issues, including connection timeouts, and service heartbeat / health-check failures.

The implementation of this uniquenss check is currently a naieve and extremely inefficient O(N^2) one, which can be trivially accelerated for common ID datatypes.

#### Related issues

N/A

### Checklist

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -17,6 +17,7 @@ var defineScope = require('./scope.js').defineScope;
 var g = require('strong-globalize')();
 var mergeQuery = utils.mergeQuery;
 var idEquals = utils.idEquals;
+var idsHaveDuplicates = utils.idsHaveDuplicates;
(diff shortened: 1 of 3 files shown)


slnode:

mgabeler-lee-6rs:
Also maybe worthy of note: While going from an ~N^2 to ~N dupe check, I suspect the big savings in wall clock here comes from not calling `JSON.stringify` 2*N^2 times (200e6 calls for my 10k strings case).  The pre-ES2015 code path could probably be improved further by directly comparing strings instead of JSON-ifying them and comparing those strings?

PR #1638, opened 2018-10-12
Report errors if where clause contains hidden properties

raymondfeng:
### Description

If we allow hidden properties to be part of the query condition, there is security risk for brute-force guessing.

It also checks circular and deep objects to avoid DoS attack.

#### Related issues

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -804,9 +804,9 @@ DataAccessObject.upsertWithWhere = function(where, data, options, cb) {
         function callConnector() {
           try {
             // Support an optional where object
-            var handleUndefined = Model._getSetting('normalizeUndefinedInQuery');
+            var normalizeUndefinedInQuery = Model._getSetting('normalizeUndefinedInQuery');
             // alter configuration of how sanitizeQuery handles undefined values
-            ctx.where = Model._sanitize(ctx.where, {handleUndefined: handleUndefined});
+            ctx.where = Model._sanitize(ctx.where, {normalizeUndefinedInQuery: normalizeUndefinedInQuery});
             ctx.where = Model._coerce(ctx.where, options);
             update = Model._sanitize(update);
             update = Model._coerce(update, options);
@@ -1579,9 +1579,9 @@ DataAccessObject._normalize = function(filter, options) {
       Object.keys(this.definition.properties), this.settings.strict);
   }
 
-  var handleUndefined = this._getSetting('normalizeUndefinedInQuery');
+  var normalizeUndefinedInQuery = this._getSetting('normalizeUndefinedInQuery');
   // alter configuration of how sanitizeQuery handles undefined values
-  filter = this._sanitize(filter, {handleUndefined: handleUndefined});
+  filter = this._sanitize(filter, {normalizeUndefinedInQuery: normalizeUndefinedInQuery});
   this._coerce(filter.where, options);
   return filter;
 };
@@ -2325,9 +2325,9 @@ DataAccessObject.destroyAll = function destroyAll(where, options, cb) {
     } else {
       try {
         // Support an optional where object
-        var handleUndefined = Model._getSetting('normalizeUndefinedInQuery');
+        var normalizeUndefinedInQuery = Model._getSetting('normalizeUndefinedInQuery');
         // alter configuration of how sanitizeQuery handles undefined values
-        where = Model._sanitize(where, {handleUndefined: handleUndefined});
+        where = Model._sanitize(where, {normalizeUndefinedInQuery: normalizeUndefinedInQuery});
         where = Model._coerce(where, options);
       } catch (err) {
         return process.nextTick(function() {
@@ -2480,9 +2480,9 @@ DataAccessObject.count = function(where, options, cb) {
   where = query.where;
 
   try {
-    var handleUndefined = Model._getSetting('normalizeUndefinedInQuery');
+    var normalizeUndefinedInQuery = Model._getSetting('normalizeUndefinedInQuery');
     // alter configuration of how sanitizeQuery handles undefined values
-    where = Model._sanitize(where, {handleUndefined: handleUndefined});
+    where = Model._sanitize(where, {normalizeUndefinedInQuery: normalizeUndefinedInQuery});
     where = this._coerce(where, options);
   } catch (err) {
     process.nextTick(function() {
@@ -2775,9 +2775,9 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
   function doUpdate(where, data) {
     try {
       // Support an optional where object
-      var handleUndefined = Model._getSetting('normalizeUndefinedInQuery');
+      var normalizeUndefinedInQuery = Model._getSetting('normalizeUndefinedInQuery');
       // alter configuration of how sanitizeQuery handles undefined values
-      where = Model._sanitize(where, {handleUndefined: handleUndefined});
+      where = Model._sanitize(where, {normalizeUndefinedInQuery: normalizeUndefinedInQuery});
       where = Model._coerce(where, options);
       data = Model._sanitize(data);
       data = Model._coerce(data, options);
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1643,6 +1643,26 @@ function coerceArray(val) {
   return arrayVal;
 }
 
+DataAccessObject._getHiddenProperties = function() {
+  var settings = this.definition.settings || {};
+  var result = settings.hiddenProperties || settings.hidden;
+  if (typeof result === 'string') {
+    result = [result];
+  }
+  result = result || [];
+  return result;
+};
+
+DataAccessObject._getProtectedProperties = function() {
+  var settings = this.definition.settings || {};
+  var result = settings.protectedProperties || settings.protected;
+  if (typeof result === 'string') {
+    result = [result];
+  }
+  result = result || [];
+  return result;
+};
+
 /*
  * Coerce values based the property types
  * @param {Object} where The where clause
@@ -1658,6 +1678,13 @@ DataAccessObject._coerce = function(where, options) {
   }
 
   options = options || {};
+
+  // Check violation of keys
+  var prohibitedKeys = this._getHiddenProperties();
+  if (options.excludeProtectedProperties) {
+    prohibitedKeys = prohibitedKeys.concat(this._getProtectedProperties());
+  }
+  utils.validateKeys(where, prohibitedKeys);
 
   var err;
   if (typeof where !== 'object' || Array.isArray(where)) {
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1477,6 +1477,7 @@ DataAccessObject._getSetting = function(key) {
 };
 
 var operators = {
+  eq: '=',
   gt: '>',
   gte: '>=',
   lt: '<',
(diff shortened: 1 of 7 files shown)


test/model-definition.test.js
```diff
@@ -334,6 +333,77 @@ describe('ModelDefinition class', function() {
     });
   });
 
+  it('should report errors if a hidden property is used in where', function(done) {
+    var Child = memory.createModel('child', {}, {hidden: ['secret']});
+    Child.create({
```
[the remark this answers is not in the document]
raymondfeng: We probably need to port these fixes to lower versions. Using callback is backward-compatible.

raymondfeng:
@bajtos PTAL

test/model-definition.test.js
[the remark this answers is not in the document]
bajtos: Yes, we should port these fixes to lower versions which are callback based.

I'd still like us to use async/await in new code, otherwise we are introducing even more work for the upcoming refactoring of juggler.

In 3.x, we should be able to use Promises instead of callbacks, it makes the test code simpler compared to callbacks.

I am not sure about 2.x, we may be able to use callbacks there too.

bajtos:
@zbarbuto thank you for chiming in.

> Silently strip the property from the filter and continue as normal - similar to what finds do with hidden fields

This proposal makes sense to me, ideally we should print a `debug` log to make troubleshooting easier.

@raymondfeng Thoughts?

zbarbuto:
Agreed. Brute force (or at least intelligent guesses) was more the case I was thinking of in my original post.

raymondfeng:
@bajtos @zbarbuto I checked in new changes to hide offending property names from the error object but allow them to be printed out with `DEBUG`. PTAL.

lib/dao.js
```diff
@@ -1814,12 +1842,28 @@ DataAccessObject._coerce = function(where, options) {
         } else if (allowExtendedOperators && typeof val === 'object') {
           // Do not coerce object values when extended operators are allowed
         } else {
+          if (!allowExtendedOperators) {
+            var extendedOperators = Object.keys(val).filter(function(k) {
+              return k.indexOf('$') === 0;
+            });
+            if (extendedOperators.length) {
+              const msg = 'Operators "' + extendedOperators.join(', ') + '" are not allowed in query';
```
bajtos: Please use `g.f` to format the message so that we can translate it.

lib/utils.js
```diff
@@ -339,6 +341,40 @@ function removeUndefined(query, handleUndefined) {
...
+  });
+  if (offendingKeys.length) {
+    const msg = 'Invalid properties are used in query';
+    const err = new Error(msg);
+    err.code = 'PROPERTY_NOT_ALLOWED_IN_QUERY';
+    err.statusCode = 400;
+    err.details = {where: where};
+    debug('Hidden or protected properties %j are used in query: %j',
+      offendingKeys, where, err);
+    throw err;
```
bajtos: By throwing an error we are signaling the client that their request contained a hidden property. As a result, a potential attacker has still a tool to discover hidden properties:
 - Send a query with a single offending key. 
 - If the server returns a `PROPERTY_NOT_ALLOWED_IN_QUERY ` error, we found a hidden model property. Bingo!
 - Otherwise our guess was wrong and we can try another name.

IMO, the only secure solution is to silently ignore hidden/protected fields (from the point of the HTTP/REST client) and log the error for the operator.

I would personally always print the error via `console.error`:
 - at development time, a console error will make it easier for developers to realize they are sending a wrong query
 - in production, a console error log should hopefully trigger investigation and make it more clear the app may be under attack

test/model-definition.test.js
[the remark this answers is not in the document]
bajtos: ☝️ 

You haven't addressed my comment above, nor responded with any counter-argument.

test/model-definition.test.js
```diff
@@ -334,6 +330,136 @@ describe('ModelDefinition class', function() {
...
+    var Child = memory.createModel('child', {}, {protected: ['secret']});
+    Parent.hasMany(Child);
+    Parent.create({
+      name: 'parent',
+    }, function(err, parent) {
+      if (err) return done(err);
+      parent.children.create({
+        name: 'child',
+        secret: 'secret',
+      }, function(err) {
```
bajtos: AFAIK, this test setup is duplicated a lot in the newly added test. Please extract it into a shared helper. I think it may be easiest to wrap the new tests in a new `describe` block, define `Parent`, `Child`, `parent, `child` as describe-level variables and initialize them from a `beforeEach` hook.

raymondfeng:
@bajtos @zbarbuto PTAL

raymondfeng:
@bajtos ping.

lib/dao.js
```diff
@@ -1579,8 +1580,8 @@ DataAccessObject._normalize = function(filter, options) {
   }
 
   var handleUndefined = this._getSetting('normalizeUndefinedInQuery');
-  // alter configuration of how removeUndefined handles undefined values
-  filter = removeUndefined(filter, handleUndefined);
+  // alter configuration of how sanitizeQuery handles undefined values
+  filter = this._sanitize(filter, {handleUndefined: handleUndefined});
```
bajtos: can we rename the option from `handleUndefined` to `normalizeUndefinedInQuery`? I find `handleUndefined` slightly confusing (what does it mean to handle undefined values?), plus the setting is already called `normalizeUndefinedInQuery` so let's be consistent.

lib/dao.js
```diff
@@ -1643,6 +1644,38 @@ function coerceArray(val) {
...
+  result = result || [];
+  return result;
+};
+
+DataAccessObject._sanitize = function(query, options) {
+  options = options || {};
+
+  // Check violation of keys
+  var prohibitedKeys = this._getHiddenProperties();
+  if (options.excludeProtectedProperties) {
```
bajtos: Should we exclude protected properties by default? I don't fully comprehend what does it mean for a property to be protected and thus whether it should be excluded from queries by default. I am raising this question so that we can consider both alternatives and make the right decision.

lib/dao.js
```diff
@@ -1814,12 +1846,28 @@ DataAccessObject._coerce = function(where, options) {
         } else if (allowExtendedOperators && typeof val === 'object') {
           // Do not coerce object values when extended operators are allowed
         } else {
+          if (!allowExtendedOperators) {
+            var extendedOperators = Object.keys(val).filter(function(k) {
+              return k.indexOf('$') === 0;
+            });
+            if (extendedOperators.length) {
```
bajtos: Have you considered using [Array.prototype.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) to save unnecessary allocation of `extendedOperators` array?

Also the check for `$` at position 0 can be simplified as `k[0] === '$'`.

```js
const hasExtendedOperators = Object.keys(val).some(function(k) {
  return k[0] === '$';
});

if (hasExtendedOperators) {
// ...
```

lib/dao.js
bajtos: > Please use `g.f` to format the message so that we can translate it.

☝️ ❗️

lib/include_utils.js
```diff
@@ -5,6 +5,8 @@
 
 'use strict';
 
+var g = require('strong-globalize')();
```
bajtos: Use `const` please, see https://loopback.io/doc/en/contrib/style-guide-es6.html#variable-declarations

There are other places in this pull request where you are using `var` instead of `const` in new code, it would be great if you could fix them too (but I am not going to insist on that).

lib/dao.js
raymondfeng: `protected` properties are hidden from included items but visible to the its own model instances.

lib/dao.js
raymondfeng: I wanted to report all violated operators instead of just a flag

lib/include_utils.js
raymondfeng: My intention to keep it consistent with existing code. I'm open to have separate PRs to use `const` over `var`.

raymondfeng:
@bajtos PTAL

lib/dao.js
bajtos: Makes sense, thanks for the explanation 👍 

Can you please add your explanation as a code comment to help future readers?

lib/dao.js
bajtos: Makes sense 👍

lib/include_utils.js
bajtos: The convention I am pushing for is to start using `const` in new code we are writing, so that the future migration from `var` to `const` can touch less files.

Not a big deal though.

lib/dao.js
bajtos: > can we rename the option from `handleUndefined` to `normalizeUndefinedInQuery`? I find `handleUndefined` slightly confusing (what does it mean to handle undefined values?), plus the setting is already called `normalizeUndefinedInQuery` so let's be consistent.

@raymondfeng ☝️ I think you forgot to address this comment.

bajtos:
@raymondfeng when you land this pull request, the fix will be available to LB4 users only. Please consider which branches to back-port it to. I think we should back-port it to all versions covered by LTS, i.e. both 3.x and 2.x.

raymondfeng:
Released as 4.1.0 and 3.25.0.

There are a lot of merging conflicts with 2.x branch. I'm not confident to back-port it to 2.x.

tvdstaaij:
Isn't this only relevant to remote queries? My internal application code (custom model logic, timed tasks etc) often queries on hidden properties and is now spamming stderr with "Potential security alert" messages. I'm a bit puzzled on why this should affect internal application queries -- isn't that the purpose of hidden properties, that you can use them internally but they aren't available on the API?

In the commits I noticed a `prohibitProtectedPropertiesInQuery` flag, does this opt out of this change completely? Where should this flag be set? The commit message of 8fa7c94 mentions it can be set on a datasource and I would greatly prefer that over individual models, but adding `prohibitProtectedPropertiesInQuery: false` it to the properties of my mongo datasource doesn't seem to have any effect.

bajtos:
@tvdstaaij thank you for the feedback! It seems to me that we have focused too much on the REST API side of things and overlooked the possibilities to query hidden properties from javascript code. Sorry for the inconvenience we have caused!

As far as I understand the changes made in this pull request, there is no way how to allow querying of hidden properties now.

I am proposing to fix this problem by introducing a new method-level `options` flag, e.g. `options.includeHiddenProperties`. Example usage:

```ts
const results = MyModel.find({where: {hiddenProp: 'some-value'}}, {includeHiddenProperties: true});
```

Thoughts?

cc @raymondfeng

raymondfeng:
I'm evaluating https://loopback.io/doc/en/lb3/Using-current-context.html#customize-the-value-provided-to-options. Here is a tentative solution:

1. Allow method level options to control such flags
2. Fix `Model. createOptionsFromRemotingContext()` to either set `remote: true` or configure such flags to best protect remote requests.

bajtos:
Using `createOptionsFromRemotingContext` makes a lot of sense to me 👍 

We would still need to modify DAO methods to accept a new `options` flag that will enable/disable querying of hidden properties, wouldn't we?

raymondfeng:
> We would still need to modify DAO methods to accept a new options flag that will enable/disable querying of hidden properties, wouldn't we?

Yes, we need to take `options` into consideration to look up such flags.

PR #1639, opened 2018-10-17
Fix crash app when include is number

nguyenphuquang:
### Description

When include parameter is number, boolean (ex: {"include": 1}), the server will crash


slnode:

raymondfeng:
@nguyenphuquang Thank you for the patch. Can you add a test?

spurreiter:
Looks like a duplicate of #1581

Issue #1640, opened 2018-10-18
If define 'scope' in mixin, the first argument should always be the corresponding query


seanxlliu:
# Description/Steps to reproduce

1. define a mixin to restrict `limit` in query as following

```
module.exports = function(Model, options) {
  Model.settings.scope = function(target, inst) {
    let limit = options.queryLimit || 100;
    if (!target.limit || target.limit > limit) {
      target.limit = limit;
      debug('limit is replaced to %d', limit);
    }
    return target;
  };
};

```

this will cause the data populate a new field
{
"limit": 100
}

# Expected result

should not include `limit`

# Additional information

in dao.js, data should not be put in defaultScope as the first argument.
```
DataAccessObject.applyProperties = function(data, inst) {
  var properties = this.definition.settings.properties;
  properties = properties || this.definition.settings.attributes;
  if (typeof properties === 'object') {
    util._extend(data, properties);
  } else if (typeof properties === 'function') {
    util._extend(data, properties.call(this, data, inst) || {});
  } else if (properties !== false) {
    var scope = this.defaultScope(data, inst) || {};
    if (typeof scope.where === 'object') {
      setScopeValuesFromWhere(data, scope.where, this);
    }
  }
};

```

seanxlliu:
Submitted a suggested PR

PR #1641, opened 2018-10-18
fix issue #1640 to avoid unexpected results with scope function

seanxlliu:
### Description

If apply default 'scope' as a function in a mixin, the query results will lead to unexpected results

#### Related issues

- connect to #1640 

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -210,7 +210,7 @@ DataAccessObject.applyProperties = function(data, inst) {
   } else if (typeof properties === 'function') {
     util._extend(data, properties.call(this, data, inst) || {});
   } else if (properties !== false) {
-    var scope = this.defaultScope(data, inst) || {};
+    var scope = this.defaultScope({}, inst) || {};
     if (typeof scope.where === 'object') {
       setScopeValuesFromWhere(data, scope.where, this);
     }


slnode:

seanxlliu:
Can someone please review? @bajtos

raymondfeng:
@seanxlliu Thank you for the patch. Can you explain why the fix is needed? Some tests will be really helpful.

seanxlliu:
@raymondfeng thanks for checking. I was considering an UT but I found the major scenario has been covered by tests name like 'should apply default scope'. The code logic has been covered like following,

```
    var scopeFn = function(target, inst) {
      return {where: {kind: this.modelName}};
    };
```

This issue can only be triggered by a corner case, but I think the original code itself should be unexpected 'typo' or something. please check the code snippet in the issue #1640 .

supposing we have a mixin like following, Model.settings.scope is set to a function instead of a normal query option, and the expected input parameter `target` is a query option so that the query cannot modified accordingly, however, the bug causes the target to be a model data, so if data will be populated with `limit` and returned in the query result.

Note: in the loopback code, `defaultScope` is called in two locations,

[Expected behavior](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L199)

[Unexpected behavior](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L213)

```
module.exports = function(Model, options) {
  Model.settings.scope = function(target, inst) {
    let limit = options.queryLimit || 100;
    if (!target.limit || target.limit > limit) {
      target.limit = limit;
      debug('limit is replaced to %d', limit);
    }
    return target;
  };
};

```

If it is still confusing or any UT suggestions, please let me know.

seanxlliu:
@raymondfeng ping...

raymondfeng:
@seanxlliu Is it caused by the fact that your mixin update `target` in place instead of returning a copy? 

```js
module.exports = function(Model, options) {
  Model.settings.scope = function(target, inst) {
   // What about clone target here?
    let limit = options.queryLimit || 100;
    if (!target.limit || target.limit > limit) {
      target.limit = limit;
      debug('limit is replaced to %d', limit);
    }
    return target;
  };
};
```

bajtos:
Raymond's conclusion seems reasonable to me. I think it means you need to change your code as follows:

```diff
 if (!target.limit || target.limit > limit) {
+  target = Object.assign({}, target);
   target.limit = limit;
 }
```

seanxlliu:
@bajtos @raymondfeng 
Sorry for late response.  the suggestion looks like a workaround and I have done similar to the suggestion in the first place. However, I just don't get the reason that the framework code passes `data` as the first argument. 
[here](https://github.com/strongloop/loopback-datasource-juggler/blob/70b3bb6aae8e93536076ff5ca6f742c5fb19bdda/lib/dao.js#L214), `var scope = this.defaultScope(data, inst) || {};`

PR #1642, opened 2018-10-18
Report errors for missing id properties for include

raymondfeng:
https://github.com/strongloop/loopback/issues/4028

#### Related issues

### Checklist

### DIFF
--- a/test/include.test.js
+++ b/test/include.test.js
@@ -46,6 +46,15 @@ describe('include', function() {
     });
   });
 
+  it('does not return included item if FK is excluded', function(done) {
+    Passport.find({include: 'owner', fields: 'number'}, function(err, passports) {
+      if (err) return done(err);
+      var owner = passports[0].toJSON().owner;
+      should.not.exist(owner);
+      done();
+    });
+  });
+
   it('should fetch hasMany relation', function(done) {
     User.find({include: 'posts'}, function(err, users) {
       should.not.exist(err);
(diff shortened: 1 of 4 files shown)


test/include.test.js
```diff
@@ -579,6 +595,15 @@ describe('include', function() {
           });
         });
 
+        it('does not return included item if hasOne is missing the id property', function(done) {
+          User.findOne({include: {relation: 'profile'}, fields: 'name'}, function(err, user) {
+            should.not.exist(err);
+            should.exist(user);
+            should.not.exist(user.toJSON().profile);
```
b-admike: I'm wondering if we should error out here as well and say that the `id` property is missing. Thoughts?

raymondfeng: For belongsTo and hasOne, the FK can be null from business function perspective. We can have an option to report as errors if necessary.

test/include.test.js
```diff
@@ -579,6 +595,15 @@ describe('include', function() {
           });
         });
 
+        it('does not return included item if hasOne is missing the id property', function(done) {
+          User.findOne({include: {relation: 'profile'}, fields: 'name'}, function(err, user) {
+            should.not.exist(err);
```
bajtos:
```js
if (err) return done(err);
```

"should not exist" check discards the stack trace from the original error, making the test failure difficult to troubleshoot.

PR #1643, opened 2018-10-19
Clear cache after destroying related model (hasOne)

ivanschwarz:
### Description

Related model is still accessible after its destruction in the `hasOne` relationship. I believe the cache should be cleared to avoid that. I've extended existing tests to demonstrate the problem.

Can someone please review? Maybe @bajtos ?

### Checklist

### DIFF
--- a/lib/relation-definition.js
+++ b/lib/relation-definition.js
@@ -1862,9 +1862,11 @@ HasOne.prototype.destroy = function(options, cb) {
     options = {};
   }
   cb = cb || utils.createPromiseCallback();
+  var self = this;
   var definition = this.definition;
   this.fetch(function(err, targetModel) {
     if (targetModel instanceof ModelBaseClass) {
+      self.resetCache();
       targetModel.destroy(options, cb);
     } else {
       cb(new Error(g.f('{{HasOne}} relation %s is empty', definition.name)));
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -3715,7 +3715,11 @@ describe('relations', function() {
         should.exist(supplier);
         supplier.account.destroy(function(err) {
           should.not.exist(e);
-          done();
+          supplier.account(function(err, act) {
+            if (err) return done(err);
+            should.not.exist(act);
+            done();
+          });
         });
       });
     });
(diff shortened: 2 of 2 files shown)


slnode:

bajtos:
@slnode ok to test

bajtos:
@ivanschwarz thank you for the pull request. I am wondering whether we should be clearing the cache for other relation methods too, e.g. when a target model in `HasMany` relation was deleted? Also how about other methods beside delete, should we clear the cache also when a related model is updated or created?

Have you considered to call the relation API in such way that the cache is bypassed?

```diff
- supplier.account(function(err, act) {
+ supplier.account(true, function(err, act) {
```

To be honest, I never really liked this caching feature and was bitten by it multiple times myself too. Having said that, I am more concerned about the impact of caching changes on existing application and the difficulty of implementing a robust cache invalidation inside our relations.

ivanschwarz:
@bajtos : 
If a target model in `HasMany` relationship is deleted, the cache is adjusted with `HasMany.prototype.removeFromCache`. I guess I could extend other tests to see, if this this is an issue in other relations.

If related model is created or updated, cache should be already in a correct state. I think it's only an issue if related model is destroyed. That also includes a situation when related mode is replaced. That's where I stumbled over it originally. Again, I guess more tests could tell us more. I'm fairly new to the codebase, so I might be definitely making wrong assumptions.

I didn't know about the possibility to bypass cache. That could be indeed a workaround. How would that work if I would like to use promises?

bajtos:
A apologize for the long silence on my side, @ivanschwarz 

> I didn't know about the possibility to bypass cache. That could be indeed a workaround. How would that work if I would like to use promises?

Use `myModel.myRelation.get()` API, it will always bypass the cache.

```js
await supplier.account.get();
```

The implementation: https://github.com/strongloop/loopback-datasource-juggler/blob/47ef38e9f1c71592a4a89a7e0dd176fec1fcac91/lib/relation-definition.js#L2033-L2041

Would you still like to proceed with this pull request? Should we close it?

bajtos:
Closing this pull request as abandoned. Feel free to reopen if you get time to work on this again.

Issue #1646, opened 2018-10-27
prohibitedKeys is sometimes an object instead of an array


joecarney:
# Description/Steps to reproduce

# Link to reproduction sandbox

# Expected result

# Additional information

Opening /explorer and running a query works first time but subsequent times it crashes with the following error:

```
Error-logger request  : GET /email-campaigns: TypeError: prohibitedKeys.indexOf is not a function
    at Object.<anonymous> (/node_modules/loopback-datasource-juggler/lib/utils.js:356:42)
```

I added the following line of debug to /lib/utls:

```
    /**
     * Make sure prohibited keys are removed from the query to prevent
     * sensitive values from being guessed
     */
    console.log('prohibitedKeys', prohibitedKeys);
    if (
      prohibitedKeys &&
      prohibitedKeys.indexOf(this.key) !== -1
    ) {
      offendingKeys.push(this.key);
      this.remove();
      return;
    }
```

First time I run the query...

```
2018-10-27T09:12:39.322Z prohibitedKeys []
2018-10-27T09:12:39.322Z prohibitedKeys []
2018-10-27T09:12:39.324Z prohibitedKeys []
```

Second time I run the query...

```
2018-10-27T09:12:40.312Z prohibitedKeys []
2018-10-27T09:12:40.312Z prohibitedKeys []
2018-10-27T09:12:40.313Z prohibitedKeys {}
```

As you can see, prohibitedKeys is changing from an array to object.

I've added a check for an Array just so I can continue working on my project...

```
    /**
     * Make sure prohibited keys are removed from the query to prevent
     * sensitive values from being guessed
     */
    console.log('prohibitedKeys', prohibitedKeys);
    if (
      prohibitedKeys &&
      Array.isArray(prohibitedKeys) &&
      prohibitedKeys.indexOf(this.key) !== -1
    ) {
      offendingKeys.push(this.key);
      this.remove();
      return;
    }
```

However, I think this problem is deeper inside loopback.  

My current config...

```
node -v
v11.0.0
```

```
mysql -V
mysql  Ver 8.0.12 for osx10.14 on x86_64 (Homebrew)
```

```
node -e 'console.log(process.platform, process.arch, process.versions.node)'
npm ls --prod --depth 0 | grep loopback

darwin x64 11.0.0
├── loopback@3.23.2
├── loopback-boot@2.27.1
├── loopback-component-explorer@6.3.1
├── loopback-connector-mysql@5.3.1
[...]

PR #1647, opened 2018-10-27
Allow hidden/protected props as an object

raymondfeng:
### Description

Model settings hidden/protected properties can be an array or object, such as:

- hidden: ['secret']
- hidden: {secret: true}

#### Related issues

#1646 

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1646,22 +1646,32 @@ function coerceArray(val) {
 
 DataAccessObject._getHiddenProperties = function() {
   var settings = this.definition.settings || {};
-  var result = settings.hiddenProperties || settings.hidden;
+  var result = settings.hiddenProperties || settings.hidden || [];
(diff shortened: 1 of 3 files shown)


lib/dao.js
```diff
@@ -1646,22 +1646,32 @@ function coerceArray(val) {
...
     result = [result];
   }
-  result = result || [];
-  return result;
+  if (Array.isArray(result)) {
+    return result;
+  } else {
+    // See https://github.com/strongloop/loopback-datasource-juggler/issues/1646
+    // `ModelBaseClass` normalize the properties to an object such as `{secret: true}`
+    return Object.keys(result);
```
zbarbuto: I think doing a straight `Object.keys` is a bit misleading without doing a bit of filtering. 

The way it stands it seems to me `{secret: false}` would still be filtered as hidden.

Issue #1651, opened 2018-10-30
Regression in multi-level related queries


ashearer:
The recent commit to prevent "deeply nested or circular" queries breaks existing well-behaved apps that chain more than 3 relationships in a programmatic query.

The new code, in addition to checking whether the traverse package marks the query `circular`, also checks the number of levels used to represent the query against an arbitrary limit of 10. Each chained relationship in the query consumes 2 or 3 levels in the object representation, so the 4th relationship can fail.

There's a new field in the "options" parameter to change the maximum depth to something other than 10, but it's not clear how to set it or whether it's possible to set, since the existing callers don't appear to pass it through.

At a minimum, the limit should be raised (for example, to allow at least 10 chained relationships instead of 3). Even better, the arbitrary depth limit should be removed entirely, since circularity is the important criterion.

The regression occurred in commit 0ce3f4ead9914951aec275c1a4829d2bf5afb56d

# Description/Steps to reproduce

Made-up example:

```
const storeReport = models.Store.findById(id, {
  include: [
    {relation: "retailChain"},
    {relation: "receivables",
     scope: {
       include: [
         {relation: "brand"},
         {
           relation: "sender",
           scope: {
             include: {
               relation: "region",
               scope: [
                 {
                   include: {relation: "country"}
                 }
               ]
             }
           }
         }
       ]
     }
   }
 ]
});
```
# Expected result
Success

# Actual result
exception `The query object is too deep or circular`

# Additional information

darwin x64 8.11.3
```
├── loopback@3.23.2
├── loopback-boot@2.27.1
├── loopback-component-explorer@6.3.1
├── loopback-connector-mysql@5.3.1
├── loopback-console@1.1.0
```

dhmlau:
@raymondfeng, could you PTAL? Thanks.

raymondfeng:
@ashearer What do you think about https://github.com/strongloop/loopback-datasource-juggler/commit/0d84b55f908e46f21eb921fd00667fecd8020ff0?

raymondfeng:
Please use 3.26.0/4.2.0 and follow instructions at https://github.com/strongloop/loopback.io/blob/gh-pages/pages/en/lb3/Querying-data.md#sanitizing-filter-objects.

ashearer:
Thanks for the commit! This makes it possible to work around the problem.

[...]

PR #1652, opened 2018-10-30
Allow configuration of maxDepthOfQuery

raymondfeng:
https://github.com/strongloop/loopback-datasource-juggler/issues/1651

#### Related issues

- connect to #1653 #1651

### Checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -336,18 +336,25 @@ function sanitizeQuery(query, options) {
   const prohibitedKeys = options.prohibitedKeys;
   const offendingKeys = [];
   const normalizeUndefinedInQuery = options.normalizeUndefinedInQuery;
-  const maxDepth = options.maxDepth || 10;
+  const maxDepth = options.maxDepth || 12;
   // WARNING: [rfeng] Use map() will cause mongodb to produce invalid BSON
   // as traverse doesn't transform the ObjectId correctly
   const result = traverse(query).forEach(function(x) {
     /**
      * Security risk if the client passes in a very deep where object
      */
-    if (this.level > maxDepth || this.circular) {
-      const msg = g.f('The query object is too deep or circular');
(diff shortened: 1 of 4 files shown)


test/model-definition.test.js
```diff
@@ -257,6 +257,35 @@ describe('ModelDefinition class', function() {
     done();
   });
 
+  it('should report errors for deep query than maxDepthOfQuery', function(done) {
+    var MyModel = memory.createModel('my-model', {}, {
+      maxDepthOfQuery: 5,
+    });
+
+    var filter = {where: {and: [{and: [{and: [{and: [{and: [{and:
```
b-admike: nitpick: In the future, if we have more tests like these, we can extract this part out to a helper function like `givenFilterWithDepth()` or something like that for better readability.

lib/utils.js
```diff
@@ -336,18 +336,25 @@ function sanitizeQuery(query, options) {
...
      * Security risk if the client passes in a very deep where object
      */
+    if (this.circular) {
+      const msg = g.f('The query object is circular');
+      const err = new Error(msg);
+      err.statusCode = 400;
+      err.code = 'QUERY_OBJECT_IS_CIRCULAR';
+      throw err;
+    }
     if (this.level > maxDepth || this.circular) {
```
b-admike: Can we remove the check for `this.circular` in this if condition here since we've separated the checks?

b-admike:
@raymondfeng I think we should also include some docs about the usage of this new flag for users opting to customize it for their applications.

raymondfeng:
Docs - https://github.com/strongloop/loopback.io/pull/760

lib/utils.js
raymondfeng: fixed

raymondfeng:
@b-admike PTAL

PR #1654, opened 2018-11-05
fix: update typescript declarations

b-admike:
### Description
Fix return type of PersistedModel's `findOrCreate` function so that it conforms to the return contract of connectors which support it (took mongodb and memory as examples). Also, add a string index signature to the `Connector` interface to allow checks for atomic functions.

Connect to https://github.com/strongloop/loopback-next/issues/1956. 

### Checklist

### DIFF
--- a/types/connector.d.ts
+++ b/types/connector.d.ts
@@ -15,6 +15,7 @@ export interface Connector {
   disconnect(callback?: Callback): PromiseOrVoid; // Disconnect from the underlying system
   ping(callback?: Callback): PromiseOrVoid; // Ping the underlying system
   execute?(...args: any[]): Promise<any>;
+  [property: string]: any; // Other properties that vary by connectors
 }
 
 /**
--- a/types/persisted-model.d.ts
+++ b/types/persisted-model.d.ts
@@ -154,7 +154,7 @@ export declare class PersistedModel extends ModelBase {
     data: PersistedData,
     options?: Options,
     callback?: Callback<PersistedModel>,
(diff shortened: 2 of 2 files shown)


types/persisted-model.d.ts
```diff
@@ -154,7 +154,7 @@ export declare class PersistedModel extends ModelBase {
     data: PersistedData,
     options?: Options,
     callback?: Callback<PersistedModel>,
-  ): PromiseOrVoid<PersistedModel>;
+  ): PromiseOrVoid<[PersistedModel, boolean]>;
```
bajtos: :+1:

types/connector.d.ts
```diff
@@ -15,6 +15,7 @@ export interface Connector {
   disconnect(callback?: Callback): PromiseOrVoid; // Disconnect from the underlying system
   ping(callback?: Callback): PromiseOrVoid; // Ping the underlying system
   execute?(...args: any[]): Promise<any>;
+  [property: string]: any; // Other properties that vary by connectors
```
bajtos: At minimum please use `Function` instead of `any`.

Personally, I'd prefer to see a list of well-known properties declared as optional (see `execute` above).

bajtos: We cannot use function for the indexed properties because this type must support all well-defined properties too.

bajtos:
@b-admike what's the status of this pull request? Can you rebase it on top of the latest master, fix Travis CI build and get it landed?

PR #1656, opened 2018-11-06
Fix createModelClassCtor instanceof + new for models named “data” or “options”

ryan-open:
### Description

Building a function out of a string with `new Function` isn’t necessary to set its name.

This isn’t strictly backwards-compatible because names that aren’t valid identifiers after replacing `[-.:]` won’t become `ModelConstructor` anymore. That’s in order to avoid something silly like:

```js
try {
  void new Function(`
    void function ${name}(data, options) {
      if (!(this instanceof ${name})) {}
    };
    ${JSON.stringify(name)};`);
} catch (error) {
  name = 'ModelConstructor';
}
```

(which still doesn’t cover everything from the original if you take into account names designed to break it.)

### Checklist

- [X] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide.html)

### DIFF
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -656,41 +656,19 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
 };
 
 function createModelClassCtor(name, ModelBaseClass) {
-  // A simple sanitization to handle most common characters
-  // that are used in model names but cannot be used as a function/class name.
-  // Note that the rules for valid JS indentifiers are way too complex,
-  // implementing a fully spec-compliant sanitization is not worth the effort.
-  // See https://mathiasbynens.be/notes/javascript-identifiers-es6
+  // Backwards-compatible replacements of common characters.
+  // https://github.com/strongloop/loopback-datasource-juggler/pull/1534
   name = name.replace(/[-.:]/g, '_');
-  let constructor;
 
-  try {
-    // It is not possible to access closure variables like "ModelBaseClass"
-    // from a dynamically defined function. The solution is to
-    // create a dynamically defined factory function that accepts
-    // closure variables as arguments.
-    const factory = new Function('ModelBaseClass', `
-      // every class can receive hash of data as optional param
-      return function constructor(data, options) {
-        if (!(this instanceof constructor)) {
-          return new constructor(data, options);
-        }
-        if (constructor.settings.unresolved) {
-          throw new Error(g.f('Model %s is not defined.', ${JSON.stringify(name)}));
-        }
-        ModelBaseClass.apply(this, arguments);
-      };`);
-
-    constructor = factory(ModelBaseClass);
-  } catch (err) {
-    // modelName is not a valid function/class name, e.g. 'grand-child'
-    // and our simple sanitization was not good enough.
-    // Falling back to legacy 'ModelConstructor' name.
-    if (err.name === 'SyntaxError') {
-      return createModelClassCtor('ModelConstructor', ModelBaseClass);
-    } else {
-      throw err;
+  // every class can receive hash of data as optional param
+  function constructor(data, options) {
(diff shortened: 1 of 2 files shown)


slnode:

jannyHou:
@slnode test please

lib/model-builder.js
```diff
@@ -656,41 +656,27 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
...
+    if (!(this instanceof constructor)) {
+      return new constructor(data, options);
+    }
+    if (constructor.settings.unresolved) {
+      throw new Error(g.f('Model %s is not defined.', name));
     }
+    ModelBaseClass.apply(this, arguments);
   }
+
+  Object.defineProperty(constructor, 'name', {
```
bajtos: I am surprised this works. Can you please verify that the `name` property is used by error stack traces too? In the past, it was not possible to override function name this way.

bajtos: It's also to verify this works on all Node.js versions we support, I believe that means Node.js 8.x, 10.x and 11.x for the `master` branch.

ryan-open: > Can you please verify that the `name` property is used by error stack traces too?

It is.

> It's also important to verify this works on all Node.js versions we support, I believe that means Node.js 8.x, 10.x and 11.x for the master branch.

It has worked since Node 4.

ryan-open:
This is primarily a bug fix, not refactoring.

bajtos:
@ryan-open I am afraid your proposal does not work as intended. Consider the following code snippet:

```js
function createModelClassCtor(name, ModelBaseClass) {
  name = name.replace(/[-.:]/g, '_');
  // every class can receive hash of data as optional param
  function constructor(data, options) {
    if (!(this instanceof constructor)) {
      return new constructor(data, options);
    }
    ModelBaseClass.apply(this, arguments);
  }

  Object.defineProperty(constructor, 'name',
    {
      configurable: true,
      value: name,
    });

  return constructor;
}

const modelCtor = createModelClassCtor('User', Object);
modelCtor.prototype.fail = function() {
  throw new Error('bang!');
}

new modelCtor().fail();
```

It prints the following error:

```
$ node ctor.js
/private/tmp/ctor.js:22
  throw new Error('bang!');
  ^

Error: bang!
    at constructor.modelCtor.fail (/private/tmp/ctor.js:22:9)
    at Object.<anonymous> (/private/tmp/ctor.js:30:17)
```

The model name `User` is lost from the stack trace.

Compare this with the original approach we have in juggler now:

```js
function createModelClassCtor(name, ModelBaseClass) {
  name = name.replace(/[-.:]/g, '_');

  // It is not possible to access closure variables like "ModelBaseClass"
  // from a dynamically defined function. The solution is to
  // create a dynamically defined factory function that accepts
  // closure variables as arguments.
  const factory = new Function('ModelBaseClass', `
    // every class can receive hash of data as optional param
    return function ${name}(data, options) {
      if (!(this instanceof ${name})) {
        return new ${name}(data, options);
      }
      ModelBaseClass.apply(this, arguments);
    };`);

  return factory(ModelBaseClass);
}

const modelCtor = createModelClassCtor('User', Object);
modelCtor.prototype.fail = function() {
  throw new Error('bang!');
}

new modelCtor().fail();
```

It prints the following output:

```
js:22
  throw new Error('bang!');
  ^

Error: bang!
    at User.modelCtor.fail (/private/tmp/ctor.js:22:9)
    at Object.<anonymous> (/private/tmp/ctor.js:30:17)
```

bajtos:
Let's find a different solution how to fix the issue you are experiencing. I am proposing to use different variable names when we detect clash with the model name.

ryan-open:
I’m no longer interested in contributing to LoopBack.

PR #1660, opened 2018-11-09
chore: update file list for apidocs

b-admike:
### Description

Add in some missing files from our `lib` directory, so their JSDocs show up at https://apidocs.strongloop.com/loopback-datasource-juggler.

Question: should we include files under `lib/connectors` and `lib/kvao`?

### Checklist

- [ ] New tests added or existing tests modified to cover all changes *N/A*

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -1,7 +1,12 @@
 {
     "content": [
+        "lib/dao.js",
+        "lib/scope.js",
+        "lib/relation-definition.js",
         "lib/datasource.js",
         "lib/model.js",
+        "lib/model-definition.js",
+        "lib/mixins.js",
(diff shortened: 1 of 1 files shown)


dhmlau:
so we need to list all the files one by one?

b-admike:
@dhmlau yeah AFAICT

bajtos:
> Question: should we include files under `lib/connectors` and `lib/kvao`?

My take: `lib/kvao` should be included, those are defining APIs available on models attached to KeyValue connectors (Redis, WebSphere eXtremeScale).

`lib/connectors` provide implementation of connectors, I think they don't have much (if any) public API besides the API contract expected by juggler.

PR #1662, opened 2018-11-12
Use options.remotingContext for conditional settings

raymondfeng:
### Description

This PR uses `options` from remote requests to decide default values for certain security-related settings

#### Related issues

See https://github.com/strongloop/loopback-datasource-juggler/pull/1638#issuecomment-437407868
https://github.com/strongloop/loopback-datasource-juggler/issues/1663

### Checklist

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -323,7 +323,7 @@ Memory.prototype.upsertWithWhere = function(model, where, data, options, callbac
   }
 };
 
-Memory.prototype.findOrCreate = function(model, filter, data, callback) {
+Memory.prototype.findOrCreate = function(model, filter, data, options, callback) {
   var self = this;
   var nodes = self._findAllSkippingIncludes(model, filter);
   var found = nodes[0];
@@ -345,7 +345,7 @@ Memory.prototype.findOrCreate = function(model, filter, data, callback) {
     });
   }
 
-  self._models[model].model.include(nodes[0], filter.include, {}, function(err, nodes) {
+  self._models[model].model.include(nodes[0], filter.include, options, function(err, nodes) {
     process.nextTick(function() {
       if (err) return callback(err);
       callback(null, nodes[0], false);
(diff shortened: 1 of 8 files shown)


lib/connectors/memory.js
```diff
@@ -345,7 +345,7 @@ Memory.prototype.findOrCreate = function(model, filter, data, callback) {
     });
   }
 
-  self._models[model].model.include(nodes[0], filter.include, {}, function(err, nodes) {
+  self._models[model].model.include(nodes[0], filter.include, options, function(err, nodes) {
```
jannyHou: `options || {}` in case the option is `undefined`.

raymondfeng:
@bajtos PTAL

lib/utils.js
```diff
@@ -336,7 +336,7 @@ function sanitizeQuery(query, options) {
   const prohibitedKeys = options.prohibitedKeys;
   const offendingKeys = [];
   const normalizeUndefinedInQuery = options.normalizeUndefinedInQuery;
-  const maxDepth = options.maxDepth || 12;
+  const maxDepth = options.maxDepth || Number.MAX_SAFE_INTEGER;
```
bajtos: Is it a good idea to allow arbitrary deep queries? Or is this reverting back to the old behavior we used to have in juggler before recent security-related fixes?

bajtos:
> coverage/coveralls — Coverage decreased (-0.01%) to 83.954%

Please check the coverage data, there may be an important use case (test scenario) that you forgot to cover.

lib/utils.js
raymondfeng: This is a default value for the utility method itself. Juggler already sets `32` as the default value.

bajtos: So can we use `32` as the default value in this place too? I'd rather avoid all possibilities of implicitly infinite depth.

raymondfeng:
> I don't think it's a good idea to move DataAccessObject._normalize into a different file:
The format of Filter object is to specific to DAO only, KeyValue APIs use different filter properties.
It is pretty much impossible to review changes made in that method.
Such huge change will be difficult to back-port to older versions, because there is no automated way how to ensure the backport is not accidentally back-porting changes in _normalize implementation that are not supposed to be back-ported.

The refactoring is to fix https://github.com/strongloop/loopback-datasource-juggler/issues/1663.

PR #1666, opened 2018-11-20
Fix Promisification of DAO create 

mazesoul:
### Description

The issue can be exhibited by using  
```
  Model.create({date: 'bad data that'})
```

When calling DAO.create without a callback,

- the DAO tries to `applyProperties` [lib/dao.js ln294](https://github.com/strongloop/loopback-datasource-juggler/blob/2.x/lib/dao.js#L294).
- this trigger the DateType to throw in its constructor [lib/dao.js ln1518](https://github.com/strongloop/loopback-datasource-juggler/blob/2.x/lib/dao.js#L1518)
- because of the error, the create method returns early without ever 
returning the promise.

The fix ensures that when this occurs, we always return the promise.
This should fix any other instances of the same bug.

#### Related issues

I'm sure I've read issues related to this previously but can't find 
them anywhere. 
The issues usually refer to custom types introduced with `registerType`

### Checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -493,7 +493,11 @@ function createPromiseCallback() {
 
   var promise = new Promise(function(resolve, reject) {
     cb = function(err, data) {
-      err ? reject(err) : resolve(data);
+      if (err) {
+        reject(err);
+      } else {
+        resolve(data);
+      }
       return promise;
     };
   });
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -493,8 +493,8 @@ function createPromiseCallback() {
 
   var promise = new Promise(function(resolve, reject) {
     cb = function(err, data) {
-      if (err) return reject(err);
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -493,8 +493,12 @@ function createPromiseCallback() {
...
   var promise = new Promise(function(resolve, reject) {
     cb = function(err, data) {
-      if (err) return reject(err);
-      return resolve(data);
+      if (err) {
+        reject(err);
+      } else {
+        resolve(data);
+      }
+      return promise;
```
bajtos: This change is not fixing the problem. The return value of a callback is almost always discarded, therefore it does not matter what this function is returning.

bajtos:
I am closing this pull request as abandoned. Feel free to reopen it if you get time to work on this change again.

PR #1671, opened 2018-12-06
feat: dataSource.execute(cmd, args, opts, cb)

bajtos:
Implement a new helper API for calling connector's "execute" method
in a promise-friendly way.

Example usage:

```ts
await MyModel.dataSource.execute(
  'SELECT * FROM Users WHERE email = ?',
  ['alice@example.com'],
  {/* options */});
```

While it's possible to call `MyModel.dataSource.execute` directly, this new API adds promise support to existing callback-based connectors. When we upgrade connector API to use promises under the hood, consumers of this new API won't be affected by that change, as opposed to consumers of `connector.execute` that will have to upgrade their code.

/cc @koalabi 

#### Related issues

- connect to strongloop/loopback-next#2053
- StackOverflow: [How to execute arbitrary SQL query](https://stackoverflow.com/q/53338050/69868)

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -2594,6 +2594,54 @@ DataSource.prototype.ping = function(cb) {
   return cb.promise;
 };
 
+/**
+ * Execute an arbitrary command. The commands are connector specific,
+ * please refer to the documentation of your connector for more details.
+ *
+ * @param command String|Object The command to execute, e.g. an SQL query.
+ * @param [args] Array Parameters values to set in the command.
+ * @param [options] Object Additional options, e.g. the transaction to use.
+ * @returns Promise A promise of the result
+ */
+DataSource.prototype.execute = function(command, args = [], options = {}) {
+  assert(typeof command === 'string' || typeof command === 'object',
+    '"command" must be a string or an object.');
+  assert(typeof args === 'object',
+    '"args" must be an object, an array or undefined.');
+  assert(typeof options === 'object',
+    '"options" must be an object or undefined.');
+
+  if (!this.connector) {
+    return Promise.reject(errorNotImplemented(
+      `DataSource "${this.name}" is missing a connector to execute the command.`
+    ));
+  }
+
+  if (!this.connector.execute) {
+    return Promise.reject(new errorNotImplemented(
+      `The connector "${this.connector.name}" used by dataSource "${this.name}" ` +
+      'does not implement "execute()" API.'
+    ));
+  }
+
+  return new Promise((resolve, reject) => {
+    this.connector.execute(command, args, options, onExecuted);
+    function onExecuted(err, result) {
+      if (err) return reject(err);
+      if (arguments.length > 2) {
+        result = Array.prototype.slice.call(arguments, 1);
+      }
+      resolve(result);
+    }
+  });
+
+  function errorNotImplemented(msg) {
+    const err = new Error(msg);
+    err.code = 'NOT_IMPLEMENTED';
+    return err;
+  }
+};
+
 /*! The hidden property call is too expensive so it is not used that much
 */
 /**
--- a/test/datasource.test.js
+++ b/test/datasource.test.js
@@ -352,4 +352,100 @@ describe('DataSource', function() {
         .should.not.containEql('TestModel');
     });
   });
+
+  describe('execute', () => {
+    let ds;
+    beforeEach(() => ds = new DataSource('ds', {connector: 'memory'}));
+
+    it('calls connnector to execute the command', async () => {
+      let called = 'not called';
+      ds.connector.execute = function(command, args, options, callback) {
+        called = {command, args, options};
+        callback(null, 'a-result');
+      };
+
+      const result = await ds.execute(
+        'command',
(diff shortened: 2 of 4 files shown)


bajtos:
A related question on StackOverflow: [How to execute arbitrary SQL query](https://stackoverflow.com/q/53338050/69868)

lib/datasource.js
```diff
@@ -2594,6 +2594,74 @@ DataSource.prototype.ping = function(cb) {
...
+ * Execute an arbitrary command. The commands are connector specific,
+ * please refer to the documentation of your connector for more details.
+ *
+ * @param command String|Object The command to execute, e.g. an SQL query.
+ * @param [args] Array Parameters values to set in the command.
+ * @param [options] Object Additional options, e.g. the transaction to use.
+ * @param [callback] Function The callback to call when done.
+ * @promise
+ */
+DataSource.prototype.execute = function(command, args, options, cb) {
```
bajtos: Considering that this new API will be available for LB4 apps only (no 2.x/3.x), do we still need to keep supporting callback style?

I would prefer to provide Promise mode only, it will greatly simplify the implementation.

Of course, connector's `execute` API is staying callback-based, that's not going to change.

bajtos: Considering that this new API will be available to LB4 only (no LB3/LB2), can we make it promise-only? By removing the callback variant, the implementation can be greatly simplified.

Of course, the API contract of connector's `execute` method remains unchanged & callback-based.

bajtos:
Added a new commit to remove callback style and keep Promise style only.

test/datasource.test.js
```diff
@@ -352,4 +352,86 @@ describe('DataSource', function() {
...
+
+    it('throws NOT_IMPLEMENTED when no connector is provided', () => {
+      ds.connector.execute = undefined;
+      return ds.execute('command').should.be.rejectedWith({
+        code: 'NOT_IMPLEMENTED',
+      });
+    });
+
+    it('throws NOT_IMPLEMENTED for connectors  not implementing execute', () => {
+      ds.connector = undefined;
```
jannyHou: I think the name of this test should switch with the one above :)

- this one tests connector not provided.
- the one above tests execute not implemented.

lib/datasource.js
```diff
@@ -2594,6 +2594,54 @@ DataSource.prototype.ping = function(cb) {
...
+  if (!this.connector) {
+    return Promise.reject(errorNotImplemented(
+      `DataSource "${this.name}" is missing a connector to execute the command.`
+    ));
+  }
+
+  if (!this.connector.execute) {
+    return Promise.reject(new errorNotImplemented(
+      `The connector "${this.connector.name}" used by dataSource "${this.name}" ` +
+      'does not implemented "execute()" API.'
```
jannyHou: nitpick: does not implement

lib/datasource.js
```diff
@@ -2594,6 +2594,54 @@ DataSource.prototype.ping = function(cb) {
...
+/**
+ * Execute an arbitrary command. The commands are connector specific,
+ * please refer to the documentation of your connector for more details.
+ *
+ * @param command String|Object The command to execute, e.g. an SQL query.
+ * @param [args] Array Parameters values to set in the command.
+ * @param [options] Object Additional options, e.g. the transaction to use.
+ * @returns Promise A promise of the result
+ */
+DataSource.prototype.execute = function(command, args = [], options = {}) {
```
jannyHou: @bajtos I found the command type here a bit confusing: the function types defined at the bottom of this PR only takes `string` type as command. Any particular reason to allow an object?

bajtos: > Any particular reason to allow an object?

Some community connectors are using object value for the comman and I want to preserve support for those connectors.

See e.g. https://www.npmjs.com/package/loopback-connector-neo4j-graph

```js
var cypher = {
    query: "MATCH (u:User {email: {email}}) RETURN u",
    params: {
        email: 'alice@example.com',
    }
};
// ...
ModelClass.dataSource.connector.execute(
    cypher,
    [],
    options,
    callback
);
```

I think the connector authors misunderstood how command & args are meant to be used together. I think the following would be a better usage:

```ts
ModelClass.dataSource.connector.execute(
  `MATCH (u:User {email: {email}}) RETURN u`,
   {email: 'alice@example.com'},
   options,
   callback);
```

Based on this analysis, I am going to change the type of `args` from array to object. (Note that arrays are objects too.)

bajtos:
I have addressed the review comments and made the following extra changes:
- Enabled ES2017 (async functions) in eslint config
- Rewrote new tests using async/await. This is ok because juggler v4 requires Node.js 8+

@jannyHou LGTY now?

dougal83:
Is this supported for mongodb connector?

jannyHou:
@dougal83 I believe so, mongodb has its [`execute`](https://github.com/strongloop/loopback-connector-mongodb/blob/07a634c796e007ddf6f024ff01d8fae93f92509a/lib/mongodb.js#L467) implementation.

dougal83:
@jannyHou Do you know if there are any usage examples?

PR #1672, opened 2018-12-06
feat: add unique index support on create for in-memory connector

b-admike:
### Description
Connect to https://github.com/strongloop/loopback-next/pull/2126. Goal is to add support for unique indexes for `in-memory` connector. So far I've added logic to throw on duplicate entries for unique index properties. I'm sure there is more to add as I'm not familiar with how the connector works.

#### Related issues

### Checklist

### DIFF
--- a/lib/connectors/memory.js
+++ b/lib/connectors/memory.js
@@ -255,12 +255,14 @@ Memory.prototype._createSync = function(model, data, fn) {
     this.collection(model, {});
   }
 
-  for (var p in props) {
-    if (props[p].index && props[p].index.unique === true) {
+  var indexes = this._models[model].model.definition.indexes();
+  for (var idx in indexes) {
+    var idxPropName = idx.substring(0, idx.indexOf('_index'));
+    if (indexes[idx].unique === true) {
       for (var instId in this.cache[model]) {
         var inst = JSON.parse(this.cache[model][instId]);
-        if (inst[p] === data[p]) {
-          var duplicateIndexError = new Error(g.f('Duplicate entry for %s.%s', model, p));
+        if (inst[idxPropName] === data[idxPropName]) {
+          var duplicateIndexError = new Error(g.f('Duplicate entry for %s.%s', model, idxPropName));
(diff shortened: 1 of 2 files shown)


lib/connectors/memory.js
```diff
@@ -255,6 +255,19 @@ Memory.prototype._createSync = function(model, data, fn) {
...
+        var inst = JSON.parse(this.cache[model][instId]);
+        if (inst[p] === data[p]) {
+          var duplicateIndexError = new Error(g.f('Duplicate entry for %s.%s', model, p));
+          duplicateIndexError.statusCode = duplicateIndexError.status = 409;
+          return fn(duplicateIndexError);
+        }
+      }
+    }
+  }
+
```
raymondfeng: Please use https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-definition.js#L194

b-admike: Good catch!

raymondfeng:
What about `update` operations? There is a chance to violate unique constraints.

b-admike:
> What about `update` operations? There is a chance to violate unique constraints.

Good catch. Looks like we don't have that for case of updating PKs as well when there is an existing PK with the same value :-(. I've written the test cases, and will add the logic to handle that use case 👍

lib/connectors/memory.js
```diff
@@ -255,6 +255,21 @@ Memory.prototype._createSync = function(model, data, fn) {
     this.collection(model, {});
   }
 
+  var indexes = this._models[model].model.definition.indexes();
+  for (var idx in indexes) {
+    var idxPropName = idx.substring(0, idx.indexOf('_index'));
+    if (indexes[idx].unique === true) {
+      for (var instId in this.cache[model]) {
+        var inst = JSON.parse(this.cache[model][instId]);
```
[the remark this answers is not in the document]
b-admike: The reason I have to use JSON parse is because we stringify the model instances to store in the cache :(. However, I like that approach and I'm trying it out. I think it will work well with multiple-key unique indexes too.

PR #1673, opened 2018-12-07
Upgrade dependencies to their latest versions

bajtos:
- Upgrade bson to 4.x (latest)
- Update eslint-config-loopback to latest
- Update remaining dependencies

Skip update of shouldjs to avoid breaking tests in connectors.

### Checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -624,6 +624,12 @@ function createPromiseCallback() {
   return cb;
 }
 
+function isBsonType(value) {
+  // bson@1.x stores _bsontype on ObjectID instance, bson@4.x on prototype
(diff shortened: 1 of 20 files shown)


lib/utils.js
```diff
@@ -636,7 +642,7 @@ function uniq(a) {
   }
   assert(Array.isArray(a), 'array argument is required');
   var comparableA = a.map(
-    item => item.hasOwnProperty('_bsontype') ? item.toString() : item
+    item => isBsonType(item) ? item.toString() : item
```
bajtos: This change is important - it's fixing compatibility with new version of `bson`.

PR #1685, opened 2019-01-24
Improve Model type definitions

bajtos:
- Define two new types: ModelSettings and ModelProperties to describe the objects describing model properties and settings.
- Add static property `base` and static method `extend` to the definition
of `ModelBase` class.
- Add property `defaultModelBaseClass` to `ModelBuilder` class definition.
- Fix `PropertyDefinition` interface: remove `name`, add optional `id`.

These issues were discovered while working on https://github.com/strongloop/loopback-next/pull/2274.

### Checklist

### DIFF
--- a/types/model.d.ts
+++ b/types/model.d.ts
@@ -19,8 +19,8 @@ export type PropertyType =
  * Property definition
  */
 export interface PropertyDefinition extends AnyObject {
-  name: string;
   type?: PropertyType;
+  id?: boolean | number;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


types/model.d.ts
```diff
@@ -19,17 +19,17 @@ export type PropertyType =
  * Property definition
  */
 export interface PropertyDefinition extends AnyObject {
-  name: string;
```
bajtos: AFAICT, the property name is always stored as the key in property map (name => definition).

types/model.d.ts
```diff
@@ -19,17 +19,17 @@ export type PropertyType =
...
   type?: PropertyType;
+  id?: boolean;
```
raymondfeng: In LB3, `id` is a `boolean` or `number`. See https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-definition.js#L132

bajtos: Good catch! Fixed.

bajtos:
@raymondfeng LGTY now?

PR #1690, opened 2019-02-12
Fix callback for find queries with near filters

b-admike:
### Description

When using [near](https://loopback.io/doc/en/lb2/Where-filter.html#near) operator in a filter for find operations, The code path taken for memory connector is different than other connectors which support this feature. 

For memory connector, the call stack is as follows: DAO.find() -> invokeConnectorMethod() with the callback provided as `allCb` -> (after connector's all/find method returns) allCb() does some processing (converting model PDO (plain data object) to model instance) and calls utils.createPromiseCallback(). 

For MySQL/MongoDB, the call stack is: DAO.find() -> invokeConnectorMethod() with `geoCallBack` being the cb function -> eventually geoCallBackWithoutNotify() is called which creates a model instance with a memory connector and calls `memory.all()` but this time supplies `utils.createPromiseCallback()` as the callback and thus does not get to change the model PDO to model instance when it returns. 

In this PR, the first commit moves the functions `geoCallBackWithoutNotify` and `geoCallBackWithNotify` outside `geoQuery` so `allCb` can be called from within `geoCallBackWithoutNotify` (refactor). The second commit is changing the callback provided to `Memory.all()` call within `geoCallBackWithoutNotify` from `utils.createPromiseCallBack()` (plain cb) to `allCb`. I have a test case for this (https://github.com/strongloop/loopback-next/blob/bcaa0e4935ab5268f2dd2c0240ba5478dd2200f2/packages/repository/test/unit/repositories/legacy-juggler-bridge.unit.ts#L186-L206) which I would like to move to a common test suite in juggler to verify the fix. ~Thoughts on where it should go?~

EDIT: In https://github.com/strongloop/loopback-datasource-juggler/pull/1690/commits/47bd60b50487dd1de2c8194564b7b8c8f59389d0 I've added assertions in our common test suite for geo queries to verify the fix.

Fixes https://github.com/strongloop/loopback-next/issues/1419
Related to https://github.com/strongloop/loopback-next/pull/2299/files

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1531,6 +1531,7 @@ DataAccessObject.find = function find(query, options, cb) {
 
   const near = query && geo.nearFilter(query.where);
   const supportsGeo = !!connector.buildNearFilter;
+  let geoQueryObject;
 
   if (near) {
     if (supportsGeo) {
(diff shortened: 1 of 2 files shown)


test/basic-querying.test.js
```diff
@@ -654,6 +654,7 @@ describe('basic-querying', function() {
             if (err) done(err);
             users.should.have.property('length', 3);
             users[0].name.should.equal('John Lennon');
+            should(users[0].toObject()).be.a.function;
```
bajtos: This is weird. `users[0].toObject()` calls `toObject` function and then you are asserting that the retval of that function is a function again? Also I think that `be.a.function` is a no-op, you need to call the assertion - `be.a.function()`.

Since your intention is to verify that returned items are instances of the model class (and not just plain-data objects), maybe you can use `instanceof` check against a base model class?

```js
users[0].should.be.instanceOf(User);
```

(The `User` model is defined at the top of the test file and is accessible to all tests. In this particular test, you start by calling `User.find`, a method on that model.)

bajtos:
I think you have nailed the root cause of geo queries returning plain data objects instead of model instances. Well done! 👏 

When I disable whitespace changes, the patch looks very innocent. I appreciate that you made the changes in such way that almost all of the code remains in the same place and thus it's easy to spot what is actually changing 👍 

Let's improve the tests now, see my comment above.

test/basic-querying.test.js
b-admike: Oops I meant to do `should(users[0].toObject).be.a.function()` 😅. Thank you for pointing it out and will use the `instanceOf` assertion which is exactly what I am looking for!

b-admike:
Thanks @bajtos, I've applied your feedback. LGTY now?

lib/dao.js
```diff
@@ -1558,65 +1559,64 @@ DataAccessObject.find = function find(query, options, cb) {
...
+      data.forEach(function(obj) {
+        memory.create(modelName, obj, options, function() {
+          // noop
+        });
+      });
 
-        function geoCallbackWithNotify(err, data) {
-          if (err) return cb(err);
+      // FIXME: apply "includes" and other transforms - see allCb below
+      memory.all(modelName, geoQueryObject, options, allCb);
```
jannyHou: Great catch 👍

lib/dao.js
```diff
@@ -1558,65 +1559,64 @@ DataAccessObject.find = function find(query, options, cb) {
           queryGeo(ctx.query);
         });
       }
+      // already handled
+      return cb.promise;
```
jannyHou: Sorry I am a little bit confused, my understanding is the new code returns `cb.promise` outside of the `if...else` statement, so what is the difference between returning it outside and inside?

test/basic-querying.test.js
bajtos: > It's weird when I was debugging it, `User` was undefined even though like you said it's defined globally at the top.

I think `User` was undefined in the debugger because your callback was not accessing it and thus the variable was not captured in callback function's scope.

lib/dao.js
b-admike: Actually the `cb.promise` call is in the same spot as before (inside the `if(supportsGeo)...else if (query.where)` block). It's hard to see it from the changes (Disabling white-space changes in the Diff settings helps)

jannyHou: Thank you I see, much better after disabling white-space!

b-admike:
Thank you all for the reviews. The coverage decrease shows lines which are covered but only when the tests are run downstream with connectors other than in-memory (which is not the case for the tests run in CI for this PR). Going to land as is.

PR #1693, opened 2019-02-20
feat: Support "type" key in sub-properties

hacksparrow:
### Description

Currently, a model definition like this:

```
  "properties": {
    "items": {
      "type" : {
        "itemname": { "type": "string"},
        "type": { "type": "string" }
      }
    }
  }
```
results in invalid data.

`{ items: [object Object] }`

It is caused by the sub-property `type` conflicting with the type resolution object.

This PR, allows sub-property `type` to be used without any problem.

#### Related issues

https://github.com/strongloop/loopback/issues/4131

### Checklist

### DIFF
--- a/test/datatype.test.js
+++ b/test/datatype.test.js
@@ -35,6 +35,24 @@ describe('datatypes', function() {
     db.automigrate(['Model'], done);
   });
 
+  it('should resolve top-level "type" property correctly', function() {
+    const Account = db.define('Account', {
+      type: String,
+      id: String,
+    });
+    Account.definition.properties.type.type.should.equal(String);
+  });
+
+  it('should resolve "type" sub-property correctly', function() {
+    const Account = db.define('Account', {
+      item: {type: {
+        itemname: {type: String},
+        type: {type: String},
+      }},
+    });
+    Account.definition.properties.item.type.should.not.equal(String);
+  });
+
   it('should return 400 when property of type array is set to string value',
     function(done) {
       const myModel = db.define('myModel', {
--- a/lib/model-builder.js
+++ b/lib/model-builder.js
@@ -856,47 +856,46 @@ ModelBuilder.prototype.getSchemaName = function(name) {
 /**
  * Resolve the type string to be a function, for example, 'String' to String.
  * Returns {Function} if the type is resolved
- * @param {String} type The type string, such as 'number', 'Number', 'boolean',
- * or 'String'. This parameter is case insensitive.
+ * @param {String | Object | Array} prop The object whose type is to be resolved
  */
-ModelBuilder.prototype.resolveType = function(type) {
-  if (!type) {
-    return type;
+ModelBuilder.prototype.resolveType = function(prop, isSubProperty) {
+  if (!prop) {
+    return prop;
   }
-  if (Array.isArray(type) && type.length > 0) {
+  if (Array.isArray(prop) && prop.length > 0) {
     // For array types, the first item should be the type string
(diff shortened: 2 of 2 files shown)


hacksparrow:
I also made some cosmetic changes to the codebase for clarity.

test/datatype.test.js
```diff
@@ -35,6 +35,16 @@ describe('datatypes', function() {
...
   });
 
+  it('should resolve "type" sub-property', function() {
+    const Account = db.define('Account', {
+      item: {type: {
+        itemname: {type: String},
+        type: {type: String},
+      }},
+    });
+    Account.definition.properties.item.type.should.not.equal(String);
```
raymondfeng: Can we add a test case to cover the following?
```js
const Account = db.define('Account', {
      type: String, // account type - checking/saving
      id: String
    });
```

hacksparrow: Sure, thanks.

hacksparrow:
Valid cases. They will likely be covered while fixing https://github.com/strongloop/loopback/issues/4132.

For this one, let's just focus on the issue that was reported in https://github.com/strongloop/loopback/issues/4131.

arukris-cdw:
@hacksparrow , I am having the same issue. 

when my data has property name as "type", it is considered as key and not writing properly to mongodb. Can you please let me know the solution to this problem. I went through all reference links shared on the fix, but couldn't understand the details. Appreciate if you could share the steps. 

**Issue:**
"orderInstallNotes" : [
            "[object Object]"
        ],

I am using below version of datasource juggler, not sure if this version has the fix for this issue. Please let me know. 

C:\ProgramData\IBM\MQSI\node_modules>npm ls loopback-datasource-juggler
C:\ProgramData\IBM\MQSI
`-- loopback@3.25.0
  +-- loopback-connector-remote@3.4.1
  | `-- loopback-datasource-juggler@3.31.1  deduped
  `-- loopback-datasource-juggler@3.31.1

thanks
Arun S.

hacksparrow:
@arukris-cdw can you reproduce your issue in a very basic app and share?

arukris-cdw:
[Observation.docx](https://github.com/strongloop/loopback-datasource-juggler/files/3503463/Observation.docx)
[mongocloud.zip](https://github.com/strongloop/loopback-datasource-juggler/files/3503465/mongocloud.zip)

@hacksparrow uploaded the application as a zip file and also documented my observation. Please review and let me know. 

 Hope you remember that I use IBM ACE to interact with loopback. I am not able to reproduce the same exact error (having [object object] in the output) by standalone loopback application. But i have recorded my observation and there seems some issue in loopback itself while property name itself is "type"

thanks
Arun s.

hacksparrow:
@arukris-cdw if the issue cannot be reproduced with a standalone LoopBack app, then it is an ACE issue. I will be talking to the ACE team regarding another problem, hopefully, we'll get some insight into what is going on.

arukris-cdw:
@hacksparrow , did you get a chance to look into my Observation document.. There is an issue with standalone loopback app. The datatypes of other elements are not retained if there is an element with the name "type" in it.. 

Request you to verify that once as well. it would help your further discussion with IBM ACE team. 

My experience in the past is, ACE team wants us to confirm that there are no issues with loopback before they get to work on it. 

thanks
Arun S.

hacksparrow:
@arukris-cdw can you include instructions to reproduce your findings? Which model, what request body, which request method?

hacksparrow:
Also, please use the latest version of LoopBack 3 `3.26.0`.

PR #1694, opened 2019-02-23
feat: call after operation hook in case of errors

spurreiter:
### Description

The connector hook "after operation"  is not called in case of an error caused by the connector.
This for example does not allow to change the HTTP status code for specific errors. Instead HTTP 500 Internal Server Error is returned. 
This PR calls the "after operation" even in such situations. The original error is passed with `ctx.res.err` to allow inspection and manipulation as described in https://loopback.io/doc/en/lb3/Connector-hooks.html#error-handling.

### Checklist

### DIFF
--- a/lib/observer.js
+++ b/lib/observer.js
@@ -7,6 +7,7 @@
 
 const async = require('async');
 const utils = require('./utils');
+const debug = require('debug')('loopback:observer');
 
 module.exports = ObserverMixin;
 
@@ -232,7 +233,24 @@ ObserverMixin.notifyObserversAround = function(operation, context, fn, callback)
 
       function cbForWork(err) {
         const args = [].slice.call(arguments, 0);
-        if (err) return callback.apply(null, args);
+        if (err) {
+          // call observer in case of error to hook response
+          context.error = err;
+          self.notifyObserversOf('after ' + operation + ' error', context,
+            function(_err, context) {
+              if (_err && err) {
+                debug(
+                  'Operation %j failed and "after %s error" hook returned an error too. ' +
+                    'Calling back with the hook error only.' +
+                    '\nOriginal error: %s\nHook error: %s\n',
+                  err.stack || err,
+                  _err.stack || _err
+                );
+              }
+              callback.call(null, _err || err, context);
+            });
+          return;
+        }
         // Find the list of params from the callback in addition to err
         const returnedArgs = args.slice(1);
         // Set up the array of results
--- a/test/async-observer.test.js
+++ b/test/async-observer.test.js
@@ -325,6 +325,86 @@ describe('async observer', function() {
       }
     );
   });
+
+  it('should call after operation hook on error', function(done) {
+    const context = {
+      req: {},
+    };
+    const operationError = new Error('The operation failed without result');
+    let callCount = 0;
+
+    function fail(context, done) {
+      process.nextTick(() => {
+        done(operationError);
+      });
+    }
+
+    TestModel.observe('after execute error', function(ctx, next) {
+      callCount++;
+      next();
+    });
+
+    TestModel.notifyObserversAround('execute', context, fail, (err, ctx) => {
+      callCount.should.eql(1);
+      err.message.should.eql(operationError.message);
+      ctx.error.message.should.eql(operationError.message);
+      done();
+    });
+  });
+
+  it('should call after operation hook on error while overwriting error', function(done) {
+    const context = {
+      req: {},
+    };
+    const operationError = new Error('The operation failed without result');
+    const overwriteError = new Error('Overwriting the original error');
+    let callCount = 0;
+
+    function fail(context, done) {
+      process.nextTick(() => {
+        done(operationError);
+      });
+    }
+
+    TestModel.observe('after execute error', function(ctx, next) {
+      callCount++;
+      next(overwriteError);
+    });
+
+    TestModel.notifyObserversAround('execute', context, fail, (err, ctx) => {
+      callCount.should.eql(1);
+      err.message.should.eql(overwriteError.message);
+      ctx.error.message.should.eql(operationError.message);
+      done();
+    });
+  });
+
+  it('should call after operation hook on error while allowing to change err', function(done) {
+    const context = {
+      req: {},
+    };
+    const operationError = new Error('The operation failed without result');
+    let callCount = 0;
+
+    function fail(context, done) {
+      process.nextTick(() => {
+        done(operationError);
+      });
+    }
+
+    TestModel.observe('after execute error', function(ctx, next) {
+      callCount++;
+      const err = ctx.error;
+      next(err, ctx);
+    });
+
+    TestModel.notifyObserversAround('execute', context, fail, (err, ctx) => {
+      callCount.should.eql(1);
+      err.message.should.eql(operationError.message);
+      ctx.error.message.should.eql(operationError.message);
+      done();
+    });
+  });
 });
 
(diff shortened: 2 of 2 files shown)


slnode:

jannyHou:
@slnode test please

lib/observer.js
```diff
@@ -231,7 +231,16 @@ ObserverMixin.notifyObserversAround = function(operation, context, fn, callback)
 
       function cbForWork(err) {
         const args = [].slice.call(arguments, 0);
-        if (err) return callback.apply(null, args);
+        if (err) {
+          // call observer in case of error to hook response
+          err.status = 500;
+          context.res = {err};
```
hacksparrow: We should probably omit out `err.status = 500` and let the handler decide what to do with the error.

And the error object should be contained in `context.error` instead of `context.res`.

@bajtos @raymondfeng, thoughts?

bajtos: Yes please, it should be `context.error`.

bajtos:
Thank you @spurreiter for the pull request. 

Just to built a shard context: the hooks `before execute` and `after execute` are supposed to be implemented by individual connectors. For example, the base SQL connector does the following:

https://github.com/strongloop/loopback-connector/blob/5ab9c9c76e32c3acb29c99ed0e55e839852a39d9/lib/sql.js#L632-L645
```js
  this.notifyObserversAround('execute', context, function(context, done) {
    self.executeSQL(context.req.sql, context.req.params, context.options,
      function(err, info) {
        if (err) {
          debug('Error: %j %j %j', err, context.req.sql, context.req.params);
        }


        if (!err && info != null) {
          context.res = info;
        }
        // Don't pass more than one args as it will confuse async.waterfall
        done(err, info);
      });
  }, callback);
```

Now if we look at the current implementation of `notifyObserversAround`, it's clear that it does not call any observer when the operation failed with an error.

I have few concerns regarding your implementation:
- By calling the same observer for both successful and unsuccessful results, we make it more difficult for app developers to subscribe to one type only (e.g. observe only errors).
- With your change in place, we start calling `after execute` for failed requests. This may break existing applications that are built on the assumption that `after execute` is invoked only on success.
- All other LoopBack hooks use a different name for success and failure, e.g. `beforeRemote`, `afterRemote` and `afterRemoteError`. Your proposal is inconsistent with that.

I am proposing to use the following event name: `'after ' + operation + ' error`, e.g. `after execute error`.

Also please note that LoopBack 3.x is in LTS mode and no new features are allowed (see our [Long Term Support policy](https://loopback.io/doc/en/contrib/Long-term-support.html)). The new hook will not be back-ported to `loopback-datasource-juggler@3`, it will become available to LB4 users only.

spurreiter:
The update contains the following changes
- error moved into context
- err.status removed
- next(err) may overwrite original error - original error remains in context

lib/observer.js
```diff
@@ -231,7 +231,15 @@ ObserverMixin.notifyObserversAround = function(operation, context, fn, callback)
 
       function cbForWork(err) {
         const args = [].slice.call(arguments, 0);
-        if (err) return callback.apply(null, args);
+        if (err) {
+          // call observer in case of error to hook response
+          context.error = err;
+          self.notifyObserversOf('after ' + operation + ' error', context,
+            function(_err, context) {
+              callback.call(null, _err || err, context);
```
bajtos: I am not sure if it's a good idea to give priority to the error reported by the hook over the actual error in the database operation. @raymondfeng @strongloop/loopback-maintainers thoughts?

Either way, we can report only one error and the other error will be discarded. To make troubleshooting easier, it's important to print the discarded error to the debug log.

E.g.

```js
if (_err && err) {
  debug(
    'Operation %j failed and "after %s error" hook returned an error too. ' +
      'Calling back with the hook error only.' +
      '\nOriginal error: %s\nHook error: %s\n',
    err.stack || err,
    _err.stack || _err);
}
```

test/async-observer.test.js
```diff
@@ -324,6 +324,86 @@ describe('async observer', function() {
       }
     );
   });
+
+  it('should call after operation hook on error', function(done) {
+    const context = {
+      req: {},
+    };
+    const operationError = new Error('The operation failed without result');
+    let passesBy = 0;
```
bajtos: Let's use a more accurate name, e.g.

```suggestion
    let callCount = 0;
```

The same comment applies to other newly added tests below too.

test/async-observer.test.js
```diff
@@ -324,6 +324,86 @@ describe('async observer', function() {
...
+
+    function fn(context, done) {
```
bajtos: Let's use a more descriptive name please, e.g.

```suggestion
    function failingOperation(context, done) {
```

or simply:

```suggestion
    function fail(context, done) {
```

The same comment applies to other newly added tests below too.

test/async-observer.test.js
```diff
@@ -324,6 +324,86 @@ describe('async observer', function() {
...
+
+    TestModel.notifyObserversAround('execute', context, fn, (err, ctx) => {
+      passesBy.should.eql(1);
+      err.message.should.eql(operationError.message);
+      ctx.error.message.should.eql(operationError.message);
+      done();
+    });
+  });
+
+  it('should call after operation hook on error while overwriting error', function(done) {
```
bajtos: +1 for adding a test to verify what happens when both the operation and the hook report an error 👍

test/async-observer.test.js
```diff
@@ -324,6 +324,86 @@ describe('async observer', function() {
...
+    function fn(context, done) {
+      process.nextTick(() => {
+        done(operationError);
+      });
+    }
+
+    TestModel.observe('after execute error', function(ctx, next) {
+      passesBy++;
+      const err = ctx.error;
+      ctx.end(err, ctx);
```
bajtos: I don't remember what is `ctx.end` doing, could you please help me to understand what's the expected outcome of this call?

Also AFAICT, you are not changing the error, you keep using the same `operationError` instance. What am I missing?

spurreiter: From https://loopback.io/doc/en/lb3/Connector-hooks.html#before-execute 
> To terminate the invocation, call ctx.end(err, result)
Imho it should early terminate the request with a custom response.

bajtos: I see, thank you for the link. The documentation is for "before execute" hook where it's possible to skip execution of the actual command. In "after execute" and "after execute error" hooks, the command was already executed, thus I don't see what would be the point of calling `ctx.end`?

Anyhow, I don't mind to test this scenario, but please rework the test to make it more clear what is the use case being tested.

spurreiter:
Hi, any update on this?

bajtos:
@spurreiter Most of us don't receive notification about new commits pushed to pull requests, it's important to post a comment to let us know when there are new changes to be reviewed.

AFAICT, you haven't addressed https://github.com/strongloop/loopback-datasource-juggler/pull/1694/files#r270470236 yet, PTAL again.

test/async-observer.test.js
bajtos: IMO, you should call `next` instead of `ctx.end` to signal the error.

```suggestion
      next(err);
```

spurreiter:
Requested change to "call next instead of ctx.end" in test is made.
Please review.

spurreiter:
Commit is squashed. Thanks.

dhmlau:
@slnode  test please

bajtos:
Landed 🎉  Thank you for the contribution! ❤️

PR #1699, opened 2019-03-07
feat: support array within array

hacksparrow:
### Description

POST request body:
```
 { items": [
    [
      {"name": "APPLE"},
      {"serial": 67}
    ]
  ]
}
```

Currently stored as: 
```
 { "items": [
    {
      "0": {
        "name": "APPLE"
      },
      "1": {
        "serial": 67
      }
    }
  ]
}
```

Will be stored as:

```
 { items": [
    [
      {
        "name": "APPLE"
      },
      {
        "serial": 67
      }
    ]
  ]}
```
#### Related issues

https://github.com/strongloop/loopback/issues/4132

### Checklist

### DIFF
--- a/lib/list.js
+++ b/lib/list.js
@@ -64,7 +64,12 @@ function List(items, itemType, parent) {
   }
 
   List.prototype.toItem = function(item) {
-    return isClass(this.itemType) ? new this.itemType(item) : this.itemType(item);
+    if (isClass(this.itemType)) {
+      return new this.itemType(item);
+    } else {
+      if (Array.isArray(item)) return item;
+      else return this.itemType(item);
+    }
   };
 
   items.forEach(function(item, i) {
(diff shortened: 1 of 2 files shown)


hacksparrow:
@raymondfeng @bajtos please review.

hacksparrow:
Agreed. I am aware that there can be other conditions when the resulting object will be unexpected. Let's just focus on the reported bug at the moment.

lib/list.js
```diff
@@ -64,7 +64,12 @@ function List(items, itemType, parent) {
   }
 
   List.prototype.toItem = function(item) {
-    return isClass(this.itemType) ? new this.itemType(item) : this.itemType(item);
+    if (isClass(this.itemType)) {
+      return new this.itemType(item);
+    } else {
+      if (Array.isArray(item)) return item;
```
raymondfeng: Do we want to test if `item instanceof List` to convert `List`, which is a special array?

hacksparrow: In that case it is breaking some test cases. The data structure which elicited the bug is a very unlikely one. This PR takes care of that scenario without side effects.

I think we should land the PR and move on.

raymondfeng: ok

PR #1702, opened 2019-03-21
fix: nested property coercion and resolution

b-admike:
### Description
Ensure that nested array properties maintain connector-specific data type metadata and also coerce string values for array of primitive properties into their type. Connect to https://github.com/strongloop/loopback-connector-mongodb/issues/493.

#### Related issues

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -2301,6 +2301,7 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
         if (!(data instanceof Model)) {
           try {
             inst = new Model(data, {applyDefaultValues: false});
+            ctx.data = inst.toObject(true);
           } catch (err) {
             return cb(err);
           }
--- a/lib/list.js
+++ b/lib/list.js
@@ -67,8 +67,14 @@ function List(items, itemType, parent) {
     if (isClass(this.itemType)) {
       return new this.itemType(item);
     } else {
-      if (Array.isArray(item)) return item;
-      else return this.itemType(item);
+      if (Array.isArray(item)) {
+        return item;
+      } else if (this.itemType === Date) {
+        if (item === null) return null;
+        return new Date(item);
+      } else {
+        return this.itemType(item);
+      }
     }
   };
 
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -1055,6 +1055,40 @@ describe('basic-querying', function() {
     });
   });
 
+  describe('updateAll', function() {
+    it('coerces primitive datatypes on update', async () => {
+      const numAndDateModel = db.define('numAndDateModel', {
+        dateProp: Date,
+        dateArray: [Date],
+        numProp: Number,
+        numArray: [Number],
+      });
+      const createDate = new Date('2019-02-21T12:00:00').toISOString();
+      const createData = {
+        dateProp: createDate,
+        dateArray: [createDate, createDate],
+        numProp: '1',
+        numArray: ['1', '2'],
+      };
+      const updateDate = new Date('2019-04-15T12:00:00').toISOString();
+      const updateData = {
+        dateProp: updateDate,
+        dateArray: [updateDate, updateDate],
+        numProp: '3',
+        numArray: ['3', '4'],
+      };
+      const created = await numAndDateModel.create(createData);
+      const updated = await numAndDateModel.updateAll({id: created.id}, updateData);
+      const found = await numAndDateModel.findById(created.id);
+      found.dateProp.should.deepEqual(new Date(updateDate));
+      found.dateArray[0].should.deepEqual(new Date(updateDate));
+      found.dateArray[1].should.deepEqual(new Date(updateDate));
+      found.numProp.should.equal(3);
+      found.numArray[0].should.equal(3);
+      found.numArray[1].should.equal(4);
+    });
+  });
+
   context('regexp operator', function() {
     const invalidDataTypes = [0, true, {}, [], Function, null];
 
--- a/test/datatype.test.js
+++ b/test/datatype.test.js
@@ -52,6 +52,54 @@ describe('datatypes', function() {
     });
     Account.definition.properties.item.type.should.not.equal(String);
   });
+  it('should resolve array prop with connector specific metadata', function() {
+    const model = db.define('test', {
+      randomReview: {
+        type: [String],
+        mongodb: {
+          dataType: 'Decimal128',
+        },
+      },
+    });
+    model.definition.properties.randomReview.type.should.deepEqual(Array(String));
+    model.definition.properties.randomReview.mongodb.should.deepEqual({dataType: 'Decimal128'});
+  });
+
+  it('should coerce array of dates from string', async () => {
+    const dateArrayModel = db.define('dateArrayModel', {
+      bunchOfDates: [Date],
+      bunchOfOtherDates: {
+        type: [Date],
+      },
+    });
+    const dateVal = new Date('2019-02-21T12:00:00').toISOString();
+    const created = await dateArrayModel.create({
+      bunchOfDates: [dateVal,
+        dateVal,
+        dateVal],
+      bunchOfOtherDates: [dateVal,
+        dateVal,
+        dateVal],
+    });
+    created.bunchOfDates[0].should.be.an.instanceOf(Date);
+    created.bunchOfDates[0].should.deepEqual(new Date(dateVal));
+    created.bunchOfOtherDates[0].should.be.an.instanceOf(Date);
+    created.bunchOfOtherDates[0].should.deepEqual(new Date(dateVal));
+  });
+
+  it('should coerce array of numbers from string', async () => {
+    const numArrayModel = db.define('numArrayModel', {
+      bunchOfNums: [Number],
+    });
+    const dateVal = new Date('2019-02-21T12:00:00').toISOString();
+    const created = await numArrayModel.create({
+      bunchOfNums: ['1',
+        '2',
+        '3'],
+    });
+    created.bunchOfNums[0].should.be.an.instanceOf(Number);
+    created.bunchOfNums[0].should.equal(1);
+  });
 
   it('should return 400 when property of type array is set to string value',
     function(done) {
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -2366,7 +2366,7 @@ describe('manipulation', function() {
     it('should not coerce invalid values provided in where conditions', function(done) {
       Person.update({name: 'Brett Boe'}, {dob: 'notadate'}, function(err) {
         should.exist(err);
-        err.message.should.equal('Invalid date: notadate');
+        err.message.should.equal('Invalid date: Invalid Date');
         done();
       });
     });


lib/model-definition.js
```diff
@@ -238,9 +238,17 @@ ModelDefinition.prototype.build = function(forceRebuild) {
       };
       if (typeof prop === 'object' && prop !== null) {
         for (const a in prop) {
+          const isNestedArrayProp = Array.isArray(prop[a]);
           // Skip the type property but don't delete it Model.extend() shares same instances of the properties from the base class
-          if (a !== 'type') {
+          if (a !== 'type' && !isNestedArrayProp) {
```
hacksparrow: I have landed a PR handling a condition like this - https://github.com/strongloop/loopback-datasource-juggler/pull/1699.

It makes me wonder if some architectural optimization can be done around this structure, instead of checking if a structure is an array within an array (including my implementation).

@bajtos @raymondfeng

lib/model-definition.js
```diff
@@ -238,9 +238,17 @@ ModelDefinition.prototype.build = function(forceRebuild) {
...
             typeDef[a] = prop[a];
+          } else if (isNestedArrayProp) {
+            const propDef = prop[a][0];
+            for (const a in propDef) {
+              if (a !== 'type') {
+                typeDef[a] = propDef[a];
```
bajtos: What if `propDef[a]` is another array? I think we need to recursively apply the same logic you are applying at line 243 above.

Please add a unit test to cover this scenario.

lib/model.js
```diff
@@ -333,6 +333,13 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
     if (propVal === undefined && persistUndefinedAsNull) {
       self.__data[p] = propVal = null;
     }
+    // coerce array values of primitives from string
+    if (Array.isArray(type) && type[0].name) {
+      const actualType = type[0];
+      if (BASE_TYPES[actualType.name] && Array.isArray(self.__data[p])) {
+        self.__data[p].forEach((ele, idx) => self.__data[p][idx] = new actualType(ele));
```
bajtos: `ele` sounds weird to me, can we use `elem` instead? Or even `it` (standing for _array item_).

Also I think this can be more elegantly expressed using `map` function?

```ts
self._data[p] = self._data[p].map(item => new actualType(item));
```

lib/model.js
```diff
@@ -333,6 +333,13 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
     if (propVal === undefined && persistUndefinedAsNull) {
       self.__data[p] = propVal = null;
     }
+    // coerce array values of primitives from string
+    if (Array.isArray(type) && type[0].name) {
```
bajtos: What if a non-primitive type is nested deeper? I am thinking about something like this:

```js
type = [
  // an array of anonymous objects with the following structure:
  {
    // name is a string
    name: String,
    // roles hold an array of Role instances
    roles: [Role]
  }
]
```

Do we need to handle this case?

If anonymous objects are converted into named models, then how about array of arrays?

```js
// Cell is a model, let's say it has one property "color"

// 2D array of Cell instances
type = [[Cell]]; 

// Example value:
propValue = [
 [{color: 'black'}, {color: 'white'}],
 [{color: 'white'}, {color: 'black'}],
];
```

lib/model-definition.js
bajtos: +1 to find a way how to address this problem in a wider context.

I am proposing the following:
- Land this PR first, to fix the immediate problem
- Create a follow-up spike story to investigate better solutions. Time-box it to 2-3 days only. I don't want to spend too much time on improving this legacy codebase, we are likely to throw most of it away as part of juggler refactor in the future.

b-admike: Actually, I've talked to @raymondfeng and have concluded to revert this change. Right now the following two flavours of an array property work:
```javascript
   const model = db.define('test', {
      randomReview: {
        type: [
          {
            type: 'string',
          },
        ],
        mongodb: {
          dataType: 'Decimal128',
        },
      },
    });
```
or
```javascript
   const model = db.define('test', {
      randomReview: {
        type: [String ],
        mongodb: {
          dataType: 'Decimal128',
        },
      },
    });
```

lib/model-definition.js
b-admike: Reverting these changes as mentioned above. Let's discuss.

lib/model.js
b-admike: I would like to tackle this in a follow-up PR. Will create an issue for it.

bajtos: @b-admike have you created that issue yet? Please post a link to it here.

lib/model.js
bajtos: > I think this can be more elegantly expressed using `map` function?
>
> ```ts
> self._data[p] = self._data[p].map(item => new actualType(item));
> ```

☝️

bajtos:
> have concluded to revert this change. Right now the following two flavours of an array property work

Sounds good to me 👍

test/datatype.test.js
```diff
@@ -52,6 +52,36 @@ describe('datatypes', function() {
...
+    });
+    const dateVal = new Date('2019-02-21T12:00:00').toISOString();
+    const created = await dateArrayModel.create({
+      bunchOfDates: [dateVal,
+        dateVal,
+        dateVal],
+    });
+    created.bunchOfDates[0].should.be.an.instanceOf(Date);
+    created.bunchOfDates[0].should.equal(Date(dateVal));
+  });
```
bajtos: I think it would be good to test coercion for array types defined using the approach shown in the first test too.

```ts
const dateArrayModel = db.define('dateArrayModel', {
  bunchOfDates: {
    type: [Date],
  },
});

// ...
created.bunchOfDates[0].should.be.an.instanceOf(Date);
created.bunchOfDates[0].should.equal(Date(dateVal));
```     

Unless that's something we have already covered by other tests?

Similarly, test how connector-specific properties are preserved for the second flavor.

```ts
const model = db.define('model', {
   bunchOfDates: [
     {
       type: Date,
        // is this the right syntax?
        mongodb: {
           dataType: 'Decimal128',
        },  
     },
   ],
 });
```

test/datatype.test.js
```diff
@@ -52,6 +52,36 @@ describe('datatypes', function() {
...
+
+  it('should coerce array of primitive types from string', async () => {
+    const dateArrayModel = db.define('dateArrayModel', {
+      bunchOfDates: [
+        {
+          type: Date,
+        },
+      ],
+    });
```
bajtos: On the second thought, I am not sure if this is a valid syntax. I remember we were recently fixing a problem where properties called `type` were treated incorrectly.

What are the rules for deciding whether this model contains an array of Date instances (`[dateVal, dateVal]`), vs. an array of anonymous objects that have a `type` property containing a date instance (`[{type: dateVal}, {type: dateVal}]`)?

lib/model.js
b-admike: Yes I've created it now :) https://github.com/strongloop/loopback-datasource-juggler/issues/1716.

test/datatype.test.js
b-admike: Okay I can try this out as well, I'll check if there is existing coverage.

test/datatype.test.js
b-admike: Good question. I think the fix we made recently allows users to use `type` name for properties IIRC. In my head, it'd be something like:
```js
prop: {
 type: [Date]
}
```
versus
```js
prop: {
 type: {
   type: [Date]
 }
}
```

test/datatype.test.js
b-admike: >I think it would be good to test coercion for array types defined using the approach shown in the first test too.
Agreed, will add the test case in my next commit.
> Similarly, test how connector-specific properties are preserved for the second flavor.
Should we allow this? I don't think we handle this case and this was what I tried to fix with my first commit, but I feel like it can be addressed properly in a wider context and we should advocate our users use first flavour (`prop: {type:[String], mongodb: ...}`. FYI this is how we interpret that definition:
def:
```
model.definition.properties.customMongoProp2
Object {0: Object, type: Array(1)}
should:Assertion
type:Array(1) []
0:Object {type: , mongodb: Object}
__proto__:Object {constructor: , __defineGetter__: , __defineSetter__: , …}
```
result:
```
model.definition.properties.customMongoProp2
Object {0: Object, type: Array(1)}
should:Assertion
type:Array(1) []
0:Object {type: , mongodb: Object}
__proto__:Object {constructor: , __defineGetter__: , __defineSetter__: , …}
```

test/basic-querying.test.js
```diff
@@ -1031,6 +1031,38 @@ describe('basic-querying', function() {
...
+        dateProp: updateDate,
+        dateArray: [updateDate, updateDate],
+        numProp: '3',
+        numArray: ['3', '4'],
+      };
+      const created = await numAndDateModel.create(createData);
+      const updated = await numAndDateModel.updateAll({id: created.id}, updateData);
+      const found = await numAndDateModel.findById(created.id);
+      found.dateProp.should.deepEqual(new Date(updateDate));
+      found.dateArray[1].should.deepEqual(new Date(updateDate));
```
jannyHou: nitpick: shall we assert `found.dateArray[0]` and `found.numArray[0]` as well?

lib/dao.js
```diff
@@ -2293,6 +2293,7 @@ DataAccessObject.updateAll = function(where, data, options, cb) {
         if (!(data instanceof Model)) {
           try {
             inst = new Model(data, {applyDefaultValues: false});
+            ctx.data = inst.toObject(true);
```
jannyHou: what does `true` mean here?

test/datatype.test.js
jannyHou: Just followed the discussion above, I think the definition is missing an object property name:

```
const model = db.define('model', {
   bunchOfDates: [
     {
       // Please note the object should contain each property name and its definition as a key-value pair
       date: {
         type: Date,
         // is this the right syntax?
         mongodb: {
           dataType: 'Decimal128',
         },  
       }
     },
   ],
 });
```

jannyHou: FYI This is a list of nested property definitions I've tested with loopback-connector-mongodb@3.x:
https://github.com/strongloop/loopback-connector-mongodb/blob/3.x/docs/decimal128.md

lib/model.js
b-admike: I think this case should be fixed now since the actual bug was the fact that we didn't coerce array types of dates properly, but we do traverse through nested properties properly. As part of #1716, we should probably write tests first to verify before working on a fix.

test/datatype.test.js
b-admike: Thanks Janny. I think the use cases you pointed out are covered in the PR I created for mongodb (see https://github.com/strongloop/loopback-connector-mongodb/pull/501). We are looking at specifying an array type like `{"emails": [{"type": "string", "length": 64}]}` (see https://loopback.io/doc/en/lb3/LoopBack-types.html#array-types). I see the same results for even non-connector specific properties, so I think it's possibly a bug or we support it and maybe I need to change my assertions:

```
model.definition.properties.emails
Object {0: Object, type: Array(1)}
should:Assertion
type:Array(1) []
0:Object {type: "string", length: 64}
__proto__:Object {constructor: , __defineGetter__: , __defineSetter__: , …}
```
I'm going to leave it out of this PR since the fix in this PR is for a) making sure Date array properties are properly coerced, and b) making sure we coerce primitive types on update operation.

lib/dao.js
b-admike: Good question. From the API documentation, it'll convert model instance to Plain Data Object for properties defined in the schema:
```
/**
 * Convert model instance to a plain JSON object.
 * Returns a canonical object representation (no getters and setters).
 *
 * @param {Boolean} onlySchema Restrict properties to dataSource only.  Default is false.  If true, the function returns only properties defined in the schema;  Otherwise it returns all enumerable properties.
```
I followed what was done for create (see https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L353) for this statement.

test/manipulation.test.js
```diff
@@ -2366,7 +2366,7 @@ describe('manipulation', function() {
     it('should not coerce invalid values provided in where conditions', function(done) {
       Person.update({name: 'Brett Boe'}, {dob: 'notadate'}, function(err) {
         should.exist(err);
-        err.message.should.equal('Invalid date: notadate');
+        err.message.should.equal('Invalid date: Invalid Date');
```
b-admike: FYI this change is because we coerce date values in update now, and then validate them using `Model._coerce` in https://github.com/strongloop/loopback-datasource-juggler/pull/1702/files#diff-d717a01e40d8164976e4468af1bce663R2330. I was going to return the invalid string value for the date instead in the first pass (during model instantiation), but we have other tests which expect `Invalid Date` to be returned for those values (https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/manipulation.test.js#L2226), and I feel like that'd not be backwards compatible as opposed to a not so helpful message here.

lib/list.js
```diff
@@ -67,8 +67,14 @@ function List(items, itemType, parent) {
     if (isClass(this.itemType)) {
       return new this.itemType(item);
     } else {
-      if (Array.isArray(item)) return item;
-      else return this.itemType(item);
+      if (Array.isArray(item)) {
+        return item;
+      } else if (this.itemType === Date) {
+        if (item === null) return null;
```
b-admike: FYI keeping the same conversion we do in https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/model-builder.js#L704-L708 for non-array Date properties here.

Issue #1692, opened 2019-02-18
DataSource juggler is adding the model property with its default value in response


sharathmuthu6:
## Description/Steps to reproduce
I created a simple loopback application with model name `testDefault` with its model defintion as below : 
```
{
  "name": "testDefault",
  "base": "PersistedModel",
  "idInjection": true,
  "options": {
    "validateUpsert": true
  },
  "properties": {
    "name": {
      "type": "string"
    },
    "age": {
      "type": "number",
      "default": 12
    }
  },
  "validations": [],
  "relations": {},
  "acls": [],
  "methods": {}
}
```

DataSource of `testDefault` is as follows : 
```
  "testDefault": {
    "dataSource": "db",
    "public": true
  }
```

Since it is an  in-memory connector, i changed the `.all` method of memory connector (https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/connectors/memory.js#L504)
 to return `callback(null,[{"name":"john"}])`

Once i start the app and do a GET on /testDefaults, i am getting the following response : 
```
[
  {
    "name": "john",
    "age": 12
  }
]
```

## Expected result

My connector response(from endpoint) doesn't have age property, so i am expecting an output as below
```
[
  {
    "name": "john"
  }
]
```

In dao.js, https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1637
A new model is being created, thats where i could see that 
```
ModelConstructor {
  __cachedRelations: {},
  __data: { name: 'john', age: 12 },
  __dataSource: undefined,
  __strict: false,
  __persisted: true }
``` 
`age` property is getting added. 

## Task list

- [x] Fix in `master` (4.x): #1704, #1707 
- [x] Fix in `3.x`: #1705
- [x] Fix in `2.x`: #1706

hacksparrow:
@sharathmuthu6 you are modifying the source code and expecting it to behave in a certain way, which is not a valid issue in terms of the user API.

I can help you achieve what you want to do, if you can describe your requirements with the use of the API exposed to the user.

sharathmuthu6:
I have added the testcase in loopback-sandbox repo
https://github.com/sharathmuthu6/loopback-sandbox/tree/issue1692

hacksparrow:
Hi @sharathmuthu6 you are overwriting the behavior of the connector with your own implementation.

```js
Memory.prototype.all = function all(model, data, options, callback) {
[...]

PR #1704, opened 2019-04-01
fix: do not apply default values on data from database

bajtos:
Before this change, when a property was configured with a default value at LoopBack side and the database was returned a record with a missing value for such property, then we would supply use the configured default.

This behavior is problematic for reasons explained in #1692.

~~In this commit, we are introducing a new model-level setting called `applyDefaultsOnReads`, which is enabled by default for backwards compatibility.~~

~~When this setting is set to `false`, operations like `find` and `findOrCreate` will NOT apply default property values on data returned by the database (connector).~~

In this pull request, we are fixing DAO operations like `find` and `findOrCreate` so that they do NOT apply default property values on data returned by the database (connector).

Please note that most of the other CRUD methods were already not applying default values on database data as long as the connector provided native implementation of the operation.

#### Related issues

- connect to #1692

/cc @sharathmuthu6 @snarjuna
### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1109,8 +1109,14 @@ DataAccessObject.findOrCreate = function findOrCreate(query, data, options, cb) 
         let obj;
 
         if (data) {
-          obj = new Model(data, {fields: query.fields, applySetters: false,
-            persisted: true});
+          const ctorOpts = {
+            fields: query.fields,
+            applySetters: false,
+            persisted: true,
+            // see https://github.com/strongloop/loopback-datasource-juggler/issues/1692
+            applyDefaultValues: false,
+          };
+          obj = new Model(data, ctorOpts);
         }
 
         if (created) {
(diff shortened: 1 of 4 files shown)


bajtos:
Discussion points:
- My intention is to backport this fix to 3.x and 2.x too. Because those versions are in LTS, it's important to preserve backwards compatibility so that we don't break existing applications. That's why I am introducing a feature flag.
- Should we enable this feature flag by default in juggler 4.x (LB4)? Or perhaps remove it entirely and never apply default values on data from database?

bajtos:
> I tend to "remove it entirely and never apply default values on data from database", so that the data returned from juggler preserves the original data retrieved from backend

Agreed. I think such behavior is also more consistent across different connectors.

I have reworked the pull request accordingly, @jannyHou PTAL again

bajtos:
Hmm, the CI build is failing, I'll need to investigate before landing.

bajtos:
It turns out we have tests to verify that default values are applied on reads, I fixed them in bf63682 AFAICT, these tests were added in the original juggler codebase before we forked it, see https://github.com/strongloop/loopback-datasource-juggler/commit/fcc2869806c2d680074766982515b672770fcc4c

PR #1705, opened 2019-04-04
Do not apply default values on data from database [3.x]

bajtos:
Before this change, when a property was configured with a default value at LoopBack side and the database was returned a record with a missing value for such property, then we would supply use the configured default.

This behavior is problematic for reasons explained in #1692.

In this patch, we are introducing a new model-level setting called `applyDefaultsOnReads`, which is enabled by default for backwards compatibility.

When this setting is set to `false`, operations like `find` and `findOrCreate` will NOT apply default property values on data returned by the database (connector).

Please note that most of the other CRUD methods did not apply default values on database data as long as the connector provided native implementation of the operation, that aspect is not changing.

#### Related issues

This is a back-port of #1704. Compared to #1704 (master/4.x), we are preserving backwards compatibility by introducing a new feature flag.

See also #1692.

/cc @sharathmuthu6 @snarjuna

### Checklist


bajtos:
> [cis-jenkins] downstream: loopback-connector-mongodb@master — Failed! (be1c365)

The failure seems unrelated, see https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-mongodb%7Emaster/35/console

> [cis-jenkins] downstream: loopback-connector-mysql@master — Failed! (be1c365)

This is a valid failure. Because the new tests don't run autoupdate/automigrate, the database schema is not matching model schema. I'll investigate.

bajtos:
In 6e7aee4, I added `automigrate` calls that fixed failing tests for SQL connectors.

However, there are still few failures that look relevant to me.

@jannyHou could you please help me to investigate the Cloudant failure? 

**Cloudant**
https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-cloudant%7Emaster/40/console
```
3) cloudant imported features basic-querying find applies default values by default:
07:50:37      TypeError: Cannot read property 'use' of undefined
07:50:37       at Cloudant.CouchDB.selectModel (node_modules/loopback-connector-couchdb2/lib/couchdb.js:750:29)
07:50:37       at Cloudant.destroyAll (node_modules/loopback-connector-couchdb2/lib/couchdb.js:423:17)

4) cloudant imported features basic-querying find preserves empty values from the database when "applyDefaultsOnReads" is false:
07:50:37      TypeError: Cannot read property 'use' of undefined
07:50:37       at Cloudant.CouchDB.selectModel (node_modules/loopback-connector-couchdb2/lib/couchdb.js:750:29)
07:50:37       at Cloudant.destroyAll (node_modules/loopback-connector-couchdb2/lib/couchdb.js:423:17)
07:50:37       at node_modules/loopback-connector-couchdb2/lib/migrate.js:42:14
```

The other two test failures reported by Cloudant seem unrelated,  they are reported by CouchDB connector too.

@raymondfeng could you please help me with MSSQL? It looks like a bug in the `autoupdate` implementation. Other SQL connectors are passing these tests.

**mssql**
https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-mssql%7Emaster/39/console
```
1) mssql imported features
07:58:29        basic-querying
07:58:29          find
07:58:29            applies default values by default:
07:58:29      RequestError: Could not drop object 'dbo.Product' because it is referenced by a FOREIGN KEY constraint

2) mssql imported features
07:58:29        basic-querying
07:58:29          find
07:58:29            preserves empty values from the database when "applyDefaultsOnReads" is false:
07:58:29      RequestError: Could not drop object 'dbo.Product' because it is referenced by a FOREIGN KEY constraint.

3) mssql imported features
07:58:29        manipulation
07:58:29          findOrCreate
07:58:29            applies default values on returned data:
07:58:29      RequestError: Could not drop object 'dbo.Product' because it is referenced by a FOREIGN KEY constraint.

4) mssql imported features
07:58:29        manipulation
07:58:29          findOrCreate
07:58:29            preserves empty values from the database when "applyDefaultsOnReads" is false:
07:58:29      RequestError: Could not drop object 'dbo.Product' because it is referenced by a FOREIGN KEY constraint.
```

---

Other test failures seem unrelated:

**dashdb**
https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-dashdb%7Emaster/42/console

Seems unrelated:

```
07:48:54 > node pretest.js
07:48:54 
07:48:55 > Seeding the database ...
07:49:55 
07:49:55 /home/jenkins/workspace/ds/loopback-connector-dashdb~master/pretest.js:53
07:49:55   if (err) throw err;
07:49:55            ^
07:49:55 Error: [IBM][CLI Driver] SQL30081N  A communication error has been detected. Communication protocol being used: "TCP/IP".  Communication API being used: "SOCKETS".  Location where the error was detected: "9.30.218.42".  Communication function detecting the error: "selectForConnectTimeout".  Protocol specific error code(s): "115", "*", "*".  SQLSTATE=08001
07:49:55 
07:49:55     at Error (native)
```

Same error for "db2" .

**kv-extreme-scale**
https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-kv-extreme-scale%7Emaster/40/console

Cannot connect to the testing database :(

**mongodb@3.x**
https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-mongodb%7E3.x/20/console

Fails in `lazyConnect` and `ping` test cases, seems unrelated to me.

Similarly for mongodb@master.

**rest**
https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-rest%7Emaster/21/console

```
Uncaught AssertionError: You must use an API key to authenticate each request to Google Maps Platform APIs. For additional information, please refer to http://g.co/dev/maps-no-account
07:57:27       + expected - actual
07:57:27 
07:57:27       -REQUEST_DENIED
07:57:27       +OK
07:57:27       
```

jannyHou:
Sure I am looking a the Cloudant failure.

bajtos:
> [cis-jenkins] downstream: loopback-connector-mssql@master — Failed! (6e7aee4)
> `RequestError: Could not drop object 'dbo.Product' because it is referenced by a FOREIGN KEY constraint.`

Apparently, there are other tests (or our setup script) using `Product` model and setting up a foreign key constraint(s) for it. I am not sure why our mssql connector is not dropping those constraints.

Anyhow, I managed to fix the problem by renaming my test model from `Product` to `Player`.

bajtos:
> [cis-jenkins] downstream: loopback-connector-cloudant@master — Failed! (6e7aee4)
> `07:50:37      TypeError: Cannot read property 'use' of undefined`

I managed to fix this problem with a small tweak in Cloudant's test suite, see https://github.com/strongloop/loopback-connector-cloudant/pull/202

bajtos:
I consider this pull request as done, I'll clean up the git history before landing.

@jannyHou @raymondfeng please review

test/basic-querying.test.js
```diff
@@ -917,6 +917,71 @@ describe('basic-querying', function() {
...
+      let created;
+      return db.automigrate('Player')
+        .then(() => Player.create({name: 'Pen'}))
+        .then(result => {
+          created = result;
+
+          // Later on, we decide to introduce `active` property
+          Player.defineProperty('active', {
+            type: Boolean,
+            default: false,
```
hacksparrow: Why not check with a string or a number? The fact that `null` is an acceptable common result for both `applyDefaultsOnReads = true` and `applyDefaultsOnReads  = false` makes the tests a little unreliable in my opinion.

A `. oneOf()` condition might be not be a good choice in this case.

bajtos: > Why not check with a string or a number?

I think it does not really matter. `undefined` or `null` is not a valid value for a boolean in the same way as it is for a string or a number.

> The fact that `null` is an acceptable common result for both `applyDefaultsOnReads = true` and `applyDefaultsOnReads  = false` makes the tests a little unreliable in my opinion.

Please note that SQL databases cannot represent `undefined`, they always use `null`. When `applyDefaultOnReads` is enabled (e.g. by default), then we return the default value, not `null`. Unless you have configured your model property with `default: null`, of course.

Could you please elaborate more on this? I don't understand your concerns :(

> A `.oneOf()` condition might be not be a good choice in this case.

Why not? How do you propose to test handling of default values, considering that different databases are returning different values for properties that have been added to existing model instances (database rows)? SQL databases return `null`, no-SQL databases (memory, MongoDB, etc.) return `undefined`.

bajtos:
@slnode test please

test/basic-querying.test.js
hacksparrow: > Why not? How do you propose to test handling of default values, considering that different databases are returning different values for properties that have been added to existing model instances (database rows)? SQL databases return null, no-SQL databases (memory, MongoDB, etc.) return undefined.

If we don't use a Boolean data type we won't have to deal with the differences in the behavior of different databases. Why bring in a third-party factor in the test case for a subject? If we want to factor in the differences in the behavior of different databases, it should be covered in a separate test case.

hacksparrow: To sum it up, it looks like to me, that we are testing the behavior of `applyDefaultsOnReads` as well as the database translation of a Boolean value, inside a singe test case. Shouldn't test cases be confined to a single condition, unless it is an acceptance test?

bajtos: I had a chat with @hacksparrow and now I understand the problem - I was using a wrong assertion for the tests checking the legacy behavior where default values are applied. I fixed the problem in f498c26.

bajtos: Actually, it turns out my original test assertions were correct. When the connector returns property value `null`, we don't consider it to be `undefined` and thus the default value is not applied. How confusing! I have updated the comments in the tests to make this more clear.

bajtos:
> I had a chat with @hacksparrow and now I understand the problem - I was using a wrong assertion for the tests checking the legacy behavior where default values are applied. I fixed the problem in f498c26.

Actually, it turns out my original test assertions were correct. When the connector returns property value null, we don't consider it to be undefined and thus the default value is not applied. How confusing! I have reverted f498c26 and updated the comments in the tests to make this more clear.

PR #1707, opened 2019-04-09
Improve tests for empty vs. default values

bajtos:
### Description

In #1705, I had to modify the tests added by #1704 to make them pass on different databases (SQL vs NoSQL, etc.)

This pull request is cherry-picking the changes from #1705 into master.

#### Related issues

-  #1704, #1705 and #1692

### Checklist

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -2615,25 +2615,25 @@ describe('manipulation', function() {
     it('preserves empty values from the database', async () => {
       // https://github.com/strongloop/loopback-datasource-juggler/issues/1692
 
-      // Initially, all Products were always active, no property was needed
-      const Product = db.define('Product', {name: String});
+      // Initially, all Players were always active, no property was needed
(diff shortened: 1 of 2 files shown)


test/manipulation.test.js
```diff
@@ -2615,25 +2615,25 @@ describe('manipulation', function() {
...
 
       // And upsertWithWhere an existing record
-      const found = await Product.upsertWithWhere({id: created.id}, {name: 'updated'});
-      found.toObject().should.eql({
-        id: created.id,
-        name: 'updated',
-        active: undefined,
-      });
+      const found = await Player.upsertWithWhere({id: created.id}, {name: 'updated'});
+      should(found.toObject().active).be.oneOf([
```
b-admike: Should we want to preserve the checks for `id` and `name` as well?

bajtos: Those properties are not really affected by default values, I feel it's ok to not check them - we have other tests in place to verify.

Issue #1724, opened 2019-04-23
EmbedsMany delete methods fail to pass options


bmatson:
# Description/Steps to reproduce

We use the options inside of our loopback app to enforce client tenancy.  Since this is a security feature we require the options and the associated security token to be passed through to all methods, and fail if it is not passed.  Since the options also contain transaction information, this can cause concurrency problems as well for people who are doing transactions.

In the EmbedsMany.prototype.destroyById and destroyAll, the calls to modelInstance.updateAttribute or modelInstance.setAttribute fail to pass through the incoming options.

We're using the 3.x version of the loopback-datasource-juggler, however the problem exists in master as well.  

PR coming shortly against master with fix.

dhmlau:
PRs for both master and 3.x have landed. Closing as resolved. @bmatson, thanks again for your contribution!

PR #1725, opened 2019-04-23
Fix for #1724 - Added options to attribute updates

bmatson:
### Description

Adds options to calls to attribute updates in EmbedsMany relations

#### Related issues

- connect to #1724 

### Checklist

### DIFF
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -4980,6 +4980,72 @@ describe('relations', function() {
       });
     });
 
+    it('should pass options when removed by id', function(done) {
+      const verifyOptions = function(ctx, next) {
+        if (!ctx.options || !ctx.options.verify) {
+          return next(new Error('options or options.verify is missing'));
+        }
+        return next();
+      };
+      Person.observe('before save', verifyOptions);
+      Person.findOne(function(err, p) {
+        p.addressList.create({street: 'options 1'}, {verify: true}, function(err, address) {
+          if (err) {
+            Person.clearObservers('before save');
+            return done(err);
+          }
+          p.addressList.destroy(address.id, {verify: true}, function(err) {
+            if (err) {
+              Person.clearObservers('before save');
+              return done(err);
+            }
+            Person.findById(p.id, function(err, verify) {
+              if (err) {
+                Person.clearObservers('before save');
+                return done(err);
+              }
+              verify.addresses.should.have.length(1);
+              Person.clearObservers('before save');
+              done();
+            });
+          });
+        });
+      });
+    });
+
+    it('should pass options when removed by where', function(done) {
+      const verifyOptions = function(ctx, next) {
+        if (!ctx.options || !ctx.options.verify) {
+          return next(new Error('options or options.verify is missing'));
+        }
+        return next();
+      };
+      Person.observe('before save', verifyOptions);
+      Person.findOne(function(err, p) {
+        p.addressList.create({street: 'options 2'}, {verify: true}, function(err, address) {
+          if (err) {
+            Person.clearObservers('before save');
+            return done(err);
+          }
+          p.addressList.destroyAll({street: 'options 2'}, {verify: true}, function(err) {
+            if (err) {
+              Person.clearObservers('before save');
+              return done(err);
+            }
+            Person.findById(p.id, function(err, verify) {
+              if (err) {
+                Person.clearObservers('before save');
+                return done(err);
+              }
+              verify.addresses.should.have.length(1);
+              Person.clearObservers('before save');
+              done();
+            });
+          });
+        });
+      });
+    });
+
     // eslint-disable-next-line mocha/no-identical-title
     it('should create embedded items on scope', function(done) {
       Person.findOne(function(err, p) {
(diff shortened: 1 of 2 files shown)


slnode:

dhmlau:
@bmatson, thanks for the PR.  Could you please add some tests?  Thanks.

dhmlau:
@slnode test please

bmatson:
@dhmlau Not quite sure how I'd even test this change, since all I did was add an existing passed-in parameter to an existing call that expects it as an optional parameter I believe the fact that the existing test cases don't break covers the change.  If I'd added new logic or code paths I could see the need for a unique test case, but the way that options got worked into the existing code base makes writing test cases for it not being passed very difficult.  Looking at the coverage report, it looks like the only "new" branch is the non-taken side of ```options || {}```.

bmatson:
@dhmlau - Added some tests, added an observer for before save that throws if options aren't passed.

bmatson:
@dhmlau - Any progress on getting this approved and merged into the 3.X branch?  Due to this bug we're having to run forked versions of both loopback and juggler in production.

dhmlau:
@bmatson, sorry about the delay. Could you please rebase your PR? Thanks.
@strongloop/sq-lb-apex, could you PTAL?

dhmlau:
@slnode test please

bmatson:
@dhmlau - Need to make sure this change gets cherrypicked back into the 3.x branch, too, since the issue we're having isn't really going to be resolved until loopback 3.x gets fixed.

dhmlau:
After this got approved and merged, you can cherry pick the changes to the 3.x branch.

bmatson:
@dhmlau - I can't do that, we're running off our fork of the 3.x branch, which is behind yours and already has the change merged in.  The cherry picking of these two commits needs to happen in the strongloop repository, which I don't have access to.

dhmlau:
@b-admike mentioned he has some questions about the tests and will look into it some more. I'd like to wait for his review comments.

test/relations.test.js
```diff
@@ -4980,6 +4980,72 @@ describe('relations', function() {
...
     });
 
+    it('should pass options when removed by id', function(done) {
+      const verifyOptions = function(ctx, next) {
+        if (!ctx.options || !ctx.options.verify) {
+          return next(new Error('options or options.verify is missing'));
+        }
+        return next();
+      };
+      Person.observe('before save', verifyOptions);
```
b-admike: I'm not too familiar with the observer codebase, so sorry if I misunderstand. We make the changes in `before delete` hook above, but we are not using it in the tests at all. Can you please explain the flow of the test and why we don't use it to verify the options object passed into it?

bmatson:
When you delete from an embedded list it saves a new version without the deleted value.  The fix is for this save not passing the options value if you look at the code change.  Because options isn’t passed it isn’t there for the hooks to use.  To verify that it’s passed you have to hook the before save event on the containing model.  If you edit the fake options you can see the test fail.

Bill

b-admike:
> When you delete from an embedded list it saves a new version without the deleted value. The fix is for this save not passing the options value if you look at the code change. Because options isn’t passed it isn’t there for the hooks to use. To verify that it’s passed you have to hook the before save event on the containing model. If you edit the fake options you can see the test fail.

Thank you for the explanation @bmatson. I was able to test it out locally as well.

test/relations.test.js
```diff
@@ -4980,6 +4980,72 @@ describe('relations', function() {
...
+          if (err) {
+            Person.clearObservers('before save');
+            return done(err);
+          }
+          p.addressList.destroy(address.id, {verify: true}, function(err) {
+            if (err) {
+              Person.clearObservers('before save');
+              return done(err);
+            }
+            Person.findById(p.id, function(err, verify) {
```
b-admike: nitpick: let's call `verify` either `person` or `p`.

test/relations.test.js
```diff
@@ -4980,6 +4980,72 @@ describe('relations', function() {
...
+              if (err) {
+                Person.clearObservers('before save');
+                return done(err);
+              }
+              verify.addresses.should.have.length(1);
```
b-admike: Nitpick: from a test perspective, it would be good if you deleted all the address instances from the `Person` model before your two new tests so that here we expect none. If not, then can we leave a comment here saying there was one already?

test/relations.test.js
```diff
@@ -4980,6 +4980,72 @@ describe('relations', function() {
...
+          if (err) {
+            Person.clearObservers('before save');
+            return done(err);
+          }
+          p.addressList.destroyAll({street: 'options 2'}, {verify: true}, function(err) {
+            if (err) {
+              Person.clearObservers('before save');
+              return done(err);
+            }
+            Person.findById(p.id, function(err, verify) {
```
b-admike: ditto

test/relations.test.js
```diff
@@ -4980,6 +4980,72 @@ describe('relations', function() {
...
+              if (err) {
+                Person.clearObservers('before save');
+                return done(err);
+              }
+              verify.addresses.should.have.length(1);
```
b-admike: ditto

dhmlau:
@bmatson, thanks for your contribution. Your PR has landed! 🎉 
To backport the changes to 3.x, can you submit a PR against the 3.x branch?

bmatson:
Ok, just created PR 1748 with the commit.

Bill

PR #1726, opened 2019-04-25
fix: allow coercion of nested properties [3.x]

### LINKED ISSUE #1723: Mongo Extended operators in updateAll not working

Merge request #1702 broke functionality of  extended operators in `updateAll` operation on the models.

The new line added in `dao.js` [here](https://github.com/strongloop/loopback-datasource-juggler/blob/df4366eb3114a7e16673ce1b4d29044ab33bf094/lib/dao.js#L2306) stores in `ctx.data` the **object** representation of the created instance from the above line, so this is the final value passed in the connector update method which handles the data update.

This has the effect that any passed extended operators such as in **MongoDB** `$set` is stripped from the update data because `.toObject` does not take into account any extended operators that might have been given in the update data, which is referenced here: https://loopback.io/doc/en/lb3/MongoDB-connector.html

So basically you cannot use extended operators in `updateAll` function.

To find out the result just issue an update on a model that has enabled the use of `allowExtendedOperators` such as  `Model.updateAll({}, { $set: { prop: 'sth'} })`

# Expected result

It should pass the given extended operators in the update object and let the MongoDB Connector handle them appropriately.

# Additional information

linux x64 10.15.3
├── loopback@3.25.1
├── loopback-connector-mongodb@4.1.0

b-admike:
### Description

Fixes #1723. Make sure we coerce nested property values for DAO methods like `updateAll`.

### Checklist

### DIFF
--- a/lib/model-utils.js
+++ b/lib/model-utils.js
@@ -350,11 +350,12 @@ ModelUtils._sanitizeData = function(data, options) {
  * Coerce values based the property types
  * @param {Object} where The where clause
  * @options {Object} [options] Optional options to use.
+ * @param {Object} Optional model definition to use.
  * @property {Boolean} allowExtendedOperators.
  * @returns {Object} The coerced where clause
  * @private
  */
-ModelUtils._coerce = function(where, options) {
+ModelUtils._coerce = function(where, options, modelDef) {
   const self = this;
   if (where == null) {
     return where;
@@ -367,8 +368,13 @@ ModelUtils._coerce = function(where, options) {
     err.statusCode = 400;
     throw err;
   }
+  let props;
+  if (modelDef && modelDef.properties) {
+    props = modelDef.properties;
+  } else {
+    props = self.definition.properties;
+  }
 
-  const props = self.definition.properties;
   for (const p in where) {
     // Handle logical operators
     if (p === 'and' || p === 'or' || p === 'nor') {
@@ -393,7 +399,7 @@ ModelUtils._coerce = function(where, options) {
     if (!DataType) {
       continue;
     }
-    if (Array.isArray(DataType) || DataType === Array) {
+    if ((Array.isArray(DataType) || DataType === Array) && !isNestedModel(DataType)) {
       DataType = DataType[0];
     }
     if (DataType === Date) {
@@ -410,10 +416,6 @@ ModelUtils._coerce = function(where, options) {
     }
 
     if (!DataType) {
-      continue;
-    }
-
-    if (DataType.prototype instanceof BaseModel) {
       continue;
     }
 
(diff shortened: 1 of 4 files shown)


b-admike:
@strongloop/sq-lb-apex @mitsos1os PTAL

b-admike:
@slnode test please

lib/model-utils.js
```diff
@@ -551,3 +568,15 @@ ModelUtils._coerce = function(where, options) {
...
+/**
+* A utility function which checks for nested property definitions
+*
+* @param {*} propType Property type metadata
+*
+*/
+function isNestedModel(propType) {
+  if (!propType) return false;
+  if (Array.isArray(propType)) return isNestedModel(propType[0]);
+  return propType.definition && propType.definition.properties;
```
mitsos1os: Here an actual boolean could be returned to keep function signature consistent. ex:
`return !!(propType.definition && propType.definition.properties;` or 
`return propType.hasOwnProperty('definition') && propType.definition.hasOwnProperty('properties');`

lib/model-utils.js
```diff
@@ -391,6 +397,17 @@ ModelUtils._coerce = function(where, options) {
...
+    if (isNestedModel(DataType)) {
+      if (Array.isArray(DataType)) {
+        if (value === null || value === undefined) continue;
+        for (const it of value) {
+          self._coerce(it, options, DataType[0].definition);
+        }
+      } else {
+        self._coerce(value, options, DataType.definition);
+      }
+    }
```
mitsos1os: I might be missing something here, but by taking a quick look, if **nested** model is found, couldn't it just `continue` on to next iteration instead of following the rest of the logic?

b-admike: If the nested model is not an array, then we are recursively sending its `definition` to use with the `_coerce` function here.

mitsos1os: Yes I understand that. I was talking about using a `continue`  before closing the **NestedModel** is between line 409 and 410

b-admike: I see what you mean now thank you for clarifying. If the model is nested, then it either checks for array or not but either way it calls `self._coerce()`. Would it be better if I re-wrote this as follows (without the `else` block)?
```js
    if (isNestedModel(DataType)) {
      if (Array.isArray(DataType)) {
        if (value === null || value === undefined) continue;
        for (const it of value) {
          self._coerce(it, options, DataType[0].definition);
        }
      } 
        self._coerce(value, options, DataType.definition);
    }
```

b-admike:
Thanks for your feedback @mitsos1os. LGTY now?

mitsos1os:
@b-admike LGTM! Thanks for reviewing the feedback!

b-admike:
@slnode test please

PR #1728, opened 2019-04-29
Add support for Node.js 12.x

bajtos:
- Fix code to pass all tests on Node.js 12
- Add Node.js 12 to the list of Travis CI platforms

#### Related issues

### Checklist

### DIFF
--- a/lib/model-utils.js
+++ b/lib/model-utils.js
@@ -389,6 +389,8 @@ ModelUtils._coerce = function(where, options) {
         self._coerce(clauses[k], options);
       }
 
+      where[p] = clauses;
+
       continue;
     }
     let DataType = props[p] && props[p].type;
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,6 @@ language: node_js
 node_js:
   - "8"
   - "10"
+  - "12"
 
 after_success: npm run coverage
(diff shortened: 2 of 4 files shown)


bajtos:
Any objections against back-porting this change to 3.x?

lib/model-utils.js
```diff
@@ -389,6 +389,8 @@ ModelUtils._coerce = function(where, options) {
         self._coerce(clauses[k], options);
       }
 
+      where[p] = clauses;
```
bajtos: This is kind of weird. Before my change, the block "handle logical operators" did not apply updates made by `coerceArray` to the original `where[p]` value. As a result, array-like object `{1: {field: 'value'}}` was not converted into an array `[{field: 'value'}]`. I am not sure why this was not caught earlier and surfaced only with Node.js 12. Is it perhaps a change in how deep-equality checks work on the new Node version? IDK.

bajtos: Here are the tests that are failing without my fix:

https://github.com/strongloop/loopback-datasource-juggler/blob/47addd943adfd7ace37cd52d2d4dcbd04ee7c0d8/test/loopback-dl.test.js#L1498-L1503

bajtos:
@slnode test please

b-admike:
@slnode test please

PR #1736, opened 2019-05-06
[DO NOT MERGE] trigger downstream builds of 4.x

bajtos:
Test strongloop/loopback-connector-postgresql#376 in practice.


bajtos:
@rmg This pull request was supposed to trigger a downstream build of `loopback-datasource-juggler@master`, but didn't.

IIUC, downstream builds are running on Node.js 6.x (see e.g. https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-postgresql%7Emaster/64/console). This version of juggler does not support Node.js 6.x, is that perhaps the reason why the downstream build was not triggered?

How can we upgrade cis-jenkins to use Node.js 10.x for downstream builds?

rmg:
@bajtos the version of node used for downstream tests is picked from an ordered list. The first hit that the module claims to support is what gets used. If none match, downstream testing is skipped for that dependant. I've opened up a PR on the CI system to change the downstream version list to prefer node 10.

rmg:
@slnode test please

rmg:
I didn't look closing at the timing earlier. You created this pull request very soon after merging strongloop/loopback-connector-postgresql#376. The job configuration that is generated for the upstream/downstream testing is only run every 2 hours, so when you opened this PR to test the change it had not yet been applied to the job config. Before triggering another build I took a look and verified that the job now includes a downstream build trigger that was not part of the config when the previous build ran.

bajtos:
> I didn't look closing at the timing earlier. You created this pull request very soon after merging strongloop/loopback-connector-postgresql#376. The job configuration that is generated for the upstream/downstream testing is only run every 2 hours, so when you opened this PR to test the change it had not yet been applied to the job config. Before triggering another build I took a look and verified that the job now includes a downstream build trigger that was not part of the config when the previous build ran.

Makes sense, thank for explaining.

The remaining problem is that cis-jenkins is picking Node.js 6 runtime, despite the fact that the build was triggered by a change in a module that does not support Node.js 6 and requires 8+.

Quoting from https://cis-jenkins.swg-devops.com/job/ds/job/loopback-connector-postgresql%7Emaster/65/console:
```
Now using node v6.17.1 (npm v3.10.10)
```

Fortunately, Node.js 6 has reached EOL, so we can solve this problem by removing Node.js 6 support from the downstream packages.

bajtos:
@slnode test please

bajtos:
@slnode test please

bajtos:
> [cis-jenkins] downstream: loopback-connector-postgresql@master — Success! (ac1a640)

Nice, cis-jenkins executed connector tests including juggler-v4 tests on Node.js 8.16.0 👍

PR #1739, opened 2019-05-13
Fix implementation of Operation Hooks for "near" (geo) queries

bajtos:
The first commit is reworking the setup of the Operation Hooks test suite for memory connector so that the "unoptimized" variant disables not only atomic CRUD operations, but also geo queries.

This way we can test both "near" querying implementations:
 - When the connectors supports "near" queries natively.
 - When "near" queries are executed at LoopBack side in-memory.

The second commit fixes "loaded" hook for unoptimized "near" queries (as demonstrated by a failing test).

The third commit fixes "access" hook for unoptimized "near" queries (as demonstrated by a failing test).

These changes are needed to make MongoDB connector tests pass when running against juggler v4.

#### Related issues

- https://github.com/strongloop-internal/scrum-apex/issues/416

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1569,7 +1569,8 @@ DataAccessObject.find = function find(query, options, cb) {
       return cb.promise;
     }
   }
-  function geoCallbackWithoutNotify(err, data) {
+
+  function geoCallback(err, data) {
     const memory = new Memory();
     const modelName = self.modelName;
 
@@ -1588,37 +1589,14 @@ DataAccessObject.find = function find(query, options, cb) {
         });
       });
 
-      // FIXME: apply "includes" and other transforms - see allCb below
       memory.all(modelName, geoQueryObject, options, allCb);
     } else {
       cb(null, []);
     }
   }
 
-  function geoCallbackWithNotify(err, data) {
-    if (err) return cb(err);
-
-    async.map(data, function(item, next) {
-      const context = {
-        Model: self,
-        data: item,
-        isNewInstance: false,
-        hookState: hookState,
-        options: options,
-      };
-
-      self.notifyObserversOf('loaded', context, function(err) {
-        if (err) return next(err);
-        next(null, context.data);
-      });
-    }, function(err, results) {
-      if (err) return cb(err);
-      geoCallbackWithoutNotify(null, results);
-    });
-  }
   function queryGeo(query) {
     geoQueryObject = query;
-    const geoCallback = options.notify === false ? geoCallbackWithoutNotify : geoCallbackWithNotify;
     invokeConnectorMethod(connector, 'all', self, [{}], options, geoCallback);
   }
 
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -1535,7 +1535,7 @@ DataAccessObject.find = function find(query, options, cb) {
 
   this.applyScope(query);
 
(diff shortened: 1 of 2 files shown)


lib/dao.js
```diff
@@ -1588,38 +1589,15 @@ DataAccessObject.find = function find(query, options, cb) {
...
       });
 
-      // FIXME: apply "includes" and other transforms - see allCb below
       memory.all(modelName, geoQueryObject, options, allCb);
     } else {
       cb(null, []);
     }
   }
 
-  function geoCallbackWithNotify(err, data) {
```
b-admike: I'm wondering what was the original intent for `geoCallbackWithNotify` and why we are removing it now?

lib/dao.js
```diff
@@ -1597,6 +1597,7 @@ DataAccessObject.find = function find(query, options, cb) {
 
   function queryGeo(query) {
     geoQueryObject = query;
+    near = query && geo.nearFilter(query.where);
```
b-admike: 👍

lib/dao.js
b-admike: e.g. we have it in [3.x](https://github.com/strongloop/loopback-datasource-juggler/blob/3.x/lib/dao.js#L1596-L1619) branch, so maybe we need to revert changes introduced by #1690?

bajtos: In https://github.com/strongloop/loopback/pull/4194, you changed the way how query results are handled when the geo query is handled in-memory.

Before your changes, `geoCallbackWithNotify` was notifying "loaded" observers and calling `cb` directly.

After your change, `geoCallbackWithNotify` was calling `allCb` that notifies "loaded" observers, but at the same time it was  notifying "loaded" observers too, which caused double notifications.

My change is fixing the problem by removing `geoCallbackWithNotify` and always relying on `allCb` to notify "loaded" observers.

I believe the 3.x branch is not affected, because it still calls `cb` instead of `allCb`, see here:
https://github.com/strongloop/loopback-datasource-juggler/blob/548f3018a92b17e2e206971bcfdf541191a1410e/lib/dao.js#L1590

rmg:
@slnode test please

lib/dao.js
b-admike: Ok thanks for the clarification @bajtos. Do you intend to backport this change into `3.x` so we have consistent behaviour too?

bajtos: I am thinking about back-porting the following two commits:
 - Test "near" queries executed in memory
 - Fix "access" hook for unoptimized "near" queries
 - Make any other adjustments to make the tests pass

I believe that the commit _Fix "loaded" hook for unoptimized "near" queries_ is required only because of  #1690. AFAICT, #1690 was not back-ported to 3.x. WDYT, should we try to back-port it together with my fixes?

bajtos: On the second though, I am also going to rename `geoCallbackWithoutNotify` to `geoCallback` to avoid possible confusion. Even though `geoCallback` does not notify hook observers directly, it calls `allCb` which does trigger notifications.

bajtos: Backport: https://github.com/strongloop/loopback-datasource-juggler/pull/1742

b-admike: oh I see, in that case those two commits should be fine 👍. Agreed with the rename too.

PR #1752, opened 2019-06-24
chore: add beginTransaction API to declaration file

b-admike:
### Description
Connect to https://github.com/strongloop/loopback-next/issues/2614. We need to add `beginTransaction()` to the persisted model declaration file in order to utilize it with LB4 repositories.

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

- [ ] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style
  guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #1753, opened 2019-06-27
Add `beginTransaction` method for datasources

b-admike:
### Description

Enable `beginTransaction()` API for datasources to call Transaction's `begin()` method in `loopback-connector`. This returns a transaction instance that can be used independently by LB4's repositories.

#### Related issues

- connect to https://github.com/strongloop/loopback-next/issues/2614
- blocked on https://github.com/strongloop/loopback-connector/pull/147

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -2648,6 +2648,17 @@ DataSource.prototype.execute = function(command, args = [], options = {}) {
   }
 };
 
+/**
+ * Begin a new Transaction.
+ *
+ *
+ * @param [options] Options {isolationLevel: '...', timeout: 1000}
+ * @returns Promise A promise which resolves to a Transaction object
+ */
+DataSource.prototype.beginTransaction = function(options) {
+  return Transaction.begin(this.connector, options);
+};
+
 /*! The hidden property call is too expensive so it is not used that much
 */
 /**
--- a/test/transaction.test.js
+++ b/test/transaction.test.js
@@ -71,6 +71,13 @@ describe('Transactions on test connector without execute()', () => {
     }, done);
   });
 
+  it('beginTransaction returns a transaction', async () => {
+    const promise = db.beginTransaction(Transaction.READ_UNCOMMITTED);
+    promise.should.be.Promise();
+    const transaction = await promise;
+    transaction.should.be.instanceof(EventEmitter);
+  });
+
   it('exposes and caches slave models', done => {
     testModelCaching(tx.models, db.models);
     done();
--- a/types/datasource.d.ts
+++ b/types/datasource.d.ts
@@ -12,6 +12,7 @@ import {
   PropertyDefinition,
 } from './model';
 import {EventEmitter} from 'events';
+import {IsolationLevel, Transaction} from './transaction-mixin';
 
 /**
  * LoopBack models can manipulate data via the DataSource object.
@@ -185,4 +186,15 @@ export declare class DataSource extends EventEmitter {
     args?: any[] | object,
     options?: Options
   ): Promise<any>;
+
+/**
+ * Begin a new transaction.
+ *
+ *
+ * @param [options] Options {isolationLevel: '...', timeout: 1000}
+ * @returns Promise A promise which resolves to a Transaction object
+ */
+  beginTransaction(
+    options?: IsolationLevel | Options,
+  ): PromiseOrVoid<Transaction>;
 }
--- a/types/persisted-model.d.ts
+++ b/types/persisted-model.d.ts
@@ -6,7 +6,6 @@
 import {Callback, Options, PromiseOrVoid} from './common';
 import {ModelBase, ModelData} from './model';
 import {Filter, Where} from './query';
-import {Transaction} from './transaction-mixin';
 
 /**
  * Data object for persisted models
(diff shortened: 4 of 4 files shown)


lib/datasource.js
```diff
@@ -2648,6 +2648,17 @@ DataSource.prototype.execute = function(command, args = [], options = {}) {
...
 
+/**
+ * Begin a new Transaction calling the underlying connector's begin method.
+ *
+ *
+ * @param [options] Options {isolationLevel: '...', timeout: 1000}
+ * @returns Promise A promise which resolves to a Transaction object
+ */
+DataSource.prototype.beginTransaction = async function(options) {
+  return await Transaction.begin(this.connector, options);
```
bajtos: `return await` is an anti-pattern, just return the promise returned by `begin`.

```ts
DataSource.prototype.beginTransaction = function(options) {
  return Transaction.begin(this.connector, options);
}
```

test/transaction.test.js
```diff
@@ -71,6 +71,17 @@ describe('Transactions on test connector without execute()', () => {
     }, done);
   });
 
+  it('beginTransaction returns a transaction', done => {
+    const promise = db.beginTransaction(Transaction.READ_UNCOMMITTED);
+    promise.should.be.Promise();
+    promise.then(transaction => {
```
bajtos: Let's use await to keep the code simpler.

```suggestion
it('beginTransaction returns a transaction', async () => {
  // ...
  const transaction = await promise;
  // ...
}
```

types/datasource.d.ts
```diff
@@ -185,4 +186,15 @@ export declare class DataSource extends EventEmitter {
     args?: any[] | object,
     options?: Options
   ): Promise<any>;
+
+/**
+ * Begin a new Transaction calling the underlying connector's begin method.
```
bajtos: `underlying connector's begin method` - that's not an accurate description, is it? We call `Transaction.begin` that calls connector's method `beginTransaction`.

```suggestion
 * Begin a new transaction.
 ```

File: types/persisted-model.d.ts [hunk 4]
```diff
@@ -478,58 +477,6 @@ export declare class PersistedModel extends ModelBase {
...
-  * Valid values of `isolationLevel` are:
-  *
-  * - Transaction.READ_COMMITTED = 'READ COMMITTED'; // default
-  * - Transaction.READ_UNCOMMITTED = 'READ UNCOMMITTED';
-  * - Transaction.SERIALIZABLE = 'SERIALIZABLE';
-  * - Transaction.REPEATABLE_READ = 'REPEATABLE READ';
-  * @callback {Function} cb Callback function.
-  * @returns {Promise|undefined} Returns a callback promise.
-  */
-  beginTransaction(
```
bajtos: I have a question. Does `beginTransaction` exist in `PersistedModel` constructor or not? If it exists, then I think we should keep this definition, as removing it could be considered a breaking change. If it does not exist, then I am fine to remove this definition because it's not working anyways.

b-admike: I added this change in https://github.com/strongloop/loopback-datasource-juggler/pull/1752 and no it doesn't (see https://github.com/strongloop/loopback/blob/master/lib/persisted-model.js), which is why I'm reverting the changes made in that PR. IIUC it exists on a persisted model once an application is booted and `TransactionMixin` is applied.

types/datasource.d.ts
b-admike: That is correct - good catch!

lib/datasource.js
```diff
@@ -2648,6 +2648,17 @@ DataSource.prototype.execute = function(command, args = [], options = {}) {
...
 };
 
+/**
+ * Begin a new Transaction calling the underlying connector's begin method.
+ *
+ *
+ * @param [options] Options {isolationLevel: '...', timeout: 1000}
+ * @returns Promise A promise which resolves to a Transaction object
+ */
+DataSource.prototype.beginTransaction = async function(options) {
```
jannyHou: What is the relation between the new `beginTransaction` function and the existing `DataSource.prototype.transaction`(https://github.com/strongloop/loopback-datasource-juggler/blob/36135df4c9b5575821a6922316b972c1cc23f6cc/lib/datasource.js#L2235)?

b-admike: The `beginTransaction` one just gives you a Transaction instance with a uuid you can use for your model instance to make certain db queries with whereas `transaction` gives you back transaction object that is tied to all the models attached to your datasource. IIUC we don't want a certain repository class which ties one model to a datasource to be able to use the higher level `transaction` API for this reason (shouldn't be accessing other models). See https://loopback.io/doc/en/lb3/Using-database-transactions.html#transaction-apis for more info.

types/persisted-model.d.ts
bajtos: It seems that `beginTransaction` is added to models when they are attached to a (compatible) data-source. 

Initial state - `beginTransaction` is not available:

```
$ node
const {PersistedModel} = require('loopback');
undefined
> PersistedModel.beginTransaction
undefined
```

Using https://github.com/strongloop/loopback-sandbox, adding `console.log` statement after the app has booted (https://github.com/strongloop/loopback-sandbox/blob/96e99c26ae7ff07ca0d97c073b795bb6fcdf987b/server/server.js#L28):

```js
$ node .
[Function]
Web server listening at: http://localhost:3000
Browse your REST API at http://localhost:3000/explorer
```

Maybe we should keep `beginTransaction` but make it optional?

```suggestion
  beginTransaction?(
```

@raymondfeng what's your take?

I don't have a strong opinion, I am ok to remove this recently added typedef.

b-admike: Thank you for investigating @bajtos. Yea we can definitely do that - I don't have a strong opinion either. If we remove it from the declaration file does that constitute a breaking change? I.e. what do you think the semver version change should be depending on removal or making it optional?

b-admike:
Downstream postgres build is failing because of linter issues (not related to this PR).

types/persisted-model.d.ts
raymondfeng: Let's remove it as a decision to not expose it to TS.

Issue #1754, opened 2019-06-27
v3.31.0 introduces bug


ataft:
The following code works fine in v3.30.0:

```
const query = {
    where: {
        or: [
            { username: username },
            { email: username }
        ]
    }
};
self.userModel.findOne(query, function (err, user) {
```

As of v.3.31.0, I get the following error:
```
Error: The where clause "abc@xyz.com" is not an object
    at Function.ModelUtils._coerce (C:\apps\flexit\node_modules\loopback-datasource-juggler\lib\model-utils.js:369:11)
    at Function.ModelUtils._coerce (C:\apps\flexit\node_modules\loopback-datasource-juggler\lib\model-utils.js:552:20)
    at Function.ModelUtils._coerce (C:\apps\flexit\node_modules\loopback-datasource-juggler\lib\model-utils.js:394:14)
    at Function.ModelUtils._normalize (C:\apps\flexit\node_modules\loopback-datasource-juggler\lib\model-utils.js:177:8)
    at Function.find (C:\apps\flexit\node_modules\loopback-datasource-juggler\lib\dao.js:1529:10)
    at Function.findOne (C:\apps\flexit\node_modules\loopback-datasource-juggler\lib\dao.js:1782:8)
    at Strategy._verify (C:\apps\flexit\server\boot\authentication.js:288:36)
    at Strategy.authenticate (C:\apps\flexit\node_modules\passport-local\lib\strategy.js:90:12)
    at attempt (C:\apps\flexit\node_modules\passport\lib\middleware\authenticate.js:361:16)
    at authenticate (C:\apps\flexit\node_modules\passport\lib\middleware\authenticate.js:362:7)
```

Looks like something broke in the _coerce function.

raymondfeng:
It seems to be related to https://github.com/strongloop/loopback-datasource-juggler/commit/4f31db1b64efa384e2c46c73bd34b97463db78f1.

b-admike:
@ataft Can you check #1755 to see if I'm missing something for a test I wrote using your scenario? I am not able to reproduce your issue with it:
```
biniams-mbp:loopback-datasource-juggler badmike$ npm t

> loopback-datasource-juggler@3.31.1 test /Users/badmike/loopback/loopback-datasource-juggler
> nyc mocha



  ․

  1 passing (46ms)
```

b-admike:
@ataft can you share a sandbox app to reproduce the failure if that's easier? Please also include the steps you took to see the error.

[...]

PR #1755, opened 2019-06-27
Add a test for findOne with or operator

b-admike:
### Description

connect to #1754.

### Checklist

### DIFF
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -1067,6 +1067,19 @@ describe('basic-querying', function() {
         });
       });
     });
+
+    it('should work for or operator', function() {
+      const query = {
+        where: {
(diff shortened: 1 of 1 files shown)


test/basic-querying.test.js
```diff
@@ -1067,6 +1067,19 @@ describe('basic-querying', function() {
...
     });
+
+    it('should work for or operator', function() {
+      const query = {
+        where: {
+          or: [{name: 'Paul McCartney'}, {email: 'abc@xyz.com'}],
+        },
+      };
+      return User.findOne(query, function(err, user) {
+        should.not.exist(err);
```
bajtos:
```suggestion
        if (err) return done(err);
```

Even better, write the test as an `async` function and use `await User.findOne(query)` instead.

PR #1756, opened 2019-07-02
fix: report errors from automigrate/autoupdate

bajtos:
Defer automigrate/autoupdate until we are connected, so that connection errors can be reported back to callers.

Fix `postInit` handler to not report connection error to `console.log` and via dataSource "error" event in case there is already an operation queued. When this happens, we want the error to be handled by the queued operation and reported to its caller.

#### Related issues

- fixes https://github.com/strongloop/loopback-next/issues/3044

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -136,6 +136,7 @@ function DataSource(name, settings, modelBuilder) {
   this.models = this.modelBuilder.models;
   this.definitions = this.modelBuilder.definitions;
   this.juggler = juggler;
+  this._queuedInvocations = 0;
 
   // operation metadata
   // Initialize it before calling setup as the connector might register operations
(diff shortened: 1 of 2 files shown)


lib/datasource.js
```diff
@@ -2575,7 +2601,11 @@ function(obj, args) {
 
   if (!this.connecting) {
     debug('Connecting datasource %s to connector %s', this.name, this.connector.name);
-    this.connect();
+    // When no callback is provided tyo connect(), it returns a promise.
```
hacksparrow: tyo -> to

bajtos: Good catch! Fixed.

bajtos:
@slnode test please

PR #1759, opened 2019-07-18
feat: add DataSource.deleteAllModels() API

bajtos:
When writing tests, for performance reasons we often want to reuse the same data-source instance for many tests suites. At the same time, we want to keep such test suites independent and allow them to reuse the same model name for different model classes.

Juggler does support redefinition of a model with the same name.

This change is adding a new API called `dataSource.reset()` that allows tests to remove all old models before creating new ones (typically from a `before` hook).

#### Related issues

This is very loosely related to https://github.com/strongloop/loopback-next/pull/3387.

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -845,6 +845,16 @@ DataSource.prototype.deleteModelByName = function(modelName) {
 };
 
 /**
+ * Remove all models from the registry, but keep the connector instance
+ * (including the pool of database connections).
+ */
+DataSource.prototype.deleteAllModels = function() {
+  for (const m in this.modelBuilder.models) {
+    this.deleteModelByName(m);
+  }
+};
+
+/**
  * Mixin DataAccessObject methods.
  *
  * @param {Function} ModelCtor The model constructor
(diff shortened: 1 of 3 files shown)


bajtos:
> Do we also want to update types/datasource.d.ts file as well for this API?

Good catch! Of course we want to update typings too.

bajtos:
Amended the commit to include changes in `types/datasource.d.ts` too.

lib/datasource.js
```diff
@@ -844,6 +844,16 @@ DataSource.prototype.deleteModelByName = function(modelName) {
   delete this.connector._models[modelName];
 };
 
+/**
+ * Remove all models from the registry, but keep the connector instance
+ * (including the pool of database connections).
+ */
+DataSource.prototype.reset = function() {
```
raymondfeng: Let's rename it to `deleteAllModels` as `reset` hints the connection pool will be reset too.

bajtos: Yeah, I was aware of that possibility. In the future, we may want to add more steps than just deleting all models, that's why I wanted to use a more generic name than `deleteAllModels`. 

Let's follow YAGNI principle and don't worry about things we may need in the future.

I am going to rename the method to `deleteAllModels` as suggested.

bajtos:
@raymondfeng LGTY now?

PR #1763, opened 2019-07-25
Fix coercion of PK value in `replaceById` method

bajtos:
Before this change, when both the PK value (`id`) and the `data` object were provided as plain-data values (e.g. as received in a JSON request), and the connector was using a complex PK type (e.g. `ObjectID` in MongoDB), then `replaceById` operation was printing confusing warnings:

    WARNING: id property cannot be changed from 5d39775a59f5f541513c5e05
        to 5d39775a59f5f541513c5e05 for model:Post
        in 'before save' operation hook

    WARNING: id property cannot be changed from 5d39775a59f5f541513c5e05
        to 5d39775a59f5f541513c5e05 for model:Post
        in 'loaded' operation hook

This commit fixes the problem by applying the same type coercion on the PK value (`id`) as has been applied by the model constructor on the PK property (`data.id`).

#### Related issues

I discovered this problem while working on the spike for Inclusion of related models (see https://github.com/strongloop/loopback-next/pull/3387), but the relation is not really relevant.

### Checklist

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -1576,6 +1576,33 @@ describe('manipulation', function() {
         done();
       });
     });
+
+    it('correctly coerces the PK value', async () => {
+      const created = await Post.create({
+        title: 'a title',
+        content: 'a content',
+      });
+
+      // Emulate what happens when model instance is received by REST API clients
+      const data = JSON.parse(JSON.stringify(created));
+
+      // Modify some of the data
+      data.title = 'Draft';
+
+      // Call replaceById to modify the database record
+      await Post.replaceById(data.id, data);
+
+      // Verify what has been stored
+      const found = await Post.findById(data.id);
+      found.toObject().should.eql({
+        id: created.id,
+        title: 'Draft',
+        content: 'a content',
+      });
+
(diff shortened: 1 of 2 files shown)


test/manipulation.test.js
```diff
@@ -1576,6 +1576,34 @@ describe('manipulation', function() {
...
+      // Verify what has been stored
+      const found = await Post.findById(data.id);
+      found.toObject().should.eql({
+        id: created.id,
+        title: 'Draft',
+        content: undefined,
+      });
+
+      // Verify that no warnings were triggered
+      Object.keys(Post._warned).should.be.empty();
```
bajtos: Without my fix in place, this checks passes when we run the tests in juggler, but fails against the MongoDB connector. I have verified that myself by running the MongoDB test suite against `npm link`-ed version of juggler.

bajtos:
> [cis-jenkins] downstream: loopback-connector-mysql@master — Failed! (f1fa976)
>  [cis-jenkins] downstream: loopback-connector-oracle@master — Failed

I checked the console logs and these downstream failures do not seem to be caused by my changes.

PR #1767, opened 2019-08-01
fix: prevent max listeners warning

emonddr:
### Description

If establishing a database connection is slow and database migration runs and there are many models, sql operations are queued up and this leads to the node.js max emitters exceeded warning.

A default value for max emitters has now been introduced, and it can also be configured in datasources.json.

#### Related issues

- connect to https://github.com/strongloop/loopback-next/issues/2198

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -195,6 +195,11 @@ util.inherits(DataSource, EventEmitter);
 DataSource.DataAccessObject = DataAccessObject;
 
 /**
+ * Global maximum number of event listeners
+ */
+DataSource.DEFAULT_MAX_OFFLINE_REQUESTS = 16;
+
+/**
  * Set up the connector instance for backward compatibility with JugglingDB schema/adapter
  * @private
  */
@@ -207,6 +212,10 @@ DataSource.prototype._setupConnector = function() {
       this.connector.dataSource = this;
     }
     const dataSource = this;
+
+    // Set max listeners to a default/configured value
+    dataSource.setMaxListeners(dataSource.getMaxOfflineRequests());
+
     this.connector.log = function(query, start) {
       dataSource.log(query, start);
     };
@@ -2697,6 +2706,27 @@ DataSource.prototype.execute = function(command, args = [], options = {}) {
  */
 DataSource.prototype.beginTransaction = function(options) {
   return Transaction.begin(this.connector, options);
+};
+
+/**
+ * Get the maximum number of event listeners
+ */
+DataSource.prototype.getMaxOfflineRequests = function() {
+  // Set max listeners to a default value
+  // Override this default value with a datasource setting
+  // 'maxOfflineRequests' from an application's datasources.json
+
+  let maxOfflineRequests = DataSource.DEFAULT_MAX_OFFLINE_REQUESTS;
+  if (
+    this.settings &&
+    this.settings.maxOfflineRequests
+  ) {
+    if (typeof this.settings.maxOfflineRequests !== 'number')
+      throw new Error('maxOfflineRequests must be a number');
+
+    maxOfflineRequests = this.settings.maxOfflineRequests;
+  }
+  return maxOfflineRequests;
 };
 
 /*! The hidden property call is too expensive so it is not used that much
--- a/test/datasource.test.js
+++ b/test/datasource.test.js
@@ -545,6 +545,28 @@ describe('DataSource', function() {
       ds.connector.should.equal(connector);
     });
   });
+
+  describe('getMaxOfflineRequests', () => {
+    let ds;
+    beforeEach(() => ds = new DataSource('ds', {connector: 'memory'}));
+
+    it('sets the default maximum number of event listeners to 16', () => {
+      ds.getMaxOfflineRequests().should.be.eql(16);
+    });
+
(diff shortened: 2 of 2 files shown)


lib/datasource.js
```diff
@@ -2699,6 +2704,22 @@ DataSource.prototype.beginTransaction = function(options) {
...
 
+/**
+ * Get the maximum number of event listeners
+ */
+DataSource.prototype.getMaxOfflineRequests = function() {
+  // Set max listeners to a default value
+  // Override this default value with a datasource setting
+  // 'maxOfflineRequests' from an application's datasources.json
+  // solves bug https://github.com/strongloop/loopback-next/issues/2198
+  let maxOfflineRequests = 16;
```
raymondfeng: We can expose `DEFAULT_MAX_OFFLINE_REQUESTS` as a property of `DataSource` class itself so that the global default can be changed as well.

lib/datasource.js
```diff
@@ -2699,6 +2704,22 @@ DataSource.prototype.beginTransaction = function(options) {
...
+  if (this.settings && this.settings.maxOfflineRequests) {
```
raymondfeng: `typeof his.settings.maxOfflineRequests === 'number'`?

lib/datasource.js
```diff
@@ -194,6 +194,11 @@ util.inherits(DataSource, EventEmitter);
 // allow child classes to supply a data access object
 DataSource.DataAccessObject = DataAccessObject;
 
+/**
+ * Global maximum number of event listeners
+ */
+DataSource.prototype.DEFAULT_MAX_OFFLINE_REQUESTS = 16;
```
raymondfeng: You can set at class level and reference it as `DataSource.DEFAULT_MAX_OFFLINE_REQUESTS`.

```js
DataSource.DEFAULT_MAX_OFFLINE_REQUESTS = 16;
```

test/datasource.test.js
```diff
@@ -545,6 +545,30 @@ describe('DataSource', function() {
...
+      ds.getMaxOfflineRequests().should.be.eql(16);
+    });
+
+    it('uses provided number of listeners if it is greater than 16', () => {
+      ds.settings.maxOfflineRequests = 17;
+      ds.getMaxOfflineRequests().should.be.eql(17);
+    });
+
+    it('uses default if a non-number is provided for the max number of listeners', () => {
+      ds.settings.maxOfflineRequests = '17';
```
raymondfeng: I think we should report it as an error.

test/datasource.test.js
```diff
@@ -545,6 +545,30 @@ describe('DataSource', function() {
...
+    });
+
+    it('uses the default max number of event listeners if the provided one is less than 16', () => {
+      ds.settings.maxOfflineRequests = 15;
+      ds.getMaxOfflineRequests().should.be.eql(16);
+    });
+
+    it('uses provided number of listeners if it is greater than 16', () => {
+      ds.settings.maxOfflineRequests = 17;
+      ds.getMaxOfflineRequests().should.be.eql(17);
```
raymondfeng: The `settings.maxOfflineRequests` should override the default, IMO.

emonddr: @raymondfeng , sure...np.

lib/datasource.js
nabdelgadir: I tried this in [this commit](https://github.com/strongloop/loopback-datasource-juggler/pull/1767/commits/b14875da2a24b81821135075d4753d5e1092bef5) but now I can only access it through `this.juggler.DataSource.DEFAULT_MAX_OFFLINE_REQUESTS`/`this.juggler.Schema.DEFAULT_MAX_OFFLINE_REQUESTS` instead of `this.DEFAULT_MAX_OFFLINE_REQUESTS`. I think we should keep it with `prototype`, but what do you think?

raymondfeng: You should be able to access the setting as `DataSource.DEFAULT_MAX_OFFLINE_REQUESTS`. What's your concern?

nabdelgadir: So we don't use `this.DEFAULT_MAX_OFFLINE_REQUESTS`, we use `DataSource.DEFAULT_MAX_OFFLINE_REQUESTS`? I thought `this` was equivalent to `DataSource` in this file, which is where my confusion was stemming from. I understand now though, thanks!

test/datasource.test.js
```diff
@@ -545,6 +545,33 @@ describe('DataSource', function() {
...
+    it('sets the default maximum number of event listeners to 16', () => {
+      ds.getMaxOfflineRequests().should.be.eql(16);
+    });
+
+    it('uses the provided max number of listeners even if it is less than 16', () => {
+      ds.settings.maxOfflineRequests = 15;
+      ds.getMaxOfflineRequests().should.be.eql(15);
+    });
+
+    it('uses provided number of listeners if it is greater than 16', () => {
```
jannyHou: nitpick: I think the value of `settings.maxOfflineRequests` doesn't matter, so 15 and 17 looks like a dup test to me. (or did I miss something?)

nabdelgadir: It was because before we were only assigning it if it was greater, but now that that's removed I can remove this test, thanks!

PR #1768, opened 2019-08-03
Add 'false' as expected for defaults value

loryk:
### Description
Due to possible set default value for boolean types, fix test to check result value

#### Related issues
- connect to strongloop/loopback-connector-mysql#380

### DIFF
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -983,6 +983,7 @@ describe('basic-querying', function() {
           should(found.toObject().active).be.oneOf([
             undefined, // databases supporting `undefined` value
             null, // databases representing `undefined` as `null`
+            false, // database supports default for boolean
           ]);
         });
     });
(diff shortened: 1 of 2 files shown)


slnode:

test/basic-querying.test.js
```diff
@@ -983,6 +983,7 @@ describe('basic-querying', function() {
           should(found.toObject().active).be.oneOf([
             undefined, // databases supporting `undefined` value
             null, // databases representing `undefined` as `null`
+            false, // database supports default for boolean
```
bajtos: I am afraid this addition is breaking the test scenario and rendering the test useless.

If we have a connector that's treating a property with a `default` value set as if it was non-nullable, then I am proposing to apply the following changes:

1. On line 974, remove `default: false`. The `active` property should be declared as optional with no default value.
2. After calling `db.autoupdate('Player')`, redefine the `active` property and add `default: false`.

I think such setup should be more robust and also more closely aligned with the scenario being tested.

PR #1769, opened 2019-08-07
Add default and description in db specific column properties

thiagolinsx:
### Description

Add these two properties if the connector returns it.

### Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -1656,6 +1656,8 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
         dataPrecision: item.dataPrecision,
         dataScale: item.dataScale,
         nullable: item.nullable,
+        description: item.description,
+        default: item.default,
       };
       // merge connector-specific properties
       if (item[dbType]) {


slnode:

lib/datasource.js
```diff
@@ -1656,6 +1656,8 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
         dataPrecision: item.dataPrecision,
         dataScale: item.dataScale,
         nullable: item.nullable,
+        description: item.description,
+        default: item.default,
```
bajtos: Are you sure these two new items should be stored in db-specific metadata?

IMO, the `description` is not specific to the database being used and should be stored in `schema.properties[propName].description`, so that tooling like loopback-swagger or LB4' JSON/OpenAPI schema builders can use it.

I feel the same argument can be made for the `default` field too, but I am not so sure here.

@raymondfeng thoughts?

bajtos:
@thiagolinsx We also need you to sign our Contributor License Agreement, see https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler

Last but not least, the CI builds are failing, make sure that `npm` test is passing on your machine! From what I can tell, the failing test is actually showing you the place where to modify the test suite to verify your changes too.

PR #1770, opened 2019-08-09
feat: add applyDefaultOnWrites property [3.x]

hacksparrow:
### Description

Adds the ability to ignore writing default values to the database.

#### Related issues

- connect to https://github.com/strongloop/loopback-connector-mongodb/issues/534

### Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -431,6 +431,8 @@ DataAccessObject.create = function(data, options, cb) {
           });
         }
 
+        val = applyDefaultsOnWrites(val, Model.definition);
+
         context = {
           Model: Model,
           data: val,
@@ -451,6 +453,19 @@ DataAccessObject.create = function(data, options, cb) {
 
   return cb.promise;
 };
+
+// Implementation of applyDefaultOnWrites property
+function applyDefaultsOnWrites(obj, modelDefinition) {
+  for (const key in modelDefinition.properties) {
+    const prop = modelDefinition.properties[key];
+    if ('applyDefaultOnWrites' in prop && !prop.applyDefaultOnWrites &&
+      prop.default !== undefined && prop.default === obj[key]) {
+      delete obj[key];
+    }
+  }
+
+  return obj;
+}
 
 function stillConnecting(dataSource, obj, args) {
   if (typeof args[args.length - 1] === 'function') {
--- a/test/defaults.test.js
+++ b/test/defaults.test.js
@@ -76,4 +76,29 @@ describe('defaults', function() {
       });
     });
   });
+
+  context('applyDefaultOnWrites', function() {
+    it('does not affect default behavior when not set', async () => {
+      const Apple = db.define('Apple', {
+        color: {type: String, default: 'red'},
+        taste: {type: String, default: 'sweet'},
+      }, {applyDefaultsOnReads: false});
+
+      const apple = await Apple.create();
+      apple.color.should.equal('red');
+      apple.taste.should.equal('sweet');
+    });
+
+    it('removes the property when set to `false`', async () => {
+      const Apple = db.define('Apple', {
+        color: {type: String, default: 'red', applyDefaultOnWrites: false},
+        taste: {type: String, default: 'sweet'},
+      }, {applyDefaultsOnReads: false});
+
+      const apple = await Apple.create({color: 'red', taste: 'sweet'});
+      const found = await Apple.findById(apple.id);
+      should(found.color).be.undefined();
+      found.taste.should.equal('sweet');
+    });
+  });
 });


lib/dao.js
```diff
@@ -452,6 +454,21 @@ DataAccessObject.create = function(data, options, cb) {
   return cb.promise;
 };
 
+// Implementation of defaultValueHandling property
+function applyDefaultValueHandling(obj, modelDefinition) {
+  Object.keys(modelDefinition.properties).forEach(function(key) {
```
raymondfeng: A `for` loop will be more efficient.

lib/dao.js
```diff
@@ -452,6 +454,21 @@ DataAccessObject.create = function(data, options, cb) {
...
+    const prop = modelDefinition.properties[key];
+    if (prop.default &&
+      prop.default === obj[key] &&
+      prop.defaultValueHandling &&
+      prop.defaultValueHandling === 'ignore') {
+      delete obj[key];
```
raymondfeng: Can it be simplified as the following?
```js
if (prop.defaultValueHandling === 'ignore' && prop.default !== undefined && prop.default === obj[key]) {
   delete obj[key];
}
```

test/defaults.test.js
```diff
@@ -76,4 +76,34 @@ describe('defaults', function() {
...
+        apple.taste.should.equal('sweet');
+        done();
+      });
+    });
+
+    it('if set to "ignore", should remove the property', function(done) {
+      const Apple = db.define('Apple', {
+        color: {type: String, default: 'red', defaultValueHandling: 'ignore'},
+        taste: {type: String, default: 'sweet'},
+      }, {applyDefaultsOnReads: false});
```
raymondfeng: Should we use the same name `applyDefaultsOnReads` at property level instead of introducing `defaultValueHandling`? For example:
```
const Apple = db.define('Apple', {
        color: {type: String, default: 'red', applyDefaultsOnWrites: false},
        taste: {type: String, default: 'sweet'},
      }, {applyDefaultsOnReads: false});
```

hacksparrow: `applyDefaultsOnReads` for reading, this feature is related to writing. So maybe `applyDefaultsOnWrite`?

hacksparrow:
@raymondfeng I have applied your suggestions. PTAL.

test/defaults.test.js
```diff
@@ -76,4 +76,34 @@ describe('defaults', function() {
       });
     });
   });
+
+  context('applyDefaultOnWrites', function() {
+    it('if not set, should not affect default behavior', function(done) {
```
bajtos: - Our test names are usually following the convention "subject, verb/effect, condition".
- We use arrow functions in new tests and async/await.

```suggestion
    it('does not affect default behavior when not set', async () => {
```

test/defaults.test.js
```diff
@@ -76,4 +76,34 @@ describe('defaults', function() {
...
+      const Apple = db.define('Apple', {
+        color: {type: String, default: 'red'},
+        taste: {type: String, default: 'sweet'},
+      }, {applyDefaultsOnReads: false});
+
+      Apple.create(function(err, apple) {
```
bajtos:
```suggestion
      const apple = await Apple.create();
```

test/defaults.test.js
```diff
@@ -76,4 +76,34 @@ describe('defaults', function() {
...
+        apple.color.should.equal('red');
+        apple.taste.should.equal('sweet');
+        done();
+      });
+    });
+
+    it('if set to `false`, should remove the property', function(done) {
```
bajtos: Similarly here:

```suggestion
    it('removes the property when set to `false`', async () => {
```

test/defaults.test.js
```diff
@@ -76,4 +76,34 @@ describe('defaults', function() {
...
+      const Apple = db.define('Apple', {
+        color: {type: String, default: 'red', applyDefaultOnWrites: false},
+        taste: {type: String, default: 'sweet'},
+      }, {applyDefaultsOnReads: false});
+
+      Apple.create({color: 'red', taste: 'sweet'}, function(err, result) {
```
bajtos: You forgot to handle `err`.

Please use `await Apple.create(...)` instead.

hacksparrow:
Thanks for the feedback @bajtos. Applied all of them.

What do we do about the downstream errors?

dhmlau:
Not sure about `loopback`. For `loopback-connector-oracle`, we have a ticket to track that: https://github.com/strongloop/loopback-connector-oracle/issues/183.

bajtos:
> What do we do about the downstream errors?

The loopback build failed because a network error while downloading Node and/or dependencies.

```
06:56:20 Now using node v8.16.1 (npm v6.4.1)
06:56:20 Installing default global packages from /home/jenkins/.nvm/default-packages...
06:57:30 npm ERR! code ECONNREFUSED
06:57:30 npm ERR! errno ECONNREFUSED
06:57:30 npm ERR! FetchError: request to http://10.91.203.23:48584/semver failed, reason: connect ECONNREFUSED 10.91.203.23:48584
```

Can you try to re-run the tests?

test/defaults.test.js
```diff
@@ -76,4 +76,29 @@ describe('defaults', function() {
...
+
+    it('removes the property when set to `false`', async () => {
+      const Apple = db.define('Apple', {
+        color: {type: String, default: 'red', applyDefaultOnWrites: false},
+        taste: {type: String, default: 'sweet'},
+      }, {applyDefaultsOnReads: false});
+
+      const apple = await Apple.create({color: 'red', taste: 'sweet'});
+      const found = await Apple.findById(apple.id);
+      (found.color === undefined).should.be.ok();
```
bajtos: I think the following may produce a better assertion error message:

```suggestion
      should(found.color).be.undefined()
```

dhmlau:
@slnode test please

PR #1771, opened 2019-08-12
chore: remove irrelevant test

hacksparrow:
`db.beginTransaction` does not return a promise. Why is the code testing for promise?

### DIFF
--- a/test/transaction.test.js
+++ b/test/transaction.test.js
@@ -71,13 +71,6 @@ describe('Transactions on test connector without execute()', () => {
     }, done);
   });
 
-  it('beginTransaction returns a transaction', async () => {
-    const promise = db.beginTransaction(Transaction.READ_UNCOMMITTED);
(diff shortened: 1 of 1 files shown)


dhmlau:
Looking through the commit history, I think this test case was introduced relatively recent. https://github.com/strongloop/loopback-datasource-juggler/commit/1ed385e393375cddac7529e683fb2cca2ed6afd2 by @b-admike @bajtos.

dhmlau:
@b-admike @bajtos, could you please address @hacksparrow 's question?  Since the test was added recently for the transaction work, i don't think it's irrelevant.  Thanks.

raymondfeng:
@hacksparrow IIRC, `db.beginTransaction()` should return a `Promise` now.

hacksparrow:
Thanks @raymondfeng. All tests pass now.

PR #1775, opened 2019-09-03
chore: reduce identifier length [4.x]

hacksparrow:
Existing identifier length exceeds Oracle DB's limit.

#### Related issues

https://github.com/strongloop/loopback-connector-oracle/issues/183


hacksparrow:
> BTW the Oracle downstream CI build has failed, is that expected?

Yes. There are other test failures, I will be going through them and fixing them.

PR #1782, opened 2019-10-14
Make sure invalid RegExp string is properly sanitized in query

### LINKED ISSUE #1781: Unsanitized RegExp special content crashes application server
## Steps to reproduce

Just issue a query using any of the availalbe regexp operators: **like, nlike, ilike, nilike, regexp** .
For example lets say we have a model **Person** as follows:

```json
name: 'string',
lastName: 'string',
description: 'string'
```
If we issue a query of the form `Person.find({ where: { name: '[' } });` the application will crash

## Current Behavior

Issuing an improper query (also form remote API request) will result i an application crash with the Error **SyntaxError: Invalid regular expression: /[/: Unterminated character class`**

## Expected Behavior

The query should be sanitized first in order to be able to use special characters such as `[`

## Additional information

linux x64 10.15.3
loopback-datasource-juggler@4.13.0 /home/mitsos/Documents/Projects/loopback-datasource-juggler
├── loopback-connector@4.9.0

## Related Issues

Same issue here: https://github.com/strongloop/loopback-datasource-juggler/issues/1570
It was closed as  stale, but bug still persists

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

mitsos1os:
Fix the case where an invalid RegExp string is queried on a model. Instead of crashing the app,
we first check if the query can be successfully converted to RegExp and if not, it is sanitized.

Fixes #1781 

## Checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -29,6 +29,7 @@ exports.idName = idName;
 exports.rankArrayElements = rankArrayElements;
 exports.idsHaveDuplicates = idsHaveDuplicates;
 exports.isClass = isClass;
+exports.escapeRegExp = escapeRegExp;
 
 const g = require('strong-globalize')();
 const traverse = require('traverse');
@@ -318,6 +319,36 @@ function isProhibited(key, prohibitedKeys) {
 }
 
 /**
+ * Accept an operator key and return whether it is used for a regular expression query or not
+ * @param {string} operator
+ * @returns {boolean}
+ */
+function isRegExpOperator(operator) {
+  return ['like', 'nlike', 'ilike', 'nilike', 'regexp'].includes(operator);
+}
+
+/**
+ * Accept a RegExp string and make sure that any special characters for RegExp are escaped in case they
+ * create an invalid Regexp
+ * @param {string} str
+ * @returns {string}
+ */
+function escapeRegExp(str) {
+  assert.strictEqual(typeof str, 'string', 'String required for regexp escaping');
+  try {
+    new RegExp(str); // try to parse string as regexp
+    return str;
+  } catch (unused) {
+    console.warn(
+      'Auto-escaping invalid RegExp value %j supplied by the caller. ' +
+      'Please note this behavior may change in the future.',
+      str
+    );
+    return str.replace(/[\-\[\]\/\{\}\(\)\+\?\.\\\^\$\|]/g, '\\$&');
+  }
+}
+
+/**
  * Sanitize the query object
  * @param query {object} The query object
  * @param options
@@ -390,6 +421,10 @@ function sanitizeQuery(query, options) {
       // This object is not a plain object
       this.update(x, true); // Stop navigating into this object
       return x;
+    }
+
+    if (isRegExpOperator(this.key) && typeof x === 'string') { // we have regexp supporting operator and string to escape
+      return escapeRegExp(x);
     }
 
     return x;
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -23,6 +23,7 @@ describe('basic-querying', function() {
       birthday: {type: Date, index: true},
       role: {type: String, index: true},
       order: {type: Number, index: true, sort: true},
+      tag: {type: String, index: true},
       vip: {type: Boolean},
       address: {
         street: String,
@@ -591,6 +592,12 @@ describe('basic-querying', function() {
             });
         });
 
+      it('should sanitize invalid usage of like', async () => {
+        const users = await User.find({where: {tag: {like: '['}}});
+        users.should.have.length(1);
+        users[0].should.have.property('name', 'John Lennon');
+      });
+
       it('should support "like" that is not satisfied',
         function(done) {
           User.find({where: {name: {like: 'Bob'}}},
@@ -615,6 +622,11 @@ describe('basic-querying', function() {
           users.length.should.equal(0);
           done();
         });
+      });
+
+      it('should properly sanitize invalid ilike filter', async () => {
+        const users = await User.find({where: {name: {ilike: '['}}});
+        users.should.be.empty();
       });
     });
 
@@ -825,7 +837,7 @@ describe('basic-querying', function() {
 
       sample({name: true}).expect(['name']);
       sample({name: false}).expect([
-        'id', 'seq', 'email', 'role', 'order', 'birthday', 'vip', 'address', 'friends', 'addressLoc',
+        'id', 'seq', 'email', 'role', 'order', 'birthday', 'vip', 'address', 'friends', 'addressLoc', 'tag',
       ]);
       sample({name: false, id: true}).expect(['id']);
       sample({id: true}).expect(['id']);
@@ -1311,6 +1323,7 @@ function seed(done) {
       birthday: new Date('1980-12-08'),
       order: 2,
       vip: true,
+      tag: '[singer]',
       address: {
         street: '123 A St',
         city: 'San Jose',
--- a/test/memory.test.js
+++ b/test/memory.test.js
@@ -201,6 +201,7 @@ describe('Memory connector', function() {
       birthday: {type: Date, index: true},
       role: {type: String, index: true},
       order: {type: Number, index: true, sort: true},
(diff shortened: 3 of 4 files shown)


lib/utils.js
```diff
@@ -317,6 +318,32 @@ function isProhibited(key, prohibitedKeys) {
   return false;
 }
 
+/**
+ * Accept an operator key and return whether it is used for a regular expression query or not
+ * @param {string} operator
+ * @returns {boolean}
+ */
+function isRegExpOperator(operator) {
```
emonddr: what about other operators?

https://loopback.io/doc/en/lb4/apidocs.repository.operators.html

image

test/basic-querying.test.js
```diff
@@ -591,6 +592,12 @@ describe('basic-querying', function() {
             });
         });
 
+      it('should sanitize invalid usage of like', async () => {
```
emonddr: Why should an invalid usage of find with '[' return an actual result?

emonddr: I am confused because another test below returns `empty`. This is what I would expect here as well.

e.g.

image

mitsos1os:
Hi @bajtos and thank you for your quick reactions to the pull request. 

I understand your expressed concern about functionality.

To be sincere, this has come to our attention because we are using a web app which in a user view provides some DB lists.
In that view, there is a field for *easy-search* that accepts user input and searches all available list fields in the database using **like** operator to be able to partial match field contents.
This is where we caught the bug. There was a list element with field content starting with **[type]**. So when we tried inserting `[` in the **easy-search** input, the query sent was `like: '['` which caused the aforementioned application crash in the issue report. 

In our case, it is a matter of user-spec if we want to enable RegExp special characters from **easy-search**, or provide an easier way of direct partial-searching the user input by escaping all user input. I leave this up to you to decide what you want me to implement.

What **needs** to be done, is for the application not to crash.
So we could just sent a desired message and let the actual handling in the user application

lib/utils.js
raymondfeng: @emonddr This is for regexp related operators only.

mitsos1os: @emonddr as @raymondfeng said this targets the crash caused by Invalid regular expression operators

test/basic-querying.test.js
mitsos1os: These tests were made under the context described in my reponse to @bajtos comment.  I approached it by choosing to interpret an invalid RegExp string as a literal (escaped) one, which seraches for the actual characters provided.

The first test returns the result because of the regex search on new **tag** field which matches the `[singer]` content, while the `ilike` is on the name field which does not match.

If you guys decide that you just simply want to return the error as discussed with @bajtos then I will have to change these tests.

mitsos1os:
any news with the pull request? Should I proceed with the forementioned functionality?

lib/utils.js
```diff
@@ -317,6 +318,32 @@ function isProhibited(key, prohibitedKeys) {
...
+ * @param {string} str
+ * @returns {string}
+ */
+function escapeRegExp(str) {
+  assert.strictEqual(typeof str, 'string', 'String required for regexp escaping');
+  let success = true;
+  try {
+    new RegExp(str); // try to parse string as regexp
+  } catch (unused) {
+    success = false;
```
bajtos: While I am not very happy about the current implementation, I also don't want to block fixing the source of crashes. 

To make it easier for us to change this behavior in the future and start rejecting requests with unescaped regexp values, I am proposing to print a warning here. For example:

```js
  try {
     new RegExp(str); // try to parse string as regexp
   } catch (unused) {
     console.warn(
       'Auto-escaping invalid RegExp value %j supplied by the caller. ' + 
         'Please note this behavior may change in the future.', 
       str
     );
     success = false;
    }
```

While making changes in this function, can we simplify the implementation as follows?

```js
function escapeRegExp(str) {
  assert.strictEqual(typeof str, 'string', 'String required for regexp escaping');
  try {
    new RegExp(str); // try to parse string as regexp
    return str;
  } catch (unused) {
    console.warn(/*...*/);
    return str.replace(/[\-\[\]\/\{\}\(\)\+\?\.\\\^\$\|]/g, '\\$&');
  }
}
```

test/basic-querying.test.js
```diff
@@ -616,6 +623,11 @@ describe('basic-querying', function() {
           done();
         });
       });
+
+      it('should properly sanitize invalid ilike filter', async () => {
+        const users = await User.find({where: {name: {ilike: '['}}});
+        users.should.be.empty;
```
bajtos:
```suggestion
        users.should.be.empty();
```

lib/utils.js
mitsos1os: I can change it right now, because we also want this fix to land.
Would you prefer, just catching the error, and returning it as a SyntaxError: Invalid regular expression to the request?

mitsos1os: well? 😃

dhmlau:
@slnode test please

lib/utils.js
jannyHou: I was following the chat here and +1 for @bajtos 's simplified implementation:

```js
function escapeRegExp(str) {
  assert.strictEqual(typeof str, 'string', 'String required for regexp escaping');
  try {
    new RegExp(str); // try to parse string as regexp
    return str;
  } catch (unused) {
    // we can reject with proper message in a follow-up PR
    console.warn(/*...*/);
    return str.replace(/[\-\[\]\/\{\}\(\)\+\?\.\\\^\$\|]/g, '\\$&');
  }
}
```

test/basic-querying.test.js
```diff
@@ -23,6 +23,7 @@ describe('basic-querying', function() {
       birthday: {type: Date, index: true},
       role: {type: String, index: true},
       order: {type: Number, index: true, sort: true},
+      notes: {type: String, index: true},
```
jannyHou: @mitsos1os Wouldn't the property here be `tag` instead of `notes`? The connector's tests fail due to `tag` is not defined in the model(therefore no mapping column in the mysql/mssql schema).

test/memory.test.js
```diff
@@ -201,6 +201,7 @@ describe('Memory connector', function() {
       birthday: {type: Date, index: true},
       role: {type: String, index: true},
       order: {type: Number, index: true, sort: true},
+      notes: {type: String, index: true},
```
jannyHou: ditto, I think the property should be `tag` not `notes`

lib/utils.js
mitsos1os: ok @jannyHou! changing it now!

test/memory.test.js
mitsos1os: Yes you are right! fixed!

test/basic-querying.test.js
mitsos1os: Fixed!

mitsos1os:
@jannyHou @bajtos All changes made! Please check again so we can proceed

dhmlau:
@mitsos1os, could you please squash your commits? Thanks! 

@jannyHou @emonddr, could you please review again?

mitsos1os:
> @mitsos1os, could you please squash your commits? Thanks!
> 
> @jannyHou @emonddr, could you please review again?

@dhmlau done as well!

mitsos1os:
> @mitsos1os, could you please squash your commits? Thanks!
> 
> @jannyHou @emonddr, could you please review again?

@dhmlau done as well!

jannyHou:
@slnode test please

dhmlau:
@mitsos1os, thanks for your contribution.  Your PR has landed! 🎉

dhmlau:
`loopback-datasource-juggler@4.14.0` has been released.

mitsos1os:
> `loopback-datasource-juggler@4.14.0` has been released.

Will this be backported to 3.x as well? We are using it in Loopback v3 currently

dhmlau:
Could you please submit a patch against the 3.x branch?

mitsos1os:
@dhmlau There you go:
https://github.com/strongloop/loopback-datasource-juggler/pull/1785

PR #1801, opened 2019-11-30
Count issue with related models using though model #1795

### LINKED ISSUE #1795: Count issue with related models using though model
## Steps to reproduce

Create a model called Book with the properties id,author,title 
Add a hasMany relation between the default User model and the Book model using a through Model called UserToBook which has the following properties: userId,bookId

The relation is defined in the User model like so.
```json
    "books": {
      "type": "hasMany",
      "model": "Book",
      "foreignKey": "userId",
      "through": "UserToBook",
      "keyThrough": "bookId"
    },
```

## Current Behavior

Querying the user for his books using a where filter works as expected (there is a bug there as well which I will address in a different issue). The reason it works is the `related` function
https://github.com/strongloop/loopback-datasource-juggler/blob/d54d76947724d75a0f46b151c064742c678b0bbe/lib/scope.js#L61
Which, after querying the through model, also queries the related model directly.

When trying to count the number of books of a user, using a where filter (on the author for example), the result is the entire number of books the user has and not the number of books from that author. The reason for that is that the count function
https://github.com/strongloop/loopback-datasource-juggler/blob/d54d76947724d75a0f46b151c064742c678b0bbe/lib/scope.js#L509-L528
Only queries the though model, and not the related model directly.

## Expected Behavior

Return the actual count, taking the filter under account

## Link to reproduction sandbox

Working on it

### LINKED ISSUE #1796: Querying related models make unnecessary requests
## Steps to reproduce

Create a model called Book with the properties id,author,title 
Add a hasMany relation between the default User model and the Book model using a through Model called UserToBook which has the following properties: userId,bookId

The relation is defined in the User model like so.
```json
    "books": {
      "type": "hasMany",
      "model": "Book",
      "foreignKey": "userId",
      "through": "UserToBook",
      "keyThrough": "bookId"
    },
```

## Current Behavior

Querying the user for his books using a where filter performs 2 requests - one to the through model and one to the actual related model. The query to the through model is adding an `include` to the related model, which causes a fetch of a lot of data from the related model. 
The reason this data is needed, is to extract the related model ids, so we can use it to filter out when performing the actual call the related model.

## Expected Behavior

There is no need at all to fetch the related model data in the first call - we can simply get the ids from the through model, and use them

## Link to reproduction sandbox

Working on it

## Additional information

Since I'm using (sadly) loopback 2.x I have added the fix to my own fork - https://github.com/regevbr/loopback-datasource-juggler/blob/2.x/lib/scope.js
You can understand and base the fix based on the changes there

https://github.com/regevbr/loopback-datasource-juggler/blob/2cf8269e41a153d01e4d505af586bad9bb9110dd/lib/scope.js#L100-L138

regevbr:
Fixes #1795 - fixed wrong behaviour of `count`, `findOne`, `updateAll`, `destroyAll` on related models that use through models.

Fixes #1796 - Optimized the bloated query being used on through models when performing `find`, making it more efficient

Added optimization when querying related models using the id in the filter, to make the call to the through model more efficient.

Applied the fix from strongloop/loopback/issues/1076 to the explicit call of `find` on a related model

## Checklist

### DIFF
--- a/test/relations.test.js
+++ b/test/relations.test.js
@@ -605,6 +605,424 @@ describe('relations', function() {
       db.automigrate(['Physician', 'Patient', 'Appointment', 'Address'], done);
     });
 
+    describe('scoped queries', function() {
+      describe('find', function() {
+        it('should allow to use fields on related model', async function() {
+          const physician = await Physician.create();
+          const patient = await physician.patients.create({name: 'a', age: 5});
+          const address = await Address.create({name: 'z'});
+          patient.address(address);
+          await patient.save();
+          const ch = await physician.patients({fields: 'age'});
+          should.exist(ch);
+          ch.should.have.lengthOf(1);
+          ch[0].age.should.eql(5);
+          should.not.exist(ch[0].name);
+        });
+
+        it('should find related model using model id', async function() {
+          const physician = await Physician.create();
+          const patient = await physician.patients.create({name: 'a', age: 5});
+          await physician.patients.create({name: 'b', age: 5});
+          await physician.patients.create({name: 'c', age: 5});
+          const ch = await physician.patients({where: {id: patient.id}});
+          should.exist(ch);
+          ch.should.have.lengthOf(1);
+          ch[0].age.should.eql(5);
+        });
+
+        it('should find related model using model id list', async function() {
+          const physician = await Physician.create();
+          const patient = await physician.patients.create({name: 'a', age: 5});
+          await physician.patients.create({name: 'b', age: 5});
+          await physician.patients.create({name: 'c', age: 5});
+          const ch = await physician.patients({where: {id: {inq: [patient.id]}}});
+          should.exist(ch);
+          ch.should.have.lengthOf(1);
+          ch[0].age.should.eql(5);
+        });
+
+        it('should find scoped record with promises based on related model properties', async function() {
+          const physician = await Physician.create();
+          await physician.patients.create({name: 'a'});
+          await physician.patients.create({name: 'z'});
+          await physician.patients.create({name: 'c'});
+          const patients = await physician.patients({
+            where: {
+              name: 'a',
+            },
+          });
+          should.exist(patients);
+          patients.should.have.lengthOf(1);
+          patients[0].name.should.equal('a');
+        });
+
+        it('should find scoped record with promises based on related model' +
+          ' properties with empty results', async function() {
+          const physician = await Physician.create();
+          await physician.patients.create({name: 'a'});
+          await physician.patients.create({name: 'z'});
+          await physician.patients.create({name: 'c'});
+          const patients = await physician.patients({
+            where: {
+              id: 'bar',
+            },
+          });
+          should.exist(patients);
+          patients.should.have.lengthOf(0);
+        });
+
+        it('should find scoped record with promises based on related model id', async function() {
+          const physician = await Physician.create();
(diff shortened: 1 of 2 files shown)


slnode:

regevbr:
@bajtos can you please review?
I added 4 tests that failed before the fix (TDD style) and fixed all problematic functions - count,findOne,updateAll and destroyAll

regevbr:
@bajtos I combined here the fix of #1796 as well and they are closely related and  they use the same logic.

dhmlau:
@slnode test please

bajtos:
Thank you @regevbr for the pull request. This is a large change, I need to set aside some time to review it in detail. I'll try to get it done by the end of next week, before I leave for vacation.

@raymondfeng can you please take a look too?

test/relations.test.js
```diff
@@ -605,6 +605,864 @@ describe('relations', function() {
       db.automigrate(['Physician', 'Patient', 'Appointment', 'Address'], done);
     });
 
+    describe('scoped queries', function() {
+      describe('find', function() {
+        it('should allow to use fields on related model', function(done) {
+          Physician.create(function(err, physician) {
```
bajtos: Unhandled `err`.

Please use async/await style in new tests, juggler supports only Node.js 8+ version now.

```ts
const physician = await Physician.create();
```

test/relations.test.js
```diff
@@ -605,6 +605,864 @@ describe('relations', function() {
...
+              patients.should.have.lengthOf(0);
+              done();
+            });
+          }
+        });
+
+        it('should find scoped record with promises based on related model id', function(done) {
+          let id;
+          Physician.create()
+            .then(function(physician) {
```
bajtos: Similarly here. Please use `await` instead of Promise/`.then()` style.

```ts
const physician = await Physician.create();
```

test/relations.test.js
bajtos: (The same comment applies to all other places where you are invoking a function or a method and providing a callback.)

test/relations.test.js
bajtos: (The same comment applies to all other places where you are using `.then()` or `.catch()`.)

regevbr:
@bajtos sure I will try to get on it soon, I have a lot on my plate :-)

test/relations.test.js
regevbr: Fixed

test/relations.test.js
regevbr: Fixed

regevbr:
@bajtos I fixed the tests (which actually got me to find bugs I introduced so I fixed them as well)

bajtos:
Thank you @regevbr for the updates, the tests look much better now!

I am afraid I got stuck in other tasks and did not find enough time to review your changes. If nobody from @strongloop/loopback-3x-maintainers and/or @strongloop/loopback-maintainers steps up to review this pull request, then I am afraid you will have to wait until January.

Sorry for the delay 😢

bajtos:
@slnode test please

bajtos:
Sorry @regevbr for the long silence on our side. The pull request is quite involved, I find it difficult to find enough time (and energy) to dive into the changes. Is there any chance to split this patch into smaller pull requests, where each pull request would be a meaningful change on its own? The smaller the patch, the easier it is to review for us.

PR #1805, opened 2019-12-03
fix(security): avoid prototype pollution on model create and update

STRML:
Without this protection, incoming objects from the network can overwrite
the ModelConstructor prototype.

## Checklist

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -63,6 +63,8 @@ function getIdValue(m, data) {
 
 function copyData(from, to) {
   for (var key in from) {
+    // Don't traverse down prototype chain
+    if (!from.hasOwnProperty(key)) continue;
     to[key] = from[key];
   }
 }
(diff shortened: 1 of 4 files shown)


STRML:
@bajtos @raymondfeng  I can't see the failing tests. But this is high priority and your team should ensure this does not affect 3.x and 4.x. We were able to crash real servers with this.

raymondfeng:
I wonder if we can have a middleware/body parser that prevents `__proto__` from being accepted from incoming json requests.

raymondfeng:
@STRML See https://github.com/strongloop/loopback-next/blob/202b2c76b77122654ed81e60afcd2e2a2985de91/packages/rest/src/parse-json.ts

jannyHou:
The 3 failures in 0.12 are not related to this PR, I will fix them in https://github.com/strongloop/loopback-datasource-juggler/pull/1806

PR #1807, opened 2019-12-05
Fix Promise/Callback variants in datasource types

bajtos:
Remove usage of `PromiseOrVoid`, it makes it difficult to consume the API via `await`. For each async method, define multiple variants to correctly describe what's returned when a callback arg was provided (`void`) and what's returned when called with no callback (`Promise<T>`).

See https://github.com/strongloop/loopback-next/pull/4235 for background information.

## Checklist

### DIFF
--- a/types/datasource.d.ts
+++ b/types/datasource.d.ts
@@ -3,7 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {AnyObject, Callback, Options, PromiseOrVoid} from './common';
+import {AnyObject, Callback, Options} from './common';
 import {Connector} from './connector';
 import {
   ModelBaseClass,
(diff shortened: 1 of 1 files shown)


types/datasource.d.ts
```diff
@@ -3,7 +3,7 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {AnyObject, Callback, Options, PromiseOrVoid} from './common';
+import {AnyObject, Callback, Options} from './common';
```
emonddr: Commit message has typo:

> corretly describe 

correctly describe

bajtos: Good catch!

types/datasource.d.ts
```diff
@@ -140,66 +140,156 @@ export declare class DataSource extends EventEmitter {
...
     options?: Options,
-    callback?: Callback<PropertyDefinition[]>,
-  ): PromiseOrVoid<PropertyDefinition[]>;
+  ): Promise<PropertyDefinition[]>;
+  // legacy callback style (no options)
+  discoverForeignKeys(
+    modelName: string,
+    callback: Callback<PropertyDefinition[]>,
+  ): void;
+  // legacy callback style (no options)
```
jannyHou: this should be with options I think

bajtos: Good catch! Will fix.

Issue #1809, opened 2019-12-09
loopback-connector-ibmi needs modifications to some query tests


markdirish:
## Steps to reproduce

Run datasource tests on [loopback-connector-ibmi](https://github.com/strongloop/loopback-connector-ibmi)

## Current Behavior

There are 2 issues with the tests found in `test/basic-querying.test.js`:

1. The test `should support string "gt" that is not satisfied` assumes that the table is encoded in ASCII or a related encoding (lowercase letters come after capital letters). IBM i tables are by default encoded in EBCDIC (lowercase letters come before capital letters). The test as written doesn't pass on IBM i, so need to check the db adapter and adjust the test accordingly.

2. Like many connectors (including `db2`), the IBM i connector does not have GeoPoint support. Currently, GeoPoint tests in the file fail when using `loopback-connector-ibmi`. Need to add `ibmi` to the list of adapters that do not support GeoPoint.

3. When checking for explicit error messages, be sure to check `error.odbcErrors[0].message` instead of `error.message`. The underlying ODBC connection used by this connector can return multiple errors, so `error.message` is usually a generic message saying where the error happened, while the `error.odbcErrors[x].message` is an error directly from the database.

## Expected Behavior

There is nothing wrong with the connector in regards to these tests, just have to make sure the juggler tests are aware of IBM i's limitations. Need to adjust tests to ensure they pass.

## Additional information

These changes are needed for the `loopback-connector-ibmi` to pass all tests. Will make PRs for both v3 (lts) and v4 (current)

## Related Issues

dhmlau:
@markdirish, assigning this to you since you have a PR in progress #1810 .

dhmlau:
@markdirish, are you still pursuing this? Thanks.

PR #1810, opened 2019-12-09
Modify logic around some query tests to fix test failures

markdirish:
Several `basic-querying.test.js` tests fail when running with `loopback-connector-ibmi`. Explanations can be found in issue #1809. This PR fixes the tests on v3 (LTS) branch.

Part of #1809 

## Checklist

### DIFF
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -481,23 +481,21 @@ describe('basic-querying', function() {
       });
 
     it('should support string "gt" that is not satisfied', function(done) {
-      // IBM i tables are coded in EBCDIC by default, need to modify the query
-      // so that it returns no results
-      if (db.adapter.name === 'ibmi') {
-        User.find({where: {name: {'gt': 'XYZ'},
-        }}, function(err, users) {
-          should.not.exist(err);
-          users.should.have.property('length', 0);
-          done();
-        });
-      } else {
-        User.find({where: {name: {'gt': 'xyz'},
-        }}, function(err, users) {
-          should.not.exist(err);
-          users.should.have.property('length', 0);
-          done();
-        });
+      // IBM i tables are coded in EBCDIC by default.
+      // In EBCDIC, capital letters come after lowercase letters (opposite of
+      // ASCII). This test expects no results but assumes ASCII encoding. Need
+      // to have a special case for ibmi, otherwise "'gt: 'xyz'" will return
+      // every name beginning with a capital letter, and the test will fail.
+      let compareString = 'xyz';
+      if (db.adapter.name === 'ibmi') { // i.e. "if EBCDIC encoding"
+        compareString = 'XYZ';
       }
+      User.find({where: {name: {'gt': compareString},
+      }}, function(err, users) {
+        should.not.exist(err);
+        users.should.have.property('length', 0);
+        done();
+      });
     });
 
     bdd.itIf(connectorCapabilities.cloudantCompatible !== false,
--- a/test/basic-querying.test.js
+++ b/test/basic-querying.test.js
@@ -49,7 +49,8 @@ describe('basic-querying', function() {
 
     db = getSchema();
     // connectors that do not support geo-point types
-    connectorCapabilities.geoPoint = (db.adapter.name != 'dashdb') && (db.adapter.name != 'db2') &&
+    connectorCapabilities.geoPoint = (db.adapter.name != 'dashdb') &&
+    (db.adapter.name != 'db2') && (db.adapter.name != 'ibmi') &&
     (db.adapter.name != 'informix') && (db.adapter.name != 'cassandra');
     if (connectorCapabilities.geoPoint) userModelDef.addressLoc = {type: 'GeoPoint'};
     User = db.define('User', userModelDef);
@@ -480,12 +481,23 @@ describe('basic-querying', function() {
       });
 
     it('should support string "gt" that is not satisfied', function(done) {
-      User.find({where: {name: {'gt': 'xyz'},
-      }}, function(err, users) {
-        should.not.exist(err);
-        users.should.have.property('length', 0);
-        done();
-      });
+      // IBM i tables are coded in EBCDIC by default, need to modify the query
+      // so that it returns no results
+      if (db.adapter.name === 'ibmi') {
+        User.find({where: {name: {'gt': 'XYZ'},
+        }}, function(err, users) {
+          should.not.exist(err);
+          users.should.have.property('length', 0);
+          done();
+        });
+      } else {
+        User.find({where: {name: {'gt': 'xyz'},
+        }}, function(err, users) {
+          should.not.exist(err);
+          users.should.have.property('length', 0);
+          done();
+        });
+      }
     });
 
     bdd.itIf(connectorCapabilities.cloudantCompatible !== false,
@@ -653,7 +665,7 @@ describe('basic-querying', function() {
       });
     });
 
-    describe('geo queries', function() {
+    bdd.describeIf(connectorCapabilities.geoPoint, 'geo queries', function() {
       describe('near filter', function() {
         it('supports a basic "near" query', function(done) {
           User.find({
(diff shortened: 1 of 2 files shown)


dhmlau:
@markdirish, thanks for the PR.  Could you please fix the commit linter error? Thanks. 
```
**************************************************
**
**  Linting commit logs
**
**  1 problems found:
**    6dd0640 - Modify logic around some query tests to avoid false: First line should be 50 characters or less (saw 88)
**
**************************************************
```

For details, please see the commit message guidelines: https://loopback.io/doc/en/contrib/git-commit-messages.html.

dhmlau:
And please sign the CLA as well: https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler. Thanks!

markdirish:
> And please sign the CLA as well: https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler. Thanks!

Still trying to get a little guidance on this... kind of a weird situation since I'm an IBM employee, and so are StrongLoop employees(?), so trying to figure out if there is any issues.

dhmlau:
@markdirish, from my understanding, IBM employees still need to sign CLA, but I thought we only need to sign once within the `strongloop` org. Since you're the IBM i connector maintainers, i'm not sure why it's happening.  Do you have concerns in signing this one? 

I remember @rmg explained what's happening in the CLA land, but I cannot recall the details. Maybe @rmg would be able to shed some lights on this.

dhmlau:
@slnode test please

rmg:
No need to _sign_ any CLA, but you do have to _sign in_ to the CLA app so that it has permission to check your org membership, which is then used to bypass the signature requirement.

test/basic-querying.test.js
```diff
@@ -480,12 +481,23 @@ describe('basic-querying', function() {
...
-      User.find({where: {name: {'gt': 'xyz'},
-      }}, function(err, users) {
-        should.not.exist(err);
-        users.should.have.property('length', 0);
-        done();
-      });
+      // IBM i tables are coded in EBCDIC by default, need to modify the query
+      // so that it returns no results
+      if (db.adapter.name === 'ibmi') {
+        User.find({where: {name: {'gt': 'XYZ'},
```
emonddr: @markdirish , any reason why we need to test for 'XYZ' as well as 'xyz' for 'gt'?

The tests for 'lt' don't test for all lower and then all upper case. Just wondering

markdirish: As mentioned in the comment above the test (although it is cryptic looking at it with new eyes), the tables that are created in the Db2 for i DBMS are encoded in EBCDIC by default. In EBCDIC, capital letters come after lowercase letters (have a greater value)

In the tables this test is using, the names begin with a capital, like "Mark". So in ASCII, where capitals come first, the test passes fine (as 'M' has a decimal value of 77, and 'x' has a value of 120, and we are looking for strings > 'xyz'). But in EBCDIC the opposite is true ('M' has a value 212, 'x' has a value of 167), so the `User.find` will return results, and then the test fails (because it expects _no_ results).

The `lt` test as written passes in either scenario, because it is comparing capital letters to capital letters, and A has a lower value than Z in both EBCDIC and ASCII

bajtos: Uh oh, this is tricky. 

> the tables that are created in the Db2 for i DBMS are encoded in EBCDIC by default.

Would it make sense to configure the database (or at least the `User` table) to use ASCII-based encoding instead of the default EBCDIC? I am imagining that with that change we won't need to modify test code. The downside is that we will no longer test integration of LoopBack with EBCDIC-encoded tables - is that a problem? What encoding is typically used by Db2 for i DBMS customers?

If we decide it's better to test against EBCDIC, then I would like the code comment to capture the detailed explanation from https://github.com/strongloop/loopback-datasource-juggler/pull/1810/files#r356624276.

test/manipulation.test.js
```diff
@@ -382,7 +382,11 @@ describe('manipulation', function() {
             if (!err) {
               return done(new Error('Create should have rejected duplicate id.'));
             }
-            err.message.should.match(/duplicate/i);
+            if (db.adapter.name === 'ibmi') {
+              err.odbcErrors[0].message.should.match(/duplicate/i);
```
bajtos: What is `err.message` in this case? I think this test is intentionally checking `err.message` to ensure the framework is reporting consistent errors independently of the connector used. I would expect the connector to normalize the error object from ODBC style (in the case of `ibmi`) to LoopBack style.

Thoughts?

markdirish: `err.message` in this case would be something like `Error executing sql statement`. 

I know Sequelize has their own error class that is implemented by a number of sub classes (So `SequelizeError` is the parent class of `UniqueConstraintError` or `DuplicateKeyError`), and each connector intercepts any error output for the connector, parses the message/SQLState/code, and determines for itself which type of error it threw. That might be something to look into for the future.

For now, I'd say the vast majority of errors coming from the odbc connector contain just one error, so I could capture the error and just manually do something like:

```
// at the bottom of executeSQL... given a callback in invocation
this.pool.query(sql, params, (error, results) => {
  if (error) {
    error = error.odbcErrors[0];
  }
  callback(error, results);
});
```

test/basic-querying.test.js
markdirish: I'd say 99% or greater of IBM i users keep the standard EBCDIC encoding. It has implications for all of the code written over the last 30+ years, so it would be hard to get them to switch now.

I can update the code comment to be a little more descriptive

test/manipulation.test.js
markdirish: This works, will push changes to both PRs to remove the changes to this test.

test/basic-querying.test.js
bajtos: > I'd say 99% or greater of IBM i users keep the standard EBCDIC encoding. It has implications for all of the code written over the last 30+ years, so it would be hard to get them to switch now.

Makes sense.

What if we changed the value compared against from letters to numbers, would that allow us to use the same query for both ASCI and EBCDIC encoding?

If not, then I am fine to accept your current proposal as long as the code comment is updated to be more descriptive.

test/manipulation.test.js
bajtos: > For now, I'd say the vast majority of errors coming from the odbc connector contain just one error, so I could capture the error and just manually do something like:
> (...)
> This works, will push changes to both PRs to remove the changes to this test.

Excellent :+1:

Could you please verify that the stack traces reported for the inner errors (`error.odbcErrors[0]`) are offering enough information about the place where the error originated?

test/basic-querying.test.js
markdirish: Nope! In ASCII, numbers precede all letters, while in EBCDIC, number are after all letters. Looks like there is no set of character that would work for a "gt" that is not satisfied. On the more recent commit, I believe I have a much more descriptive comment (and reduced code duplication)

test/basic-querying.test.js
```diff
@@ -653,7 +663,7 @@ describe('basic-querying', function() {
       });
     });
 
-    describe('geo queries', function() {
+    bdd.describeIf(connectorCapabilities.geoPoint, 'geo queries', function() {
```
bajtos: Hmm, I believe we are intentionally testing geo queries for all connectors, including those that don't provide native GeoPoint & near query support. IIRC, when the connector does not implement near queries, juggler will perform the search in memory.

Take `db2` and `dashdb` for an example. According to the check on the line 52 above, these connectors don't support GeoPoint type, yet they still pass all `describe('geo queries')` tests.

Let's find a way how to allow the `ibmi` connector pass these test too.

Does the connector support object-values in columns? I think that most of our SQL connectors are storing objects (and arrays) as JSON strings in the database columns. Is this perhaps a missing feature of the `ibmi` connector?

test/manipulation.test.js
markdirish: I'm not sure that there is a good stack trace on those errors, unfortunately. The errors are created in N-API, and it doesn't look like I have ensured a stack track is created. I have modified the connector so that it creates a stack trace to the pool.query call, but below that there isn't a good trace. Have added it to my TODOs for the `odbc` package.

dhmlau:
@markdirish, are you still working on this? Thanks.

PR #1813, opened 2019-12-13
feat: added persistDefaultValues

hacksparrow:
Ignores value if it matches default value.

**Use case**: User wants to reduce noise in the database, since the defaults will be filled in by LoopBack or the front-end framework. Makes a big difference if there are thousands of such properties in millions of records.

## Checklist

### DIFF
--- a/lib/model.js
+++ b/lib/model.js
@@ -299,6 +299,10 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
       }
     }
 
+    if (ignoresMatchedDefault(properties[p]) && properties[p].default === propVal) {
+      delete self.__data[p];
+    }
+
     // Set default value using a named function
     if (applyDefaultValues && propVal === undefined) {
       const defn = properties[p].defaultFn;
@@ -361,6 +365,13 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
   }
   this.trigger('initialize');
 };
+
+// Implementation of persistDefaultValues property
+function ignoresMatchedDefault(property) {
+  if (property && property.persistDefaultValues === false) {
+    return true;
+  }
+}
 
 // Helper function for determing the applyDefaultOnWrites value of a property
 function appliesDefaultsOnWrites(property) {
--- a/test/defaults.test.js
+++ b/test/defaults.test.js
@@ -130,4 +130,48 @@ describe('defaults', function() {
       found.qualities.length.should.equal(1);
     });
   });
+
+  context('persistDefaultValues', function() {
+    it('removes property if value matches default', async () => {
+      const Apple = db.define('Apple', {
+        color: {type: String, default: 'red', persistDefaultValues: false},
+        taste: {type: String, default: 'sweet'},
+      }, {applyDefaultsOnReads: false});
+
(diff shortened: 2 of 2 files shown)


lib/model.js
```diff
@@ -299,6 +299,10 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
       }
     }
 
+    if (ignoresMatchedDefault(properties[p]) && properties[p].default === propVal) {
```
jannyHou: Nitpick: The value comparison here works for primitive types like number/string/etc...but I am not sure would it also return true for data in complicated types, like Date or other models.

emonddr: Looks good Yaapa, but I agree with @jannyHou 's question. Is there recursive code that handles nested object properties so that all properties are investigated?

hacksparrow:
@emonddr `ignoresMatchedDefault()` gets called in inside a loop recursing through the properties. There are two tests to check if it works inside objects and arrays.

lib/model.js
```diff
@@ -362,6 +366,13 @@ ModelBaseClass.prototype._initProperties = function(data, options) {
   this.trigger('initialize');
 };
 
+// Implementation of ignoreMatchedDefault property
+function ignoresMatchedDefault(property) {
+  if (property && property.ignoreMatchedDefault) {
```
bajtos: I find the option name a bit difficult to understand. "ignore" can be interpreted in many different ways.

How about reworking `ignoreMatchedDefault: true` to `persistDefaultValues: false`?

```suggestion
  if (property && property.persistDefaultValues === false) {
```

See also the existing model-level property `persistUndefinedAsNull` that's documented in https://loopback.io/doc/en/lb3/Synchronization.html#enable-change-tracking.

hacksparrow: Model-level  and datasource-level config are good suggestions. We'll still need the ability to exclude some properties from the blanket setting, so property-level configurability is required anyway.

bajtos: Sure. Our usual approach to this problem is to allow the flag to be set at multiple levels, where model-level flag overrides datasource-level settings and propery-level flag overrides model/datasource levels.

test/defaults.test.js
```diff
@@ -130,4 +130,48 @@ describe('defaults', function() {
...
+    it('removes property if value matches default in an array', async () => {
+      const Apple = db.define('Apple', {
+        name: {type: String},
+        qualities: [
+          {color: {type: String, default: 'red', persistDefaultValues: false}},
+          {taste: {type: String, default: 'sweet'}},
+        ],
+      }, {applyDefaultsOnReads: false});
+
+      const apple = await Apple.create({name: 'Honeycrisp', qualities: [{taste: 'sweet'}]});
```
nabdelgadir: I might be misunderstanding, but how is it removing the `color` property if it's never added when it's created?

hacksparrow: We have set a default value in the model (line 166), if we don't manually set the value manually, Juggler sets it for us.

lib/model.js
jannyHou: fair enough 👍

test/defaults.test.js
nabdelgadir: Ah I see 👍 thanks!

PR #1817, opened 2020-01-16
fix: warn about strict cannot be false with relational dbs

agnes512:
Fixes https://github.com/strongloop/loopback-next/issues/4042

## Checklist

### DIFF
--- a/docs/definition-language.md
+++ b/docs/definition-language.md
@@ -178,11 +178,11 @@ There are a set of options to control the model definition.
 
 - strict:
     - true: Only properties defined in the model are accepted. Use this
-    mode if you want to make sure only predefined properties are accepted.
+    mode if you want to make sure only predefined properties are accepted. Relational databases only support this setting.
     - false: The model will be an open model. All properties are accepted,
     including the ones that not predefined with the model. This mode is useful
     if the mobile application just wants to store free form JSON data to
-    a schema-less database such as MongoDB.
+    a schema-less database such as MongoDB. For relational databases, the value will be converted back to true.
     - undefined: Default to false unless the data source is backed by a
     relational database such as Oracle or MySQL.
 
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -826,7 +826,9 @@ DataSource.prototype.define = function defineClass(className, properties, settin
       settings.strict = true;
     }
     if (settings.strict === false) {
-      settings.strict = 'throw';
+      console.warn("WARNING: relational database doesn't support {strict: false} mode." +
+      ` {strict: true} mode will be set for model ${className} instead.`);
+      settings.strict = true;
     }
   }
 


bajtos:
Looks mostly good so far. 

Is it possible to write a test to verify the behavior by spying on `console.warn` methods? I quickly checked juggler and LB3 codebase and could not find any test to show as an example. I also did a quick search on the internet and it looks like intercepting `console` methods is not trivial because Mocha may be using them too. I guess it's ok to make these changes without test coverage.

> I don't know how we warn users in Juggler. 

Using `console.warn` is fine.

Can you please verify that your fix is covering all different ways we have for defining a model? Most importantly, is the warning triggered when models are created before they are attached to a datasource? That's what we are doing in LoopBack 3 applications.

Here is the code creating a new model class via `Model.extend`:
https://github.com/strongloop/loopback/blob/6a7fc529fcbf219d3a2f6b77cdf221ee5c3b1dbf/lib/registry.js#L98-L128

And here we are attaching a model to a datasource - I think that's the point where your new warning should be printed.
https://github.com/strongloop/loopback/blob/6a7fc529fcbf219d3a2f6b77cdf221ee5c3b1dbf/lib/registry.js#L235

agnes512:
The reason why I don't have a test for the changes is because I am thinking to add them in lb4 tests since we can easily test them on real DBs. Thoughts?

> Can you please verify that your fix is covering all different ways we have for defining a model? Most importantly, is the warning triggered when models are created before they are attached to a datasource? That's what we are doing in LoopBack 3 applications.

Good point, I didn't consider is before. I just chasing through and found where we check `isRelational`. 

I did some debug. I didn't create a LB3 app and check it with relational connectors( I don't know how :see_no_evil:). I used LB4 code and MySQL as dataSource (I think it's fine since both of them use connectors and Juggler?).

It shows in LB4 that it creates models with [DataSource.prototype.createModel](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/datasource.js#L804),`strict` i checked in this step, then attach to DataSource with [DataSource.prototype.attach](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/datasource.js#L982). If LB3 follows the same path, does that mean the change matches our expectation?

lib/datasource.js
```diff
@@ -826,7 +826,9 @@ DataSource.prototype.define = function defineClass(className, properties, settin
       settings.strict = true;
     }
     if (settings.strict === false) {
-      settings.strict = 'throw';
+      console.warn('WARNING: relational database doesn\'t support {strict: false} mode.' +
```
nabdelgadir: Use double quotes to avoid escaping: "WARNING: relational database doesn't support {strict: false} mode."

docs/definition-language.md
```diff
@@ -178,11 +178,11 @@ There are a set of options to control the model definition.
 
 - strict:
     - true: Only properties defined in the model are accepted. Use this
-    mode if you want to make sure only predefined properties are accepted.
+    mode if you want to make sure only predefined properties are accepted. Relational databases only support this setting.
```
emonddr: > Relational databases only support this setting

Only relational databases support this setting.

agnes512: NoSQL DBs support this setting as well.

bajtos:
> I did some debug. I didn't create a LB3 app and check it with relational connectors( I don't know how :see_no_evil:). I used LB4 code and MySQL as dataSource (I think it's fine since both of them use connectors and Juggler?).
>
> It shows in LB4 that it creates models with [DataSource.prototype.createModel](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/datasource.js#L804),`strict` i checked in this step, then attach to DataSource with [DataSource.prototype.attach](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/datasource.js#L982). If LB3 follows the same path, does that mean the change matches our expectation?

AFAIK, LB3 uses a different code path for creating models. Creating a LB3 app is easy, see https://loopback.io/lb3/getting-started. You can also use our Coffee Shop example https://github.com/strongloop/loopback-getting-started.

As for adding a relational connector, just install the connector and change the datasource configuration accordingly. In LB3, datasources are configured in `server/datasources.json` file, see https://github.com/strongloop/loopback-getting-started/blob/master/server/datasources.json

AFAICT, the Coffee Shop example is already using MySQL to store the data! I think it should be enough to `git clone` the app, add `strict: false` model setting to [common/models/coffee-shop.json](https://github.com/strongloop/loopback-getting-started/blob/master/common/models/coffee-shop.json) and verify that your warning is printed. 

On the other hand, considering that LB3 is in LTS, maybe it's enough to implement the warning for LB4 only?

agnes512:
> AFAIK, LB3 uses a different code path for creating models. 

You're right. I ran the coffee shop example with MySQL. The change only affects LB4. I think LB3 uses `model-builder.js` instead.
> On the other hand, considering that LB3 is in LTS, maybe it's enough to implement the warning for LB4 only?

I think it's fair to just have the warning and tests for LB4. I will add them once the change gets released.

agnes512:
@slnode test please

agnes512:
@slnode test please

PR #1819, opened 2020-01-27
feat: allows different naming conventions for db  columns for discover

agnes512:
Connects to https://github.com/strongloop/loopback-next/issues/3343

Allows users to keep the property name as the same as the db columns names. It can be set by `options.useDefaultCase` to false.

To close #3343, we need to:
1. modify Juggler (this PR)
2. update LB3 docs ( https://github.com/strongloop/loopback.io/pull/928)
3. update `lb4 discover` CLI and docs ( need this PR to be landed first)

## Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -28,6 +28,8 @@ const g = require('strong-globalize')();
 const juggler = require('..');
 const deprecated = require('depd')('loopback-datasource-juggler');
 const Transaction = require('loopback-connector').Transaction;
+const pascalCase = require('change-case').pascalCase;
+const camelCase = require('change-case').camelCase;
 
 if (process.env.DEBUG === 'loopback') {
   // For back-compatibility
@@ -1450,17 +1452,22 @@ function capitalize(str) {
   return str.charAt(0).toUpperCase() + ((str.length > 1) ? str.slice(1).toLowerCase() : '');
 }
 
-function fromDBName(dbName, camelCase) {
-  if (!dbName) {
-    return dbName;
+/**
+ * Renames db column names with different naming conventions:
+ * camelCase for property names as it's LB default naming convention for properties,
+ * or keep the name the same if needed.
+ *
+ * @param {*} name name defined in database
+ * @param {*} caseFunction optional. A function to convert the name into different case.
+ */
+function fromDBName(name, caseFunction) {
+  if (!name) {
+    return name;
   }
-  const parts = dbName.split(/-|_/);
-  parts[0] = camelCase ? parts[0].toLowerCase() : capitalize(parts[0]);
-
-  for (let i = 1; i < parts.length; i++) {
-    parts[i] = capitalize(parts[i]);
+  if (typeof caseFunction === 'function') {
+    return caseFunction(name);
   }
-  return parts.join('');
+  return name;
 }
 
 /**
@@ -1567,13 +1574,14 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
     cb = options;
     options = {};
   }
-
   cb = cb || utils.createPromiseCallback();
 
   const self = this;
   const dbType = this.connector.name;
 
   let nameMapper;
+  const disableCamelCase = !!options.disableCamelCase;
+
   if (options.nameMapper === null) {
     // No mapping
     nameMapper = function(type, name) {
@@ -1586,11 +1594,19 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
     // Default name mapper
     nameMapper = function mapName(type, name) {
       if (type === 'table' || type === 'model') {
-        return fromDBName(name, false);
+        return fromDBName(name, pascalCase);
       } else if (type == 'fk') {
-        return fromDBName(name + 'Rel', true);
+        if (disableCamelCase) {
+          return fromDBName(name + 'Rel');
+        } else {
+          return fromDBName(name + 'Rel', camelCase);
+        }
       } else {
-        return fromDBName(name, true);
+        if (disableCamelCase) {
+          return fromDBName(name);
+        } else {
+          return fromDBName(name, camelCase);
+        }
       }
     };
   }
@@ -1770,12 +1786,17 @@ DataSource.prototype.discoverSchemasSync = function(modelName, options) {
   if (!columns || columns.length === 0) {
     return [];
   }
+  const disableCamelCase = !!options.disableCamelCase;
 
   const nameMapper = options.nameMapper || function mapName(type, name) {
     if (type === 'table' || type === 'model') {
-      return fromDBName(name, false);
+      return fromDBName(name, pascalCase);
     } else {
-      return fromDBName(name, true);
+      if (disableCamelCase) {
+        return fromDBName(name);
+      } else {
+        return fromDBName(name, camelCase);
+      }
     }
   };
 
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
   },
   "dependencies": {
     "async": "^3.1.0",
+    "change-case": "^4.1.1",
     "debug": "^4.1.0",
     "depd": "^2.0.0",
     "inflection": "^1.6.0",
--- a/test/discovery.test.js
+++ b/test/discovery.test.js
@@ -73,7 +73,7 @@ describe('Memory connector with mocked discovery', function() {
     };
   });
 
-  it('should convert table/column names to camel cases', function(done) {
+  it('should convert table names to pascal cases and column names to camel case', function(done) {
     ds.discoverSchemas('INVENTORY', {}, function(err, schemas) {
       if (err) return done(err);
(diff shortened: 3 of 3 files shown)


lib/datasource.js
```diff
@@ -1450,17 +1452,21 @@ function capitalize(str) {
...
-    return dbName;
-  }
-  const parts = dbName.split(/-|_/);
-  parts[0] = camelCase ? parts[0].toLowerCase() : capitalize(parts[0]);
-
-  for (let i = 1; i < parts.length; i++) {
-    parts[i] = capitalize(parts[i]);
+/**
+ * Renames db column names with differernt naming conventions:
+ * camelCase for property names as it's LB default naming converntion for properties;
```
emonddr: > converntion

convention

lib/datasource.js
```diff
@@ -1450,17 +1452,21 @@ function capitalize(str) {
...
+ * or keep the name the same if needed.
+ *
+ * @param {*} colName column name defined in database
+ * @param {*} defCefcamelCase boolean. Converts the name to camelcase if true as
```
emonddr: > defCefcamelCase

defCamelCase

lib/datasource.js
```diff
@@ -1450,17 +1452,21 @@ function capitalize(str) {
...
-  if (!dbName) {
-    return dbName;
-  }
-  const parts = dbName.split(/-|_/);
-  parts[0] = camelCase ? parts[0].toLowerCase() : capitalize(parts[0]);
-
-  for (let i = 1; i < parts.length; i++) {
-    parts[i] = capitalize(parts[i]);
+/**
+ * Renames db column names with differernt naming conventions:
```
nabdelgadir: typo: `differernt` -> `different`

lib/datasource.js
nabdelgadir: `;` -> `,`

lib/datasource.js
```diff
@@ -1450,17 +1452,21 @@ function capitalize(str) {
...
+ * camelCase for property names as it's LB default naming converntion for properties;
+ * or keep the name the same if needed.
+ *
+ * @param {*} colName column name defined in database
+ * @param {*} defCefcamelCase boolean. Converts the name to camelcase if true as
+ * its the default case for variables / properties in LoopBack.
```
nabdelgadir: `its` -> `it's`

lib/datasource.js
```diff
@@ -1586,11 +1594,11 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
     // Default name mapper
     nameMapper = function mapName(type, name) {
       if (type === 'table' || type === 'model') {
-        return fromDBName(name, false);
+        return pascalCase(name);
```
jannyHou: @agnes512 A question about the behaviour change:

Before we allowed camel case for table name, and IIUC this PR converts all the table names to be pascal case, there is no option to config it to be other naming conventions including camel case. 
Is it intended?

lib/datasource.js
```diff
@@ -1770,12 +1778,13 @@ DataSource.prototype.discoverSchemasSync = function(modelName, options) {
   if (!columns || columns.length === 0) {
     return [];
   }
+  const defCamelCase = options.useDefaultCase === undefined ? true : options.useDefaultCase;
 
   const nameMapper = options.nameMapper || function mapName(type, name) {
     if (type === 'table' || type === 'model') {
-      return fromDBName(name, false);
+      return pascalCase(name);
```
jannyHou: same question as https://github.com/strongloop/loopback-datasource-juggler/pull/1819/files#r372036050 :)

lib/datasource.js
agnes512: I don't think we allowed table/model names to be camelCase before because every time we checked the type:

```ts
if (type === 'table' || type === 'model') {
      return fromDBName(name, false); // this is saying not to use camelCase, use pascal instead
```
the `false` in the second param was set, no options were allowed to configure it.

And since it was converting _all_ table/model names to pascalCase, I replace it with a formal function `pascalCase()` to make it more legit.

Before the function `fromDBName` handled cases change for type `table`, `model`, `fk`, and `column`. ~I change it to just handle column names only.~ I change it to just handle **column names** and **fk** only
( I realized that I made a mistake here, I should let it handle `fk` as well as it's part of property names).

Another reason I think table and model names should just be pascal case is because lots of our naming conventions are based on the **model** name. We assume model names will be Pascal case, and file names that are generated  based on the model names will be snake case, etc. I think allow customized table/model might cause some unexpected issues.

lib/datasource.js
```diff
@@ -1450,17 +1452,21 @@ function capitalize(str) {
...
+/**
+ * Renames db column names with different naming conventions:
+ * camelCase for property names as it's LB default naming convention for properties,
+ * or keep the name the same if needed.
+ *
+ * @param {*} colName column name defined in database
+ * @param {*} defcamelCase boolean. Converts the name to camelcase if true as
+ * it's the default case for variables / properties in LoopBack.
+ */
+function fromDBName(colName, defCamelCase) {
```
raymondfeng: It's probably better to make 2nd argument an optional function (default to camelCase) for case conversion. Using a `boolean` argument is not very friendly for JS.

- fromDBName(colName) or fromDBName(colName, camelCase) // convert to camel case
- fromDBName(colName, noCase) // no conversion
- fromDBName(colName, pascalCase)   // convert to pascal case

lib/datasource.js
```diff
@@ -1574,6 +1580,8 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
   const dbType = this.connector.name;
 
   let nameMapper;
+  const defCamelCase = options.useDefaultCase === undefined ? true : options.useDefaultCase;
```
raymondfeng: Maybe a better way is `options.disableCamelCase` (default to `false`).

lib/datasource.js
```diff
@@ -1586,11 +1594,19 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
     // Default name mapper
     nameMapper = function mapName(type, name) {
       if (type === 'table' || type === 'model') {
-        return fromDBName(name, false);
+        return fromDBName(name, pascalCase);
       } else if (type == 'fk') {
-        return fromDBName(name + 'Rel', true);
+        if (disableCamelCase === true) {
```
jannyHou: any reason we cannot use `if(disableCamelCase)`?

lib/datasource.js
```diff
@@ -1586,11 +1594,19 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
...
+          return fromDBName(name + 'Rel');
+        } else {
+          return fromDBName(name + 'Rel', camelCase);
+        }
       } else {
-        return fromDBName(name, true);
+        if (disableCamelCase === true) {
```
jannyHou: ditto https://github.com/strongloop/loopback-datasource-juggler/pull/1819/files#r372519980

jannyHou:
And please make sure the CI passes. Oracle seems random, postgresql seems related...😞

lib/datasource.js
agnes512: 👍 I thought I need to check the type as well turns out I didn't need to, changed.

lib/datasource.js
```diff
@@ -1567,13 +1574,15 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
...
   }
-
   cb = cb || utils.createPromiseCallback();
 
   const self = this;
   const dbType = this.connector.name;
 
   let nameMapper;
+  const disableCamelCase = options.disableCamelCase === undefined ? false : options.disableCamelCase;
+  console.log(disableCamelCase === true);
```
raymondfeng: Please remove console.log.

lib/datasource.js
```diff
@@ -1567,13 +1574,15 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
...
     options = {};
   }
-
   cb = cb || utils.createPromiseCallback();
 
   const self = this;
   const dbType = this.connector.name;
 
   let nameMapper;
+  const disableCamelCase = options.disableCamelCase === undefined ? false : options.disableCamelCase;
```
raymondfeng:
```ts
const disableCamelCase = !!options.disableCamelCase;
```

lib/datasource.js
agnes512: 🙈 Right

agnes512:
The failure of Postgres is because of [this line](https://github.com/strongloop/loopback-connector-postgresql/blob/master/test/postgresql.discover.test.js#L357), 
```ts
 db.discoverAndBuildModels('GeoPoint', {schema: 'strongloop'}, function(err, schema) {
      schema.Geopoint.find(function(err, data) {
```
With the change, the schema name is supposed to be `GeoPoint` instead of `Geopoint`. Can we land this PR first and I will fix it in Postgres connector?

PR #1820, opened 2020-02-24
feat: add OperationHookContext interface

hacksparrow:
Added typings for supporting operation hooks in LB4. Part of addressing https://github.com/strongloop/loopback-next/issues/3952.

Signed-off-by: Hage Yaapa <hage.yaapa@in.ibm.com>

## Checklist

### DIFF
--- a/types/observer-mixin.d.ts
+++ b/types/observer-mixin.d.ts
@@ -4,8 +4,22 @@
 // License text available at https://opensource.org/licenses/MIT
 
 import {Callback, PromiseOrVoid} from './common';
+import {PersistedModel, PersistedModelClass} from './persisted-model';
 
-export type Listener = (ctx: object, next: (err?: any) => void) => void;
+export interface OperationHookContext<T extends typeof PersistedModel> {
+  /**
+   * The constructor of the model that triggered the operation.
+   */
+  Model: T;
+
+  /**
+   * Additional context properties, not typed yet.
+   * See https://loopback.io/doc/en/lb3/Operation-hooks.html#hooks
+   */
+  [property: string]: any;
+}
+
+export type Listener<Ctx> = (ctx: Ctx, next: (err?: any) => void) => void;
 
 export interface ObserverMixin {
   /**
@@ -36,7 +50,7 @@ export interface ObserverMixin {
    * `this` set to the model constructor, e.g. `User`.
    * @end
    */
-  observe(operation: string, listener: Listener): void;
+  observe(operation: string, listener: Listener<OperationHookContext<PersistedModelClass>>): void;
 
   /**
    * Unregister an asynchronous observer for the given operation (event).
@@ -54,7 +68,7 @@ export interface ObserverMixin {
    * @callback {function} listener The listener function.
    * @end
    */
-  removeObserver(operation: string, listener: Listener): Listener | undefined;
+  removeObserver(operation: string, listener: Listener<OperationHookContext<PersistedModelClass>>): Listener<OperationHookContext<PersistedModelClass>> | undefined;
 
   /**
    * Unregister all asynchronous observers for the given operation (event).


types/persisted-model.d.ts
```diff
@@ -519,6 +520,10 @@ export declare class PersistedModel extends ModelBase {
    * @returns {String} The `id` property name
    */
   static getIdName(): string;
+
+  definePersistedModel<T>(entityClass: T): PersistedModel;
```
bajtos: Where is this method implemented? I could not find it in the juggler codebase `grep definePersistedModel -r lib/`.

types/persisted-model.d.ts
```diff
@@ -519,6 +520,10 @@ export declare class PersistedModel extends ModelBase {
...
+
+  static observe(operation: string, handler: Listener): void;
```
bajtos: While you are adding `observe` method, can you please add typings for other methods related to operation hooks?

**SHOULD HAVE**

https://github.com/strongloop/loopback-datasource-juggler/blob/36303b3b63932478cb429b5c5af6fec7f09b602c/lib/observer.js#L76

https://github.com/strongloop/loopback-datasource-juggler/blob/36303b3b63932478cb429b5c5af6fec7f09b602c/lib/observer.js#L99

**NICE TO HAVE**

I think these APIs are mostly internal. Do they support promise-based invocations?

https://github.com/strongloop/loopback-datasource-juggler/blob/36303b3b63932478cb429b5c5af6fec7f09b602c/lib/observer.js#L131

https://github.com/strongloop/loopback-datasource-juggler/blob/36303b3b63932478cb429b5c5af6fec7f09b602c/lib/observer.js#L221

IIRC, we often copy API documentation from jsdocs into tsdoc comments, so that consumers of the TypeScript typedefs can get API docs rendered in their IDE like VSCode.

types/observer-mixin.d.ts
```diff
@@ -5,7 +5,11 @@
 
 import {Callback, PromiseOrVoid} from './common';
 
-export type Listener = (ctx: object, next: (err?: any) => void) => void;
+export interface OperationHookContext {
+  [property: string]: any;
```
bajtos: For longer term, I'd like us to provide strongly typed context types based on the hook name. See this doc page for more details about context objects provided by different hooks: https://loopback.io/doc/en/lb3/Operation-hooks.html#hooks

For example:

```ts
export interface AccessHookContext<T extends typeof PersistedModel> {
  Model: T;
  query: Filter<T>;
}

export type Listener<Ctx> = (ctx: Ctx, next: (err?: any) => void) => void;

class PersistedModel {
// ...
static observe(operation: 'access', handler: Listener<AccessHookContext>): void;
static observe(operation: 'before save', handler: Listener<BeforeSaveHookContext>): void;
// etc.
}
```

I think it's ok to leave that as a long term goal, out of scope of this pull request.

Having said that, I believe that all operation hooks are always providing `Model` property. Let's include it in the type definition please.

```ts
export interface OperationHookContext<T extends typeof PersistedModel> {
  /**
   * The constructor of the model that triggered the operation.
   */
  Model: T;

  /**
   * Additional context properties, not typed yet.
   * See https://loopback.io/doc/en/lb3/Operation-hooks.html#hooks
   */
  [property: string]: any;
}
```

types/observer-mixin.d.ts
```diff
@@ -5,7 +5,11 @@
 
 import {Callback, PromiseOrVoid} from './common';
 
-export type Listener = (ctx: object, next: (err?: any) => void) => void;
+export interface OperationHookContext {
```
agnes512: Can we have API docs, example, r comments on what this new interface is for?  Same for the rest of changes.

types/persisted-model.d.ts
hacksparrow: Errata, it's not required.

types/observer-mixin.d.ts
hacksparrow: 👍

types/persisted-model.d.ts
hacksparrow: `notifyObserversOf ` and `notifyObserversAround ` are for internal uses, let's omit them.

types/observer-mixin.d.ts
hacksparrow: Added.

types/observer-mixin.d.ts
hacksparrow: Created an issue to note the suggestion and follow through - https://github.com/strongloop/loopback-datasource-juggler/issues/1822.

dhmlau: @bajtos @hacksparrow, would it be a `tech-debt` for Q2? or much longer term?

hacksparrow: Wouldn't call it a tech debt, exactly; more like something very good to have.

types/persisted-model.d.ts
```diff
@@ -519,6 +520,67 @@ export declare class PersistedModel extends ModelBase {
...
+  * MyModel.removeObserver('before save', function removedObserver(ctx, next) {
+    // some logic user want to apply to the removed observer...
+    next();
+  });
+  * ```
+  *
+  * @param {String} operation The operation name.
+  * @callback {function} listener The listener function.
+  */
+  static removeObserver(operation: string, handler: Listener<OperationHookContext<PersistedModelClass>>): void;
```
jannyHou: I am a bit confused about this function, the function name gives me an impression that it removes a registered observer(registered by model.observe) by name, but it takes in a handler as the 2nd parameter...

Or does it remove ALL the observers bound to the operation?

hacksparrow: It removes all listeners for an operation.

This PR is not creating the method, though. It is just providing the TypeScript definitions for something that already existed. So any feedback on the looks and behavior of the method is for some other time and place.

types/observer-mixin.d.ts
bajtos: Perhaps `developer-experience` + `2020Q2`?

types/persisted-model.d.ts
bajtos: @jannyHou 

To register an Operation Hook observer, we need two pieces of information:
- what hook to observe (the operation name - a string like `before save`) and 
- the handler function to call. 

Similarly, to unregister an observer, we need the same two pieces.

Consider the following code:

```js
const logHook(ctx, next) {
  debug(ctx);
  next();
}

MyModel.observe('before save', logHook);
MyModel.observe('after save', logHook);

// now we want to remove 'after save' hook but preserve 'before save' hook
MyModel.removeObserver('after save', logHook);
```

types/observer-mixin.d.ts
```diff
@@ -36,7 +50,7 @@ export interface ObserverMixin {
    * `this` set to the model constructor, e.g. `User`.
    * @end
    */
-  observe(operation: string, listener: Listener): void;
+  observe(operation: string, listener: Listener<OperationHookContext<PersistedModelClass>>): void;
```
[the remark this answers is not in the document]
bajtos: It makes me wonder, can we use `this` as the value of the generic argument?

```suggestion
  observe(operation: string, listener: Listener<OperationHookContext<this>>): void;
```

If yes, then I think such typing will express exactly what we need (see my other comment https://github.com/strongloop/loopback-datasource-juggler/pull/1820#pullrequestreview-365481860 for more context):
> if a class T extends Model, then it will include ObserverMixin<T> methods

bajtos:
Regarding the commit message:

> chore: typings for supporting operation hooks in LB4 

Commit type `chore` is used for changes that are not affecting users. The changelog generator we use in loopback-next is excluding such commits from the log.

Since you are improving type definitions, you should use `feat` type.

hacksparrow:
Will edit in the final commit message after squashing the commits.

types/observer-mixin.d.ts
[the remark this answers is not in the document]
bajtos: ok, so it's not possible to use `this` in the context of a static method, but there is a workaround described in https://github.com/microsoft/TypeScript/issues/5863#issuecomment-410887254.

```ts
observe<T>(this: T, operation: string, listener: Listener<OperationHookContext<T>>): void;
```

Here a standalone example I used to verify this approach, it compiles cleanly for me.

```ts
interface Context<T> {
  target: T;
}

type Listener<T> = (ctx: Context<T>) => void;

class Base {
  static observe<T extends Base>(this: T, listener: Listener<T>): void {
    listener({target: this});
  }
}

class Foo extends Base {
  static isFoo = true;
}

Foo.observe(ctx => {
  console.log('target is foo', ctx.target.isFoo);
});
```

Anyhow, I think this part may be out of scope of this pull request, I have some ideas how to address it in https://github.com/strongloop/loopback-datasource-juggler/pull/1823.

types/persisted-model.d.ts
jannyHou: thanks @hacksparrow I see, didn't notice that the function already exists. 
and thanks for the explanation from @bajtos , very clear to me 👍

PR #1823, opened 2020-02-28
Add ObserverMixin to ModelBase typings

bajtos:
See https://github.com/strongloop/loopback-datasource-juggler/pull/1820 and https://github.com/strongloop/loopback-next/pull/4730.

## Model APIs

- [x] `ModelBaseClass`
- [x] `PersistedModelClass`
- [x] `typeof KeyValueModel`

## Checklist

### DIFF
--- a/types/__test__.ts
+++ b/types/__test__.ts
@@ -1,0 +1,90 @@
+// Copyright IBM Corp. 2020. All Rights Reserved.
+// Node module: loopback-datasource-juggler
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+// A test file to verify types described by our .d.ts files.
+// The code in this file is only compiled, we don't run it via Mocha.
+
+import {
+  DataSource,
+  KeyValueModel,
+  ModelBase,
+  ModelBaseClass,
+  PersistedModel,
+  PersistedModelClass,
+} from '..';
+
+const db = new DataSource('db', {connector: 'memory'});
+
+//-------
+// ModelBase should provide ObserverMixin APIs as static methods
+//-------
+//
+(function() {
+  const Data = db.createModel('Data');
+
+  // An operation hook can be installed
+  Data.observe('before save', async ctx => {});
+
+  // Context is typed and provides `Model` property
+  Data.observe('before save', async ctx => {
+    console.log(ctx.Model.modelName);
+  });
+
+  // ModelBaseClass can be assigned to `typeof ModelBase`
+  // Please note that both `ModelBaseClass` and typeof ModelBase`
+  // are different ways how to describe a class constructor of a model.
+  // In this test we are verifying that the value returned by `createModel`
+  // can be assigned to both types.
+  const modelTypeof: typeof ModelBase = Data;
+  const modelCls: ModelBaseClass = modelTypeof;
+});
+
+//-------
+// PersistedModel should provide ObserverMixin APIs as static methods
+//-------
+(function () {
+  const Product = db.createModel<PersistedModelClass>(
+    'Product',
+    {name: String},
+    {strict: true}
+  );
+
+  // It accepts async function
+  Product.observe('before save', async ctx => {});
+
+  // It accepts callback-based function
+  Product.observe('before save', (ctx, next) => {
+    next(new Error('test error'));
+  });
+
+  // ctx.Model is a PersistedModel class constructor
+  Product.observe('before save', async ctx => {
+    await ctx.Model.findOne();
+  });
+
+  // PersistedModelClass can be assigned to `typeof PersistedModel`
+  // Please note that both `PersistedModelClass` and typeof PersistedModel`
+  // are different ways how to describe a class constructor of a model.
+  // In this test we are verifying that the value returned by `createModel`
+  // can be assigned to both types.
+  const modelTypeof: typeof PersistedModel = Product;
+  const modelCls: PersistedModelClass = modelTypeof;
+});
+
+//-------
+// KeyValueModel should provide ObserverMixin APIs as static methods
+//-------
+(function () {
+  const kvdb = new DataSource({connector: 'kv-memory'});
+  const CacheItem = kvdb.createModel<typeof KeyValueModel>('CacheItem');
+
+  // An operation hook can be installed
+  CacheItem.observe('before save', async ctx => {});
+
+  // ctx.Model is a KeyValueModel class constructor
+  CacheItem.observe('before save', async ctx => {
+    await ctx.Model.expire('key', 100);
+  });
+});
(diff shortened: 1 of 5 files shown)


bajtos:
This is actually a breaking change.

Before, a value of type `typeof PersistedModel` was assignable to `PersistedModelClass`.

With my change in place, this is no longer possible, because `typeof PersistedModel` is missing `ObserverMixin` methods.

bajtos:
> This is actually a breaking change.
> 
> Before, a value of type `typeof PersistedModel` was assignable to `PersistedModelClass`.
>
> With my change in place, this is no longer possible, because `typeof PersistedModel` is missing `ObserverMixin` methods.

As I was thinking about this problem, I came to the conclusion that we should prefer ease of use and ergonomics of our typings over implementation simplicity.

The original implementation in https://github.com/strongloop/loopback-datasource-juggler/commit/fff92b87418c27e8817826c1e013ce3a49533175 was based on changing `{Model}Class` type aliases to include `& ObserverMixin`. While this is quick fix in juggler that can reuse typings in `ObserverMixin`, it introduces a breaking change because `typeof PersistedModel` is no longer assigne-able to `PersistedModelClass`, which I would find very surprising as a user.

I have reworked this pull request to simply copy typings of the public members of `ObserverMixin` into `ModelBase` class as static members. This is not perfect because we are duplicating typings, but it should be find in practice, because the API is pretty much stable by now.

I have verified that `@loopback/next` compiles cleanly against the updated typings, therefore I believe this new version should be fully backwards-compatible 🎉 

The code here is prepared to support better-typed Operation Hook Context introduced in https://github.com/strongloop/loopback-datasource-juggler/pull/1820 🚀 

I consider this PR as ready for review. @emonddr @zbarbuto @jannyHou @dhmlau PTAL.

/cc @hacksparrow @raymondfeng - I'd like to hear your opinions too.

hacksparrow:
@bajtos there is at least one place where `PersistedModel & ObserverMixin` is used instead of `typeof PersistedModel & ObserverMixin`. Would that affect anything in the PR?

bajtos:
> there is at least one place where `PersistedModel & ObserverMixin` is used instead of `typeof PersistedModel & ObserverMixin`. Would that affect anything in the PR?

As I understand the API, `PersistedModel & ObserverMixin` is incorrect. There are no observer-related methods on model instance, those methods are always static.

I think you need to fix those places and modify `PersistedModel & ObserverMixin` to `typeof PersistedModel & ObserverMixin`

bajtos:
AFAICT, your PR https://github.com/strongloop/loopback-next/pull/4730 uses the following two variants, both of them are correct:

- `typeof PersistedModel & ObserverMixin`
- `juggler.PersistedModelClass & juggler.ObserverMixin` 

Notice that in the second case, we are using `PersistedModelClass`.

types/__test__.ts
```diff
@@ -0,0 +1,67 @@
...
+//-------
+//
+(function() {
+  const Data = db.createModel('Data');
+
+  // An operation hook can be installed
+  Data.observe('before save', async ctx => {});
+
+  // ModelBaseClass can be assigned to `typeof ModelBase`
+  const modelTypeof: typeof ModelBase = Data;
```
emonddr: Please add a comment that `ModelBaseClass` is an alias of `typeof ModelBase` and we are testing the alias here. thx.

types/model.d.ts
```diff
@@ -243,6 +244,85 @@ export declare class ModelBase {
     anotherClass: string | ModelBaseClass | object,
     options?: Options,
   ): ModelBaseClass;
+
```
emonddr: why just the `model.d.ts` file, why not `model.ts` ?

bajtos: I am afraid I don't understand the question. I am editing an existing TypeScript declaration file which is providing type information for artifacts implemented by JavaScript code in `lib`. There is no `model.ts` file in juggler.

See "What is a Declaration File?" in https://medium.com/razroo/what-are-declaration-files-in-typescript-476c1c06f353#936b for a decent introduction into type declarations, and [Introduction to Declaration Files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) in TypeScript HandBook for content going more deep.

bajtos:
Rebased on top of https://github.com/strongloop/loopback-datasource-juggler/pull/1820, added few more tweaks in 6c38347.

types/__test__.ts
bajtos: See 6c38347, is my comment good enough?

bajtos:
I linked my modified version of juggler into `packages/repository` of loopback-next and run `npm run build` to verify that my changes are not breaking anything. I did this for both `master` and `feat/operation-hooks` (https://github.com/strongloop/loopback-next/pull/4730) branches.

bajtos:
Created https://github.com/strongloop/loopback-datasource-juggler/pull/1825 to fast-track landing of `Listener` fix.

bajtos:
_Cross-posting from Slack._

> More of a general TS question:
> Why do we do `removeObserver<T extends typeof ModelBase>`  and `not removeObserver<T extends ModelBase>`?
> What does `this: T` do and why do we need to do that? 


Please start here:
```
  // See also https://github.com/microsoft/TypeScript/issues/5863#issuecomment-410887254
  // for more information about using `this` in static members.
```

Here is what we want to describe: for a model extending from `BaseModel` - let's say `Product`, when the static method `removeObserver` is invoked, the context argument is of type `OperationHookContext<Product>`.

If `removeObserver` was a prototype method, then we could use `OperationHookContext<this>`. Because it's a static method, we need to find a different solution - see https://github.com/microsoft/TypeScript/issues/5863#issuecomment-410887254

Regarding `this: T`, see this docs for introduction to how this is handled: https://www.typescriptlang.org/docs/handbook/functions.html#this
`this` parameters (function arguments) are described later in that doc, see https://www.typescriptlang.org/docs/handbook/functions.html#this-parameters

> You can provide an explicit `this` parameter. `this` parameters are fake parameters that come first in the parameter list of a function.

For example, a mocha test function returning a promise is typed as `function (this: Mocha.Context): Promise<void>`, which allows you to access API like `this.timeout()` and `this.skip()` inside that function.

Now in our case, by using `this: T`, we allow the compiler to infer what's the target model class we are going to observe, so that can describe the type of context argument using the target model class.

> Why do we do `removeObserver<T extends typeof ModelBase>`  and not `removeObserver<T extends ModelBase>`

Because observer API is mixed into the class constructor (`Product.observe()`), not the prototype (`new Product().observer()`).

bajtos:
Added b01cd74 to capture some of the information from my comment above.

types/__test__.ts
emonddr: The comment is great. thx. But a tiny typo: `verifying that the value retunred `

types/model.d.ts
emonddr: @bajtos , I haven't worked with `.d.ts` files yet. I know that they are one of the file types generated by a transpilation of Typescript to Javascript. I will read the article you mentioned.

PR #1831, opened 2020-04-03
Update node versions

bajtos:
- Add support for Node.js 13.x
- Drop support for Node 8.x

## Checklist

- Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-  - "8"
   - "10"
   - "12"
   - "13"
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ node_js:
   - "8"
   - "10"
   - "12"
+  - "13"
 
 after_success: npm run coverage
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   ],
   "author": "IBM Corp.",
   "engines": {
-    "node": ">=8"
+    "node": ">=10"
   },
   "repository": {
     "type": "git",


bajtos:
We will release this change as semver-minor. Cross-posting from https://github.com/strongloop/loopback-connector/issues/172#issuecomment-597009643:

>Ideally, when we remove support for a major Node.js version (e.g. `8.x`), then we should publish a major version of our module too.
> 
> Unfortunately, this does not work well with our LTS support policy. If we keep supporting EOL versions of Node, and there is a security vulnerability discovered in one of our dependencies, and this vulnerability is fixed in the latest version only, and that latest version no longer supports EOLed version of Node, then we cannot get the fix. If we are not able to fix security vulnerabilities in our LTS versions, then it's IMO no long-term support at all.
> 
> We have discussed this for a bit around Node.js 4.x release (pretty much the first relevant release after 2-3 years of having everybody on Node.js 0.10). I believe we reached the conclusion that when having to choose between security fixes and strict adherence to semver rules, we prefer security over user convenience.
> 
> BTW this gets even more complicated for major versions of the framework itself (e.g. `loopback` and `loopback-datasource-juggler` module). When LB 3.x is Current and LB 2.x is Active LTS, and Node.js 6.x goes EOL, then how should be drop support for Node.js 6.x from LB 2.x? LB 3.x is already taken at that point...
> 
> My recommendation is to use package-lock and npm-shrinkwrap functionality to lock down dependencies in your project to the latest versions that still supports the end-of-life Node.js version you are using (e.g. 8.x).

PR #1844, opened 2020-07-18
Add definition attribute to datasource

khlilturki97:
this fix is needed in this [PR](https://github.com/strongloop/loopback-next/pull/5769)

### DIFF
--- a/types/datasource.d.ts
+++ b/types/datasource.d.ts
@@ -9,7 +9,7 @@ import {
   ModelBaseClass,
   ModelBuilder,
   ModelDefinition,
-  PropertyDefinition,
+  PropertyDefinition
 } from './model';
 import {EventEmitter} from 'events';
 import {IsolationLevel, Transaction} from './transaction-mixin';
@@ -83,6 +83,8 @@ export declare class DataSource extends EventEmitter {
   connecting?: boolean;
 
   connector?: Connector;
+
+  definitions: {[modelName: string]: ModelDefinition};
 
   DataAccessObject: AnyObject & {prototype: AnyObject};
 


types/datasource.d.ts
```diff
@@ -5,12 +5,7 @@
 
 import {AnyObject, Callback, Options} from './common';
 import {Connector} from './connector';
-import {
-  ModelBaseClass,
-  ModelBuilder,
-  ModelDefinition,
-  PropertyDefinition,
-} from './model';
+import {ModelBaseClass, ModelBuilder, ModelDefinition, PropertyDefinition} from './model';
```
bajtos: Let's revert these unrelated formatting changes.

types/datasource.d.ts
```diff
@@ -84,6 +79,8 @@ export declare class DataSource extends EventEmitter {
 
   connector?: Connector;
 
+  definitions?: AnyObject;
```
bajtos: Please note `definitions` are always set to `modelBuilder.definitions`:

https://github.com/strongloop/loopback-datasource-juggler/blob/8d1690e9b2c76e3d0a8336b47f133e0605cbcc9a/lib/datasource.js#L139

We should use the same type for this property as we do in `ModelBuilder`:

https://github.com/strongloop/loopback-datasource-juggler/blob/8d1690e9b2c76e3d0a8336b47f133e0605cbcc9a/types/model.d.ts#L339

bajtos:
@rmg could you PTAL why `cis-jenkins` and `clahub` checks are not executed for this pull request?

Screen Shot 2020-07-24 at 17 13 53

bajtos:
@slnode ok to test

bajtos:
Landed, thank you for the contribution ❤️

PR #1855, opened 2020-07-27
types: support non-SQL styles of `ds.execute`

bajtos:
Add new overloads for `DataSource.execute` method to support non-SQL connectors like MongoDB.

See https://github.com/strongloop/loopback-next/issues/3342 for background and https://loopback.io/doc/en/lb3/Executing-native-SQL.html for the existing documentation for `db.execute` API.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-datasource-juggler) 👈

- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [x] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)

PR #1857, opened 2020-08-03
feat: improve `dataSource.execute` to support more flavors

bajtos:
Add support for the following variants:

- `execute(collection, command, ...params, options)` (MongoDB)
- `execute(...params)` (forward-compatibility & other databases)

This is a follow-up for #1855, which added type descriptions for `execute` variants that turned out to be rejected by the actual (runtime) implementation. See also https://github.com/strongloop/loopback-next/issues/3342

The changes is pushing the responsibility for processing optional arguments (`(cmd)`, `(cmd, args)`, `(cmd, args, options)`) to the connector. This should be fine because the base `SqlConnector` already handles optional arguments (see [lib/sql.js#L551-L563](https://github.com/strongloop/loopback-connector/blob/bce6fa8c1f498340a5185dcd21545972a3cce7a7/lib/sql.js#L551-L563)), and we weren't really supporting non-SQL `execute` before.

## Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -2690,19 +2690,10 @@ DataSource.prototype.ping = function(cb) {
  * Execute an arbitrary command. The commands are connector specific,
  * please refer to the documentation of your connector for more details.
  *
- * @param command String|Object The command to execute, e.g. an SQL query.
- * @param [args] Array Parameters values to set in the command.
- * @param [options] Object Additional options, e.g. the transaction to use.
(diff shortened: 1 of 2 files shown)


bajtos:
> coverage/coveralls — Coverage decreased (-0.0005%) to 84.722%

I believe this is caused by reducing the number of lines in `lib/datasource.js` - i.e. the number of lines executed by our tests is staying the same, but the total number of lines is lower now, therefore the coverage ratio went slightly down. Nothing to worry about.

PR #1861, opened 2020-08-13
SPIKE: New objectid

hacksparrow:
## Checklist

👉 [How to sign off your commits with DCO](https://github.com/strongloop/loopback-datasource-juggler#contributing) 👈

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -380,7 +380,15 @@ DataAccessObject.create = function(data, options, cb) {
       obj.trigger('save', function(saveDone) {
         const _idName = idName(Model);
         let val = Model._sanitizeData(obj.toObject(true), options);
-        function createCallback(err, id, rev) {
+
+        // SPIKE: changed the signature
+        function createCallback(options) {
+
+          const { err, id, rev, propertyCoercer } = options;
+
+          // SPIKE: apply coercion
+          propertyCoercer(obj.__data, Model.definition.properties);
+
           if (id) {
             obj.__data[_idName] = id;
             defineReadonlyProp(obj, _idName, id);
@@ -1628,7 +1636,7 @@ DataAccessObject.find = function find(query, options, cb) {
         } else {
           withNotify(item, next);
         }
-
+        
         function buildResult(data, callback) {
           const ctorOpts = {
             fields: query.fields,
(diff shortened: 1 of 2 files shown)


lib/dao.js
```diff
@@ -380,7 +380,15 @@ DataAccessObject.create = function(data, options, cb) {
...
         let val = Model._sanitizeData(obj.toObject(true), options);
-        function createCallback(err, id, rev) {
+
+        // SPIKE: changed the signature
+        function createCallback(options) {
+
+          const { err, id, rev, propertyCoercer } = options;
+
+          // SPIKE: apply coercion
+          propertyCoercer(obj.__data, Model.definition.properties);
```
bajtos, line 390: I find this approach as too complex and error-prone. We have to remember to call coercion in every DAO method that's dealing with model data coming from the connector.

Have you considered applying the coercion rules inside [ModelBaseClass.prototype._initProperties](https://github.com/strongloop/loopback-datasource-juggler/blob/261dd1c865e8b197882a161b909b90b0e6c3c2a3/lib/model.js#L202-L207), as I explained in https://github.com/strongloop/loopback-connector-mongodb/pull/591#discussion_r473908112?

lib/dao.js
```diff
@@ -1628,18 +1636,28 @@ DataAccessObject.find = function find(query, options, cb) {
...
           try {
+            // SPIKE: This converts string-id back to ObjectID
             obj = new Model(data, ctorOpts);
+
+const propDefs = self.dataSource.definitions[self.modelName].properties;
+if (propDefs.id.type.name === 'ObjectID') {
+  // SPIKE: to change the constructor type
+  const strId = new String(obj.id.toString());
+  obj.id = strId;
+}
```
bajtos, line 1659: Is it necessary to have an object-id specific coercion rule? Can we have a more generic rule that will coerce all non-string values to a string for properties defined with `type: string`? See https://github.com/strongloop/loopback-connector-mongodb/pull/591#discussion_r473908112

It's also important to coerce all properties, not just `id`:
1. The `id` property may not even exist, models can use a different name for the primary key.
2. We need to coerce foreign key values too (e.g. `Order.customerId`), otherwise query inclusion resolvers won't work.

bajtos (changes requested):
Thank you @hacksparrow for a PoC showing how you are envisioning ObjectID coercion.

I have few comments regarding the implementation details you choose, see below.

In the light of the second part of https://github.com/strongloop/loopback-connector-mongodb/pull/591#discussion_r473908112, I think it would be better to deal with the coercion inside connectors, so that no changes in juggler are needed.

Let's discuss in that other pull request first.

lib/dao.js
jannyHou: @bajtos I got your concern. 
The `toDatabase` coercion is already handled in mongodb connector(for both decimal128 and objectid), but for the other direction `fromDatabase`, what's the current situation? 
If currently an objectid property passed from mongodb will be automatically converted to string in juggler, and it's a common logic for any custom type from any connectors, then I agree we should keep juggler untouched.

bajtos: MongoDB connector already implements a similar functionality for `decimal128` type. It uses the following property definition:

```js
{ 
  type: 'string', 
  mongodb: {dataType: 'decimal128'},
}
```

The connector takes care of converting string values coming from juggler to MongoDB's `Decimal128` instances, and converting `Decimal128` instances coming from MongoDB back to string before they are passed to juggler.

PR #1865, opened 2020-09-04
array_type_field_filter: Adds contains filter for postgresDB

### LINKED ISSUE #342: Filter by array of values
I have following records:

```
{
  "_id": ObjectId("544971f9f144b60000a50bea"),
  "name": "test",
  "description": null,
  "contacts": "Test contacts",
  "url": null,
  "dateList": [
    "2014-11-08T05:00:00.000Z",
    "2014-11-15T05:00:00.000Z"
  ],
  "status": "new",
  "collectionId": ObjectId("54459818953c3c341812df8a"),
  "fileId": null
}
```

How can I filter by date? possibly with range, something like 

```
Model.find({
 where: {
  dateList: "...."
 }
}
```

or 
```
Model.find({
 where: {
  dateList: has([..., ...])
 }
}
```

shubhisood:
Adds contains filter to make searches on Array type field.

Fixes #342 of [loopback-postgres connector](https://github.com/strongloop/loopback-connector-postgresql/pull/460)

## Checklist

- [X] [Sign off your commits](https://loopback.io/doc/en/contrib/code-contrib.html) with DCO (Developer Certificate of Origin)

### DIFF
--- a/lib/model-utils.js
+++ b/lib/model-utils.js
@@ -88,6 +88,7 @@ const operators = {
   ilike: 'ILIKE',
   nilike: 'NOT ILIKE',
   regexp: 'REGEXP',
+  contains: '@>',
 };
 
 /*
@@ -449,6 +450,7 @@ ModelUtils._coerce = function(where, options, modelDef) {
             case 'inq':
             case 'nin':
             case 'between':
+            case 'contains':
               try {
                 val = coerceArray(val);
               } catch (e) {


raymondfeng:
@shubhisood Thank you for the PR. Please sign DCO (https://loopback.io/doc/en/contrib/code-contrib.html).

lib/model-utils.js
```diff
@@ -88,6 +88,7 @@ const operators = {
   ilike: 'ILIKE',
   nilike: 'NOT ILIKE',
   regexp: 'REGEXP',
+  contains: '@>',
```
raymondfeng, line 91: Please note that the operator needs to be supported by other LoopBack db connectors beyond Postgres. Do we have some ideas about the feasibility of other DBs?

rgupta2: @raymondfeng This feature is so much needed. We literally had to hack our way into Loopback as we needed it extensively. Please figure out a way to get this merged.

shubhisood: @raymondfeng currently i have added it only for postgres db, not sure how we may implement it for other db connectors as well, i may need to look into this..

shubhisood:
Travis build is failing because the following PR related to juggler needs to be merged first:
https://github.com/strongloop/loopback-datasource-juggler/pull/1865

lib/model-utils.js
bajtos: I agree with @raymondfeng that we must keep other connectors in mind when making any changes in juggler.

Is it possible to implement array coercion of `contains` filter inside the postgresql connector?

In MongoDB, we have the concept of extended operators, see line 507 later in this file.

https://github.com/strongloop/loopback-datasource-juggler/blob/d953eabe30d793642b9268014efc5f4da4633e29/lib/model-utils.js#L507

Can we perhaps use the same mechanism for PostgreSQL?

bajtos (changes requested):
Thank you @shubhisood for the pull request. It makes a lot of sense to me to support additional database-specific operators like `contains` :+1:

However, we need to find a solution that will work for all juggler users, regardless of the connector they are using.

lib/model-utils.js
shubhisood: I need to check into this @bajtos will inform you after analysis.

shubhisood: @bajtos you are right after using allowExtendedOperators option within postgres connector model, I am able to handle the case for array field filter.

bajtos:
Closing per https://github.com/strongloop/loopback-datasource-juggler/pull/1865#discussion_r488069639

> after using allowExtendedOperators option within postgres connector model, I am able to handle the case for array field filter, we can close this PR after postgres connector specfic PR gets approved.

PR #1864, opened 2020-09-02
add tests for atomic upsertWithWhere

mrbatista:
See https://github.com/strongloop/loopback-connector-mongodb/pull/563

## Checklist

- [x] [Sign off your commits](https://loopback.io/doc/en/contrib/code-contrib.html) with DCO (Developer Certificate of Origin)
- [x] `npm test` passes on your machine
- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [x] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)

PR #1871, opened 2020-10-15
chore: add test for atomic upsertWithWhere (#1864)

jannyHou:
Introduce the new property atomicUpsertWithWhere for
connector that implement specific method.
See https://github.com/strongloop/loopback-connector-mongodb/pull/563
for mongodb implementation.

Backport #1864 

Signed-off-by: Matteo Padovano <mrbatista@users.noreply.github.com>

## Checklist

### DIFF
--- a/test/manipulation.test.js
+++ b/test/manipulation.test.js
@@ -2620,22 +2620,43 @@ describe('manipulation', function() {
         });
     });
 
-    it('fails the upsertWithWhere operation when multiple instances are ' +
-        'retrieved based on the filter criteria', function(done) {
-      Person.create([
-        {id: '2', name: 'Howie', city: 'Florida'},
-        {id: '3', name: 'Kevin', city: 'Florida'},
-      ], function(err, instance) {
-        if (err) return done(err);
-        Person.upsertWithWhere({city: 'Florida'}, {
-          id: '4', name: 'Brian',
-        }, function(err) {
-          err.message.should.equal('There are multiple instances found.' +
+    bdd.itIf(!connectorCapabilities.atomicUpsertWithWhere,
+      'fails the upsertWithWhere operation when multiple instances are ' +
+      'retrieved based on the filter criteria', function(done) {
+        Person.create([
+          {id: '2', name: 'Howie', city: 'Florida'},
+          {id: '3', name: 'Kevin', city: 'Florida'},
+        ], function(err, instance) {
+          if (err) return done(err);
+          Person.upsertWithWhere({city: 'Florida'}, {
+            id: '4', name: 'Brian',
+          }, function(err) {
+            err.message.should.equal('There are multiple instances found.' +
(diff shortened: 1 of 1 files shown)


test/manipulation.test.js
```diff
@@ -2620,22 +2620,43 @@ describe('manipulation', function() {
...
-      Person.create([
-        {id: '2', name: 'Howie', city: 'Florida'},
-        {id: '3', name: 'Kevin', city: 'Florida'},
-      ], function(err, instance) {
-        if (err) return done(err);
-        Person.upsertWithWhere({city: 'Florida'}, {
-          id: '4', name: 'Brian',
-        }, function(err) {
-          err.message.should.equal('There are multiple instances found.' +
+    bdd.itIf(connectorCapabilities.atomicUpsertWithWhere !== true,
```
bajtos, line 2623: When adding a new connector capability flag, we should pick a reasonable default so that we don't have to update all connectors with the new flag. Considering that DAO does not require connectors to implement atomic `upsertWithWhere` and provides a non-atomic implementation in such case, I think we should assume connectors don't support atomic version by default.

We need to ensure one of the `bdd.itIf` cases is executed when `connectorCapabilities.atomicUpsertWithWhere` is not set to any value (it's `undefined`). I think it will be this particular case, right?

I think the condition here could be simplified as follows:

```suggestion
    bdd.itIf(!connectorCapabilities.atomicUpsertWithWhere,
```

test/manipulation.test.js
```diff
@@ -2620,22 +2620,43 @@ describe('manipulation', function() {
...
+          });
         });
       });
-    });
+
+    bdd.itIf(connectorCapabilities.atomicUpsertWithWhere === true,
+      'upsertWithWhere update the first matching instance when multiple instances are ' +
+      'retrieved based on the filter criteria', async () => {
+        // The first matching instance is determinate from specific connector implementation
+        // For example for mongodb connector the sort parameter is used (default to _id asc)
```
bajtos, line 2645: I find it problematic that connectors implement different behavior than our default non-atomic implementation. Are we sure that SQL connectors like MySQL and PostgreSQL will update the first record only, the same way as MongoDB connector works? It would be great if we could align juggler-provided non-atomic version with the behavior of real connectors (both SQL and noSQL). If different connectors behave differently, then IMO we need to clearly document these differences and probably need to add a new `connectorCapability` flag to allow the test suite to execute the right set of expectations?

jannyHou: > Are we sure that SQL connectors like MySQL and PostgreSQL will update the first record only, the same way as MongoDB connector works?

@bajtos Currently only mongodb supports that, no other connectors have atomic implementation.

Hmm isn't `atomicUpsertWithWhere` already a connector flag? I will document the behaviour after PRs merge :)

bajtos (approved):
Fair enough :+1:

PR #1878, opened 2020-11-05
feat: consistent behavior of generated id and forceId

hacksparrow:
Currently, if PK property `generated` is `false` and user does not supply a value, Juggler is autogenerating the primary key (id). I think this is an unexpected behavior. I encountered this issue while writing tests for the new [string interface for ObjectId](https://github.com/strongloop/loopback-connector-mongodb/pull/601) in the MongoDB connector.

In this PR, the `test/id.test.js` file defines the rules of interaction between `generated` and `forceId` and `lib/model-builder.js` is modified to make id value mandatory if `generated` is `false`.

Consequent of the changes, 14 existing test cases fail with `'id' can't be blank` errors.

What do you think of the proposes change? It is obviously a breaking change.

## Checklist


test/id.test.js
```diff
@@ -0,0 +1,151 @@
...
+      });
+    });
+  });
+
+  context('generated:false (default)', function() {
+    context('with forceId:true (default)', function() {
+      let IdCheckC;
+      before(async () => {
+        IdCheckC = ds.define('IdCheckC', {
+          name: String,
```
bajtos, line 69: I think you are missing the definition if `id` property here? Otherwise it's not clear to me how is this test context (`generated:false`) different from the previous one (`unspecified (defaults to generated:true)`).

test/id.test.js
```diff
@@ -0,0 +1,151 @@
...
+        });
+        await ds.automigrate(IdCheckC.modelName);
+        await IdCheckC.destroyAll();
+      });
+
+      // If id is not autogenerated and forceId is true, a value must be specified
```
bajtos, line 75: IMO, the combination of id with `generated: false` and model setting `forceId: true` is not valid and we should reject it early on. Probably when the model is defined or when it's attached to the DB - I don't remember where exactly is the logic computing `forceId` flag from id property settings.

bajtos: Another way how to look at this is that when id is not autogenerated, then `forceId` is ignored. In that case, we should clarify the comment.

```suggestion
      // If id is not autogenerated, forceId is not applicable and a value must be specified
```

test/id.test.js
```diff
@@ -0,0 +1,151 @@
...
+const uid = require('./helpers/uid-generator');
+
+let ds;
+const ValidationError = require('..').ValidationError;
+
+describe.only('id property', function() {
+  before(function(done) {
+    ds = getSchema();
+    done();
+  });
```
bajtos, line 22: Nitpick:

```suggestion
  before(function() {
    ds = getSchema();
  });
```

test/id.test.js
```diff
@@ -0,0 +1,151 @@
...
+        await IdCheckC.destroyAll();
+      });
+
+      // If id is not autogenerated and forceId is true, a value must be specified
+      it('should require id value', async () => {
+        await IdCheckC.create({name: 'Pablo'}).should.be.rejected();
+      });
+
+      // If id is not autogenerated and forceId is true, a value can be specified
+      it('should allow id value to be specified', async () => {
```
bajtos, line 81: Ditto.

test/id.test.js
```diff
@@ -0,0 +1,151 @@
...
+      });
+    });
+  });
+
+  context('generated:true', function() {
+    context('with forceId:true (default)', function() {
+      let IdCheckE;
+      before(async () => {
+        IdCheckE = ds.define('IdCheckE', {
+          name: String,
```
bajtos, line 113: I think you need to define the id property here? See my comment above for `generated:false` test context.

test/id.test.js
```diff
@@ -0,0 +1,151 @@
...
+        });
+        await ds.automigrate(IdCheckE.modelName);
+        await IdCheckE.destroyAll();
+      });
+
+      // If id is autogenerated and forceId is true, a value need not be specified
```
bajtos, line 119: When forceId is true, id value *must* not be specified - the client must omit it.

bajtos (changes requested):
The tests look reasonable. 

I find it a bit difficult to spot the difference between `.not.be.rejected()` and `.be.rejected()`. Have you considered to use `.be.fulfilled()` or `.be.resolved()` for success cases? See http://shouldjs.github.io/#assertion-fulfilled

For all `.be.rejected()` checks, please add a constraint to ensure the test passes only when the expected error was reported, the test should fail when a different error was reported. Ideally, the error reported by `forceId` handler should have a unique `err.code` string that we can check in the tests. If that's not feasible, then I am ok to use a RegExp to match the error message, just extract it to a shared constant please.

lib/model-builder.js
```diff
@@ -355,6 +355,11 @@ ModelBuilder.prototype.define = function defineClass(className, properties, sett
       forceId = false;
     }
 
+    // Make id value mandatory if it is not autogenerated
+    if (!idProp.generated) {
+      ModelClass.validatesPresenceOf(idName);
+    }
```
bajtos, line 361: There are three possible values of `generated` field:

- `undefined` - the user did not indicate whether the PK values should be generated by the DB or provided by the caller, expecting the framework to pick a reasonable default
- `true` - the user wants PK values to be generated by the database
- `false` - the user wants PK values to be provided by the caller

As LB user, I would expect the default behavior to be `generated: true` for primary keys. I.e. when I don't provide any `generated` flag, the framework should assume `generated: true`.

Would that reduce the number of tests broken by your change(s)?

bajtos (commented):
> Currently, if PK property generated is false and user does not supply a value, Juggler is autogenerating the primary key (id). I think this is an expected behavior. I encountered this issue while writing tests for the new string interface for ObjectId in the MongoDB connector.

I am little bit confused. When I configure the PK property with `generated: false`, then I would expect juggler to reject requests without any PK value provided. Why do you think it's an expected behavior that we are generating PK values even when `generated` is explicitly set to `false`?

lib/model-builder.js
bajtos: Proposed implementation to try:

```js
if (idProp.generated === false) {
  ModelClass.validatesPresenceOf(idName);
}
```

bajtos:
@hacksparrow PTAL at my comment above, I think the current juggler version treats PKs as `generated: true` by default and if we preserve that behavior, then your patch may not break any tests.

Regardless of which default we choose (`generated: true` or `generated: false`), I need you to describe what's changing (breaking) in a way that will allow reviewers of your pull request to asses impact on our users, and later allow LB users to asses the impact of the upgrade on their projects.

bajtos:
I am looking for information in the following format: "When you do this, then juggler v4 behaves like this but the new version v5 does that instead." 

A made up example:

> When you have a model with a PK defined with `id: true` but no `generated` field specified, we will assume the PK is NOT autogenerated and must be provided by the caller. As a result, the following requests that were valid in juggler v4 will fail with a validation error in juggler v5:
> - `create` with no PK value provided
> - `replaceOrCreate` with no PK value provided

hacksparrow:
> Why do you think it's an expected behavior that we are generating PK values even when generated is explicitly set to false?

Sorry, typo. I meant "unexpected".

hacksparrow:
@bajtos thanks for you review.

> I think the current juggler version treats PKs as generated: true by default and if we preserve that behavior, then your patch may not break any tests.

It is a little tricky.

If the PK property itself is not specified, then it is `generated: true`, but if it is specified (without `generated`) then it is `generated: false`.

Issue #1866, opened 2020-09-15
yarn 2 Qualified path resolution failed - none of the candidates can be found on the disk.


mdbetancourt:
![image](https://user-images.githubusercontent.com/22435967/93167921-af7c5880-f6ef-11ea-88e1-1dbd77a7c1c7.png)
looks like the problem is in the following code
```ts

function connectorModuleNames(name) {
  const names = []; // Check the name as is
  if (!name.match(/^\//)) {
    names.push('./connectors/' + name); // Check built-in connectors
    if (name.indexOf('loopback-connector-') !== 0) {
      names.push('loopback-connector-' + name); // Try loopback-connector-<name>
    }
  }
  // Only try the short name if the connector is not from StrongLoop
  if (['mongodb', 'oracle', 'mysql', 'postgresql', 'mssql', 'rest', 'soap', 'db2', 'cloudant']
    .indexOf(name) === -1) {
    names.push(name);
  }
  return names;
}
```
trying to check for rest file in connector folder but there no exists

i fixed it with

```ts
function connectorModuleNames(name) {
  const names = []; // Check the name as is
  if (!name.match(/^\//)) {
    if(['kv-memory', 'memory', 'transient'].indexOf(name) !== -1) {
      console.log(name)
      names.push('./connectors/' + name); // Check built-in connectors

    } else if (name.indexOf('loopback-connector-') !== 0) {
      names.push('loopback-connector-' + name); // Try loopback-connector-<name>
    }
  }
  // Only try the short name if the connector is not from StrongLoop
  if (['mongodb', 'oracle', 'mysql', 'postgresql', 'mssql', 'rest', 'soap', 'db2', 'cloudant']
    .indexOf(name) === -1) {
    names.push(name);
  }
  return names;
}
```

mdbetancourt:
a workaround is use the full name 
{
connector: "loopback-connector-rest" instead of "rest"
}

ferrantejake:
I am having the same issue with the mongodb connector. It is not resolved by using the connectors full name.

```shell
Cannot start the application. Error: Qualified path resolution failed - none of the candidates can be found on the disk.

Source path: C:\Users\jakef\code\examplerepo\.yarn\cache\loopback-datasource-juggler-npm-4.26.0-67d3cf0749-75cc866df5.zip\node_modules\loopback-datasource-juggler\lib\connectors\loopback-connector-mongodb
[...]

PR #1883, opened 2020-12-04
Remove check for error message matching connector file path.

ferrantejake:
Signed-off-by: ferrantejake <jake.ferrante@hotmail.com>

Removes impossible-to-fulfill condition which expects the error message to match the name of the path of a connector.

Fixes 
- #1866 

## Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -259,8 +259,7 @@ function tryModules(names, loader) {
     try {
       mod = loader(names[m]);
     } catch (e) {
-      const notFound = e.code === 'MODULE_NOT_FOUND' &&
-        e.message && e.message.indexOf(names[m]) > 0;
+      const notFound = e.code === 'MODULE_NOT_FOUND' && e.message;
 
       if (notFound) {
         debug('Module %s not found, will try another candidate.', names[m]);


bajtos (changes requested):
The current implementation is intentional, to prevent juggler from swallowing an error thrown when the connector exists but cannot load one of its dependencies.

See https://github.com/strongloop/loopback-datasource-juggler/pull/1241

bajtos:
Thank you @ferrantejake for the pull request. I appreciate your effort to improve LoopBack ❤️ and apologize for a long silence on our side.

I am fine to change the way how we deal with loading connectors in juggler as long as we don't make it confusing for LB users to troubleshoot the situation when a connector is installed and cannot be landed, as described e.g. in https://github.com/strongloop/loopback-datasource-juggler/pull/1241.

Our current implementation relies on the behavior of Node.js built-in module loader and I believe it works well. As you can see in the screenshot below, the error message reported by node includes the module name:

```
$ node
Welcome to Node.js v14.16.0.
Type ".help" for more information.
> require('loopback-connector-foobar')
Uncaught Error: Cannot find module 'loopback-connector-foobar'
```

The issue you are trying to fix - https://github.com/strongloop/loopback-datasource-juggler/issues/1866 - mentions Yarn2, I assume we are talking about Yarn's [plug'n'play](https://next.yarnpkg.com/features/pnp)? I like the idea of supporting that feature, but only if we can find a reasonable solution that would not degrade the user experience for other users.

One option that comes to my mind is to leverage `require.resolve` to check if the connector exists in project dependencies. 

Current solution (at high level):

```js
try {
  require(names[m]);
} catch (err) {
  // inspect err to find out if the module was not found
  // or whether there was some other error
}
```

Proposed solution (at high level) assuming Yarn2 works as expected:

```js
try {
  require.resolve(names[m]);
} catch (err) {
  // connector not found
}

// no try-catch block, we want to report
// any errors to the caller
mod = require(names[m]);
```

What do you think?

ferrantejake:
Thanks @bajtos for the response! I'm excited to be able to contribute!

Ah, that's interesting. I see what you mean where the node require error message includes the module name. Yarn2's error message does not include the module name. Instead it has a generic message:
```
Qualified path resolution failed - none of the candidates can be found on the disk.
```
see: https://github.com/strongloop/loopback-datasource-juggler/issues/1866#issuecomment-738238578

So the goal of this check is to determine if the connector is not found or if one of the dependencies the connector relies on is not found. If there is a way to determine the differece when using Yarn, I am no aware such a method. Given that Yarn2's error message does not contain the name of the module it tried to find, might I propose the following: 

_if the loader is using Yarn2, then skip the check for the error message containing the module name._

I cannot find authoritative information from Yarn on how to determine if the script is run using yarn, however many packages geared towards figuring this out refer to `process.env.npm_config_user_agent` to determine if yarn is being used. 

A quick test:
```shell
$ node
Welcome to Node.js v14.11.0.
Type ".help" for more information.
> process.env.npm_user_agent_version
undefined

$ yarn node
Welcome to Node.js v14.11.0.
Type ".help" for more information.
> process.env.npm_config_user_agent
'yarn/2.3.2 npm/? node/14.11.0 win32 x64'
```


So this would look something like:
```typescript
// user agent might not be present
let usingYarn = false;
if(process.env.npm_config_user_agent && process.env.npm_config_user_agent .contains('yarn') {
   usingYarn = true;
}

try{
   require(module)
} catch(e) {
   if(usingYarn) {
      // perform notfound check without check for module name in error message.
   } else {
      // perform notfound check with check fo module name in error message.
   }
}
```

What do you think about this approach?

Issue #1886, opened 2020-12-10
bug EmbedsMany.prototype.create


lidafeo:
## Steps to reproduce

I have 3 models:

1. clone the sandbox https://github.com/lidafeo/loopback-sandbox/tree/bug-embedded-relations
2. add a data source for Project Model (I used mongodb)
3. start app
4. run POST /Projects/setStages

## Current Behavior

If I call the function EmbedsMany.prototype.create, 12 executives added successfully, but on the thirteenth I get an error:
node_modules\loopback-datasource-juggler\lib\relation-definition.js:2626
inst.trigger = function(actionName, work, data, callback) {
^
RangeError: Maximum call stack size exceeded
at Stage.inst.trigger

## Link to reproduction sandbox

https://github.com/lidafeo/loopback-sandbox/tree/bug-embedded-relations

## Additional information

**I think the problem is in file loopback-datasource-juggler\lib\relation-definition.js in line:**
`if (inst && inst.triggerParent !== 'function') {`
missing typeof

Version loopback-datasource-juggler@3.28.0
Version loopback@3.28.0
node.js v10.17.0
Operating system - Windows 10

lidafeo:
## Link to reproduction sandbox
https://github.com/lidafeo/loopback-sandbox/tree/bug-embedded-relations

PR #1887, opened 2020-12-14
fix prepareEmbeddedInstance

lidafeo:
Issue #1886: bug EmbedsMany.prototype.create


dhmlau:
@slnode test please

bajtos (changes requested):
Thank you @lidafeo for the pull request. The fix looks reasonable to me. Could you please add a test to verify the new version works as expected and to prevent regressions in the future?

PR #1892, opened 2021-05-26
Fixed bug preventing includes when related id = 0

mvertopoulos:
This fixes an issue that was responsible for preventing includes on related records which have an id (targetId) of zero (0), similarly to this fix:
https://github.com/strongloop/loopback/pull/3719

ping @bajtos

### DIFF
--- a/lib/include.js
+++ b/lib/include.js
@@ -903,7 +903,7 @@ Inclusion.include = function(objects, include, options, cb) {
           }
         }
         const targetId = obj[relation.keyFrom];
-        if (targetId) {
+        if (targetId != null) {
           targetIds.push(targetId);
           const targetIdStr = targetId.toString();
           objTargetIdMap[targetIdStr] = objTargetIdMap[targetIdStr] || [];


dhmlau:
@mvertopoulos, thanks for the PR. 
Could you please add some tests to verify your change? And also take care of the commit linter error too? 
```
**************************************************
**
**  Linting commit logs
**
**  1 problems found:
**    49c8d679c5 - Fix for allowing includes where related id (targetI: First line should be 50 characters or less (saw 61)
**
**************************************************
```

Thanks.

lib/include.js
```diff
@@ -903,7 +903,7 @@ Inclusion.include = function(objects, include, options, cb) {
           }
         }
         const targetId = obj[relation.keyFrom];
-        if (targetId) {
+        if (targetId != null) {
```
marioestradarosa, line 906: Could you please use ``!==`` as such. The !== has faster performance, than the != operator because the Strict Equality Comparison Algorithm doesn't typecast values.

```ts
if (targetId !== null) 

```

marioestradarosa (changes requested):
Thank you for bringing this up, could you please a test?

PR #1904, opened 2021-09-22
feat: add connector TS types

achrinza:
Signed-off-by: Rifa Achrinza <25147899+achrinza@users.noreply.github.com>

## Checklist

### DIFF
--- a/types/connector.d.ts
+++ b/types/connector.d.ts
@@ -3,12 +3,58 @@
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
 
-import {Callback, DataSource, Options, PromiseOrVoid} from '..';
+import {Callback, DataSource, ModelBaseClass, ModelDefinition, ModelProperties, Options, PromiseOrVoid, PropertyDefinition, PropertyType, Schema} from '..';
 
 // Copyright IBM Corp. 2018. All Rights Reserved.
 // Node module: loopback-datasource-juggler
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
+
+export type DiscoveryScopeOptions = {
+  owner?: string,
+  all?: boolean,
+  views?: boolean,
+  limit?: number,
+  offset?: number,
+}
+
+export type SchemaDiscoveryOptions = {
+  owner?: string,
+  relations?: boolean,
+  all?: boolean,
+  views?: boolean,
+}
+
+export type PrimaryKeysDiscoveryReturnValue = {
+  owner?: string | null,
+  tableName?: string,
+  columnName?: string,
+  keySeq?: number,
+  pkName?: string | null,
+}
+
+export type ForeignKeysDiscoveryReturnValue = {
+  fkOwner?: string | null,
+  fkName?: string | null,
+  fkTableName?: string,
+  fkColumnName?: string,
+  keySeq?: number,
+  pkOwner?: string | null,
+  pkName?: string | null,
+  pkTableName?: string,
+  pkColumnName?: string,
+}
+
+export type ModelPropertiesDiscoveryReturnValue = {
+  owner?: string,
+  tableName?: string,
+  columnName?: string,
+  dataType?: string,
+  dataLength?: number,
+  dataPrecision?: number,
+  dataScale?: number,
+  nullable?: boolean,
+}
 
 /**
  * Connector from `loopback-connector` module
@@ -16,10 +62,46 @@ import {Callback, DataSource, Options, PromiseOrVoid} from '..';
 export interface Connector {
   name: string; // Name/type of the connector
   dataSource?: DataSource;
-  connect(callback?: Callback): PromiseOrVoid; // Connect to the underlying system
-  disconnect(callback?: Callback): PromiseOrVoid; // Disconnect from the underlying system
+  connect?(callback?: Callback): PromiseOrVoid; // Connect to the underlying system
+  disconnect?(callback?: Callback): PromiseOrVoid; // Disconnect from the underlying system
   ping(callback?: Callback): PromiseOrVoid; // Ping the underlying system
   execute?(...args: any[]): Promise<any>;
+  getDefaultIdType(): object;
+  getTypes?(): string[];
+  define?(def: {model: ModelBaseClass, properties: PropertyDefinition, settings: ModelDefinition['settings']}): void;
+  defineProperty?(model: string, prop: string, params: PropertyDefinition): void;
+  defineForeignKey?(modelName: string, key: string, foreignModelName: string, cb: Callback<PropertyType>): void;
+  defineForeignKey?(modelName: string, key: string, cb: Callback<PropertyType>): void;
+  discoverModelDefinitions?(options: DiscoveryScopeOptions, cb?: Callback): Promise<ModelDefinition>;
+  /**
+   * @deprecated
+   */
+  discoverModelDefinitionsSync?(options: DiscoveryScopeOptions): ModelDefinition
+  discoverModelProperties?(modelName: string, options: DiscoveryScopeOptions, cb: Callback<ModelPropertiesDiscoveryReturnValue>): Promise<ModelPropertiesDiscoveryReturnValue>;
+  /**
+   * @deprecated
+   */
+  discoverModelPropertiesSync?(modelName: string, options: DiscoveryScopeOptions): ModelPropertiesDiscoveryReturnValue;
+  discoverPrimaryKeys?(modelName: string, options: {owner?: string}, cb: Callback<PrimaryKeysDiscoveryReturnValue>): Promise<PrimaryKeysDiscoveryReturnValue>;
+  /**
+   * @deprecated
+   */
+  discoverPrimaryKeysSync?(modelName: string, options: {owner?: string}): PrimaryKeysDiscoveryReturnValue;
+  discoverForeignKeys?(modelName: string[], options: {owner?: string}, cb: Callback<ForeignKeysDiscoveryReturnValue>): Promise<ForeignKeysDiscoveryReturnValue>;
+  /**
+   * @deprecated
+   */
+  discoverForeignKeysSync?(modelName: string[], options: {owner?: string}): ForeignKeysDiscoveryReturnValue;
+  discoverExportedForeignKeys?(modelName: string, options: {owner?: string}, cb: Callback<ForeignKeysDiscoveryReturnValue>): Promise<ForeignKeysDiscoveryReturnValue>;
+  /**
+   * @deprecated
+   */
+  discoverExportedForeignKeysSync?(modelName: string, options?: {owner?: string}): ForeignKeysDiscoveryReturnValue;
+  discoverSchemas(tableName: string, options: SchemaDiscoveryOptions, cb: Callback<Schema>): Promise<Schema>;
+
+  isActual?(models: string[], cb: Callback<boolean>): void;
+  freezeDataSource?(): void;
+  freezeSchema?(): void;
   [property: string]: any; // Other properties that vary by connectors
 }
 


types/connector.d.ts
```diff
@@ -3,28 +3,285 @@
...
   name: string; // Name/type of the connector
-  dataSource?: DataSource;
-  connect(callback?: Callback): PromiseOrVoid; // Connect to the underlying system
-  disconnect(callback?: Callback): PromiseOrVoid; // Disconnect from the underlying system
-  ping(callback?: Callback): PromiseOrVoid; // Ping the underlying system
+
+  /**
+   * @internal
+   */
+  _models?: Record<string, ModelBaseClass>;
```
achrinza, line 111: Not sure if this is correct.

types/connector.d.ts
```diff
@@ -3,28 +3,285 @@
...
+  /**
+   * Define a property on the target model.
+   * 
+   * @param model Name of model
+   * @param prop Name of property
+   * @param params Property settings
+   */
+  defineProperty?(model: string, prop: string, params: PropertyDefinition): void;
+  defineForeignKey?(modelName: string, key: string, foreignModelName: string, cb: Callback<PropertyType>): void;
+  defineForeignKey?(modelName: string, key: string, cb: Callback<PropertyType>): void;
```
achrinza, line 152: There's several of these functions that accept a callback, but I'm not sure if its optional.

types/datasource.d.ts
```diff
@@ -76,28 +86,194 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
  */
 export declare class DataSource extends EventEmitter {
   name: string;
-  settings: Options;
+  settings: ConnectorSettings;
```
achrinza, line 89: Based on the codebase, the DataSource simply passes the whole settings to the connector; Hence why the type called `ConnectorSettings`.

types/datasource.d.ts
```diff
@@ -76,28 +86,194 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
+   * {@inheritDoc Connector.getTypes}
+   * @deprecated
+   */
+  getTypes(): string[];
+
+  /**
+   * Check if the datasource supports the specified types.
+   * @param types Type name(s) to check against
+   */
+  supportTypes(types: string | string[]): boolean;
```
achrinza, line 189: Based on the original codebase comments, `getTypes()` is deprecated but not `supportTypes()`; Not sure of the reason for this. The original purpose of these functions are also not clear.

types/datasource.d.ts
```diff
@@ -76,28 +86,194 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
+  columnNames(modelName: string): string;
+
+  /**
+   * Retrieve coulmn metadata for the specified `modelName` and `propertyName`.
+   * 
+   * @param modelName Target model name
+   * @param propertyName Target property name
+   * @returns Column metadata
+   */
+  columnMetadata(modelName: string, propertyName: string): ColumnMetadata;
```
achrinza, line 233: This seems to be the only function that uses `ColumnMetadata`. However, `ColumnMetadata` typedef is non-specific. Need to find examples of the return value.

types/datasource.d.ts
```diff
@@ -76,28 +86,194 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
+   */
+  idNames(modelName: string): string[];
+
+  /**
+   * Retrieve the ID property definition.
+   * 
+   * @param modelName Target model name
+   * @returns The ID property definition
+   */
+  idProperty(modelName: string): PropertyDefinition;
```
achrinza, line 257: There's also a `IdDefinition` in `model.d.ts`. Not sure if this function is supposed to return that instead.

types/model.d.ts
```diff
@@ -77,6 +84,11 @@ export interface ModelProperties {
 export interface ModelSettings extends AnyObject {
   strict?: boolean;
   forceId?: boolean;
+  persistUndefinedAsNull?: boolean;
+  protectedProperties?: string[];
+  protected?: string[];
+  hiddenProperties?: string[];
+  hidden?: string[];
```
achrinza, line 91:
```suggestion
  hiddenProperties?: string[];
  /**
    * Overriden by {@link ModelSettings.hiddenProperties}.
    *
    * @deprecated Use {@link ModelSettings.hiddenProperties} instead
    **/
  hidden?: string[];
```

types/connector.d.ts
```diff
@@ -3,28 +3,285 @@
...
+  /**
+   * Check whether or not migrations are required for the database schema to match
+   * the model definitions attached to the {@link DataSource}.
+   * 
+   * @param models Name of models to check. If not defined, all models are checked.
+   * @param cb 
+   */
+  isActual?(models?: string | string[], cb?: Callback<boolean>): void;
+  freezeDataSource?(): void;
+  freezeSchema?(): void;
```
achrinza, line 249: Not sure the difference between these functions. `DataSource.prototype.freeze` calls both functions if they are defined.

types/connector.d.ts
```diff
@@ -3,28 +3,285 @@
...
+  Model: ModelBaseClass,
+  instance?: object,
+  query?: Filter,
+  where?: Where,
+  data?: AnyObject,
+  hookState?: AnyObject,
+  options?: Options,
+  isNewInstance?: boolean,
+  currentInstance?: Readonly<ModelBase>,
+}
```
achrinza, line 100: These are defined here:

- https://loopback.io/doc/en/lb3/Operation-hooks.html
- https://loopback.io/doc/en/lb3/Remote-hooks.html#context-object
- https://loopback.io/doc/en/lb3/Connector-hooks.html#context

Currently, this is only based on all the possible attributes for an Operation Hooks context.

Will probably need to add separate context typedefs for the other 2.

While the Operation Hooks context differs depending on the event being observed and the function called, I'm not sure if its better to create distinctive op. hook contexts for these different hooks.

types/connector.d.ts
achrinza: It seems to have been introduced by this commit:

https://github.com/loopbackio/loopback-datasource-juggler/commit/9b169efc8c7277e69947239a7aff5bc64383e870#diff-7de38f47065ef67b4936fbd6396bb71a580fe9a5a9ff2c9723c4fe859d1e2a6eR977-R979

achrinza: Based on the commit just before the one mentioned above, they should be identical, and the use of both is for backwards-compatibility with JugglingDB:

https://github.com/loopbackio/loopback-datasource-juggler/commit/6af4b1ba2f5894c32d960258f43b38f2b69a37e1

types/datasource.d.ts
```diff
@@ -76,28 +90,196 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
+   */
+  columnName(modelName: string, propertyName: string): string;
+
+  /**
+   * Retrieve the column names for the specified `modelName`.
+   * 
+   * @param modelName Target model name
+   * @returns Column names
+   */
+  columnNames(modelName: string): string;
```
achrinza, line 230: Is this correct? Plural `columnNames` does not make sense with a single `string` return type.

types/datasource.d.ts
```diff
@@ -76,28 +90,196 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
+    settings?: ConnectorSettings,
     modelBuilder?: ModelBuilder,
   );
 
+  private setup(dsName: string, settings: ConnectorSettings): void;
+  private setup(settings: ConnectorSettings): void;
+
+  private _setupConnector();
+
+  private mixin<T extends object>(ModelCtor: T): T
```
achrinza, line 173: Add tsdoc explaining purpose of each function.

types/datasource.d.ts
```diff
@@ -76,28 +90,196 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
+   * 
+   * @remarks
+   * This default can be overriden through
+   * {@link ConnectorSettings.maxOfflineRequests}.
+   */
+  static DEFAULT_MAX_OFFLINE_REQUESTS: number;
+
+  /**
+   * A hash-map of the different relation types.
+   */
```
achrinza, line 155: Let's include an example since it's not expected to change.

types/datasource.d.ts
```diff
@@ -76,28 +90,196 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
+  /**
+   * {@inheritDoc DataSource.connector}
+   * @deprecated Use {@link DataSource.connector} instead.
+   */
+  adapter?: Connector;
 
+  /**
+   * Connector instance.
+   */
   connector?: Connector;
```
achrinza, line 108: Replaced this with `BuiltConnector`?

types/datasource.d.ts
```diff
@@ -76,28 +90,196 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
 
   initialized?: boolean;
   connected?: boolean;
   connecting?: boolean;
+  
+  /**
+   * {@inheritDoc DataSource.connector}
+   * @deprecated Use {@link DataSource.connector} instead.
+   */
+  adapter?: Connector;
```
achrinza, line 103: Replace this with `BuiltConnector`?

types/datasource.d.ts
```diff
@@ -76,28 +90,196 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
  */
 export declare class DataSource extends EventEmitter {
   name: string;
-  settings: Options;
+  settings: ConnectorSettings;
 
   initialized?: boolean;
   connected?: boolean;
   connecting?: boolean;
```
achrinza, line 97: Add tsdoc explaining the circumstances for the possible values.

types/connector.d.ts
```diff
@@ -3,28 +3,348 @@
...
+  log: DataSource['log'];
+  logger(query: string, start: number): (query: string) => void;
+}
+
 /**
  * Base connector class
+ * 
+ * @internal
  */
 export declare class ConnectorBase implements Connector {
```
achrinza, line 349: May need to extract this to `loopback-connector` package along with transaction-related typedefs.

types/model.d.ts
```diff
@@ -328,6 +520,24 @@ export declare class ModelBase {
...
+    indexes?: {patch: boolean};
+    methods?: {patch: boolean};
+    mixins?: {patch: boolean};
+    relations?: {patch: boolean};
+    scope?: {replace: boolean};
+    acls?: {rank: boolean};
+    __delete?: boolean | null;
+    __default?: {replace: boolean};
+  }
 }
```
achrinza, line 541: Spin off model merge config to its own interface; Add tsdoc for each config property.

types/model.d.ts
```diff
@@ -74,9 +90,159 @@ export interface ModelProperties {
...
+   * @remarks
+   * Defaults to `true`.
+   */
+  idInjection?: boolean;
+  
+  plural?: string;
+
+  http?: {
+    path?: string;
+  }
```
achrinza, line 111: Deprecate; Add tsdoc explaining their purpose, defaults, and relationship between each other (e.g. `plural` is used for `http.path` when the latter is not set)

types/model.d.ts
```diff
@@ -74,9 +88,127 @@ export interface ModelProperties {
...
+       * @remarks
+       * Handled by {@link Connector}s directly by default.
+       * 
+       * Overrides {@link ModelSettings.indexes.columns}.
+       */
+      keys: string[];
+
+      // Postgresql-specific
+      type: string;
+      kind: string;
```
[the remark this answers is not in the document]
achrinza: Also, add tsdoc

types/model.d.ts
```diff
@@ -74,9 +90,159 @@ export interface ModelProperties {
...
+  
+  plural?: string;
+
+  http?: {
+    path?: string;
+  }
+
+  /**
+   * @remarks
+   * Alias of {@link ModelSettings.super}. Takes lower precedence.
```
achrinza, line 115:
```suggestion
   * Alias of {@link ModelSettings.super}. Takes higher precedence.
```

types/model-utils.d.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export interface ModelUtilsOptions {
+    /**
+     * Sets if non-standard {@link Where} operators are permitted in passed
+     * {@link Filter}.
+     */
+    allowExtendedOperators?: boolean;
+    maxDepthOfQuery?: number;
+    maxDepthOfData?: number;
+    prohibitHiddenPropertiesInQuery?: boolean;
```
achrinza, line 21: Explain these options with tsdoc.

types/datasource.d.ts
```diff
@@ -76,28 +90,238 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
+  /**
+   * Retrieve the column names for the specified `modelName`.
+   * 
+   * @param modelName Target model name
+   * @returns Column names
+   */
+  columnNames(modelName: string): string;
+
+  /**
+   * Retrieve coulmn metadata for the specified `modelName` and `propertyName`.
```
achrinza, line 271: Typo.
```suggestion
   * Retrieve column metadata for the specified `modelName` and `propertyName`.
```

types/datasource.d.ts
```diff
@@ -273,18 +524,24 @@ export declare class DataSource extends EventEmitter {
...
     options: Options,
     callback: Callback<AnyObject[]>,
-  ): void;
+  ): void; 
 
   buildModelFromInstance(
     modelName: string,
     jsonObject: AnyObject,
     options?: Options,
   ): ModelBaseClass;
```
achrinza, line 533: Add documentation on what (Connector?) methods this calls behind the scenes.

types/datasource.d.ts
achrinza: On second thought, we may want to create a dedicated passthrough type to avoid potential future issues due to the shared type.

types/datasource.d.ts
```diff
@@ -76,28 +90,238 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
   connecting?: boolean;
+  
+  /**
+   * {@inheritDoc DataSource.connector}
+   * @deprecated Use {@link DataSource.connector} instead.
+   */
+  adapter?: Connector;
 
+  /**
+   * Connector instance.
```
achrinza, line 106: Document that it's the preferred alias of `adapter`.

types/datasource.d.ts
```diff
@@ -76,28 +90,238 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
 
-  constructor(name: string, settings?: Options, modelBuilder?: ModelBuilder);
+  private _queuedInvocations: number;
 
-  constructor(settings?: Options, modelBuilder?: ModelBuilder);
+  private _operations: Record<string, OperationOptions>;
+
+  operations(): Record<string, OperationOptions>;
+
+  defineOperation(name: string, options: OperationOptions, fn: Function): void;
```
achrinza, line 138: These look like they're meant exclusively for "HTTP Remoting" operations as well. We may want to mark them as deprecated.

types/connector.d.ts
```diff
@@ -3,28 +3,384 @@
...
+  connectionTimeout?: number;
+  maxOfflineRequests?: number;
+  lazyConnect?: boolean;
+  debug?: boolean;
+
+  // Postgres-specific
+  defaultIdSort?: boolean | 'numericIdOnly';
+  onError?: (err: Error | unknown) => unknown | 'ignore';
+  // END Postgres-specific
+}
```
achrinza, line 29: Add `loopback-connector-mssql`-specific settings:

```suggestion
  tableNameID?: string
}
```

#TODO: Figure out what this setting does.

types/connector.d.ts
```diff
@@ -3,28 +3,384 @@
...
+   * 
+   * If the connector DDL and DML functions map directly to a hash-map of
+   * model properties and their values, this function does not need to be
+   * implemented.
+   * 
+   * @param context 
+   * @param dbResponse 
+   */
+  generateContextData?(context: Context, dbResponse: unknown): Context;
+
```
achrinza, line 370: Add missing migration functions:

dropTable(model, cb)
alterTable(model, actualFields, actualIndexes, cb)
createTable(model, cb)
addIndexes(model, actualIndexes)
showIndexes(model, cb)
showFields(model, cb)
getColumnsToAdd(model, actualFields)
getColumnsToDrop(model, actualFields)
getAddModifyColumns(model actualFields)
propertiesSQL(model) <-- Alias in PostgreSQL: buildColumnDefinitions

The above were found in MSSQL and PostgreSQL connectors.

types/model.d.ts
```diff
@@ -219,7 +405,11 @@ export declare class ModelBase {
    * If true, then protected properties should not be brought out.
    * @returns {object} returns Plain JSON object
```
achrinza, line 406: Convert to tsdoc

types/model.d.ts
```diff
@@ -74,9 +90,159 @@ export interface ModelProperties {
...
+  unresolved?: boolean;
+
+  indexes?: {
+    [indexJugglerName: string]: {
+      name: string;
+      /**
+       * Comma-separated column names
+       * 
+       * @remarks
+       * Handled by {@link Connector}s directly by default.
```
achrinza, line 138: What does "by default" mean? What is the circumstance where this is not the case, and what is the impact?

types/model.d.ts
```diff
@@ -74,9 +90,159 @@ export interface ModelProperties {
...
+       * 
+       * Overriden by {@link ModelSettings.indexes.keys}.
+       */
+      columns: string;
+
+      /**
+       * Array of column names to create an index.
+       * 
+       * @remarks
+       * Handled by {@link Connector}s directly by default.
```
achrinza, line 148: Ditto. What does "by default" mean? What is the circumstance where this is not the case, and what is the impact?

types/connector.d.ts
achrinza: Most probably `ModelDefinition`.

types/datasource.d.ts
```diff
@@ -76,28 +90,238 @@ import {IsolationLevel, Transaction} from './transaction-mixin';
...
+   * Connector instance.
+   */
   connector?: Connector;
 
-  definitions: {[modelName: string]: ModelDefinition};
+  modelBuilder: ModelBuilder;
 
+  models: Record<string, typeof ModelBase>;
+  
+  definitions: {[modelName: string]: ModelDefinition};
```
achrinza, line 114: Will need to check and document these' relationship with Connector._models

types/connector.d.ts
```diff
@@ -3,28 +3,384 @@
...
 /**
  * Connector from `loopback-connector` module
  */
 export interface Connector {
   name: string; // Name/type of the connector
-  dataSource?: DataSource;
-  connect(callback?: Callback): PromiseOrVoid; // Connect to the underlying system
-  disconnect(callback?: Callback): PromiseOrVoid; // Disconnect from the underlying system
-  ping(callback?: Callback): PromiseOrVoid; // Ping the underlying system
+
```
achrinza, line 183:
```suggestion
define?(modelDefinition: ModelDefinition): void;
```

types/datasource.d.ts
achrinza: These are populated from the same-named properties of the ModelBuilder instance passed during class initialisation.

types/model-utils.d.ts
```diff
@@ -0,0 +1,22 @@
...
+}
```
achrinza, line 22: Add missing `normalizeUndefinedInQuery`

See: https://github.com/loopbackio/loopback-datasource-juggler/blob/b3b5dc2cd58b6307d8c7b680e8cab4250dd0ac55/lib/model-utils.js#L319

types/connector.d.ts
```diff
@@ -3,28 +3,384 @@
...
 
-import {Callback, DataSource, Options, PromiseOrVoid} from '..';
+import { AnyObject } from 'strong-globalize/lib/config';
+import {Callback, DataSource, Filter, ModelBase, ModelBaseClass, ModelDefinition, ModelProperties, Options, PromiseOrVoid, PropertyDefinition, PropertyType, Schema, Where} from '..';
 
-// Copyright IBM Corp. 2018. All Rights Reserved.
-// Node module: loopback-datasource-juggler
-// This file is licensed under the MIT License.
-// License text available at https://opensource.org/licenses/MIT
+export interface ConnectorSettings extends Options {
```
achrinza, line 9: Should extend `ModelUtilsOptions`

types/model.d.ts
```diff
@@ -74,9 +90,159 @@ export interface ModelProperties {
...
+   * @remarks
+   * 
+   * This setting is used by these operations:
+   * 
+   * - {@link DataAccessObject.removeById}/{@link DataAccessObject.destroyById}/{@link DataAccessObject.deleteById}
+   * - {@link DataAccessObject.remove}/{@link DataAccessObject.delete}/{@link DataAccessObject.destroy}
+   */
+  strictDelete?: boolean;
+
+  updatOnly?: boolean;
```
achrinza, line 242: suggestion(blocking): fix typo

```suggestion
  updateOnly?: boolean;
```

types/model.d.ts
```diff
@@ -74,9 +115,135 @@ export interface ModelProperties {
...
+  /**
+   * Mapped table name for the model.
+   */
+  table?: string;
+
+  /**
+   * Sets if JavaScript {@link undefined} as an attribute value should be
+   * persisted as database `NULL`.
+   */
+  persistUndefinedAsNull?: boolean;
```
achrinza, line 186: What's the default value?

types/model.d.ts
```diff
@@ -74,9 +115,135 @@ export interface ModelProperties {
...
+   * for delete operation.
+   *
+   * @remarks
+   *
+   * This setting is used by these operations:
+   *
+   * - {@link DataAccessObject.removeById}/{@link DataAccessObject.destroyById}/{@link DataAccessObject.deleteById}
+   * - {@link DataAccessObject.remove}/{@link DataAccessObject.delete}/{@link DataAccessObject.destroy}
+   */
+  strictDelete?: boolean;
```
achrinza, line 241: What's the default value? IIRC, it's `false` as LoopBack 4's `DefaultCrudRepository` explicitly sets this to `true`.

types/model.d.ts
```diff
@@ -7,21 +7,37 @@ import {EventEmitter} from 'events';
 import {AnyObject, Options} from './common';
 import {DataSource} from './datasource';
 import {Listener, OperationHookContext} from './observer-mixin';
+import {ModelUtilsOptions} from './model-utils';
```
achrinza, line 11: `interface Schema` has an `options` object property?

https://github.com/loopbackio/loopback-next/blob/1a51f322cc618a553dda346c1fe4e3905e794746/packages/cli/lib/model-discoverer.js#L34

types/model.d.ts
```diff
@@ -115,15 +282,71 @@ export declare class ModelDefinition extends EventEmitter implements Schema {
...
+
+interface ModelMergePolicy {
+  description?: {
+    replace: boolean;
+  };
+  properties?: {
+    patch: boolean;
+  }; // Only referenced in "legacy built-in merge policy"
+  options?: {
+    path: boolean;
```
achrinza, line 300: This looks like a typo
```suggestion
    patch: boolean;
```

types/model.d.ts
```diff
@@ -74,9 +115,135 @@ export interface ModelProperties {
...
+
+  /**
+   * {@inheritDoc ModelSettings.tableName}
+   */
+  tableName?: string
+
+  /**
+   * Mapped table name for the model.
+   */
+  table?: string;
```
achrinza, line 180: Which one takes precedence?

types/datasource.d.ts
```diff
@@ -13,9 +13,36 @@ import {
...
+import { ColumnMetadata, ConnectorSettings, ModelBase, ModelSettings, PersistedModel } from '..';
+import { PersistedModelClass } from './persisted-model';
+
+export type OperationOptions = {
+  accepts: string[],
+  returns: string[],
+  http: object,
+  remoteEnabled: boolean,
+  scope: unknown,
+  fnName: string,
```
achrinza, line 25: Refine TS typedef according to https://github.com/loopbackio/loopback-connector-openapi/blob/88cbea3ebcfe0fd7634e3fb49245bf8c15aa6a7a/README.md#extend-a-model-to-wrapmediate-api-operations

achrinza: Refine `scope` according to https://github.com/loopbackio/loopback-datasource-juggler/blob/be09d5774ef9f03c2a4c9c1c422a48b17b3a89df/lib/datasource.js#L187.

```suggestion
  accepts: string[],
  returns: string[],
  http: object,
  remoteEnabled: boolean,
  scope: DataAccessObject,
  fnName: string,
```

Note that we'll need to write `DataAccessObject` itself, which is only used to be mixin-ed into `DataSource` and `BaseModel`.

PR #1993, opened 2022-10-16
fix: add missing property in model generated

### LINKED ISSUE #899: Auto-discover auto_increment / identity field into property with generated = true and not be required
Currently, when I discover a model from a SQL Server table with an identity field or a MySQL table  auto_increment field, that model will not have that field's `generated` property set to true.

I propose to change that. So that I when I discover a table with an auto numbered field into a fresh model. I can immediately do a Model.create without getting the error message _"The `model_name` instance is not valid. Details: `primary_key_name` can't be blank (value: null)."_

What we can do to accomplish that is to alter loopback-connector-xxx/lib/queryColumns to return also a `generated` boolean to capture if a column is auto numbered by the database.

For [mssql you'd add](http://stackoverflow.com/questions/87747/how-do-you-determine-what-sql-tables-have-an-identity-column-programatically):

```
COLUMNPROPERTY(object_id(table_schema+'.'+table_name), column_name, 'IsIdentity') AS generated
```

For [mysql you'd add](http://stackoverflow.com/questions/13772019/sql-check-if-a-column-auto-increments):

```
CASE WHEN extra LIKE '%auto_increment%' THEN 1 ELSE 0 END AS generated
```

Then here in loopback-datasource-juggler/lib/datasource.js method discoverSchemas at about line 1354, You'd change this

```
schema.properties[propName] = {
  type: item.type,
  required: (item.nullable === 'N' || item.nullable === 'NO'
  || item.nullable === 0 || item.nullable === false),
  length: item.dataLength,
  precision: item.dataPrecision,
  scale: item.dataScale
};
```

to

```
schema.properties[propName] = {
  type: item.type,
  required: (item.nullable === 'N' || item.nullable === 'NO'
  || item.nullable === 0 || item.nullable === false) && !item.generated,
  length: item.dataLength,
  precision: item.dataPrecision,
  scale: item.dataScale,
  generated: item.generated
};
```

(and in discoverSchemasSync too)

What do you think?

aaqilniz:
Signed-off-by: Muhammad Aaqil <aaqilniz@yahoo.com>

This PR adds a missing property `generated` to the model. The issue is happening while using lb4 to discover models. Discovering a model that has a constraint of auto_increment set to true doesn't work correctly. The discovery fails to pick up the auto_increment constraint. This was happening due to the fact that `loopback-datasource-juggler` was not attaching the property to the model properties being passed to the discover generator. This PR fixes that and attaches `generated: true/false` to the model. And sets `required` to `false` if a field has `generated` set to `true`.

Fixes #899

## Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -1666,11 +1666,12 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
       const propName = nameMapper('column', item.columnName);
       schema.properties[propName] = {
         type: item.type,
-        required: (item.nullable === 'N' || item.nullable === 'NO' ||
+        required: !item.generated && (item.nullable === 'N' || item.nullable === 'NO' ||
           item.nullable === 0 || item.nullable === false),
         length: item.dataLength,
         precision: item.dataPrecision,
         scale: item.dataScale,
+        generated: item.generated,
       };
 
       if (pks[item.columnName]) {
@@ -1683,6 +1684,7 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
         dataPrecision: item.dataPrecision,
         dataScale: item.dataScale,
         nullable: item.nullable,
+        generated: item.generated,
       };
       // merge connector-specific properties
       if (item[dbType]) {
@@ -1830,10 +1832,11 @@ DataSource.prototype.discoverSchemasSync = function(modelName, options) {
     const propName = nameMapper('column', item.columnName);
     schema.properties[propName] = {
       type: item.type,
-      required: (item.nullable === 'N'),
+      required: !item.generated && item.nullable === 'N',
       length: item.dataLength,
       precision: item.dataPrecision,
       scale: item.dataScale,
+      generated: item.generated,
     };
 
     if (pks[item.columnName]) {
@@ -1846,6 +1849,7 @@ DataSource.prototype.discoverSchemasSync = function(modelName, options) {
       dataPrecision: item.dataPrecision,
       dataScale: item.dataScale,
       nullable: i.nullable,
+      generated: i.generated,
     };
   });
 
--- a/test/discovery.test.js
+++ b/test/discovery.test.js
@@ -34,6 +34,7 @@ describe('Memory connector with mocked discovery', function() {
       dataPrecision: null,
       dataScale: null,
       nullable: 0,
+      generated: true,
     },
     {
       owner: 'STRONGLOOP',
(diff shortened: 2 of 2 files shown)


achrinza (approved):
The changes LGTM :+1: Would like another maintainer's approval on this before merging (e.g. @dhmlau).

@aaqilniz Please squash the PR into 1 commit with the DCO sign-off, without any merge commits.

lib/datasource.js
```diff
@@ -1666,11 +1666,12 @@ DataSource.prototype.discoverSchemas = function(tableName, options, cb) {
       const propName = nameMapper('column', item.columnName);
       schema.properties[propName] = {
         type: item.type,
-        required: (item.nullable === 'N' || item.nullable === 'NO' ||
+        required: item.generated ? false : (item.nullable === 'N' || item.nullable === 'NO' ||
```
samarpanB, line 1669: Just to avoid ternary operator which causes cognitive complexity and difficult to understand.

```suggestion
        required: !item.generated && (item.nullable === 'N' || item.nullable === 'NO' ||
```

lib/datasource.js
```diff
@@ -1830,10 +1832,11 @@ DataSource.prototype.discoverSchemasSync = function(modelName, options) {
     const propName = nameMapper('column', item.columnName);
     schema.properties[propName] = {
       type: item.type,
-      required: (item.nullable === 'N'),
+      required: item.generated ? false : item.nullable === 'N',
```
samarpanB, line 1835:
```suggestion
      required: !item.generated && item.nullable === 'N',
```

test/discovery.test.js
```diff
@@ -34,6 +34,7 @@ describe('Memory connector with mocked discovery', function() {
       dataPrecision: null,
       dataScale: null,
       nullable: 0,
+      generated: undefined,
```
samarpanB, line 37: Shouldn't we need to add a test case where generated is true ?

samarpanB (changes requested):

aaqilniz:
Hi @achrinza . The commits are squashed. Looking forward to getting this PR merged. 
Thanks.

samarpanB (approved):

dhmlau:
LGTM. I just kicked off CI. Let's wait to see the results. Thanks.

aaqilniz:
Thanks a lot, @dhmlau. It looks like every check has been passed and the PR can be merged.

dhmlau:
Thanks @aaqilniz for your contribution. Your PR has been merged! 🎉

PR #2039, opened 2023-03-06
fix: isolate context for each createAll call

akshatdubeysf:
Signed-off-by: akshatdubeysf <akshat.dubey@sourcefuse.com>

Multiple calls to createAll were using a global context variable so this shared variable was causing unexpected behaviour for concurrent calls. Changed the scope of this variable to maintain a different instance for each call of createAll.

Fixes https://github.com/loopbackio/loopback-datasource-juggler/issues/2038

## Checklist

- [X] [Sign off your commits](https://loopback.io/doc/en/contrib/code-contrib.html) with DCO (Developer Certificate of Origin)

### DIFF
--- a/lib/dao.js
+++ b/lib/dao.js
@@ -669,7 +669,7 @@ DataAccessObject.createAll = function(dataArray, options, cb) {
       });
     }
 
-    context = objArray.map(obj => {
+    const context = objArray.map(obj => {
       return {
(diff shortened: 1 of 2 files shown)


samarpanB (approved):
LGTM. Let the tests pass, then I will merge it.

PR #2158, opened 2023-11-24
fix: add error handling in case of unvailable cache server

aaqilniz:
In case of an unavailable datasource server (like the Redis server), the app crashes.

With this PR, the users can add an option in the datasource config named `catchFailure` and set its value to true. This would prevent the app from crashing and show the connection failure error in the console.

## Checklist

### DIFF
--- a/lib/datasource.js
+++ b/lib/datasource.js
@@ -499,7 +499,15 @@ DataSource.prototype.setup = function(dsName, settings) {
           debug('Connection fails: %s\nIt will be retried for the next request.', err);
         } else {
           g.error('Connection fails: %s\nIt will be retried for the next request.', err);
-          this.emit('error', err);
+          if (settings.catchFailure) {
+            try {
+              this.emit('error', err);
+            } catch (error) {
+              console.log(error);
+            }
+          } else {
+            this.emit('error', err);
+          }
         }
       } else {
         // Either lazyConnect or connector initialize() defers the connection


lib/datasource.js
```diff
@@ -499,7 +499,11 @@ DataSource.prototype.setup = function(dsName, settings) {
           debug('Connection fails: %s\nIt will be retried for the next request.', err);
         } else {
           g.error('Connection fails: %s\nIt will be retried for the next request.', err);
-          this.emit('error', err);
+          try {
+            this.emit('error', err);
```
samarpanB, line 503: Shouldn't we handle this in redis connector ?

aaqilniz: I am not sure how we can achieve that. 

The redis-kv connector receives a callback function [here](https://github.com/loopbackio/loopback-connector-kv-redis/blob/14439a26e4e4aa4c7bb723da0d95947c7a84f30c/lib/kv-redis.js#L50) from the juggler. The function it receives from juggler is [postInit](https://github.com/loopbackio/loopback-datasource-juggler/blob/f80a3512255a85ef4967142a19be44b8f8fd23bf/lib/datasource.js#L472) - which ultimately handles the connection/error from the connector.

dhmlau: it might make sense to put a try catch block here. But shouldn't it be throwing the error instead of printing it out to console?

aaqilniz: Printing it console would allow the app to continue working and not crash.

This is what I had in my mind. Whenever the Redis server is not available, the app crashes. 
Shouldn't it be up and running and just show an error in the console when the server is not available rather than crashing?

samarpanB: I just wanted to understand why "this.emit" throws an error in case of redis. Can you explain and add that to PR description ?

samarpanB: I think now I understand why this is added here. This issue is not specific to redis. It is happening for any connector. But this change will impact all users using any type of connector as currently there server crashes if DB or redis is not connecting. After this PR, servers will keep running which might give them a completely different behaviour than expected. For some cases, it might be undesirable as well. @aaqilniz the better way to deal with this is if we create a config param to handle this in connector. Then read that in juggler and then handle this crash. By default that config will be disabled so every existing connector will keep working as is, with an additional capability to change the behaviour. Then we can give this capability in individual connectors and consumers can also override it if needed. What say @aaqilniz ?

aaqilniz: I think you are right. Changing its behavior permanently would not be a good idea. I will add a config option to make this optional with the default being disabled.

aaqilniz: Hi, @samarpanB. Can you have a look at the PR now? I have enabled to addition of an option (`preventCrashWhenUnvailable` - any suggestions on improving this name?) to make this change optional. Let me know what you think.

aaqilniz:
Hi, @samarpanB. I've made the changes to be optional as you suggested. Can you please confirm if we can move forward with the PR?

lib/datasource.js
```diff
@@ -499,7 +499,15 @@ DataSource.prototype.setup = function(dsName, settings) {
           debug('Connection fails: %s\nIt will be retried for the next request.', err);
         } else {
           g.error('Connection fails: %s\nIt will be retried for the next request.', err);
-          this.emit('error', err);
+          if (settings.preventCrashWhenUnvailable) {
```
samarpanB, line 502: This setting need to come from connectors. So you might need to add this connector base repo and then a=enable this in redis connector repo.

samarpanB (changes requested):

lib/datasource.js
aaqilniz: This setting actually comes from the datasource config. So I don't think we need to make any changes to the connectors.
This option can be passed from the datasource config in an app.

samarpanB: Let's rename it to catchFailure

aaqilniz: Hi, @samarpanB. I just made the changes. Can we merge the PR now?

samarpanB (approved):

samarpanB:
LGTM, but I would like to get one more approval on this before merging.

aaqilniz:
Hi, @dhmlau. Can you have a look at this PR please?

patrioticcow:
Any ideas when this will get merged in?

PR #2175, opened 2023-12-28
feat: provide support for groupby

aaqilniz:
This PR enables the loopback-datasource-juggler to  add min(), max(), count(), avg() & sum() in the result.

## Checklist


lib/connectors/memory.js
```diff
@@ -502,7 +502,41 @@ Memory.prototype._findAllSkippingIncludes = function(model, filter) {
     // limit/skip
     const skip = filter.skip || filter.offset || 0;
     const limit = filter.limit || nodes.length;
+    // groupBy
     nodes = nodes.slice(skip, skip + limit);
+    if (filter.groupBy) {
```
samarpanB, line 507: this will require support for connector too

test/crud-with-options.test.js
```diff
@@ -272,6 +272,28 @@ describe('crud-with-options', function() {
       User.find({limit: 3});
     });
 
+    it('should allow filter with groupBy, count, max, min, sum & avg', function(done) {
```
samarpanB, line 275: can you add test cases for when there are other filter queries too and limit and offset is also provided ? Basically to confirm whether this feature works with other features together or not. And whether it has any impact there.

lib/connectors/memory.js
aaqilniz: Yes. And it's implemented [here](https://github.com/loopbackio/loopback-connector/pull/466).

test/crud-with-options.test.js
aaqilniz: Hi, @samarpanB. I have added one more test to verify the change with existing filters.

PR #2407, opened 2025-06-02
chore: reduce dependencies from qs

benmccann:
https://npmgraph.js.org/?q=qs - 18 dependencies
https://npmgraph.js.org/?q=neoqs - 0 dependencies

I used the legacy entry point to keep CJS support

## Checklist

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -455,7 +455,7 @@ function sanitizeQuery(query, options) {
 }
 
 const url = require('url');
-const qs = require('qs');
+const qs = require('neoqs/legacy');
 
 /**
  * Parse a URL into a settings object
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,8 +18,8 @@
         "loopback-connector": "^6.2.4",
         "minimatch": "^10.0.1",
         "nanoid": "^3.3.11",
+        "neoqs": "^6.13.0",
         "neotraverse": "^0.6.18",
-        "qs": "^6.14.0",
         "strong-globalize": "^6.0.6",
         "uuid": "^11.1.0"
       },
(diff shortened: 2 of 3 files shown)


dhmlau:
@benmccann, thanks for the PR. While I understand the reason to switch to `neoqs`, I'm concerned about the maturity and the download volume of `neoqs` comparing with `qs`. I'd prefer to keeping `qs`. 
@loopbackio/loopback-maintainers, any thoughts?

benmccann:
Yeah, you're right that `neotraverse` has much higher download numbers than `neoqs`. I wasn't sure how much that mattered since you'd made the switch from `traverse` to `neotraverse` eventhough `traverse` has 10x as many downloads.

I guess you're thinking more downloads means it's more battle tested? Though `neoqs` shares a test suite with `qs`, so I think they have basically the same level of robustness and the same history of being battle tested. The reason I pointed out that it was the same author as `neotraverse` is to share that it's a very reputable author. I've interacted with the authors of both libraries many times via other projects and have generally seen the `neotraverse` author to be far more responsive when there's any issue.

As an alternative, https://www.npmjs.com/package/qs-esm is more downloaded, but it looks like this project still supports CJS and Node 18 and you can't require an ESM library until Node 20. I see that Node 18 was just dropped in core in https://github.com/loopbackio/loopback-next/commit/88a3fe98f8ca217271eb028697278355ec7c9478, so perhaps we could bump the Node version requirement here as well and use that library?

Yeah, my issue is how many dependencies are required. It's easy to think of it as not being that big of an issue, but the problem really adds up when using many dependencies. As an example, just installing the core docusaurus library without any plugins or anything adds up to over 1000 dependencies (https://npmgraph.js.org/?q=@docusaurus/core). This makes debugging anything a nightmare because you have to go through so many layers of libraries and there's so much code in the distributable not to mention the supply chain risks. As a result, it's something I check with every new library I adopt.

